PR #18, opened 2022-03-21
grid

ararus:


packages/grid/stories/grid-vuu.stories.tsx
```diff
@@ -0,0 +1,126 @@
+import {
```
heswell, line 1: Not sure what we should do with this story. It's not going to work for most people. Also, shouldn't be displayed under Lab.

I suggest remove this one altogether.

packages/grid/src/grid/features/useColumnMove.ts
```diff
@@ -0,0 +1,68 @@
...
+      columnMoveDataRef.current!;
+    const shiftX = event.screenX - startScreenX;
+    const shiftY = event.screenY - startScreenY;
+    const x = startHeaderX + shiftX;
+    const y = startHeaderY + shiftY;
+    model.columnDragAndDrop.move(x, y);
+  }, []);
+
+  const onColumnMoveHandleMouseDown: MouseEventHandler<HTMLDivElement> =
+    useCallback((event) => {
```
heswell, line 34: There is no delay or 'drag threshold' here, so column goes into drag mode even when user is just clicking column header (eg to sort)

heswell (approved):
I haven't attempted a detailed review of this PR, just enough to familiarize myself with the component. I think we should go ahead and merge so future work can operate with smaller PRs for specific tasks

PR #83, opened 2022-05-05
Flow and Stack layouts

Fercas123:


packages/lab/src/layout/SplitLayout/SplitLayout.css
```diff
@@ -0,0 +1,34 @@
+.uitkSplitLayout {
+  justify-content: space-between;
+}
+
+.uitkSplitLayout-separator {
+  --uitkFlexLayout-gap: calc(var(--uitk-spacing-unit) * var(--uitkFlexLayout-gap-multiplier) + var(--uitkFlexLayout-separator));
+}
+
+.uitkSplitLayout-separator > *:first-child::after,
+*:last-child::before,
```
heswell, line 10: missing the first half of rule here, needs same prefix as rule abobe. As-is, has pretty wide reaching effects.

PR #95, opened 2022-05-08
Test ag grid custom theme

origami-z:
Feasibility test on custom ag grid theme

light theme grid

dark theme grid

Steps:

- Modify `packages/ag-grid-theme/css/styles.scss` (use `reference.scss` to see which variables are available)
- Run build script in `packages/package.json`, to generate `index.css`
- Run the test app `yarn start:ag-grid-test-app`

## Discussions (WIP)

We may need to generate multiple sets of CSS, and bolt on some additional CSS we need

### High levels

- Ag grid's `@include ag-theme-alpine` scss function doesn't work well with dual class names (e.g. `.uitk-light .ag-theme-uitk`), so fall back to define two themes (e.g. `.ag-theme-uitk-dark`)
- The same function's `row-height` doesn't support CSS var due to  internal function, so `.uitk-density-xyz` won't be easily integrated. This is probably less a problem, as users less likely need to switch density. We can duplicate more classes

### Minor

- `AgGridReact` need to supply `rowHeight` / `headerHeight` on top of CSS height defined
- Row hover shadow
- Selected cell focus style (i.e. box-shadow instead of out of box border)

### Others

- ~~For some reason, prettier is doing terribly wrong on the SCSS files, where some characters at the each single line comment is removed, which doesn't happen on the [prettier playground]9https://prettier.io/playground/)~~ fixed in #96 

## References 

- https://www.ag-grid.com/react-data-grid/themes-customising/#setting-colour-parameters-using-css-variables


origami-z:
#242 will cover the full implementation

PR #98, opened 2022-05-09
Form Field CSS API & QA

libertymayc:


packages/lab/src/form-field/FormActivationIndicator.css
```diff
@@ -1,25 +1,40 @@
...
+  border-bottom-style: var(--uitkFormActivationIndicator-border-bottom-style, var(--uitk-editable-border-style));
   height: auto;
   opacity: var(--uitkFormActivationIndicator-opacity, 1);
   width: 100%;
+
   /* We don't want the activation indicator to increase height of it's container */
-  position: absolute;
-  bottom: calc(var(--uitkFormFieldHelperText-height, 0px) + 1px);
+  bottom: calc(var(--uitkFormFieldHelperText-line-height, var(--uitk-typography-line-height)) + 3px);
+  position: var(--uitkFormLabel-position, absolute);
```
origami-z, line 25: `--uitkFormLabel-position` may not be the best name for the indicator

packages/lab/src/form-field/FormLabel.tsx
```diff
@@ -25,6 +27,10 @@ export interface FormLabelProps
    * The label value for the FormLabel
    */
   label?: string;
+  /**
+   * Location of label is left; default false
+   */
+  labelLeft?: boolean;
```
origami-z, line 33: Not entirely sure whether making the label aware of it's position is a good idea. If this is the only way to do it, probably keep this the similar fashion of the prop from form field.

`labelPlacement="left"`

origami-z (dismissed):

origami-z:
Was looking at preview page

Left spacing of the label is off
form field label top

Left label width
left label width

readonly underline color

readonly form

Validation state colors missing
validation states

Background misalignment
image

libertymayc:
Left label alignment has changed in agreement with design (Ben) as previously it would overlap a FormField to the left of it (see comparison screenshot in touch density)

origami-z:
Left aligned label vertical alignment with input
left aligned form field

Should the emphasis background go under helper text?
high emphasis background

Custom styling active state could have the line?
image


> Left label alignment has changed in agreement with design (Ben) as previously it would overlap a FormField to the left of it (see comparison screenshot in touch density)

Does this mean background color should also apply to left label?

libertymayc:
All done

> > Left label alignment has changed in agreement with design (Ben) as previously it would overlap a FormField to the left of it (see comparison screenshot in touch density)
> 
> Does this mean background color should also apply to left label?

No, it just doesnt overflow anymore

packages/theme/css/characteristics/editable.css
```diff
@@ -6,6 +6,7 @@
 
   --uitk-editable-border-style: solid;
 
+  --uitk-editable-border-style: solid;
```
alycrys, line 9: duplicate of line 7

PR #129, opened 2022-05-25
update gap defaults

Fercas123:
After a conversation with Darrin and Ben, we are using 3x as a default for every density but TD where the default will be 2x

### DIFF
--- a/packages/lab/src/__tests__/__e2e__/layout/Grid.cy.tsx
+++ b/packages/lab/src/__tests__/__e2e__/layout/Grid.cy.tsx
@@ -10,7 +10,8 @@ const {
   ToolkitGridLayoutResponsiveView,
 } = composedStories;
 
-const testColumnsNumber = columns => new RegExp(`^(\\d*\\.?\\d*px *){${columns }}$`);
+const testColumnsNumber = (columns) =>
+  new RegExp(`^(\\d*\\.?\\d*px *){${columns}}$`);
 
 describe("GIVEN a Grid", () => {
   checkAccessibility(composedStories);
@@ -85,7 +86,7 @@ describe("GIVEN a Grid", () => {
 
         cy.get(".uitkGridLayout")
           .invoke("css", "grid-template-columns")
-          .should( "match", testColumnsNumber(12));
+          .should("match", testColumnsNumber(12));
 
         cy.get(".uitkGridLayout").should(
           "have.css",
@@ -108,11 +109,11 @@ describe("GIVEN a Grid", () => {
 
         cy.get(".uitkGridLayout")
           .invoke("css", "grid-template-columns")
-          .should( "match", testColumnsNumber(12));
+          .should("match", testColumnsNumber(12));
 
         cy.get(".uitkGridLayout")
           .invoke("css", "grid-template-rows")
-          .should( "match", testColumnsNumber(4));
+          .should("match", testColumnsNumber(4));
       }
     );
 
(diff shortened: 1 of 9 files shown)


packages/lab/src/__tests__/__e2e__/layout/Grid.cy.tsx
```diff
@@ -81,11 +81,9 @@ describe("GIVEN a Grid", () => {
...
         );
 
-        cy.get(".uitkGridLayout").should(
-          "have.css",
-          "grid-template-columns",
-          "72px 72px 72px 72px 72px 72px 72px 72px 72px 72px 72px 72px"
-        );
+        cy.get(".uitkGridLayout")
+          .invoke("css", "grid-template-columns")
+          .should("match", /^(\d*\.?\d*px *){12}$/);
```
JoanaMMoreira, line 86: 🎉

JoanaMMoreira (approved):

packages/lab/src/__tests__/__e2e__/layout/Grid.cy.tsx
origami-z: This may cause a little confusing for others when first seen

origami-z (approved):

packages/lab/src/__tests__/__e2e__/layout/Grid.cy.tsx
Fercas123: agree @origami-z, I tried extracting it to a testColumns() where you could set the number of them but cypress was timing out whenever i tried extracting it :(

Fercas123: @origami-z I wasn't scaping  `\`s 🤦‍♀️  pushed a change to clean it :)

packages/lab/src/layout/FlexLayout/FlexLayout.css
```diff
@@ -1,3 +1,16 @@
...
+.uitk-density-medium {
+  --uitkFlexLayout-gap-density-multiplier: 3;
+}
+.uitk-density-high {
+  --uitkFlexLayout-gap-density-multiplier: 3;
+}
+
 /* VARIABLES */
 .uitkFlexLayout {
   --uitkFlexLayout-gap-multiplier: 1;
```
libertymayc, line 16: we tend to use this --uitkFlexLayout-gap-multiplier as a CSS API, and our own local variables as --flex-layout-gap-multiplier (see Button for example)

so you'll have something like...
.uitk-density-high {
--flex-layout-gap-density-multiplier: 3;
}
etc..
and
.uitkFlexLayout {
--flex-layout-gap-multiplier: var(--uitkFlexLayout-gap-multiplier, 1);
--flex-layout-gap: calc(var(--uitk-size-unit) * var(--flex-layout-gap-multiplier) * var(--uitkFlexLayout-gap-density-multiplier, --flex-layout-gap-density-multiplier));
}

Fercas123: makes sense, I'll update those

origami-z (approved):

packages/lab/src/layout/GridLayout/GridLayout.css
```diff
@@ -1,5 +1,18 @@
+.uitk-density-touch {
+  --uitkGridLayout-gap-density-multiplier: 2;
```
libertymayc, line 2: same for these ones as flex :)

JoanaMMoreira (approved):

PR #134, opened 2022-05-27
auto gen characteristics table for component docs

libertymayc:


joshwooding:
I think the initial CSS Docgen pulls out all css variables. We should either limit that or re-use it's output.

heswell (approved):

PR #148, opened 2022-06-05
Add @types/react-window to lab devDependencies

origami-z:
Without this, consumer running `tsc` would run into below error

> node_modules/@jpmorganchase/uitk-lab/dist-types/list/useListItem.d.ts:1:41 - error TS7016: Could not find a declaration file for module 'react-window'. '/Users/f692193/projects/figma-export-to-css-variables/node_modules/react-window/dist/index.cjs.js' implicitly has an 'any' type.


origami-z:
react window will soon be removed. closing this.

PR #152, opened 2022-06-06
Add layer layout component

JoanaMMoreira:


el-dav:
Not sure why but when the component is fullscreen the close button no longer seems to work

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+    useEffect(() => {
+      if (!open && disableAnimations) {
+        setShowComponent(false);
+      }
+
+      if (open && !showComponent) {
+        setShowComponent(true);
+      }
+    }, [open, showComponent]);
```
el-dav, line 72: missing disableAnimations from dependency array

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export const LayerLayout = forwardRef<HTMLDivElement, LayerLayoutProps>(
+  function LayerLayout(props, ref) {
+    const {
+      children,
+      className,
+      displayScrim = true,
+      position = "center",
+      fullScreenAtBreakpoint = "sm",
+      disableAnimations = false,
```
el-dav, line 56: Not sure if it's a big deal but I noticed that if you toggle disableAnimations on and off while the Layer is 'open' then the Layer will show the enter animation each time disableAnimations is set to false.

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+      if (!open && disableAnimations) {
+        setShowComponent(false);
+      }
+
+      if (open && !showComponent) {
+        setShowComponent(true);
+      }
+    }, [open, showComponent]);
+
+    const fullScreen = useChangeView(fullScreenAtBreakpoint);
```
el-dav, line 74: Also noticing that when the `fullscreen` value changes the enter animation plays again

packages/lab/src/layout/LayerLayout/LayerLayout.css
```diff
@@ -0,0 +1,110 @@
+.uitkLayerLayout {
```
el-dav, line 1: Maybe we should have a `@media (prefers-reduced-motion)` media query  to disable the slide animations?

JoanaMMoreira:
> Not sure why but when the component is fullscreen the close button no longer seems to work

Good spot! Should be fixed now

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
JoanaMMoreira: Is this still an issue? I can't recreate it but I've also made some changes in the meantime so I'm not sure

packages/lab/src/layout/LayerLayout/LayerLayout.css
JoanaMMoreira: This was suggested for the ParentChild component as well but then we've agreed to add that as part of the storybook example. I can add an example for this component as well if we think that's the best solution? Or should we add @media (prefers-reduced-motion) to both components?

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+        {...rest}
+      >
+        {children}
+      </div>
+    ) : null;
+
+    return displayScrim ? (
+      <Scrim
+        open={showComponent}
+        className="uitkEmphasisMedium"
```
el-dav, line 102: Is this the intended API for getting the dark version of the Scrim? Feels like we should have a prop for this

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+  "bottom",
+] as const;
+
+export type LayerPosition = typeof LAYER_POSITION[number];
+
+export interface LayerLayoutProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Display or hide the scrim.
+   */
+  displayScrim?: boolean;
```
el-dav, line 23: I'm wondering if it would be better to remove scrim from this component and get the consumer to add it via composition if they need it?

This would avoid passing `displayScrim` and `scrimProps` to this component which I think we were trying to avoid.

@heswell / @joshwooding what do you think?

packages/lab/src/layout/utils/useChangeView.ts
```diff
@@ -4,15 +4,15 @@ import {
   useOrderedBreakpoints,
 } from "@jpmorganchase/uitk-core";
 
-export const useIsStacked = (stackedAtBreakpoint: keyof Breakpoints) => {
+export const useChangeView = (stackedAtBreakpoint: keyof Breakpoints) => {
```
el-dav, line 7: For me the name  `useChangeView` isn't super intuitive as to what this hook is doing. I guess the hook tells us if the viewport is larger than the breakpoint you're querying so maybe something like `useIsViewportLargerThanBreakpoint` (but something better if you can think of it! 😊 ).

We should also update the name of `stackedAtBreakpoint` argument

packages/lab/src/layout/LayerLayout/LayerLayout.css
el-dav: Interesting, so we're leaving it up to the consumer to add reduced motion.

I think for me it would be nice to have that built into the components at least for slide animations, but maybe a good one to discuss with the group on our next standup

el-dav:
> > Not sure why but when the component is fullscreen the close button no longer seems to work
> 
> Good spot! Should be fixed now

Might be a good one to add tests for if it's not a massive pain

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
libertymayc: we removed emphasis as a prop and as a replacement the className should be added as above. this is so that we don't have to add an emphasis prop to every single component + if anyone wants to add emphasis styles to components that UITK doesn't by default then it makes it easier

the default emphasis varies by component - for scrim it's 'low', and only low and medium are defined by UITK (although someone could add high emphasis in the future if they wanted), so for now the classname only needs to be applied for medium emphasis (like here)

i will write some docs for emphasis, as i do feel it is perhaps not intuitive and writing it out might give me some ideas on anything we could do differently.. let me know any suggestions or if anything needs clarifying

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <div
+        ref={ref}
+        className={cx(withBaseName(), className, {
+          [withBaseName("anchor")]: anchored,
+          [withBaseName("fullScreen")]: fullScreen,
+          [withBaseName(position)]: !fullScreen,
+          [withBaseName("enter-animation")]: !disableAnimations && open,
+          [withBaseName("exit-animation")]: !disableAnimations && !open,
+        })}
+        onAnimationEnd={() => {
```
el-dav, line 88: Just making a note of the potential issue we discussed.

Using the onAnimationEnd callback could be brittle if the animation get's disabled by another mechanism other than props e.g. via a prefers reduced motion media query.

Unless we use some other mechanism you might need to use the onAnimationStart callback to store in state whether the component is animating and then use that to determine if the component should be shown

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
JoanaMMoreira: As discussed I'll have a catch up with Ben about the desired behaviour for this

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
JoanaMMoreira: Thank you, I'm looking into this

packages/lab/src/layout/LayerLayout/LayerLayout.css
```diff
@@ -0,0 +1,112 @@
...
+  --layer-border-color: var(--uitkLayerLayout-border-color, var(--uitk-container-border-color-medium));
+  --layer-box-shadow: var(--uitkLayerLayout-box-shadow, var(--uitk-overlayable-shadow-modal));
+  --layer-background-color: var(--uitkLayerLayout-background-color, var(--uitk-container-background-medium));
+  --layer-border-width: var(--uitkLayerLayout-border-color, var(--uitk-container-border-width));
+  --layer-padding: var(--uitkLayerLayout-padding, calc(3 * var(--uitk-size-unit)));
+  --layer-border-style: var(--uitkLayerLayout-border-style, var(--uitk-container-border-style));
+  --layer-margin: var(--uitkLayerLayout-margin, calc(3 * var(--uitk-size-unit)));
+  --uitk-animation-duration: 0.2s;
```
libertymayc, line 9: We should add this into the foundation as some sort of state/variant instead of override it here

packages/lab/src/layout/LayerLayout/LayerLayout.css
```diff
@@ -0,0 +1,112 @@
+.uitkLayerLayout {
+  --layer-border-color: var(--uitkLayerLayout-border-color, var(--uitk-container-border-color-medium));
+  --layer-box-shadow: var(--uitkLayerLayout-box-shadow, var(--uitk-overlayable-shadow-modal));
+  --layer-background-color: var(--uitkLayerLayout-background-color, var(--uitk-container-background-medium));
+  --layer-border-width: var(--uitkLayerLayout-border-color, var(--uitk-container-border-width));
+  --layer-padding: var(--uitkLayerLayout-padding, calc(3 * var(--uitk-size-unit)));
+  --layer-border-style: var(--uitkLayerLayout-border-style, var(--uitk-container-border-style));
```
libertymayc, line 7: we decided on using camel case for attributes in tokens, so e.g. --layer-borderStyle and --uitkLayerLayout-borderStyle. most components still have the old style but ready ones (e.g. Button) have adopted the camel case already which can be looked at for reference :) 

also --uitk-container-border-style uses camel case now as well as border width/color above.. perhaps it was missed in a rebase here? best to check where ever the characteristics are used

packages/lab/src/layout/LayerLayout/LayerLayout.css
```diff
@@ -0,0 +1,112 @@
+.uitkLayerLayout {
+  --layer-border-color: var(--uitkLayerLayout-border-color, var(--uitk-container-border-color-medium));
+  --layer-box-shadow: var(--uitkLayerLayout-box-shadow, var(--uitk-overlayable-shadow-modal));
+  --layer-background-color: var(--uitkLayerLayout-background-color, var(--uitk-container-background-medium));
```
libertymayc, line 4: try to leave out the -color in -background tokens and just use the background attribute too, which will allow people to use images/gradients etc if they wish

packages/lab/src/layout/LayerLayout/LayerLayout.css
```diff
@@ -0,0 +1,112 @@
...
+}
+
+.uitkLayerLayout-anchor.uitkLayerLayout-top {
+  left: 0;
+  right: 0;
+  bottom: auto;
+}
+
+.uitkLayerLayout-anchor.uitkLayerLayout-top.uitkLayerLayout-enter-animation {
+  animation: slide-in-bottom var(--uitk-animation-duration) var(--uitk-animation-timing-function);
```
libertymayc, line 64: should these values be put into the animation foundation?

packages/lab/stories/layout/styles.css
```diff
@@ -101,6 +101,19 @@
   padding: calc(var(--uitk-size-unit) * 3) 0 calc(var(--uitk-size-unit) * 2) 0;
 }
```
libertymayc, line 103: Maybe rename this file to layer-layout.stories.css

packages/lab/src/layout/LayerLayout/LayerLayout.css
JoanaMMoreira: Makes sense, I've updated these

packages/lab/stories/layout/styles.css
JoanaMMoreira: This file contains styles for multiple layout components examples so might be best to leave it as is or call it `layout.stories.css`

JoanaMMoreira:
> > > Not sure why but when the component is fullscreen the close button no longer seems to work
> > 
> > 
> > Good spot! Should be fixed now
> 
> Might be an interesting example to add tests for

I've added this

packages/theme/css/foundations/animation.css
```diff
@@ -21,10 +22,24 @@
   --uitk-animation-slide-out-left: slide-out-left var(--uitk-animation-duration) var(--uitk-animation-timing-function);
   --uitk-animation-slide-out-right: slide-out-right var(--uitk-animation-duration) var(--uitk-animation-timing-function);
   --uitk-animation-slide-out-bottom: slide-out-bottom var(--uitk-animation-duration) var(--uitk-animation-timing-function);
-  /* Fade in Animations */
+
+  /* Drawer Slide Animations */
```
Fercas123, line 26: is there a reason to add this ones here instead of overriding `--uitk-animation-duration` in the layer layout css?

packages/lab/src/layout/utils/useIsViewportLargerThanBreakpoint.ts
```diff
@@ -4,15 +4,17 @@ import {
   useOrderedBreakpoints,
 } from "@jpmorganchase/uitk-core";
 
-export const useIsStacked = (stackedAtBreakpoint: keyof Breakpoints) => {
+export const useIsViewportLargerThanBreakpoint = (
```
Fercas123, line 7: should this not be in the general utils? i guess if there is nothing else that will use it it makes sense here, not sure if there would be other use cases

packages/lab/stories/layout/layouts.doc.lab.stories.mdx
```diff
@@ -31,3 +31,13 @@ Displays a hierarchical structure comprising of a main content area and an accom
...
 </Canvas>
+
+## [Layer layout](/docs/layout-docs-layerlayout--default-layer-layout)
+
+Defines a layer above the existing layout structure for UI elements to be displayed in.
+
+It can be used to build dialog or drawer components and it can be positioned in different parts of the screen.
+
+<Canvas>
+  <Story id="layout-layerlayout--default-layer-layout" />
```
Fercas123, line 42: this will probably need updating, it is hanging at the moment

Fercas123, line 42:
```suggestion
  <Story id="lab-layout-layerlayout--default-layer-layout" />
```

packages/lab/stories/layout/layouts.doc.lab.stories.mdx
```diff
@@ -31,3 +31,13 @@ Displays a hierarchical structure comprising of a main content area and an accom
 <Canvas>
   <Story id="lab-layout-parentchildlayout--default-parent-child-layout" />
 </Canvas>
+
+## [Layer layout](/docs/layout-docs-layerlayout--default-layer-layout)
```
Fercas123, line 35: the link needs to be 
```suggestion
## [Layer layout](/docs/documentation-lab-layout-layerlayout--page)
```

packages/lab/stories/layout/layer-layout.doc.stories.mdx
```diff
@@ -0,0 +1,70 @@
...
+
+## API
+
+```
+import {LayerLayout} from "@jpmorganchase/uitk-lab";
+```
+
+## Props
+
+<ArgsTable of={LayerLayout} />
```
Fercas123, line 36: this table is not showing

Fercas123 (commented):
it's looking good! left a few comments

packages/theme/css/foundations/animation.css
JoanaMMoreira: @libertymayc suggested we move this into foundations to avoid having the override which makes sense if we ever have similar components that would use this duration instead of the default one

packages/lab/src/layout/utils/useIsViewportLargerThanBreakpoint.ts
JoanaMMoreira: Makes sense, I think this could be used in other components like `ButtonBar` so I've moved it to the general utils folder

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
JoanaMMoreira: @heswell @joshwooding @origami-z any thoughts on this please? In the spec we have "scrim displayed by default" hence I've added it as part of the component

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
JoanaMMoreira: Should be fixed now

Fercas123 (dismissed):

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    useEffect(() => {
+      if ((!open && disableAnimations) || (!open && !hasAnimations)) {
+        setShowComponent(false);
+      }
+
+      if (open && !showComponent) {
+        setShowComponent(true);
+      }
+    }, [open, showComponent, disableAnimations, hasAnimations]);
```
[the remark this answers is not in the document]
JoanaMMoreira: As discussed, I've renamed the `open` prop and will add an issue to look at this behaviour across `Layer`, `ParentChild` and `Deck`

Fercas123 (approved):

origami-z:
The docs page has some issue?
https://2781c30b.uitk.pages.dev/?path=/docs/documentation-lab-layout-layerlayout--page

JoanaMMoreira:
> The docs page has some issue? https://2781c30b.uitk.pages.dev/?path=/docs/documentation-lab-layout-layerlayout--page

Should be fixed, it was only a matter of updating the `open` prop that had been renamed but I also noticed that it wasn't still displaying the classes properly in the anatomy section so I've manually added this. I'm not sure if there's a better way to do it but it's a bit tricky with a component like this where we need to "open" it in the docs

Fercas123 (approved):

docs/components/ComponentAnatomy.jsx
```diff
@@ -57,21 +57,21 @@ export const ComponentAnatomy = ({
   const getComponentElement = () =>
     targetId
       ? document.querySelector(`#${targetId}`)
-      : root.current.querySelector(".RenderVisualiser-component > *");
+      : root.current?.querySelector(".RenderVisualiser-component > *");
+
+  const target = getComponentElement();
```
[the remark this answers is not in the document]
origami-z: Looks like Button one is broken now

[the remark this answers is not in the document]
JoanaMMoreira: oh :( I think this is a bigger issue than I expected so I'll revert these changes and will continue looking at this on a separate PR

packages/core/src/utils/useIsViewportLargerThanBreakpoint.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  Breakpoints,
+  useCurrentBreakpoint,
+  useOrderedBreakpoints,
+} from "@jpmorganchase/uitk-core";
```
Fercas123, line 5: this should 
```suggestion
import {Breakpoints} from "../breakpoints";
import {useCurrentBreakpoint, useOrderedBreakpoints} from "./useResponsiveProp";
```

origami-z (approved):

PR #166, opened 2022-06-08
Use the next version of Chromatic

joshwooding:


heswell (approved):

origami-z:
Will Chromatic ignore what's in lock file?

PR #188, opened 2022-06-13
Fix dropdown combo list zindex

origami-z:
Dropdown list currently doesn't have z index set so helper text appears in front

**Before**

https://uitk.pages.dev/?path=/story/lab-dropdown--with-form-field-label-top
dropdown helper text visible error


.changeset/ten-ties-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-lab": minor
+---
+
+Add z index to dropdown and combobox list root
```
joshwooding, line 5: Nit
```suggestion
Add z-index to dropdown and combobox list root
```

PR #189, opened 2022-06-13
FileDropZone characteristic usage

libertymayc:


origami-z (dismissed):
Do we need backward comp?

origami-z:
Should file dropzone have hover color at all ? I thought it's only the button needed.

file dropzone hover

origami-z:
So change of color on hover normally means interactivity, but clicking the light blue area doesn't do anything. Are we sure that's what we want?

libertymayc:
maybe it makes sense to have it when something is actively being dragged in. can we detect that? @origami-z

origami-z:
Not sure the best way to handle it, but listening to drag events and add class would do?

origami-z (approved):

origami-z (dismissed):

heswell (dismissed):

joshwooding (changes requested):
I'm seeing some weird flickering when hovering over the drop zone with a file

joshwooding (dismissed):
Raised https://github.com/jpmorganchase/uitk/issues/279 for the flickering issue but everything else looks good

joshwooding (approved):

PR #190, opened 2022-06-14
add variables to Button, Icon API

heswell:
small changes needed later in upcoming List and Toolbar PRs


packages/core/src/button/Button.css
```diff
@@ -54,7 +54,7 @@
 
 /* Icon API in "primary" variant @ignore */
 .uitkButton-primary {
-  --uitkIcon-color: var(--uitk-actionable-primary-foreground);
+  --uitkIcon-color: var(--uitkButton-icon-color, var(--uitk-actionable-primary-foreground));
```
origami-z, line 57: Wonder the casing should be similar to `--uitkButton-textColor`?

packages/core/src/button/Button.css
```diff
@@ -54,7 +54,7 @@
 
 /* Icon API in "primary" variant @ignore */
 .uitkButton-primary {
-  --uitkIcon-color: var(--uitk-actionable-primary-foreground);
+  --uitkIcon-color: var(--uitkButton-iconColor, var(--uitk-actionable-primary-foreground));
```
libertymayc, line 57: I was thinking to keep camel case to just CSS attributes and component names. since icon-color isn't a css attribute, as well as text-color, i was going to keep those kebab case. thoughts? (I just put up a FormField PR where you can see where I have changed to camel and where not) happy to discuss

origami-z:
Need to consider the 'one word' conversion for tooling as well

libertymayc (commented):
Could we change in this PR --button-textColor to -text-color and --uitkButton-textColor/wherever else it is used

@heswell see PR 144

joshwooding (changes requested):
The changeset needs to be updated but otherwise looks goods.

joshwooding (approved):

libertymayc (approved):

PR #191, opened 2022-06-14
FormField consuming characteristics and styling fixes

libertymayc:


packages/lab/src/form-field/FormField.css
```diff
@@ -1,67 +1,45 @@
...
-}
-.uitk-density-high {
-  --form-field-helper-text-margin-top: 2px;
-  --form-field-activation-indicator-icon-size: 3px;
-}
-
-/* above, if `fillWidth={true}` */
-.uitk-density-touch {
-  --form-field-helper-text-margin-top: 5px;
+/* Style applied to the root element */
```
libertymayc, line 1: This is an interesting one, I had to add :root here to allow for the variables to be found in :after pseudo.. there might be a better fix or perhaps im missing something @origami-z

origami-z: This will be a devtool bug, which works in Firefox.

However, i do wonder whether we should visual regression test on focus ring.

libertymayc: good idea

origami-z (approved):

heswell (approved):

PR #192, opened 2022-06-15
Only snapshot QA stories

joshwooding:


packages/lab/src/metric/MetricContent.css
```diff
@@ -19,6 +19,10 @@
   margin-right: 4px;
 } */
 
+.uitkMetricContent-value {
```
joshwooding, line 22: @alycrys This was needed to prevent Metric having a scrollbar on Windows.

heswell (approved):

heswell (approved):

PR #215, opened 2022-06-21
New storybook nav structure for docs; Checkbox, Switch docs

libertymayc:
New nav structure makes it easier to find and ensure that the following is included in all complete component docs:

- API
- Props Table
- HTML Anatomy
- CSS Class
- Characteristics Used
- Custom Property API
- Customising styles

Also changes how stories are referred to by their id rather than re-writing them within the docs. Allows us to also ensure we give supporting documentation to all the existing stories by directly referencing them. I think we should not have each of these listed in the sidebar again, but we should somehow have an 'Examples' anchor in the sidebar just before the first example

### DIFF
--- a/packages/core/stories/button.doc.stories.mdx
+++ b/packages/core/stories/button.doc.stories.mdx
@@ -1,12 +1,6 @@
 import { Meta, Story, Canvas, ArgsTable } from "@storybook/addon-docs";
 import { ComponentAnatomy } from "docs/components/ComponentAnatomy";
 import { Button, ToolkitProvider } from "@jpmorganchase/uitk-core";
-import {
-  DownloadIcon,
-  SearchIcon,
-  SendIcon,
-  SettingsSolidIcon,
-} from "@jpmorganchase/uitk-icons";
 import { withFlexGap } from "docs/decorators/withFlexGap";
 import { CustomStylingExample } from "./button.stories";
 import {
@@ -32,11 +26,9 @@ A Button executes an action when clicked. Button labels should use only one or t
 Button contains three variants `cta`, `primary` and `secondary`. If not specified, the default value is `primary`.
 
 <Canvas>
-  <Story name="Button Variants" decorators={[withFlexGap]}>
-    <Button variant="cta">CTA</Button>
-    <Button variant="primary">Primary</Button>
-    <Button variant="secondary">Secondary</Button>
-  </Story>
+  <Story id="core-button--cta" />
+  <Story id="core-button--primary" />
+  <Story id="core-button--secondary" />
 </Canvas>
 
 ## Button with icon and label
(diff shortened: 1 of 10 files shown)


packages/core/stories/button.doc.stories.mdx
```diff
@@ -100,7 +62,9 @@ import { Button } from "@jpmorganchase/uitk-core";
 
 ## Props
 
-<ArgsTable of={Button} />
+<Story name="Props">
```
joshwooding, line 65: I understand the benefits but I still don't think we should declare stories in mdx.

packages/core/stories/button.doc.stories.mdx
```diff
@@ -32,61 +26,29 @@ A Button executes an action when clicked. Button labels should use only one or t
 Button contains three variants `cta`, `primary` and `secondary`. If not specified, the default value is `primary`.
 
 <Canvas>
```
heswell, line 28: Problem with this approach is that if user view the code , they don't see anything helpful, just a bunch of these:
'''
() => {
  return <ButtonGrid variant="cta" label="CTA" />;
}
'''
We need to make sure the target Story is simpler if we're going to link like this

packages/lab/src/pill/Pill.css
```diff
@@ -0,0 +1,201 @@
...
+  --pill-text-color-active: var(--uitkPill-text-color-active, var(--uitk-taggable-foreground-active));
+  --pill-text-color-hover: var(--uitkPill-text-color-hover, var(--uitk-taggable-foreground-hover));
+}
+
+/* Button Component API - reset to default state so they don't interfere with Pill styling */
+.uitkPill {
+  --uitkButton-background-active: var(--pill-background);
+  --uitkButton-textColor-active: var(--pill-text-color);
+  --uitkButton-background-hover: var(--pill-background);
+  --uitkButton-textColor-hover: var(--pill-text-color);
```
heswell, line 37: I've lost track of what we decided on textColor vs text-color. We seem to have both

packages/lab/src/pill/Pill.css
```diff
@@ -0,0 +1,201 @@
...
+/* Style applied to Pill when active if `deletable={false}`, `selectable={false}`, and `disabled={false}` */
+.uitkPill:not(.uitkPill-deletable):not(.uitkPill-selectable):not(.uitkPill-disabled):active,
+.uitkPill:not(.uitkPill-deletable):not(.uitkPill-selectable):not(.uitkPill-disabled).uitkPill-active {
+  background: var(--pill-background-active);
+  color: var(--pill-text-color-active);
+}
+
+/* Styles applied to the Pill checkbox */
+.uitkPill-checkbox {
+  --pill-checkbox-stroke: var(--uitkPill-checkbox-stroke, var(--uitk-selectable-borderColor));
```
heswell, line 194: can't see these two variables used anywhere

packages/core/stories/button.doc.stories.mdx
```diff
@@ -111,32 +75,34 @@ export const Link = ({ children, to, ...props }) => (
...
-      elementType={Link}
-      to="/"
-      aria-label="Link as elementType"
-      role={null}
-    >
-      Custom Link
-    </Button>
-  </ComponentAnatomy>
+  <Story name="HTML Anatomy">
+    <ComponentAnatomy showLegend>
```
heswell, line 79: Not sure what anybody else thinks, but I've come to see the legend as superfluous - takes up space and adds noise without being very useful. I'd be inclined to drop it everywhere

packages/core/stories/button.doc.stories.mdx
joshwooding: To avoid stories showing up in the sidebar.

PR #216, opened 2022-06-22
Standardize backwards compat classes and QA

libertymayc:
Includes taking chromatic screenshot of backwards compat story if applicable for component

**Standardisation**:
- AllExamplesGrid: has new Odyssey styling (screenshotted)
- BackwardsCompatGrid: overrides AllExamplesGrid with backwardsCompat className (screenshotted)
- TK1 VR: Uses BackwardsCompatGrid
- Class name "backwardsCompat" should be applied to the root of the component along with it's base name. Then, CSS should be specific selecting that component i.e. .backwardsCompat.uitkCheckbox (no space). 

_Note 1_: it may be necessary to have the class applied in two places depending on how component can be used, e.g. .backwardsCompat.uitkRadioButton and .backwardsCompat.uitkRadioButtonGroup

_Note 2_: if it is a shared component, e.g. ControlLabel, it is best to use .backwardsCompat .uitkControlLabel (with space) - but this will reflect across any components using ControlLabel, which may be an undesired effect

### DIFF
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -22,13 +22,9 @@
 .uitkCard-content {
   font-size: var(--uitkCard-content-fontSize, var(--uitk-text-fontSize));
   letter-spacing: var(--uitkCard-content-letterSpacing, normal);
-  line-height: var(--uitkCard-content-lineHeight, var(--uitk-text-lineHeight));
+  line-height: var(--uitkCard-content-lineHeight, var(--uitk-typography-lineHeight));
   min-height: var(--uitkCard-content-minHeight, var(--uitk-text-minHeight));
   padding: var(--uitkCard-padding, calc(var(--uitk-size-unit) * 3));
-}
-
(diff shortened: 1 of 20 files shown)


packages/core/src/card/Card.css
```diff
@@ -22,15 +22,11 @@
 .uitkCard-content {
   font-size: var(--uitkCard-content-fontSize, var(--uitk-text-fontSize));
   letter-spacing: var(--uitkCard-content-letterSpacing, normal);
-  line-height: var(--uitkCard-content-lineHeight, var(--uitk-text-lineHeight));
+  line-height: var(--uitkCard-content-lineHeight, var(--uitk-typography-lineHeight));
```
alycrys, line 25: should we use `--uitk-typography` or `--uitk-text` within components?

alycrys (dismissed):

origami-z (approved):

origami-z (approved):

PR #219, opened 2022-06-22
Tooltip characteristics, docs, custom styling

libertymayc:


packages/core/stories/tooltip.stories.tsx
```diff
@@ -73,20 +98,67 @@ ScrollTooltip.args = {
   placement: "top",
 };
 
-export const ErrorTooltip: ComponentStory<typeof Tooltip> = Template.bind({});
```
joshwooding, line 76: What's the benefit on not using templates?

libertymayc: @joshwooding if you try putting all 3 tooltip stories into the docs page it crashes, this was to try and fix that (but it doesn't) - i believe its something to do with the tooltip hook, would you be able to take a look?

heswell (commented):
MInor thing - in the custom example with icon and text, the text is hard against the icon, could use a little space

packages/core/src/tooltip/Tooltip.css
```diff
@@ -1,39 +1,41 @@
-.uitk-density-touch .uitkTooltip {
+.uitk-density-touch {
   --icon-top: 6px;
```
alycrys, line 2: should this be '--tooltip-icon-top'?

packages/core/src/tooltip/Tooltip.css
```diff
@@ -1,39 +1,41 @@
...
-  padding: var(--tooltip-padding);
+  border-color: var(--uitkTooltip-borderColor, var(--tooltip-status-borderColor));
+  border-width: var(--uitkTooltip-borderWidth, var(--uitk-container-borderWidth));
+  border-style: var(--uitkTooltip-borderStyle, var(--uitk-container-borderStyle));
+  box-shadow: var(--uitkTooltip-shadow, var(--uitk-overlayable-shadow-popout));
+  color: var(--uitkTooltip-text-color, var(--uitk-text-primary-foreground));
+  font-size: var(--tooltip-fontSize);
+  line-height: var(--tooltip-lineHeight);
+  max-width: var(--uitkTooltip-maxWidth, 230px);
+  min-height: calc(var(--tooltip-fontSize) * var(--tooltip-lineHeight));
```
alycrys, line 35: not sure if this formula is correct. what if they set '--uitkTooltip-lineHeight' in px?

tooling/css-variable-docgen-components/src/CharacteristicUsage.tsx
```diff
@@ -53,7 +53,7 @@ export function CharacteristicUsage(
           {Object.entries(characteristicTokenMap)
             .sort((a, b) => a[0].localeCompare(b[0]))
             .map(([name, tokens]) => (
-              <CharacteristicUsageRow name={name} tokens={tokens} />
+              <CharacteristicUsageRow key={name} name={name} tokens={tokens} />
```
alycrys, line 56: 👍

tooling/css-variable-docgen-components/src/CharacteristicUsageRow.tsx
```diff
@@ -21,7 +21,7 @@ export const CharacteristicUsageRow: FC<CharacteristicUsageRowProps> = (
       <td>
         {tokens &&
           tokens.map((token) => (
-            <Description>
+            <Description key={token}>
```
alycrys, line 24: 👍

packages/core/src/tooltip/Tooltip.tsx
```diff
@@ -111,7 +111,7 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
             ) : (
               <>
                 {getIcon({})}
-                <span className={withBaseName("body")}>{title}</span>
+                <span className={withBaseName("title")}>{title}</span>
```
joshwooding, line 114: Is there a reason for this name change other than preference? I think we used to call it a tooltip body but I don't have a preference in the name.

packages/core/src/tooltip/Tooltip.css
```diff
@@ -1,39 +1,40 @@
...
 }
 
-.uitk-density-high .uitkTooltip {
-  --icon-top: 1px;
+.uitk-density-high {
+  --tooltip-icon-top: 1px;
 }
 
+/* Styles applied to root component */
 .uitkTooltip {
```
joshwooding, line 18: Not a change needed in this PR but we should standardise placement of css variables.

Should it go in a separate class, top of the old class, should we use the css var api inline vs creating a private token etc.

alycrys (approved):

alycrys (approved):

PR #221, opened 2022-06-23
Move layouts to core

Fercas123:
BorderItem, BorderLayout, ParentChildItem, ParentChildLayout and SplitLayout from lab to core


packages/core/src/layout/BorderItem/BorderItem.tsx
```diff
@@ -1,8 +1,6 @@
 import { forwardRef, HTMLAttributes, ComponentPropsWithoutRef } from "react";
 import cx from "classnames";
-import { GridItem } from "@jpmorganchase/uitk-core";
-
-import { makePrefixer } from "@jpmorganchase/uitk-core";
+import { GridItem, makePrefixer } from "@jpmorganchase/uitk-core";
```
JoanaMMoreira, line 3: I think that as these are all now in the same package they should use a relative import instead e.g. "import { GridItem, makePrefixer } from "../GridItem"

origami-z (approved):

PR #224, opened 2022-06-23
fix flex separators

Fercas123:
Fix for flex separators: 

Components built on FlexLayout have the ability yo add separators between items, that can be center, left or right align.
The separators by default (with prop: true) will render in the center of FlexLayout children.

The current issue is: 
- Separators are displaying with the width of the child element, rather than of the layout. this makes them inconsistent in size if siblings are of a different width 
- Separators are not starting from left:0 position when items are arranged in columns, this needs to be explicit in CSS 

To ensure the separators are all the same width, we are adding a wrapper to each FlexLayout children, with `display: contents`, that will only handle the separator ensuring consistency. This PR also fixes the left position of the separator.


packages/core/src/layout/FlexLayout/FlexLayout.css
```diff
@@ -23,40 +23,50 @@
   align-items: var(--flex-layout-align);
 }
 
-.uitkFlexLayout-separator {
-  --flex-layout-gap: calc(var(--uitk-size-unit) * var(--flex-layout-gap-multiplier) * 2 + var(--flex-layout-separator));
+.uitkFlexLayout-separator-wrapper {
+  display: contents;
```
el-dav, line 27: `display: contents` could potentially be an accessibility issue on Safari (see [caniuse](https://caniuse.com/?search=display%20contents)). Having said that I think it's used on the uitk-theme element which could be a bigger issue @heswell what do you think?

Fercas123: from what i understand, it is an accessibility issue that would arise from adding accessibility props directly into the wrapper, in this case the component is adding the wrapper after mapping children, so if you wanted to add for example a role to your children, those would still apply since it is only the outside wrapper that would have `display: contents` (but I'll add a test for Safari to make sure I'm understanding this correctly)

el-dav: Oh cool I had thought it applied to children of the `display: contents` element but I could well be wrong, in which case it should be fine to use :)

heswell: I've come to think that the uitk-theme element was a mistake and that the default behaviour should be to add classes to nested child, with the custom element as the configurable option . And yes, it does use display: content , which kind of does the trick but also has the potential to break users css rules if they are not expecting this element to be present.

heswell (commented):
As this component is now in Core, we could  do with more of a description as to what exactly the problem is that this fixes.

packages/core/src/layout/FlexLayout/FlexLayout.tsx
```diff
@@ -93,7 +93,11 @@ export const FlexLayout = forwardRef<HTMLDivElement, FlexLayoutProps>(
         style={flexLayoutStyles}
         {...rest}
       >
-        {children}
+        {separators
+          ? Children.map(children, (child) => (
```
heswell, line 97: does feel a shame to have to introduce this extra layer just when we have separators. Don't have a better suggestion though

Fercas123: I guess the other option would be creating an explicit separator component, which will probably the what's needed once we start thinking about the splitter, for flexlayout implementation wise we will still need to go trough each child and append it, but that would get rid of the wrapper

PR #227, opened 2022-06-27
Add emphasis prop to Panel

libertymayc:


packages/core/src/panel/Panel.tsx
```diff
@@ -16,14 +16,28 @@ import "./Panel.css";
...
-  }
-);
+export const Panel = forwardRef<HTMLDivElement, PanelProps>(function Panel(
+  { className, children, emphasis = "medium", ...restProps },
+  ref
+) {
+  return (
+    <div
+      className={cx(withBaseName(), className, {
+        uitkEmphasisHigh: emphasis === "high",
```
alycrys, line 35: I think i'd prefer to not use a generic class for emphasis. 
Imagine a situation where we have a component inside Panel which has a different emphasis, could there be any style overrides which could cause issues?

libertymayc: emphasis is always attached to the component className specifically (e.g. .uitkEmphasisHigh.uitkPanel in this case) so nesting issues shouldn't occur

packages/core/stories/panel.doc.stories.mdx
```diff
@@ -51,8 +51,8 @@ import { Panel } from "@jpmorganchase/uitk-core";
   <ComponentAnatomy>
     <Panel aria-label="Default Panel">Default</Panel>
   </ComponentAnatomy>
-  <ComponentAnatomy>
-    <Panel aria-label="Panel with high emphasis" className="uitkEmphasisHigh">
+  <ComponentAnatomy showLegend>
```
joshwooding, line 54:
```suggestion
  <ComponentAnatomy>
```
showLegend doesn't do anything

joshwooding (approved):

PR #234, opened 2022-06-29
Move Pill to Core

libertymayc:
Depends on #237, checkbox icon is needed


packages/core/src/pill/internal/DeleteButton.tsx
```diff
@@ -1,8 +1,8 @@
 import { MouseEvent } from "react";
 import cn from "classnames";
-import { makePrefixer, Button, ButtonProps } from "@jpmorganchase/uitk-core";
+import { Button, ButtonProps } from "@jpmorganchase/uitk-core";
```
alycrys, line 3: should these be imported from '../button'?

packages/core/src/pill/internal/PillBase.tsx
```diff
@@ -22,6 +13,17 @@ import {
...
+  Button,
+  ButtonProps,
+  useTooltip,
+  useTooltipContext,
+} from "@jpmorganchase/uitk-core";
+import {
+  makePrefixer,
+  useForkRef,
+  useIsomorphicLayoutEffect,
+} from "../../utils";
```
alycrys, line 26: why do we separate utils from the rest of the core imports? should we not import from core?

heswell (approved):

origami-z (approved):

PR #242, opened 2022-07-01
Ag grid theme

ararus:


packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,214 @@
+@import "./uitk-ag-theme-mixin";
+
+.uitk-light .ag-theme-uitk-high,
+.uitk-light .ag-theme-uitk-medium,
+.uitk-light .ag-theme-uitk-low,
+.uitk-light .ag-theme-uitk-touch {
+  --uitk-ag-foreground-color: #2a2c2f;
+  --uitk-ag-disabled-foreground-color: #161616;
```
libertymayc, line 8: we tend to leave out -color on the end of background and foreground values, and also any states should go on the end of the token (-disabled suffix) - can see figma or storybook docs for more info on ordering

also attributes tend to go in camel case, so e.g. --uitk-ag-row-borderColor

ararus: Thank you for reviewing this PR. I've renamed vars, please have a look.

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
+@import "./uitk-ag-theme-mixin";
+
+.uitk-light .ag-theme-uitk-high,
+.uitk-light .ag-theme-uitk-medium,
+.uitk-light .ag-theme-uitk-low,
+.uitk-light .ag-theme-uitk-touch {
+  --uitk-ag-foreground: #2a2c2f;
```
libertymayc, line 7: We start foundations and characteristics with --uitk-<characteristic name>- .. it may be better to have these as --uitkAgGrid like how other components tokens start

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
...
+  --uitk-ag-foreground-disabled: #161616;
+  --uitk-ag-header-foreground: #61656e;
+  --uitk-ag-background: white;
+  --uitk-ag-row-background-selected: #b7def6;
+  --uitk-ag-row-color-hover: white;
+  --uitk-ag-row-borderColor: #e2e2e2;
+  --uitk-ag-header-column-separator-color: #c5c9d0;
+  --uitk-ag-cell-border-color-editable: #84878e;
```
libertymayc, line 15: -borderColor

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
+@import "./uitk-ag-theme-mixin";
+
+.uitk-light .ag-theme-uitk-high,
+.uitk-light .ag-theme-uitk-medium,
+.uitk-light .ag-theme-uitk-low,
+.uitk-light .ag-theme-uitk-touch {
```
libertymayc, line 6: do we need the densitys here here or would .uitk-light be enough?

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
...
+  --uitk-ag-menu-color: #61656e;
+  --uitk-ag-filter-border-color: #eaedef;
+  --uitk-ag-filter-apply-button-color: #242526;
+  --uitk-ag-list-item-color-hover: #b7def6;
+
+  --uitk-ag-column-drop-cell-horizontal-background: #eaedef;
+  --uitk-ag-pagination-button-color: #4c505b;
+  --uitk-ag-pagination-button-color-disabled: rgba(22, 22, 22, 0.4);
+
+  --uitk-ag-placeholder-color: #61656e;
```
libertymayc, line 45: I assume these hex values will eventually move to the color foundation, is that true? it may be worth starting to split out reused colors anyway into local tokens, e.g. #f2f4f6 is used a few times, perhaps call that something with semantic meaning so it's easier to then move later

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
...
+
+  --uitk-ag-button-color: white;
+  --uitk-ag-button-background-hover: #d9dde3;
+
+  --uitk-ag-toggle-background-checked: #2670a9;
+  --uitk-ag-header-icon-color: white;
+  --uitk-ag-menu-header-icon-color: white;
+  --uitk-ag-icon-color: white;
+  --uitk-ag-menu-option-color-hover: #2f3136;
+  --uitk-ad-side-buttons-background: #2f3136;
```
libertymayc, line 79: -ad here should be -ag? also may be nicer to have this token as -sideButtons

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,229 @@
...
+  --uitk-ag-toggle-button-background: #84878e;
+
+  --uitk-ag-menu-color: white;
+  --uitk-ag-filter-border-color: #44484f;
```
libertymayc, line 83: -borderColor (maybe worth just doing a ctrl F on -border-color and changing them all to -borderColor so none are missed)

libertymayc (changes requested):

packages/ag-grid-theme/css/uitk-ag-theme.scss
ararus: We are not going to use foundations or characteristics for this theme. All colours are hardcoded to prevent any changes in the theme when the characteristics or foundations are updated.

packages/ag-grid-theme/css/uitk-ag-theme.scss
ararus: Good spot. Thanks. Renamed

packages/ag-grid-theme/css/uitk-ag-theme.scss
ararus: Done

packages/ag-grid-theme/css/uitk-ag-theme.scss
```diff
@@ -0,0 +1,313 @@
+@import "./uitk-ag-theme-mixin";
+
+.ag-theme-uitk-high-light,
+.ag-theme-uitk-medium-light,
+.ag-theme-uitk-low-light,
+.ag-theme-uitk-touch-light {
```
libertymayc, line 6: I think this can just be .ag-theme .uitk-light { ... } and repeat for dark? 
I also realised I probably commented to make the tokens --uitkAgGrid, but perhaps --agGrid is better, because we usually give --uitkAgGrid prefix for the CSS API for consumers to override

packages/ag-grid-theme/css/uitk-ag-theme.scss
ararus: Done

packages/ag-grid-theme/css/uitk-ag-theme.scss
ararus: This was the original plan. Ag-grid renders drag-and-drop ghosts outside the grid root (and outside our theme provider actually). It applies the same style to them as was applied to the grid itself. ".uitk-light .uitkAgGrid-..." won't work because there is no .uitk-light on any tag containing dnd ghost.

origami-z:
Can we have an "detail" part example for the master detail example? - https://88823fc9.uitk.pages.dev/?path=/story/ag-grid-theme--master-detail

packages/ag-grid-theme/css/uitk-ag-theme.scss
libertymayc: darn. although, if all values are same for each density, do we really need 8 values, or could we just have 2, light and dark: ag-theme-uitk-light/dark ?

origami-z:
Looking at latest [master detail](https://fc46f624.uitk.pages.dev/?path=/story/ag-grid-theme--master-detail) example, would be good to raise the question of "should things inherited from old ag grid be 'fixed' or kept the original"

packages/ag-grid-theme/css/uitk-ag-theme.scss
libertymayc: cc @ararus

libertymayc (approved):
all styling/CSS looks good to me

origami-z (commented):
A few things to consider in this PR or another one after
- changset
- Fix build by recent List change
- Include build script in GitHub action steps, and potentially some linting steps
- Some simple Chromatic snapshots for future update regression check
- Is ag grid tool panel part of design to align? e.g. background color, active bar, pivot switch

packages/ag-grid-theme/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@jpmorganchase/uitk-ag-grid-theme",
+  "version": "0.1.0",
+  "license": "Apache-2.0",
+  "style": "index.css",
+  "files": [
+    "/index.css"
+  ],
+  "scripts": {
+    "build": "yarn node ./scripts/build.js",
```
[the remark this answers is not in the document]
ararus: merged with the other script

packages/ag-grid-theme/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "build-watch": "yarn node ./scripts/build-watch.js"
```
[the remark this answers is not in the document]
ararus: Updated the script. Now it does the initial build and then watches for changes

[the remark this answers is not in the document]
origami-z: I think still needs an update to make the output `index.css`, or update `style` and `files` block to match output css name?

joshwooding:
The build is failing due to the import in AddRemoveRows  (uitk-ag-theme.css) which doesn't exist in source code.

origami-z (approved):

PR #244, opened 2022-07-04
Wrap by default on FlexLayout and tweak layout storybook examples

JoanaMMoreira:
These changes have been reviewed by Stella. They mostly include small tweaks to the storybook examples but the most important change is allowing the `FlexLayout` component to wrap by default

### DIFF
--- a/packages/core/src/layout/FlexLayout/FlexLayout.tsx
+++ b/packages/core/src/layout/FlexLayout/FlexLayout.tsx
@@ -21,11 +21,11 @@ export type FlexContentAlignment = typeof FLEX_CONTENT_ALIGNMENT_BASE[number];
 
 export interface FlexLayoutProps extends HTMLAttributes<HTMLDivElement> {
   /**
-   * Defines the default behavior for how flex items are laid out along the cross axis on the current line.
+   * Defines the default behavior for how flex items are laid out along the cross axis on the current line, default is "stretch".
    */
   align?: FlexAlignment | "stretch" | "baseline";
   /**
-   * Establishes the main-axis, defining the direction children are placed.
+   * Establishes the main-axis, defining the direction children are placed. Default is "row".
    */
   direction?: ResponsiveProp<LayoutDirection>;
   /**
@@ -33,11 +33,11 @@ export interface FlexLayoutProps extends HTMLAttributes<HTMLDivElement> {
    */
   gap?: ResponsiveProp<number>;
   /**
-   * Defines the alignment along the main axis.
+   * Defines the alignment along the main axis, default is "start".
    */
   justify?: FlexContentAlignment;
   /**
-   * Adds a separator between elements.
+   * Adds a separator between elements, default is false.
    */
   separators?: LayoutSeparator | true;
   /**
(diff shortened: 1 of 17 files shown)


packages/core/src/layout/FlexLayout/FlexLayout.tsx
```diff
@@ -57,7 +57,7 @@ export const FlexLayout = forwardRef<HTMLDivElement, FlexLayoutProps>(
       justify,
       separators,
       style,
-      wrap,
+      wrap = true,
```
joshwooding, line 60: Should this be kept as wrap or changed to disableWrap?

packages/core/src/layout/FlowLayout/FlowLayout.tsx
```diff
@@ -1,22 +1,23 @@
-import { forwardRef, HTMLAttributes } from "react";
+import { forwardRef, HTMLAttributes, ComponentPropsWithoutRef } from "react";
 import { FlexLayout } from "../FlexLayout";
-import { FlexLayoutProps } from "../types";
+
+type FlexLayoutProps = ComponentPropsWithoutRef<typeof FlexLayout>;
```
joshwooding, line 4: Can we use FlexLayoutProps from FlexLayout?

packages/core/src/layout/StackLayout/StackLayout.tsx
```diff
@@ -1,18 +1,19 @@
-import { forwardRef, HTMLAttributes } from "react";
+import { forwardRef, HTMLAttributes, ComponentPropsWithoutRef } from "react";
 import { FlexLayout } from "../FlexLayout";
-import { FlexLayoutProps } from "../types";
+
+type FlexLayoutProps = ComponentPropsWithoutRef<typeof FlexLayout>;
```
joshwooding, line 4: Same here?

joshwooding (changes requested):
See above + the conflicts please :)

packages/core/src/layout/FlexLayout/FlexLayout.tsx
JoanaMMoreira: I think `disableWrap` makes sense, I've amended the prop

joshwooding (approved):

joshwooding (approved):

origami-z:
Why was the `wrap` being set by default? This shouldn't be an accessibility decision? We changed to `nowrap` previously because that is the default value for `flex-wrap`. This now causes mind bending debug issue when using flex layout...?

JoanaMMoreira:
> Why was the `wrap` being set by default? This shouldn't be an accessibility decision? We changed to `nowrap` previously because that is the default value for `flex-wrap`. This now causes mind bending debug issue when using flex layout...?

Stella's point was around some of the examples having a double scroll bar when zoomed in so her suggestion was that the component should always wrap by default. If we think this causes debugging issues maybe we could revert the change and add an accessibility section to the docs that explains this?

origami-z:
We can make the examples wrap, but i don't think think changing default value of any CSS is a good idea. Personally, i'm expecting flex and other layouts are some drop in replacement for not writing CSS line by line myself, if we enforce different default value than what CSS gives, we're driving into corners.

JoanaMMoreira:
> We can make the examples wrap, but i don't think think changing default value of any CSS is a good idea. Personally, i'm expecting flex and other layouts are some drop in replacement for not writing CSS line by line myself, if we enforce different default value than what CSS gives, we're driving into corners.

Makes sense although I wonder if changing the examples only (particularly the "default" one) would be a bit misleading as well

PR #245, opened 2022-07-04
Fix bug in component anatomy

JoanaMMoreira:
Fix issue where the component anatomy sometimes would not display in the docs due to the root element not being rendered straight away


.changeset/violet-poets-crash.md
```diff
@@ -0,0 +1,6 @@
+---
+"docs": minor
```
origami-z, line 2: We don't want changeset 'docs'

docs/components/ComponentAnatomy.jsx
```diff
@@ -54,10 +76,16 @@ export const ComponentAnatomy = ({
     setTree(t);
   };
 
+  useEffect(() => {
+    waitForElement(".RenderVisualiser-component > *").then((element) => {
```
origami-z, line 80: `waitForElement` is querying `document`, it will get the wrong item? e.g. Looks like Button anatomy is showing the same information.

JoanaMMoreira: Good spot! While I was fixing this I realised we don't even need the `waitForElement` function so I've removed it

PR #257, opened 2022-07-13
Add grid and parent child examples

JoanaMMoreira:


.changeset/brave-grapes-hunt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-core": minor
+---
+
+Add composite storybook examples for parent/child and grid layout components
```
joshwooding, line 5: Since, it's not a change to shippable code we don't need a changeset. Although we should think of a way of allowing interesting changes that are around documentation to be shared.

packages/core/stories/layout/flex-layout.stories.tsx
```diff
@@ -50,7 +57,12 @@ export default {
       control: "boolean",
     },
   },
-  excludeStories: ["ContactDetailsExample", "FlexLayoutNestedExample"],
+  excludeStories: [
```
joshwooding, line 60: Maybe not for this PR but ContactDetailsExample isn't used anywhere in this file so not really sure if it belongs here?

.changeset/brave-grapes-hunt.md
JoanaMMoreira: Sure, I'll remove it

joshwooding (approved):

PR #260, opened 2022-07-18
Fix QA stories

alycrys:


packages/lab/src/typography/Text.css
```diff
@@ -26,17 +22,12 @@
...
-  opacity: 0;
-}
-
 /* Heading 1 */
 
 h1.uitkText,
 .uitkText-h1.uitkText {
   font-size: var(--uitk-text-h1-fontSize);
-  line-height: var(--uitk-text-h1-lineHeight);
+  line-height: var(--uitk-text-lineHeight-backwardsCompat, var(--uitk-text-h1-lineHeight));
```
libertymayc, line 30: we should have --uitkText-lineHeight as the first var and then set that in a .backwardsCompat.uitkText class

packages/lab/src/typography/Text.css
```diff
@@ -103,35 +103,42 @@ h4.uitkText small,
...
+  --uitkText-line-height: var(--uitk-text-figure3-lineHeight);
   font-size: var(--uitk-text-figure3-fontSize);
-  line-height: var(--uitk-text-lineHeight-backwardsCompat, var(--uitk-text-figure3-lineHeight));
   font-weight: var(--uitk-text-figure3-fontWeight);
 }
+
+.uitkText {
+  line-height: var(--uitkText-line-height);
+}
+.backwardsCompat .uitkText {
```
libertymayc, line 142: should have these together .backwardsCompat.uitkText

packages/lab/src/typography/Text.css
```diff
@@ -103,35 +103,42 @@ h4.uitkText small,
...
 /* Figure 3 */
 .uitkText-figure3.uitkText {
+  --uitkText-line-height: var(--uitk-text-figure3-lineHeight);
   font-size: var(--uitk-text-figure3-fontSize);
-  line-height: var(--uitk-text-lineHeight-backwardsCompat, var(--uitk-text-figure3-lineHeight));
   font-weight: var(--uitk-text-figure3-fontWeight);
 }
+
+.uitkText {
+  line-height: var(--uitkText-line-height);
```
libertymayc, line 140: to follow our conventions, we should have line-height: var(--uitkText-lineHeight, var(--text-lineHeight)) and set the local variable (--text-lineHeight) throughout this file, so consumers can override with the given API (---uitkText-lineHeight)

libertymayc: we should similarly do this for font weight and font size if we are going to do it for line height

packages/lab/src/contact-details/ContactDetails.css
```diff
@@ -43,6 +43,7 @@
 }
 
 .uitkContactDetails {
+  --contact-details-color: var(--uitk-text-primary-foreground);
```
libertymayc, line 46: not necessary for this pr, but in future these need changing to --contactDetails-

packages/lab/src/typography/Text.css
alycrys: .uitkText may be included inside main components, so we need to make sure the rule still applies

packages/lab/src/typography/Text.css
alycrys: is this something the consumers should override?

packages/lab/src/typography/Text.css
libertymayc: in that case it should be the component thats unique text is backwards compatible (for example - not a true case that ive checked but there will be similar scenarios - h1 text is the same everywhere as we have now defined, but maybe in a dialog title someone applied backwards compatibility, and that's where the inner h1 text needs adjusting, not in the text component itself, so in .backwardsCompat.uitkDialog-title the --uitkText api is used)

packages/lab/src/typography/Text.css
libertymayc: internally probably not, but i would say yes for general purpose..
@heswell what are your thoughts here?

packages/lab/src/typography/Text.css
alycrys: Normally we would apply the backwardsCompat class on the Dialog, not on DialogTitle. Even so, the h1 would be a child of DialogTitle. There is no .backwardsCompat.uitkText as there is no uitkText in TK1.
I understand your logic, but I think this case is a little different from the pattern we have on other components.
Maybe we should have a call and discuss this in more details

packages/lab/src/contact-details/ContactDetails.css
alycrys: done

packages/lab/src/typography/Text.css
libertymayc: that makes sense.. (in that case, do we call it 'backwards compatible'?) 
i agree!

alycrys: I've added `.backwardsCompat .uitkText` to make sure all the components which now include .uitkText will be backwards compatible with the TK1 version. I'm open to other suggestions :)

packages/lab/src/typography/Text.css
```diff
@@ -9,13 +9,9 @@
...
 .uitkText {
+  color: var(--uitk-text-primary-foreground);
   font-size: var(--uitk-text-fontSize);
-  line-height: var(--uitk-text-lineHeight-backwardsCompat, var(--uitk-text-lineHeight));
-}
-
-.uitkText-overflow {
-  height: 100%;
-  overflow: auto;
+  --uitkText-line-height: var(--uitk-text-lineHeight);
```
heswell, line 14: The CSS variable naming doesn't align with our conventions.
 the variable `--uitkText-line-height` is an `API` variable - denoted by the mixed case --uitkComponentName prefix.
These should only ever be consumed within the component they belong to, not assigned values. They are for consumers or consuming components to supply overrides.

alycrys: but `--uitk-text-lineHeight` comes from characteristics. What else can I use in this situation?
@heswell @libertymayc

packages/lab/src/typography/Text.css
heswell: see comment above

heswell (changes requested):
QA stories themselves looking good. 
CSS variable naming conventions need tweaking.

Apologies for the disruption that my changes to QAContainer introduced. Good to see you've added some of the missing functionality into the Odyssey version. Hopefully now they're all converted, you'll find that the reduced boilerplate we now have is better.

alycrys:
> QA stories themselves looking good. CSS variable naming conventions need tweaking.
> 
> Apologies for the disruption that my changes to QAContainer introduced. Good to see you've added some of the missing functionality into the Odyssey version. Hopefully now they're all converted, you'll find that the reduced boilerplate we now have is better.

Indeed, the new version is much simpler. Took a little while to fix, but I think we're in a good place now having all the versions and densities visible. Thanks

packages/lab/src/contact-details/ContactDetails.css
```diff
@@ -75,25 +76,26 @@
...
 }
 
 .uitkContactDetails-mini,
 .uitkContactDetails-mini-stacked {
-  --contact-details-primary-font-size: var(--contact-details-compact-primary-font-size);
-  --contact-details-secondary-font-size: var(--contact-details-compact-contact-details-secondary-font-size);
-  --contact-details-secondary-font-weight: var(--uitk-text-fontWeight);
+  --contactDetails-primary-font-size: var(--contactDetails-compact-primary-font-size);
+  --contactDetails-secondary-font-size: var(--contactDetails-compact-secondary-font-size);
+  --contactDetails-secondary-font-weight: var(--uitk-text-fontWeight);
```
libertymayc, line 94: Could we change -font-size to -fontSize and same for weight in these tokens

packages/lab/src/contact-details/ContactDetails.css
```diff
@@ -270,22 +268,56 @@
...
 .uitkContactFavoriteToggle-svg {
-  height: var(--contact-details-icon-size);
-  width: var(--contact-details-icon-size);
+  height: var(--contactDetails-icon-size);
+  width: var(--contactDetails-icon-size);
+}
+
+/* backwardsCompat */
+
+.uitk-density-touch .backwardsCompat.uitkContactDetails-default .uitkContactPrimaryInfo {
```
libertymayc, line 293: usually would have backwardsCompat just on the root class name, but this is OK as long as backwards compat class goes on the same div as all of these variants and they're not nested

alycrys: uitkContactDetails-default comes on the same level as uitkContactDetails and will never be present without the main uitkContactDetails class. Was just avoiding a very long css selector

packages/lab/src/typography/Text.css
```diff
@@ -107,35 +104,42 @@ h4.uitkText small,
...
 /* Figure 3 */
 .uitkText-figure3.uitkText {
+  --text-lineHeight: var(--uitk-text-figure3-lineHeight);
   font-size: var(--uitk-text-figure3-fontSize);
-  line-height: var(--uitk-text-lineHeight-backwardsCompat, var(--uitk-text-figure3-lineHeight));
   font-weight: var(--uitk-text-figure3-fontWeight);
 }
+
+.uitkText {
+  line-height: var(--text-lineHeight);
```
libertymayc, line 141: shall we give --uitkText-lineHeight here for overrides?

packages/lab/src/typography/Text.css
libertymayc: sounds good, let's make the API token --uitkText-lineHeight, without -backwardsCompat, and then people can set that token within backwardsCompat class - this is the convention we have taken in other places to avoid bloating css with too many different variables

packages/lab/src/breadcrumbs/Breadcrumb.css
```diff
@@ -60,13 +60,11 @@
 
 .backwardsCompat.uitkBreadcrumbs .uitkBreadcrumb {
   padding: 0px;
-  gap: 0;
+  /* gap: 0; */
```
libertymayc, line 63: needed?

packages/lab/src/typography/Text.css
alycrys: I'm not sure about this one. Do we want to let them override line-height? And if so, should we let them override for all text components on all densities?

packages/lab/src/contact-details/ContactDetails.css
```diff
@@ -144,8 +146,8 @@
 
 .uitkContactDetails-noAvatarIndicator {
   grid-area: noAvatar;
-  width: var(--contact-details-no-avatar-indicator-width);
-  background-color: var(--contact-details-no-avatar-color);
+  width: var(--contactDetails-no-avatar-indicator-width);
+  background-color: var(--contactDetails-no-avatar-color);
```
libertymayc, line 150: possibly change to -noAvatar as it is sort of a variant?

packages/lab/src/typography/Text.css
libertymayc: if we define just the token here, then they can set that within the given classnames themselves, we don't need to provide it everywhere as that will bloat it out. but whether we let them do that or not is perhaps a question for design

heswell (dismissed):

heswell (dismissed):

.changeset/chatty-bugs-perform.md
```diff
@@ -0,0 +1,6 @@
+---
+"@jpmorganchase/uitk-lab": patch
+"@jpmorganchase/uitk-theme": patch
+---
+
+Content status add typography
```
joshwooding, line 6: Can we get a better summary of the changes here please

heswell (dismissed):

heswell (approved):

PR #278, opened 2022-07-20
Modify sideEffects to css related files for icons package.json

origami-z:
This hopefully will resolve the issue that Icon CSS is not being loaded while using `react-scripts` / webpack based setup.

        <Button>
           {"No icon here ->"}
          <SearchIcon size={12} />
        </Button>

Above code in fresh CRA / modular set up will result in below render (reproduce [repo](https://github.com/origami-z/uitk-cra-icon-css-loading))

screenshot not icons in button

While debugging, no code in `node_modules/@jpmorganchase/uitk-icons/dist-es/packages/icons/src/icon/Icon.css.js` is executed at runtime while using webpack based tooling (vite is fine). There might be other ways to force code being included, but given `css.js` files are generated by build tool, we don't have much control over that.

Interestingly enough, the same code doesn't have problem on Stackblitz [example](https://stackblitz.com/edit/react-ts-uarhsw?file=App.tsx), which also uses `react-scripts`.

I tested removing/editing `sideEffects` line in the sample reproduce repo's `node_modules` `package.json`, the icon is rendering fine.

### DIFF
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -3,7 +3,6 @@
   "version": "0.2.0",
   "license": "Apache-2.0",
   "main": "src/index.ts",
-  "sideEffects": false,
   "scripts": {
     "build": "node ./scripts/generateIconComponent.js '*.svg' && node ./scripts/generateIconIndex.js"
   },
--- a/packages/icons/package.json
+++ b/packages/icons/package.json
@@ -3,6 +3,10 @@
   "version": "0.2.0",
   "license": "Apache-2.0",
   "main": "src/index.ts",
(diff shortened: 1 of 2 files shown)


packages/icons/package.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.2.0",
   "license": "Apache-2.0",
   "main": "src/index.ts",
-  "sideEffects": false,
```
joshwooding, line 6: Ah I think this is because the Icon component was moved to this package. We can also copy the sideEffects value from core

joshwooding (approved):

PR #284, opened 2022-07-27
fix(aria-announcer): prevent breaking full height layouts

### LINKED ISSUE #277: AriaAnnouncer should not break layout
The aria announcer creates a pixel sized element on the page. This can break user layouts. We should make sure it is not within the document flow.

ZarkoPernar:
Closes #277
AriaAnnouncerProvider needed a small negative margin hack to prevent affecting the document flow. Also, the styles were updated to more closely match the [tailwind visually hidden class](https://tailwindcss.com/docs/screen-readers).

### DIFF
--- a/packages/core/src/__tests__/__e2e__/aira-announcer/AriaAnnouncer.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/aira-announcer/AriaAnnouncer.cy.tsx
@@ -1,19 +1,18 @@
 import { AriaAnnouncerProvider } from "@jpmorganchase/uitk-core";
 
-describe("Given a ToolkitProvider", () => {
-  describe("with no props set", () => {
-    it("should not affect the document flow", () => {
-      cy.mount(
-        <div id="test-1" style={{ height: "100%", width: "100%" }}>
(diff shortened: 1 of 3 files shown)


origami-z:
- Need changeset
- The div is 1x1 on [this story](https://aadef320.uitk.pages.dev/?path=/story/core-ariaannouncerprovider--aria-announce-debounce-and-delay) before & after the -1 margin line is there or not?
- Is it possible to test this?

packages/core/src/__tests__/__e2e__/aira-announcer/AriaAnnouncer.cy.tsx
```diff
@@ -0,0 +1,31 @@
...
+      style.innerHTML = `
+                body, html {
+                    height: 100%;
+                    display: block;
+                    min-height: auto;
+                }
+                [data-cy-root] {
+                    height: 100%;
+                }
+            `;
```
ZarkoPernar, line 24: Do I need to clean this up, or are individual test files run in an isolated environment @joshwooding?

joshwooding (approved):
image

Chromatic detected the change :)

PR #292, opened 2022-08-01
Update layer layout and carousel storybook examples, remove border from layer and add animations to layer layout

JoanaMMoreira:
- Rename layer examples
- Remove border from layer component
- Add fade animations to scrim
- Update carousel storybook example
- Update existing layer storybook examples
- Add new layer real life examples for centre, left, top, right and bottom positions


.changeset/plenty-cows-train.md
```diff
@@ -0,0 +1,10 @@
+---
+"@jpmorganchase/uitk-lab": minor
+---
+
+Rename layer examples
```
ZarkoPernar, line 5: Do you think non-user facing changes should be added to the change log?

JoanaMMoreira: That's a good point, I don't think they should so I've updated the change log

ZarkoPernar: Although, I've seen some libraries add documentation edits in to the change log

JoanaMMoreira: @origami-z @joshwooding what do you reckon?

joshwooding: Hmm, I think for now leaving them out is fine. Potentially in the future we can add them.

.changeset/plenty-cows-train.md
```diff
@@ -0,0 +1,6 @@
+---
+"@jpmorganchase/uitk-lab": minor
+---
+
+Remove border from layer component
+Add fade animations to scrim
```
joshwooding, line 6: Scrim doesn't actually have animations, the LayerLayout does.

joshwooding (changes requested):
See comment

joshwooding (approved):

PR #298, opened 2022-08-08
Enable core src code eslint check

origami-z:
One small step towards enabling `modular lint`, to run `eslint` against source code of `uitk-core`

- ignore `/dist` folder from eslint, otherwise it will complain duplicated package names
- modify `parserOptions.project` to take `tsconfig` from each subfolder, so we won't get the error when looking at files excluded from `<root>/tsconfig.json` like tests
- added `eslint` to dependency array so `yarn eslint` can run
- added `lint:core` to check a small amount of code in core, and enable in GitHub action
- turn off `@typescript-eslint/explicit-module-boundary-types` rule as components can generate types ok

### What i'm not doing

- Lint tests because there are other config changes needed
- Fix the actual bug / typing issue

### Other considerations

modular [includes](https://github.com/jpmorganchase/modular/blob/051ecf7d257f883268abe73fd082d25615888906/packages/eslint-config-modular-app/index.js) all ts files within `__tests__` to use `jest` eslint configs, which will be a pain to manually switch off via overrides. We can consider moving `__e2e` folder out to apply our own rules.


alycrys (dismissed):

ZarkoPernar:
>modular [includes](https://github.com/jpmorganchase/modular/blob/051ecf7d257f883268abe73fd082d25615888906/packages/eslint-config-modular-app/index.js) all ts files within __tests__ to use jest eslint configs, which will be a pain to manually switch off via overrides. We can consider moving __e2e folder out to apply our own rules.

I assume the jest eslint config is being applied to cypress tests, and was not meant for it, therefore causing issues?

ZarkoPernar (dismissed):

alycrys (approved):

origami-z:
> > modular [includes](https://github.com/jpmorganchase/modular/blob/051ecf7d257f883268abe73fd082d25615888906/packages/eslint-config-modular-app/index.js) all ts files within **tests** to use jest eslint configs, which will be a pain to manually switch off via overrides. We can consider moving __e2e folder out to apply our own rules.
> 
> I assume the jest eslint config is being applied to cypress tests, and was not meant for it, therefore causing issues?

Yes, modular doesn't include use cases where non-jest tests will be in that folder. We're too advanced 😄

alycrys (approved):

PR #312, opened 2022-08-11
Ensure Panel custom styling is only in Panel example and remove undefined css var from StepperInput

libertymayc:
As we use Panel to provide light/dark backgrounds for a lot of our stories, the global custom styling made other examples look bad:

Screenshot 2022-08-11 at 13 06 25

Screenshot 2022-08-11 at 13 06 42

Using a class name for the Panel story to fix this.

Improves StepperInput custom styling, which was affected by this but also used a confusing example story in comparison to how the others are done.


packages/lab/src/stepper-input/StepperInput.css
```diff
@@ -87,5 +87,5 @@
 
 /* Styles applied to input component */
 .uitkStepperInput-input {
-  max-height: var(--uitkStepperInput-maxHeight, var(--stepperInput-input-maxHeight));
+  max-height: var(--uitkStepperInput-maxHeight, unset);
```
origami-z, line 90: This deserves a changeset?

PR #315, opened 2022-08-11
Bump modular-scripts to 3.2.0

joshwooding:


ZarkoPernar (approved):
This is the alternative to moving everything under packages?

joshwooding:
> This is the alternative to moving everything under packages?

Not an alternative just something we are doing in addition

PR #322, opened 2022-08-15
Add state icon stories, documentation and tests

JoanaMMoreira:


packages/core/stories/state-icon.stories.tsx
```diff
@@ -0,0 +1,42 @@
...
+  argTypes: {
+    size: {
+      options: ICON_NAMED_SIZES,
+      control: { type: "select" },
+      defaultValue: ICON_NAMED_SIZES[0],
+    },
+  },
+} as ComponentMeta<typeof StateIcon>;
+
+const states: ValidationState[] = ["error", "success", "warning", "info"];
```
origami-z, line 21: Feels useful to be exported from the source code

packages/core/stories/state-icon.doc.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+    viewMode: "docs",
+  }}
+/>
+
+# State icon
+
+A state icon can be used on its own or within another component (e.g. `Tooltip`) to help convey a message.
+
+<Canvas>
+  <Story id="core-stateicon--default-state-icon" />
```
origami-z, line 31: Purely from documentation perspective, could look nice if we have bigger size

packages/core/stories/state-icon.doc.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+# API
+
+```
+import { StateIcon } from "@jpmorganchase/uitk-core";
+```
+
+## Props
+
+<ArgsTable of={StateIcon} />
```
origami-z, line 48: `size` has a default value from `Icon`, not sure why it's not shown on the props table default column

packages/core/stories/state-icon.doc.stories.mdx
JoanaMMoreira: Do you mean only on the first icon or all of them?

origami-z: Could be all of the examples. Looks very tiny as a documentation example. @bhoppers2008 to comment :)

Maybe there is ways to set default props in mdx..?

packages/core/stories/state-icon.doc.stories.mdx
JoanaMMoreira: Yeah it's very odd, I'm looking into this

origami-z: I may not be possible BTW, doc gen may not be able to trace through at all. If so, may be set the default value in state icon as well.

JoanaMMoreira: It doesn't seem to be possible unfortunately, I've tried a few different ways but this seems to be a known issue in Storybook. I've exported the default `size` from `Icon` and added it to `StateIcon`

packages/core/stories/state-icon.doc.stories.mdx
JoanaMMoreira: Setting default props in mdx doesn't seem to work but I can add them individually to each example. @bhoppers2008 let me know your thoughts :)

bhoppers2008: We can increase the size for documentation but would be best to align it to the work that paul is doing to remove size from the characteristics. That will have some size options. It's WIP though and Paul is on annual leave this week.

JoanaMMoreira: Makes sense, so should we merge this as is for now and revisit later?

origami-z (approved):

bhoppers2008 (approved):

Issue #330, opened 2022-08-16
Density dependent tokens broken by nesting


libertymayc:
The ordering of density scoped variables is broken if the component name is also used and there are nested densities, i.e. in situations like:

.uitk-density-high .MyComponent {
      --myComponent-var: ;
}
.uitk-density-medium .MyComponent {
      --myComponent-var: ;
}

This can be seen when looking at 'Compare with Original Toolkit' stories if the above format is in code. Mostly, when the canvas is in medium density, everything appears OK, but switching to e.g. touch density will impact from which density the styling is from due to nesting and CSS ordering conflicts. 

The current best fix (as seen in #323) is to scope these density variables only to e.g. .uitk-density-medium, rather than .uitk-density-medium .MyComponent. The downside to this is that it explodes the amount of variables exposed within the density scopes, which also makes these variables seem like theme tokens rather than component tokens.

joshwooding:
Closing since I believe this is now fixed. Will re-open if not

PR #323, opened 2022-08-15
Characteristics: text font size token updates

libertymayc:
Changes to text characteristic fontSize tokens

TD:
H1 36px -> 42px
H2 28px -> 32px
H3 18px -> 24px
Caption 12px -> 14px
Help 12px -> 14px

LD:
Caption 11px -> 12px
Help 11px -> 12px

PR #335, opened 2022-08-18
CSS Fixes

libertymayc:
**SPLITTING INTO SEPARATE PRS:**

(#344) Standardise attributes appearing in tokens to always be kebab case, e.g. -border-color -> -borderColor _(lint rule added)_ 
(#323) Add min height to text.css help tokens
(#350) ToggleButton using characteristics: not 100% polished, but ensures lab package passes linting
(#371) Remove random values in List (e.g. —hw-… : check with @heswell about these) *noting that list/tree etc still require token alignment and so aren’t 100% currently
(#376) Remove component name classes when tokens are density dependent (see issue #330 for explanation)
(#376) Needs fixing: Grid component has special values for ultrahigh density 

To Do:

- Ensure component tokens start with their full name, e.g. —doc-grid -> —docGrid _(lint rule added)_
- For above backwards compat tokens, standardise by creating —backwardsCompat- token and then set the **local** var token (if exists, else the raw attr) within .backwardsCompat class
- Needs fixing: Grid component uses API for variables where local property should be used


docs/components/DocGrid.css
```diff
@@ -1,21 +1,21 @@
...
+  --docGrid-code-text-color: var(--uitk-text-primary-foreground);
 }
 
 .uitkDocGrid {
   align-items: center;
   background: var(--uitk-container-background-medium);
   display: grid;
   height: var(--uitkDocGrid-height, unset);
-  gap: var(--uitkDocGrid-gap, var(--doc-grid-gap));
+  gap: var(--uitkDocGrid-gap, var(--docGrid-gap));
```
alycrys, line 15: is this component overridable?

docs/components/DraggableSnapshot.css
```diff
@@ -1,16 +1,16 @@
 .uitkDraggableSnapshot {
-  --header-height: 120px;
-  --img-height: var(--uitkDraggableSnapshot-img-height, 900px);
-  --img-opacity: var(--uitkDraggableSnapshot-opacity, 0.5);
-  --img-width: var(--uitkDraggableSnapshot-img-width, 1600px);
-  --snapshot-url: var(--uitkDraggableSnapshot-url);
+  --draggableSnapshot-header-height: 120px;
+  --draggableSnapshot-img-height: var(--draggableSnapshot-uitkDraggableSnapshot-img-height, 900px);
```
alycrys, line 3: is this a typo? `--draggableSnapshot-uitkDraggableSnapshot-img-height`

packages/grid/src/grid/components/Grid.css
```diff
@@ -157,85 +157,87 @@
   --uitk-zIndex-tooltip: 1500;
```
libertymayc, line 157: what is the need for these repeated values? @ararus

packages/core/src/button/Button.css
```diff
@@ -143,24 +136,18 @@
 }
 
 /* Added .uitkButton:focus for browsers don't support :focus-visible, like Firefox */
-.uitkButton:focus {
-  outline-style: var(--button-focused-outlineStyle);
-  outline-width: var(--button-focused-outlineWidth);
-  outline-color: var(--button-focused-outlineColor);
-  outline-offset: var(--button-focused-outlineOffset);
+.uitkButton:focus,
+.uitkButton:focus-visible {
```
origami-z, line 140: Don't think this will work for browsers doesn't support `:focus-visible` per [grouping spec](https://www.w3.org/TR/selectors-3/#grouping)

packages/core/src/card/Card.css
```diff
@@ -9,7 +9,7 @@
   box-sizing: border-box;
   color: var(--uitkCard-color, var(--uitk-text-primary-foreground));
   overflow: hidden;
-  transition: box-shadow 0.3s cubic-bezier(0.4, 0, 0.2, 1);
+  transition: -boxShadow 0.3s cubic-bezier(0.4, 0, 0.2, 1);
```
origami-z, line 12: umm?

packages/core/src/layout/GridItem/GridItem.css
```diff
@@ -1,6 +1,6 @@
 /* Style applied to the root element */
 .uitkGridItem {
-  justify-self: var(--grid-item-justify-self);
+  justify-self: var(--grid-item-justifySelf);
```
origami-z, line 3: This would be same as below `align-self`? both are css property name

origami-z (commented):
Some change impacts public CSS API, need changeset

PR #338, opened 2022-08-18
Move scrim core

ZarkoPernar:


packages/lab/stories/scrim.doc.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+**Modal experience**  
+The Scrim provides a modal experience by obscuring the underlying content. It also prevents keyboard or mouse interactions from reaching this content until the Scrim and overlaying component are closed by the user or system.
+
+## Default
+
+When triggered, the basic Scrim fills the application window and blocks user interaction with the underlying content, until it’s closed. It is often displayed as the backdrop to another component, as demonstrated in this example.
+
+<Canvas>
+  <Story name="Default" decorators={[withFlexGap]}>
+    <DefaultScrim />
```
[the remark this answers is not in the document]
ZarkoPernar: Didn't know about the `id="xyz"` but sounds like a good idea.

packages/core/src/scrim/Scrim.css
```diff
@@ -1,9 +1,11 @@
+/* Style applied to the root element */
 .uitkScrim {
-  --scrim-background: var(--uitk-overlayable-background-low);
+  --scrim-background: var(--uitk-overlayable-background-medium);
 }
 
-.uitkEmphasisMedium.uitkScrim {
-  --scrim-background: var(--uitk-overlayable-background-medium);
+/* Apply this className to the element to achieve the lighter color variant */
+.uitkEmphasisLow {
```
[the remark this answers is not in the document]
ZarkoPernar: @libertymayc isn't this the pattern you were using in the "fix css PR"?

packages/lab/stories/scrim.doc.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+# API
+
+```
+import { Scrim } from "@jpmorganchase/uitk-core";
+```
+
+## Props
+
+<ArgsTable of={Scrim} />
```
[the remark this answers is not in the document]
ZarkoPernar: Good catch on the defaults 👍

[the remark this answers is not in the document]
ZarkoPernar: @origami-z As far as the prop name, yeah, I don't love `containerFix`. However, where does the scope fall in regards to api changes, especially major breaking changes, as part of this work?

packages/lab/stories/scrim.doc.stories.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+<Canvas>
+  <Story name="Default" decorators={[withFlexGap]}>
+    <DefaultScrim />
+  </Story>
+</Canvas>
+
+## Closable Scrim
+
+When triggered, the basic Scrim fills the application window and blocks user interaction with the underlying content, until it’s closed. It is often displayed as the backdrop to another component, as demonstrated in this example.
```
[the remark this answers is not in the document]
ZarkoPernar: Yep, swapped them around.

packages/core/src/scrim/Scrim.tsx
```diff
@@ -123,10 +118,7 @@ export interface ScrimProps extends HTMLAttributes<HTMLDivElement> {
 }
```
alycrys, line 118:
```
/**
   * By default, an error will be thrown if the focus trap contains no elements in its tab order.
   * With this option you can specify a fallback element to programmatically receive focus if no other tabbable elements are found.
   * For example, you may want a popover's <div> to receive focus if the popover's content includes no tabbable elements.
   */
  fallbackFocusRef?: FocusManagerProps["fallbackFocusRef"];
```
`<div>` needs to be escaped

packages/core/src/scrim/Scrim.css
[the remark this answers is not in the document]
alycrys: Maybe we should also mention in the comment that 'medium' emphasis is the default?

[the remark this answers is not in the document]
ZarkoPernar: Sounds good @alycrys

packages/core/src/scrim/Scrim.tsx
ZarkoPernar: Good catch 👍

alycrys (approved):

packages/lab/stories/scrim.doc.stories.mdx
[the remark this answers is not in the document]
joshwooding: API changes are fine. Especially renames since they have an easy migration path.

packages/core/src/focus-manager/internal/findAllTabbableElements.ts
```diff
@@ -1,4 +1,4 @@
-import { ownerWindow } from "../../utils";
+import { ownerWindow } from "../../utils/ownerWindow";
```
joshwooding, line 1: ?

package.json
```diff
@@ -24,7 +24,7 @@
     "build:ag-grid-theme": "cd packages/ag-grid-theme && yarn build",
     "test": "modular test",
     "test:components": "cypress run --component --browser chrome --headless",
-    "test:components:local": "cypress open --component --browser chrome",
+    "test:components:local": "cypress open --component --browser electron",
```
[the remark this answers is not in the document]
ZarkoPernar: I'll ask about it in tomorrows tech meeting

packages/core/src/focus-manager/internal/findAllTabbableElements.ts
ZarkoPernar: Ah yes, I auto-imported this before adding the export

packages/lab/stories/scrim.doc.stories.mdx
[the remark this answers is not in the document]
ZarkoPernar: I decided to do prop refactors in a separate PR #348

packages/core/src/scrim/Scrim.css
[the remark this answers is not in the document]
libertymayc: Not for emphasis - the problem was the space between e.g. .uitk-density-high .uitkScrim - but here we are specifically saying scrim is medium emphasis (no space between the classes), so there shouldn't be nesting problems

libertymayc:
Will need a changeset and also maybe change the pr description :)

packages/core/src/scrim/Scrim.css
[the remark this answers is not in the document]
ZarkoPernar: Ah right, that was density not emphasis 🤦 . I'll revert this change.

alycrys (approved):

origami-z (approved):

joshwooding (approved):

packages/core/src/scrim/Scrim.tsx
```diff
@@ -24,13 +18,15 @@ import "./Scrim.css";
...
+const defaultParent = () =>
+  typeof document !== "undefined" ? document.body : null;
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
 const noop = () => {};
 
 const withBaseName = makePrefixer("uitkScrim");
 
-function preventSelection(parent = defaultParent): () => void {
+function preventSelection(parent = defaultParent()): () => void {
```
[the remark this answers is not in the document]
ZarkoPernar: I wanted this to be evaluated every time the function runs and not just when the module is loaded, like it was before, in case of server side rendering.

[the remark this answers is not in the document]
EleanorHaproff: 👍 makes sense. Maybe should be getDefaultParent then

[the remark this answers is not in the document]
ZarkoPernar: Sounds good

package.json
joshwooding, line 27: Could we create a new PR for this please

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
```diff
@@ -1,12 +1,12 @@
 import { forwardRef, HTMLAttributes, useState, useEffect } from "react";
 import cx from "classnames";
 
+import { Scrim, ScrimProps } from "@jpmorganchase/uitk-core";
```
joshwooding, line 4: These imports should merging

package.json
ZarkoPernar: Ah, yes, forgot about this

packages/lab/src/layout/LayerLayout/LayerLayout.tsx
ZarkoPernar: You mean, should not be merged?

ZarkoPernar: Should I do the following?

```ts
import { Scrim } from "@jpmorganchase/uitk-core";
import type { ScrimProps } from "@jpmorganchase/uitk-core";
```

joshwooding: Sorry, as in with the other imports from the core package in the file.

ZarkoPernar: Ah, right. Im blind :'D

packages/core/src/__tests__/__e2e__/scrim/Scrim.cy.tsx
```diff
@@ -0,0 +1,278 @@
...
+          </>
+        );
+      }
+      cy.mount(<TestComponent />);
+
+      cy.findByRole("button", { name: "OPEN SCRIM" }).click();
+      cy.findByRole("button", { name: "CLOSE SCRIM" })
+        .should("have.focus")
+        .realClick();
+      cy.wait(50);
```
joshwooding, line 217: Is the wait actually needed?

joshwooding (approved):

packages/core/src/__tests__/__e2e__/scrim/Scrim.cy.tsx
ZarkoPernar: Yeah, this one is a bit hard to test because if you removed the wait it would still pass iirc. But that would be a false positive because the return of the focus takes a bit of time so we don't want to check immediately.

joshwooding: Not sure how it would pass as a false positive though?

ZarkoPernar: Because what we are trying to test here is the return of focus functionality, and that happens at a delay. If you test immediately, you are not really testing that functionality.

joshwooding: But if it returns true the element has focus which is the correct behaviour? Cypress should manage the retry or the delay if needed?

ZarkoPernar: We are testing that the previous button does **not** have focus. If we assert that immediately it will return true, because it doesn't, and then cypress will move on but at the time of the assertion the return focus hasn't done its thing yet. There is probably a better way to test this that I didn't think of but just removing the `wait()` is not it. The way I made sure this test was making the right assertions was to toggle the return focus functionality on/off while commenting out the wait command.

ZarkoPernar: Tbh, I don't even remember the specifics of that scenario just that I did play around with this and confirmed that I need the wait. I don't like adding arbitrary waits in my tests but usually it is the best way to prove a negative in a test.

ZarkoPernar: Unless you have something else in the ui to wait for. Here I don't think I have. Maybe waiting for the Scrim to disappear but I'm not sure of the timings. 🤔

joshwooding (approved):

PR #341, opened 2022-08-23
Remove elementType prop from Button.

ImmanuelBaskaran:
Changes:

- Removed polymorphic typing from Button
- Created DivButton to replace uses of button with elementType of div
- Moved keyboard handlers that trigger onClick into DivButton, since that code was created to handle interactions on elements other than button
- Edited ToolbarButton to accommodate the removal of polymorphic code ( added overflowLabel as a prop )


origami-z:
Marking draft before ready for review

packages/core/src/button/Button.tsx
```diff
@@ -36,32 +28,26 @@ export interface ButtonBaseProps<T extends ElementType> {
    * If `true`, the button will be focusable when disabled.
    */
   focusableWhenDisabled?: boolean;
-  onBlur?: (evt: FocusEvent<inferElementType<T>>) => void;
-  onClick?: (evt: MouseEvent<inferElementType<T>>) => void;
-  onKeyDown?: (evt: KeyboardEvent<inferElementType<T>>) => void;
-  onKeyUp?: (evt: KeyboardEvent<inferElementType<T>>) => void;
+
+  onBlur?: (evt: FocusEvent<HTMLButtonElement>) => void;
```
joshwooding, line 32: Do we still need these?

packages/core/src/button/Button.tsx
```diff
@@ -36,32 +28,26 @@ export interface ButtonBaseProps<T extends ElementType> {
...
   variant?: ButtonVariant;
 }
 
-export type ButtonProps<T extends ElementType = "button"> = ButtonBaseProps<T> &
-  Omit<ComponentPropsWithoutRef<T>, keyof ButtonBaseProps<T>>;
-
-type PolymorphicButton = <T extends ElementType = "button">(
-  p: ButtonProps<T> & { ref?: polymorphicRef<T> }
-) => ReactElement<ButtonProps<T>>;
+export type ButtonProps = ButtonBaseProps &
```
joshwooding, line 43: This could be simpler now and follow our usual patten of being in the interface?

packages/core/src/button/Button.tsx
```diff
@@ -70,9 +56,9 @@ export const Button = forwardRef(function Button<
     role: roleProp,
     variant = "primary",
     ...restProps
-  }: ButtonProps<T>,
-  ref?: polymorphicRef<T>
-): ReactElement<ButtonProps<T>> {
+  }: ButtonProps,
+  ref?: Ref<HTMLButtonElement>
```
joshwooding, line 60:
```suggestion
  ref?: ForwardedRef<HTMLButtonElement>
```

packages/core/src/button/Button.tsx
```diff
@@ -93,43 +79,29 @@ export const Button = forwardRef(function Button<
     };
   }, [active, keyIsDown]);
 
-  const handleKeyUp = (event: KeyboardEvent<inferElementType<T>>) => {
+  const handleKeyUp = (event: KeyboardEvent<HTMLButtonElement>) => {
```
joshwooding, line 82: If this active logic is the same between components could we create a useActiveStyle or something hook?

packages/core/src/button/Button.tsx
```diff
@@ -138,17 +110,11 @@ export const Button = forwardRef(function Button<
...
 
-  const Component: ElementType = elementType || "button";
   // Allow an explicit null value to be passed by user to suppress role
-  const role =
-    roleProp !== undefined
-      ? roleProp
-      : Component === "button" || Component === "a"
-      ? undefined
-      : "button";
+  const role = roleProp !== undefined ? roleProp : "button";
```
joshwooding, line 114: This logic isn't correct, also we shouldn't need it anymore

packages/core/src/button/DivButton.tsx
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * If `true`, the button will be disabled.
+   */
+  disabled?: boolean;
+  /**
+   * If `true`, the button will be focusable when disabled.
+   */
+  focusableWhenDisabled?: boolean;
+
+  onBlur?: (evt: FocusEvent<HTMLDivElement>) => void;
```
joshwooding, line 30: Same comments as the other file

packages/core/src/button/Button.tsx
joshwooding: We could also do a useButton hook and include focusWhenDisabled

packages/lab/src/dropdown/DropdownButton.tsx
```diff
@@ -90,7 +90,7 @@ export const DropdownButton = forwardRef(function DropdownButton(
       )}
       // We don't want the 'button' tag to be shown in the DOM to trigger some accessibility testing
```
joshwooding, line 91: This comment isn't valid now. We could either remove it or put it above the return

packages/lab/src/toolbar/ToolbarButton.tsx
```diff
@@ -1,15 +1,16 @@
...
-  };
+export type ToolbarButtonProps = ButtonProps & {
+  overflowLabel?: string;
+  label?: string;
+};
 
-export const ToolbarButton = forwardRef(function ToolbarButton<
-  T extends ElementType = "button"
->({ label, ...props }: ToolbarButtonProps<T>, forwardedRef: polymorphicRef<T>) {
+export const ToolbarButton = forwardRef(function ToolbarButton(
```
joshwooding, line 10: I've just noticed, is there a benefit to typing the parameters specifically instead of using the forwardRef's generics?

joshwooding (changes requested):
See feedback

packages/core/src/button/DivButton.tsx
```diff
@@ -0,0 +1,108 @@
...
+      "aria-disabled": ariaDisabled,
+      tabIndex,
+      onBlur: handleBlur,
+      onClick: handleClick,
+      onKeyDown: handleKeyDown,
+      onKeyUp: handleKeyUp,
+    } = buttonProps;
+
+    // Allow an explicit null value to be passed by user to suppress role
+    const role = roleProp !== undefined ? roleProp : "button";
```
joshwooding, line 86: We should probably inline this and let role be spread on div.

joshwooding (changes requested):
The docs for Button will now need to be updated

ImmanuelBaskaran:
This pull request contains temporary changes to Pill to accomodate moving the DivButton into labs as an internal component, these changes will be reverted when @joshwooding refactors Pill very soon

.changeset/yellow-tables-worry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-core": minor
+---
+
+Removed polymorphic types from Button
```
joshwooding, line 5: Can we expand on this please. Only saying the types are being removed is a bit misleading.

packages/lab/stories/toolbar.qa.stories.tsx
```diff
@@ -126,13 +126,13 @@ export const QA: Story<QAContainerProps> = ({ imgSrc }) => (
         </Button>
       </Tooltray>
       <Tooltray aria-label="buttons tooltray">
-        <ToolbarButton itemId="exportButton" overflowLabel="Export">
+        <ToolbarButton itemID="exportButton" overflowLabel="Export">
```
joshwooding, line 129: Was this change intentional?

joshwooding: Just looked through the code. This was more probably supposed to be id="" instead of itemId or itemID

joshwooding (approved):

PR #347, opened 2022-08-24
Toolbar enhancements

heswell:
fix styling on vertical toolbars and tooltrays
add stories to cover more form field controls


joshwooding:
Close as stale

PR #355, opened 2022-08-24
Fix: Bug in getCharacteristicValue

cheemcheem:
The theme's `id` is for example `uitk-dark` whereas the theme's `name` is just `dark`. 

By using `.uitk-${this.id}`, the `uitk-` portion was duplicated and the code would try to run ``document.body.querySelector(`.uitk-uitk-dark`)`` for example.

This change simply changes the `id` to `name` in that code snippet, alternatively, replacing `.uitk-${this.id}` with `${this.id}` would also suffice if that makes more sense.

While this is not fixed, you can instead call `getCharacteristicValue()` with the `scopeElement` parameter set to ``document.body.querySelector(`.uitk-${theme.name}`)``.

### DIFF
--- a/.changeset/small-starfishes-battle.md
+++ b/.changeset/small-starfishes-battle.md
@@ -1,0 +1,5 @@
+---
+"@jpmorganchase/uitk-core": patch
+---
+
+Fix theme getCharacteristicValue function.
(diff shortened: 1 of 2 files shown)


heswell (commented):
Not sure we should even support getCharacteristicValue. We stopped using it in ToolkitProvider because it does not work as you would expect when CSS Variable is assigned a computed value - i.e. it does not return the resolved value.

cheemcheem:
> Not sure we should even support getCharacteristicValue. We stopped using it in ToolkitProvider because it does not work as you would expect when CSS Variable is assigned a computed value - i.e. it does not return the resolved value.

Is there a better way to access values from a Theme? E.g. container background colour or current font colour etc.

joshwooding (approved):

joshwooding (approved):

PR #357, opened 2022-08-25
Tweaks to storybook examples and docs for FlexLayout

JoanaMMoreira:
- Rename FlexLayout responsive example
- Fix small typos/outdated info on docs
- Amend contact details storybook example so it stays the same width when expanding


packages/core/stories/layout/flex-layout.stories.tsx
```diff
@@ -147,10 +147,7 @@ export const ContactDetailsExample = ({ index }: { index: number }) => (
       <ContactMetadataItem value="London, GBR" label="Location" />
       <ContactMetadataItem value="+44 1234 123456" label="Office" />
 
-      <ContactMetadataItem
-        value="cillum.est.exercitation@cupidatat.com"
-        label="Email"
-      />
+      <ContactMetadataItem value="cillum@cupidatat.com" label="Email" />
```
joshwooding, line 150: I'm still seeing the width move? Also I don't really understand the composite usage examples?

JoanaMMoreira: Should be fixed now :) The idea behind the composite usage examples is that we combine a few TK components to showcase how easily the layout components can be used, in this case we have `FlexLayout`, `Card` and `ContactDetails` although to be fair it's very similar to the Simple Usage example

JoanaMMoreira: I can revisit the examples with Darrin and Ben if we think they don't make sense

libertymayc (approved):

joshwooding (approved):

PR #370, opened 2022-09-01
fixed Button.css so that disabled buttons do no longer appear clickable

ImmanuelBaskaran:
The button would behave as if it were still clickable.

This pull requests fixes this in the CSS by making the selectors for the active and focused styles more strict.

The active and focused styles will only be applied if the uitkButton-disabled pseudo class is not applied to the button, which occurs when the button is disabled.


packages/core/src/button/Button.css
```diff
@@ -161,20 +161,21 @@
...
+/* Pseudo-class applied to the root element on focus when Button is not active and not disabled
+   (focus ring will still be visible if button is focusable when disabled) */
+.uitkButton:focus-visible:not(:active):not(.uitkButton-active):not(.uitkButton-disabled) {
   background: var(--uitkButton-background-hover, var(--button-background-hover));
   color: var(--uitkButton-text-color-hover, var(--button-text-color-hover));
 }
 
 /* Pseudo-class applied to the root element on hover when Button is not active or disabled */
-.uitkButton:hover {
+.uitkButton:hover:not(.uitkButton-disabled) {
```
joshwooding, line 172: The usage of :not means the css is more difficult to override but for now this seems fine.

joshwooding (approved):

alycrys (approved):

PR #371, opened 2022-09-01
Minor CSS Fixes

libertymayc:
Part of CSS cleanup:

- In List/Tree: Some tokens started with e.g. --hw- which were either unused or incorrectly named (token doesn't start with component name); these have been fixed
- -bg changed to -background
- Activation Indicator tabs token change: --activation-indicator/--tabstrip-activation-indicator --> --tabs-activation-indicator
- change background-color attrs to background


packages/lab/src/tree/Tree.css
```diff
@@ -1,26 +1,24 @@
...
   --tree-node-expanded-transform: var(--uitkTree-node-expanded-transform, none);
   --tree-node-indent: 0px;
+  --tree-toggle-width: 12px;
 
-  --list-hilited-bg: var(--hw-list-hilited-bg, rgba(0, 0, 0, 0.1));
-  --list-item-height: var(--hw-list-item-height, 30px);
-  --list-item-padding: var(--hw-list-item-padding, 0 6px);
-  --list-item-header-bg: var(--hw-list-item-header-bg, black);
-  --list-item-header-color: var(--hw-list-item-header-color, white);
+  /* TODO: Use appropriate characteristics */
```
ZarkoPernar, line 9: Where do we document these requirements, an existing jira, or a new github issue? I imagine this is a todo for the designers?

ZarkoPernar (approved):

packages/lab/src/tree/Tree.css
libertymayc: this is just for us to implement the styling correctly according to the characteristics file.. the comment isn't really necessary since it applied to all lab components anyway

PR #386, opened 2022-09-06
Fix Card global css box sizing

### LINKED ISSUE #193: Card css is targeting global wildcard pseudo elements
Card css shouldn't target wildcard all pseudo elements 

```css 
.uitkCard *, *:before, *:after {
    box-sizing: inherit;
}
```

EleanorHaproff:
fixes #193 

`<Card/>` had some css which set things globally which has now been removed. (No component-level css files should do this)

In addition, `box-sizing: border-box` is now set globally properly in `/theme/global.css`, and removed from components.


joshwooding (approved):

joshwooding (changes requested):
Can we add a changeset please :)

packages/ag-grid-theme/css/_uitk-ag-theme-mixin.scss
```diff
@@ -571,7 +571,6 @@
 
   [class^="ag-"]:focus {
     outline: none;
-    box-sizing: border-box;
```
joshwooding, line 574: @ararus Do these changes make sense to you?

.changeset/healthy-coats-rule.md
```diff
@@ -0,0 +1,11 @@
+---
+"docs": minor
```
joshwooding, line 2: Can we remove the docs package from here please

packages/ag-grid-theme/css/_uitk-ag-theme-mixin.scss
ararus: It should be fine to move it to global, looks ok

PR #388, opened 2022-09-06
Fix icon size prop with number doesn't work

origami-z:
`--uitkIcon-size` was not set in Icon.css


packages/icons/src/icon/Icon.css
```diff
@@ -42,6 +38,7 @@
 }
 
 .uitkIcon .uitkIcon-content {
+  display: block;
   /* We don't want any font size impacting the size of Icon span */
   font-size: 0;
 }
```
libertymayc, line 44: we could get rid of the hardcoded 24px below, should be always defined

joshwooding:
Superseded by #564

PR #395, opened 2022-09-08
Replace default value for success aria-label in StateIcon

JoanaMMoreira:


packages/core/src/state-icon/StateIcon.tsx
```diff
@@ -37,6 +37,7 @@ export const StateIcon = forwardRef<HTMLSpanElement, StateIconProps>(
       <IconComponent
         className={classnames(withBaseName(), withBaseName(state), className)}
         size={size}
+        {...(state === "success" && { "aria-label": "success" })}
```
joshwooding, line 40: Another way of doing is this would be to use a map e.g.

```ts
const stateToAriaLabel = {
success: "success"
}

<IconComponent aria-label={stateToAriaLabel[state]} />
```

libertymayc (approved):

packages/core/src/state-icon/StateIcon.tsx
JoanaMMoreira: Cool, thanks! I had to change it slightly so we only pass `aria-label` conditionally i.e. we don't want this to be `undefined` for other states

JoanaMMoreira: I've reverted the change as discussed

origami-z (approved):

joshwooding (approved):

PR #397, opened 2022-09-08
Rename StateIcon component to StatusIcon

JoanaMMoreira:


pseys (approved):

packages/core/src/form-field/StatusIndicator.tsx
```diff
@@ -1,7 +1,7 @@
 import { IconProps } from "@jpmorganchase/uitk-icons";
 import cx from "classnames";
 import { ElementType } from "react";
-import { StateIcon } from "../state-icon";
+import { StatusIcon } from "../status-icon";
 import { Tooltip, TooltipProps, useTooltip } from "../tooltip";
 
 export type StateIndicatorState = "error" | "info" | "warning" | "success";
```
jeanlaleuf, line 7: Shouldn't this type be called StatusIndicatorState?

packages/core/src/form-field/StatusIndicator.tsx
```diff
@@ -44,14 +44,14 @@ export const StatusIndicator = (props: StatusIndicatorProps) => {
           ...TooltipProps,
         })}
       />
-      <StateIcon
-        {...getTriggerProps<typeof StateIcon>({
-          state,
+      <StatusIcon
+        {...getTriggerProps<typeof StatusIcon>({
+          status: state,
```
jeanlaleuf, line 49: Shouldn't we simply rename the prop to "status"?

packages/lab/src/dialog/internal/DialogContext.ts
```diff
@@ -1,8 +1,8 @@
 import { createContext } from "react";
-import { ValidationState } from "@jpmorganchase/uitk-core";
+import { ValidationStatus } from "@jpmorganchase/uitk-core";
 
 export const DialogContext = createContext<{
-  state?: ValidationState;
+  state?: ValidationStatus;
```
jeanlaleuf, line 5: DialogContext state prop should also be renamed

jeanlaleuf (commented):
I think we need to propagate the change from state to status more thoroughly.

JoanaMMoreira:
> I think we need to propagate the change from state to status more thoroughly.

Makes sense, I'll raise another PR

PR #404, opened 2022-09-12
Add enableContainerMode prop to Scrim

ZarkoPernar:
- I disabled all of the List.keyboardNavigation tests by renaming the file because it was randomly failing tests in react 16 and therefore blocking the branch
- add `enableContainerMode` prop in combination with existing `containerRef`


JoanaMMoreira (dismissed):

joshwooding (changes requested):
I don't think we should hide the list tests being disabled in this PR. If it's an actual issue in main other builds would be failing

packages/core/src/scrim/Scrim.tsx
```diff
@@ -92,6 +92,14 @@ export interface ScrimProps extends HTMLAttributes<HTMLDivElement> {
...
+  /**
+   * Set to true to enable the container use case. If true the Scrim will be bound to a container ref you pass via the `parentRef` prop.
+   * Default value is false, and the default behavior is for Scrim to be bound to the document viewport.
+   */
+  enableContainerMode?: boolean;
+  /**
+      * Prop necessary to enable container use case. Pass the parent element ref that you want Scrim to be bound to.
+      * Default value is undefined, and the default behavior is for Scrim to be bound to the document viewport.
   /**
    * Prop to enable container use case. Pass the parent element ref that you want Scrim to be bound to.
```
EleanorHaproff, line 104: Think you need to to remove this line? Looks like it's duplicated on line 101

ZarkoPernar:
@joshwooding I think this depends on the current state of github containers. The list tests were failing 3x in a row on 16 and now they pass. I will undo that change but these flakey tests are currently just adding noise to our checks and not much value.

packages/core/src/scrim/Scrim.tsx
ZarkoPernar: Good catch

EleanorHaproff (approved):

joshwooding (dismissed):

PR #405, opened 2022-09-12
Add @storybook/addon-storysource and reorder addons panel

origami-z:
- `@storybook/addon-storysource` would help non developer to look at source without dive deep into our code base
- Reorder addons panel, so "Controls" addon would appear by default 
- Bump storybook packages from to 6.5.12 as i couldn't find a way to use old version when `yarn add` using `^` and not manually updating `yarn.lock`.


.storybook/main.ts
```diff
@@ -20,18 +20,20 @@ const config: ExtendedConfig = {
...
     {
       name: "@storybook/addon-essentials",
       options: {
         // We don't want the backgrounds addon as our own withThemeBackground works with theme switch to apply background
         backgrounds: false,
       },
     },
+    "@storybook/addon-links",
+    "@storybook/addon-a11y",
+    "./theme-switch/preset", //👈 Custom theme switch on the toolbar
```
joshwooding, line 32: This can be removed now

joshwooding (approved):

PR #406, opened 2022-09-12
Test --docs flag on storybook

origami-z:
Separating docs and normal stories would help users to distinguish accessible docs v.s. 'develop' purpose stories


joshwooding:
Closing for now since it will be affected by Storybook 7 and the future site work. Nice experiment though!

PR #380, opened 2022-09-05
Ensure tokens start with --componentName or --uitkComponentName

libertymayc:
Part of CSS clean up:

Ensure component tokens start with their full name, e.g. —doc-grid -> —docGrid (lint rule added)
Moves subcomponents to next part of token, e.g. --uitkCheckboxIcon- -> --uitkCheckbox-icon-, --uitkAccordionSummary -> --uitkAccordion-summary which means all tokens are evidently scoped to a specific component in our collection
Tokens renamed if necessary, e.g. --formHelperText- -> --formField-helperText-
Layout component items are treated as standalone components (since Layout items can be placed within different Layout components other than their parent, this makes sense. It is also important to ensure differentiation between e.g. the Grid component --grid, the Grid Layout component --gridLayout, and Grid Layout items --gridItem)
Extra: Remove public token from focused characteristic inset value (there should not be an API on characteristics)

PR #409, opened 2022-09-13
Rename typography folder to text

joshwooding:
After some discussions with @jeanlaleuf he suggested the typography folder/stories should be renamed to Text.

Plus this fixes an issue with #380


packages/lab/stories/typography/BaseText.stories.tsx
```diff
@@ -4,7 +4,7 @@ import { Text, TextProps } from "@jpmorganchase/uitk-lab";
 import { useEffect, useState } from "react";
```
libertymayc, line 4: Should the stories folder also change to /text?

libertymayc: also does this need a changeset?

alycrys: Also the e2e tests folder

joshwooding: Whoops, I made the change locally but it wasn't picked up...

I don't think it needs a changeset as it's not changing any user facing code. Only refactoring the structure and we don't add changelogs for docs changes at the moment.

libertymayc (approved):

alycrys (approved):

PR #413, opened 2022-09-14
Revert to FlexLayout not wrapping by default

JoanaMMoreira:
- Change the default `wrap` value in `FlexLayout` to false so it matches css Flexbox
- Simplify `FlexLayout` examples

### DIFF
--- a/packages/core/src/layout/FlowLayout/FlowLayout.tsx
+++ b/packages/core/src/layout/FlowLayout/FlowLayout.tsx
@@ -3,7 +3,7 @@ import { FlexLayout, FlexLayoutProps } from "../FlexLayout";
 
 export interface FlowLayoutProps extends HTMLAttributes<HTMLDivElement> {
   /**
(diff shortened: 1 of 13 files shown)


origami-z (approved):

packages/core/src/layout/FlowLayout/FlowLayout.tsx
```diff
@@ -21,8 +21,6 @@ export interface FlowLayoutProps extends HTMLAttributes<HTMLDivElement> {
 }
 export const FlowLayout = forwardRef<HTMLDivElement, FlowLayoutProps>(
   function FlowLayout({ ...rest }, ref) {
-    return (
-      <FlexLayout direction="row" ref={ref} disableWrap={false} {...rest} />
-    );
+    return <FlexLayout direction="row" ref={ref} wrap={true} {...rest} />;
```
joshwooding, line 24:
```suggestion
    return <FlexLayout direction="row" ref={ref} wrap {...rest} />;
```

PR #414, opened 2022-09-14
Update Button docs

sion-docs:
Text updated based on doc template.

### DIFF
--- a/packages/core/stories/button.doc.stories.mdx
+++ b/packages/core/stories/button.doc.stories.mdx
@@ -24,9 +24,12 @@ import {
 
 # Button
 
-A Button executes an action when clicked. Button labels should use only one or two words, and begin with an active verb. There are three Button variants: Primary is the most commonly-used button; Secondary performs an action less likely to be triggered; Call-To-Action (a.k.a. CTA) performs a critical action on the screen.
+A Button executes an action when clicked. Button labels should use only one or two words and begin with an active verb. There are three Button variants:  
+- **Primary** is the most commonly-used button.
+- **Secondary** performs an action less likely to be triggered.
+- **Call-To-Action** (a.k.a. CTA) performs a critical action on the screen.
 
-## Button Variants
+## Button variants
 
 Button contains three variants `cta`, `primary` and `secondary`. If not specified, the default value is `primary`.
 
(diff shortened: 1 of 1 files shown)


packages/core/stories/button.doc.stories.mdx
```diff
@@ -24,9 +24,12 @@ import {
 
 # Button
 
-A Button executes an action when clicked. Button labels should use only one or two words, and begin with an active verb. There are three Button variants: Primary is the most commonly-used button; Secondary performs an action less likely to be triggered; Call-To-Action (a.k.a. CTA) performs a critical action on the screen.
+A Button executes an action when clicked. Button labels should use only one or two words and begin with an active verb. There are three Button variants:  
+- **Primary** is the most commonly-used button.
```
joshwooding, line 28: @sion-docs Do you think we should be consistent in the order of the variants when we refer to them?

packages/core/stories/button.doc.stories.mdx
```diff
@@ -89,21 +95,29 @@ export const Link = ({ children, to, ...props }) => (
...
 use hard-coded values local to Button. Both of these can be overridden, see Theme documentation for detailed usage guidance.
 
 The CSS custom properties below are consumed by Button, but not defined by Button. They can be defined via a container or
 a custom class name to override button styling. They will always take precedence over default styles, whether from theme variable
-or declared locally. Again, see Theme documentation for guidance on when to use variables from the Custom Property API.
+or declared locally. Again, see Theme documentation for guidance on when to use variables from the custom property API.
 
 <CSSVariableTable of={Button} />
+
+# Help and support
```
joshwooding, line 117: We should make this into a block so it's in one place soon

joshwooding (dismissed):

packages/core/stories/button.doc.stories.mdx
sion-docs: Yes, agreed.

packages/core/stories/button.doc.stories.mdx
sion-docs: That would be great, how can we do it?

joshwooding: It would be on our end. If I have some time I'll give it a go

joshwooding (approved):

PR #422, opened 2022-09-18
Update list ref and scrolling api ref

### LINKED ISSUE #359: VirtualizedList gives error with empty array
With below code, it will throw error

```
    <VirtualizedList
      source={[]}
    />
```

```
TypeError: Cannot read properties of undefined (reading 'id')
    at addItem (VirtualizedList.tsx:92:17)
    at renderItems (VirtualizedList.tsx:113:12)
    at renderContent (VirtualizedList.tsx:145:5)
    at List (VirtualizedList.tsx:187:28)
    at renderWithHooks (react-dom.development.js:14985:18)
    at updateForwardRef (react-dom.development.js:17044:20)
    at beginWork (react-dom.development.js:19098:16)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:3945:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:16)
    at invokeGuardedCallback (react-dom.development.js:4056:31)
```

origami-z:
- List ref is forwarded to its container HTML element, just like all other components
- Added `scrollingApiRef` for scrolling for both List and VirtualizedList
    - VirtualizedList doesn't scroll beyond buffered items, not planning to cover in this PR 
- Fix empty source passed to VirtualizedList will throw error (fixes #359)
- Fix virtualized list className uitkHighlighted

### DIFF
--- a/packages/lab/src/list/List.tsx
+++ b/packages/lab/src/list/List.tsx
@@ -1,14 +1,13 @@
-import { makePrefixer, useIdMemo } from "@jpmorganchase/uitk-core";
+import { makePrefixer, useForkRef, useIdMemo } from "@jpmorganchase/uitk-core";
+import cx from "classnames";
 import {
   cloneElement,
-  forwardRef,
   ForwardedRef,
+  forwardRef,
   isValidElement,
   ReactElement,
   useRef,
 } from "react";
-import cx from "classnames";
-import { forwardCallbackProps } from "../utils";
 import {
   CollectionIndexer,
   CollectionItem,
@@ -19,12 +18,12 @@ import {
   useCollectionItems,
   useImperativeScrollingAPI,
 } from "../common-hooks";
-
-import { useListHeight } from "./useListHeight";
+import { forwardCallbackProps } from "../utils";
 
 import { ListItem as DefaultListItem, ListItemProxy } from "./ListItem";
-import { useList } from "./useList";
 import { ListItemProps, ListProps } from "./listTypes";
+import { useList } from "./useList";
+import { useListHeight } from "./useListHeight";
 
 import "./List.css";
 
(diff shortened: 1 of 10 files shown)


packages/lab/src/list/useVirtualization.ts
```diff
@@ -36,17 +40,17 @@ export const useVirtualization = <Item>({
   const viewportMeasures = useRef({
     contentHeight: 10000,
     firstVisibleRow: 0,
-    rowCount: 1,
+    rowCount: 0,
     rowHeight: 0,
     scrollPos: 0,
   });
-  const [rows, setRows] = useState<Row<Item>[]>([[0, 0, 1, data[0]]]);
+  const [rows, setRows] = useState<Row<Item>[]>([]);
```
heswell, line 47: good catch - not sure why I set this to first row.

packages/lab/src/list/List.tsx
```diff
@@ -82,7 +82,7 @@ export const List = forwardRef(function List<
     width,
     ...htmlAttributes
   }: ListProps<Item, Selection>,
-  forwardedRef?: ForwardedRef<ScrollingAPI<Item>>
+  forwardedRef?: ForwardedRef<HTMLDivElement>
```
heswell, line 85: This makes the type of the forwardRef correct, as I'm using the forwardedRef for the imperative API. One thing to be aware of - there was some discussion about whether List should really do this, or set the forwarded ref on the html container, to make it mor similar to other components. We would then have to return the scrolling API via a custom ref.

David wanted the latter behaviour, because it would work better for App Builder. Can't remember what the consensus was, but as toy see, I didn't actually change it.

heswell (commented):
Looks good. While you're in here could you add a small additional fix to VirtualizedList - it should use uitkHighlighted as highlighted className, not uitkListItem-highlighted.

I was slowly aligning VirtualizedList with List, with a view to eventually merging them into one.

joshwooding (approved):

PR #426, opened 2022-09-20
Fix Tabs focus and animation

libertymayc:
Tabs focus (and transition, minor) broken in #380 
We don't want to provide an API directly in the characteristics; instead where needed, (like for Tabs- this PR), provide higher specificity on component with uitkFocusVisible class.

### DIFF
--- a/packages/lab/src/tabs/Tab.css
+++ b/packages/lab/src/tabs/Tab.css
@@ -28,9 +28,7 @@
 }
 
 /* Overrides characteristic used in uitkFocusVisible */
-.uitkTab.uitkFocusVisible:after,
-.uitkTab.focused:focus:after,
-.uitkTab.focused:focus-visible:after {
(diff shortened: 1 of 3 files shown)


packages/lab/src/tabs/Tab.css
```diff
@@ -27,6 +27,13 @@
   user-select: none;
 }
 
+/* Overrides characteristic used in uitkFocusVisible */
+.uitkTab.uitkFocusVisible:after,
```
joshwooding, line 31: Bit confused why .focused is being used here?

libertymayc: hmm, i was ignorant and thought it would apply here since thats how it is in the focused.css, but digging into code quickly im not sure why its there either. removed it from here at least

joshwooding (approved):

Issue #377, opened 2022-09-02
CSS Fixes and Clean up


libertymayc:
PRs created for the following (ticked when merged):

- [x]  (#344) Standardise attributes appearing in tokens to always be kebab case, e.g. -border-color -> -borderColor _(lint rule added)_ 
- [x]  (#323) Add min height to text.css help tokens
- [x] (#350) ToggleButton using characteristics: not 100% polished, but ensures lab package passes linting
- [x] (#371) Remove random values in List (e.g. —hw-… : check with @heswell about these) *noting that list/tree etc still require token alignment and so aren’t 100% currently
- [x] (#375) Updates to measured characteristic fill value
- [x] (#376) Remove component name classes when tokens are density dependent (see issue #330 for explanation)
- [x] (#376) Needs fixing: Grid component has special values for ultrahigh density 
- [x] (#380) Ensure component tokens start with their full name, e.g. —doc-grid -> —docGrid _(lint rule added)_
- [x] (#428) (#441) (#456) For backwards compat tokens, standardise by creating —backwardsCompat- token and then set the **local** var token (if exists, else the raw attr) within .backwardsCompat class
- [x] (#297) Needs fixing: Grid component uses API for variables where local property should be used

PR #428, opened 2022-09-20
Backwards compat styling standardisation 1 + fix RadioButton styles

libertymayc:
Part of CSS clean up (#377)

 For backwards compat tokens, standardise by:
- If density/theme dependent, creating —backwardsCompat- token
- Then set the local token of what needs adjusting (if exists, else the raw attr) within .backwardsCompat class to the backwardsCompat token defined or otherwise the necessary value

Components in this PR: 

> Card
> RadioButton (+ general RadioButton styling for backwards compat/nested in FormField fixed)
> Pill
> ControlLabel

(More components standardisation to come - but splitting into smaller PR's for easier reviewing)


packages/core/src/card/Card.css
```diff
@@ -65,9 +66,4 @@ a:focus .uitkCard-interactable.uitkCard-disabled {
 /* Styles applied to nested divs if `disabled={true}` */
 .uitkCard-disabled div {
   pointer-events: none;
-}
-
-/* TK1 backwards compat styling */
-.backwardsCompat.uitkCard .uitkCard-content {
-  line-height: var(--uitk-text-base-lineHeight);
-}
```
libertymayc, line 73: Unnecessary

packages/core/src/pill/internal/PillCheckbox.css
```diff
@@ -13,13 +13,18 @@
...
+.uitk-dark {
+  --backwardsCompat-pill-checkbox-icon-stroke: var(--uitk-color-grey-80);
+  --backwardsCompat-pill-checkbox-icon-tick-color: var(--uitk-color-grey-60);
 }
 
-.uitk-dark .backwardsCompat.uitkPill .uitkPill-checkbox {
-  --uitkCheckbox-icon-stroke: var(--uitk-color-grey-80);
-  --uitkCheckbox-icon-tick-color: var(--uitk-color-grey-60);
+.backwardsCompat.uitkPill .uitkPill-checkbox {
+  --uitkCheckbox-icon-stroke: var(--backwardsCompat-pill-checkbox-icon-stroke);
```
libertymayc, line 28: This is an interesting one as its back compat for a nested component and so we need to use the API but this limits consumer overrides..

packages/core/src/control-label/ControlLabel.css
```diff
@@ -64,6 +72,5 @@
 }
 
 .backwardsCompat .uitkControlLabel-label {
-  line-height: var(--uitk-text-base-lineHeight);
```
alycrys, line 67: Is this not needed anymore?

packages/core/src/pill/internal/PillCheckbox.css
libertymayc: yes, the trouble is this is a nested component (CheckboxIcon used in Pill) - so the API is used above for CheckboxIcon, and will always override the local var here even if it was set in backwards compat. so, we would have to also set the local var and ensure its higher specificity than the nested component itself, but atm we are using the API when overriding nested components, so would have to standardise that everywhere

packages/core/src/control-label/ControlLabel.css
libertymayc: thats the default anyway for the class (line 35)

packages/core/src/pill/internal/PillCheckbox.css
libertymayc: oops i accidentally deleted your comment @alycrys ! 

> "I think I'd prefer a local var. That's what we use on --formField-helperText-fontSize, don't we?"

alycrys (approved):

ImmanuelBaskaran (approved):

PR #457, opened 2022-09-26
FormField: detect not fully supported children inputs

ZarkoPernar:
### Overview
This draft is just focusing on detecting "fully supported" children. Not sure how to call this mode but right now it is just focusing on making it all look not broken. The padding needed to be adjusted on label and helper text because otherwise it looks whack. Switch looks bad either way.

### Changes when a child input is not fully supported
- remove status indicator 
- adjust padding on FormLabel and FormHelperText

### New FormField QA story

The following can be seen under the Core > Form Field > QA > Wrapping Inputs

### Before adjusted padding and bottom border

before

### After adjusted padding and bottom border

after

#### Left aligned label

Not sure what to do about left aligned labels :/

Left aligned label

### DIFF
--- a/packages/core/src/__tests__/__e2e__/form-field/FormField.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/form-field/FormField.cy.tsx
@@ -15,6 +15,44 @@ describe("GIVEN an Input", () => {
     });
   });
 
+  describe("WHEN there is a help text", () => {
+    it("THEN it should render help text", () => {
+      cy.mount(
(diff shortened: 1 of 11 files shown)


ZarkoPernar:
@pseys @bhoppers2008 I forgot one more design issue with FormField and checkboxes/radios, the left side padding on the label makes its positioning look off. In this PR i remove the padding on non-input children types but not sure that is what we want. What do you think?

joshwooding:
Closing since we are trying to achieve this a different way.

Issue #461, opened 2022-09-29
Default Button to `type="button"`


ZarkoPernar:
Currently our Button component has no default type attribute on the `button` element. By default HTML buttons are `type="submit"` and therefore submit the form if placed inside one. This usually trips up developers. I also checked and this is a breaking change from our previous version.

ZarkoPernar:
Closed by #466

PR #466, opened 2022-10-03
Make Button's default type "button" instead of submit

ImmanuelBaskaran:
This pull request fixes issue #461, by setting the attribute `type` to "button".

### DIFF
--- a/packages/core/src/button/Button.tsx
+++ b/packages/core/src/button/Button.tsx
@@ -38,7 +38,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       onBlur,
       onClick,
       role: roleProp,
-      type: typeProp,
+      type = "button",
       variant = "primary",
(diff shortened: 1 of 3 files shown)


packages/core/src/button/Button.tsx
```diff
@@ -38,6 +38,7 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
       onBlur,
       onClick,
       role: roleProp,
+      type: typeProp,
```
joshwooding, line 41:
```suggestion
      type = "button",
```
There's no point renaming the prop here and we should try and keep defaults with the props.

origami-z (changes requested):
Need some test

ZarkoPernar (approved):

joshwooding (dismissed):

joshwooding (approved):

PR #467, opened 2022-10-03
Add polymorphic types and prop to `BorderLayout` and `GridLayout`

JoanaMMoreira:
- Add polymorphic types for layout components
- Add "as" prop to `GridLayout`, `GridItem`, `BorderLayout` and `BorderItem`
- Add tests that check for type errors
- Update storybook examples


packages/core/src/layout/GridItem/GridItem.tsx
```diff
@@ -40,9 +45,14 @@ const colEnd = "auto";
...
-export const GridItem = forwardRef<HTMLDivElement, GridItemProps>(
-  function GridItem(
+type GridItemComponent = <T extends ElementType = "div">(
+  props: GridItemProps<T>
+) => ReactElement | null;
+
+export const GridItem: GridItemComponent = forwardRef(
+  <T extends ElementType = "div">(
     {
+      as,
```
Fercas123, line 55: do we limit the possible tags that can be used as `Component`? should we?

JoanaMMoreira: At the moment we accept any valid `ElementType` but it's open for discussion although I think if we come up with a list it would be quite long

Fercas123 (dismissed):

packages/core/src/__tests__/@typings/BorderItem.tsx
```diff
@@ -0,0 +1,58 @@
+import { BorderItem } from "@jpmorganchase/uitk-core";
+
+/*
+ * Important: These tests are meant for testing TypeScript type errors only
+ */
+
+/*
+ * Should allow relevant html attributes
+ */
+<BorderItem position="main" as="div">
```
JoanaMMoreira, line 10: @joshwooding I've updated these tests as discussed

Fercas123 (approved):
LGTM

PR #504, opened 2022-10-07
ADA tweaks on DeckLayout and DeckItem

JoanaMMoreira:
- Add accessibility checks to `Deck` cypress tests
- Add `tabIndex` and `id` to `DeckItem`
- Add `deckItemProps` prop to `DeckLayout`
- Make ADA tweaks on storybook examples

### DIFF
--- a/packages/core/stories/layout/deck-layout.stories.tsx
+++ b/packages/core/stories/layout/deck-layout.stories.tsx
@@ -18,7 +18,7 @@ export default {
 const deckCards = (slides: number) =>
   Array.from({ length: slides }, (_, index) => (
     <Card key={index}>
-      <h2>{`Deck Item ${index + 1}`}</h2>
+      <h2 id="deck_item_title">{`Deck Item ${index + 1}`}</h2>
       <p>
         We can implement your cross-border liquidity model in just a few months,
         depending on the options, scope and complexity.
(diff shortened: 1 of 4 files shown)


joshwooding (approved):

packages/core/stories/layout/deck-layout.stories.tsx
```diff
@@ -18,7 +18,7 @@ export default {
 const deckCards = (slides: number) =>
   Array.from({ length: slides }, (_, index) => (
     <Card key={index}>
-      <h2>{`Deck Item ${index + 1}`}</h2>
+      <h2 id="deck_item_title">{`Deck Item ${index + 1}`}</h2>
```
joshwooding, line 21: This is producing duplicate Ids

joshwooding (changes requested):
See comment

packages/core/stories/layout/deck-layout.stories.tsx
JoanaMMoreira: I've amended the Id although I've just realised we can't pass unique values to each `DeckItem` via `deckItemProps` e.g.`aria-labelledby` so if this is an ADA requirement we possibly need to extract `DeckItem` and allow it to be used separately

joshwooding: Right now it reads all DeckItem contents. Card doesn't get focus so `aria-labelledby` isn't doing anything here

JoanaMMoreira: Makes sense, I've removed this

joshwooding (approved):

PR #509, opened 2022-10-11
Improve Grid documentation

ararus:
Grid documentation


packages/grid/src/BaseCell.css
```diff
@@ -76,5 +75,5 @@
   border-top: 0 solid transparent;
   border-left: var(--grid-editableCell-cornerTag-size) solid var(--grid-cursor-borderColor);
   border-bottom: 8px solid transparent;
-  z-index: 10;
+  /*z-index: 10;*/
```
libertymayc, line 78: Needed if commented out or can we delete?

ImmanuelBaskaran (approved):

packages/grid/src/BaseCell.css
ararus: I was searching and deleting these. One left. Will remove in the next pr. Thanks

PR #517, opened 2022-10-13
Add more Grid tests

EleanorHaproff:


.changeset/fuzzy-clouds-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-grid": patch
+---
+
+Grid tests
```
joshwooding, line 5: If this is just adding tests we don't need a changeset

joshwooding (approved):

PR #537, opened 2022-10-18
Icon set update

pseys:
85 icons added or updated. Including solid variants for existing icons, progress icons, urgency icons, mode icons, and more.


packages/icons/src/SVG/api.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
joshwooding, line 1: The new SVGs have fill="none" on them which is currently causing issues :(

pseys: Is it just api.svg or are there others? Does this prevent all of them or just the effected ones?

joshwooding: Others too: 
image

pseys: That's so strange, because the all previewed fine when I exported them and added them to Github. It looks like its all the new ones.

joshwooding: It's because the svgs have a built in path fill but we remove that so we can style the icons via the theme. We currently don't remove the fill from the svg but we could.

packages/icons/scripts/generateIconComponent.js
```diff
@@ -74,6 +71,29 @@ glob(globPath, options, function (error, filenames) {
...
+        (node) => {
+          if (node.attribs.fill) {
+            delete node.attribs.fill;
+          }
+        },
+        document.children,
+        true
+      );
+
+      // This strips fill-rule and clip-rule from the paths
```
origami-z, line 85: Removing these rules may result different render, would be good to find out which icons are having these and whether they're rendering the same or not

origami-z: I think SVGO or one of the other tools are combining `path`, e.g. `zip-solid.svg` has 4 `path` but in React, only 1. So removing them could cause render being different, e.g. warning solid is showing a solid triangle for me on Storybook preview.

bhoppers2008 (approved):

Fercas123 (approved):

PR #544, opened 2022-10-18
Ensure Dropdown callback props are merged

heswell:


packages/lab/src/list/List.tsx
```diff
@@ -223,7 +222,7 @@ export const List = forwardRef(function List<
     };
     list.push(
       isChildItem ? (
-        cloneElement(value, forwardCallbackProps(value.props, listItemProps))
+        cloneElement(value, listItemProps)
```
heswell, line 225: forwardCallbackProps not required here as we do not include any callback props in the overrides

packages/lab/src/utils/forwardCallbackProps.ts
```diff
@@ -7,25 +7,27 @@
...
+  overrideProps: Props
+): Props => {
+  const props = {
+    ...overrideProps,
+    ...Object.keys(ownProps).reduce<Props>((map, name) => {
+      const ownProp = ownProps[name];
+      const overrideProp = overrideProps[name];
+      if (typeof ownProp === "function" && typeof overrideProp === "function") {
+        map[name] = (...args: unknown[]) => {
+          ownProp(...args);
```
heswell, line 24: call the clients 'ownProp' first. For those overridden callback props where defaultPrevented will cause them to bail, gives the client the ability to handle the event solo.

packages/lab/src/dropdown/DropdownBase.tsx
```diff
@@ -126,11 +126,11 @@ export const DropdownBase = forwardRef<HTMLDivElement, DropdownBaseProps>(
         id = defaultId,
         role = defaultRole,
         ...ownProps
-      } = trigger.props as MaybeProps;
+      } = trigger.props as MaybeChildProps;
 
       return cloneElement(
         trigger,
-        forwardCallbackProps(ownProps, {
+        forwardCallbackProps(ownProps as any, {
```
heswell, line 133: sorry about the any. - couldn't see what the darn thing wouldn't typecheck

joshwooding (dismissed):

PR #559, opened 2022-10-24
Grid css update

ararus:
Removing hardcoded values from grid css. Switching to characteristics where possible.

### DIFF
--- a/packages/grid/src/Grid.css
+++ b/packages/grid/src/Grid.css
@@ -24,7 +24,7 @@
 
 .uitk-density-high {
   --grid-row-height-default: 25px;
-  --grid-padding-default: 8px;
+  --grid-padding-default: 4px;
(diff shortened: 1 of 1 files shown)


packages/grid/src/Grid.css
```diff
@@ -55,7 +55,7 @@
   --grid-groupHeader-color-default: var(--uitk-color-grey-200);
   --grid-shadow-color-default: rgba(0, 0, 0, 0.2);
   --grid-columnDropTarget-color-default: var(--uitk-color-blue-500);
-  --grid-cursor-borderColor-default: #00477b;
+  --grid-cursor-borderColor-default: var(--uitk-color-blue-400);
```
libertymayc, line 58: We should use the palette and not colors directly. Do any of the palette semantics match your needs? otherwise we could look at adding appropriate tokens

PR #570, opened 2022-10-25
Update typography.stories.mdx

sion-docs:
Small update for clarity.

### DIFF
--- a/packages/theme/stories/foundations/typography.stories.mdx
+++ b/packages/theme/stories/foundations/typography.stories.mdx
@@ -7,6 +7,7 @@ import { TextBlock } from "docs/components/TextBlock";
 <Meta title="Documentation/Styles and Theming/Foundations/Typography" />
 
 # Typography
+The default font set within our foundations is a sans-serif font named UITK Sans. Monospaced text is presented using PT Mono.
 
 ## Typography
 
--- a/packages/theme/stories/foundations/typography.stories.mdx
+++ b/packages/theme/stories/foundations/typography.stories.mdx
@@ -7,6 +7,7 @@ import { TextBlock } from "docs/components/TextBlock";
 <Meta title="Documentation/Styles and Theming/Foundations/Typography" />
 
 # Typography
+
 The default font set within our foundations is a sans-serif font named UITK Sans. Monospaced text is presented using PT Mono.
 
 ## Typography


packages/theme/stories/foundations/typography.stories.mdx
```diff
@@ -7,6 +7,7 @@ import { TextBlock } from "docs/components/TextBlock";
 <Meta title="Documentation/Styles and Theming/Foundations/Typography" />
 
 # Typography
+The default font set within our foundations is a sans-serif font named UITK Sans. Monospaced text is presented using PT Mono.
```
joshwooding, line 10: I didn't think we had a mono spaced font? Is this going to be standard now?

libertymayc: We have a code variant in the text characteristic that uses this, it's been in Figma for a while - is there a reason we shouldn't have it?

ZarkoPernar: Loading additional fonts is bad for performance. Not sure if we should be too concerned with monospaced fonts. Their applications are limited.

joshwooding: Ah so it is. I was a bit surprised since we don't really use it elsewhere. We don't ship it either and rely on the browser to have it

origami-z: > Loading additional fonts is bad for performance. Not sure if we should be too concerned with monospaced fonts. Their applications are limited.

This is necessary for financial figures so numbers can be read more easily across the board.

joshwooding: @origami-z do we currently do this in our designs?

origami-z: I don't know whether uitk put this out as a rule, but will appear very obvious in data grid right aligned scenarios, and our consumers would be using this. However, Open Sans digits may be monospaced already, so may not be 100% related to the PT Mono problem we're talking here.

ZarkoPernar: >> Loading additional fonts is bad for performance. Not sure if we should be too concerned with monospaced fonts. Their applications are limited.

>This is necessary for financial figures so numbers can be read more easily across the board.

I thought monospaced fonts were all about predictable space, not readability? And that we can achieve with `font-variant-numeric : tabular-nums;` 😁

libertymayc: @pseys what are your thoughts on this?

joshwooding (approved):

PR #571, opened 2022-10-25
Improve Icon tests and documentation

Fercas123:


packages/icons/src/__tests__/__e2e__/Icon.cy.tsx
```diff
@@ -0,0 +1,17 @@
...
+import * as iconStory from "../../../stories/icon.stories";
+import { checkAccessibility } from "../../../../../cypress/tests/checkAccessibility";
+
+const composedStories = composeStories(iconStory);
+const { CustomSVGIcon } = composedStories;
+
+describe("Given an icon", () => {
+  checkAccessibility(composedStories);
+
+  it("should render", () => {
```
joshwooding, line 11: I think this should be done via visual regression

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,81 @@
...
+  component={Icon}
+  parameters={{
+    viewMode: "docs",
+  }}
+/>
+
+# Icon
+
+Icon is an important visual representation of an application, a capability, or some other concept or specific entity with meaning on a interface. It is expected to help user to recognize and understand a content or an action regardless languages or culture boundaries.
+We have 3 sizes of icon set for users to choose in terms of different platforms or device size.
```
el-dav, line 22: I guess we'll need to update the size part of this

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+</Canvas>
+
+## Use Icon when...
+
+- You need a visual representation of a concept.
+- An icon doesn't need to be interactive.
+
+## Features
+
+We have default icon set that starts from 12px for HD & MD, then 14px for LD, 16px for TD. (Since our base size of HD/default is 10px, but we don’t allowing icon size that is smaller then 12px, hence why here, the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
```
el-dav, line 36: "We have **a** default icon set"

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+Have more size choices by adding multiplies to 2x, 3x, 4x etc.
+
+<!--  <Canvas><Story id="" /></Canvas> -->
+
+We provide 2 types of icons: default and solid. Some special icon may exclueded from it like plus and loader
```
el-dav, line 41: excluded typo

packages/icons/stories/icon.stories.tsx
```diff
@@ -87,3 +100,32 @@ export const AllIcons: ComponentStory<typeof Icon> = () => {
...
+      </FormField>
+      <FlexLayout wrap gap={1}>
+        {allIconsNames
+          .filter(({ name, icon }) => new RegExp(inputText, "i").test(name))
+          .map(({ name, icon }, i) => {
+            return (
+              <StackLayout style={{ width: "150px" }} align="center">
+                {createElement(icon, {
+                  key: i,
+                  size: 24,
```
el-dav, line 122: Need to update size value here, it's current 24 x the base value :)

el-dav:
Would be good to include some documentation on a few more things:

- Using the color css variables to change Icon color e.g.
```jsx
<AddDocumentIcon style={{ '--uitkIcon-color': 'var(--uitk-status-info-foreground)' }}/>
```

- How to use the base ```jsx <Icon />``` component if you're creating a custom Icon e.g.
```jsx
 <Icon
      aria-label="add"
      viewBox="0 0 12 12"
      {...customIconProps}
    >
      <path d="M7 0H5v5H0v2h5v5h2V7h5V5H7V0z" />
    </Icon>
```

packages/icons/stories/icon.stories.tsx
Fercas123: 😆 thanks

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Props
+
+<ArgsTable of={Icon} />
+
+## Anatomy
+
+We have default and solid icons.  
+You can supply custom icons as `children` of the `Icon` component, by using `<svg>`.
```
joshwooding, line 68: This doesn't fall under Anatomy to me.
Also I believe the custom icon guidance needs to change after the other PR.

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+## Features
+
+We have a default icon set that starts from 12px for high- and medium density (MD), then 14px for low density (LD), and 16px for touch density (TD). (Since our base size of high density (HD)/default is 10px but we don’t allow icons size smaller than 12px, hence why here the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
+You have more size choices by adding multiplies to 2x, 3x, 4x etc.
+
+<Canvas>
+  <Story id="icons-icon--toolkit-icon-multiple-sizes" />
+</Canvas>
+
+We provide two types of icons: default and solid. Some special icons may be excluded from it like `plus` and `loader`.
```
sion-docs, line 44: At the moment the second sentence here is a little unclear. We should provide a full list of icons that are excluded  - 'Some' is too vague - and explain why they are excluded/special.

joshwooding: @pseys Maybe you can provide more context here

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -34,14 +34,14 @@ If an icon is interactive, use [Button](?path=/docs/documentation-core-button--b
 
 ## Features
 
-We have a default icon set that starts from 12px for HD & MD, then 14px for LD, 16px for TD. (Since our base size of HD/default is 10px, but we don’t allowing icon size that is smaller then 12px, hence why here, the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
-Have more size choices by adding multiplies to 2x, 3x, 4x etc.
+We have a default icon set that starts from 12px for high- and medium density (MD), then 14px for low density (LD), and 16px for touch density (TD). (Since our base size of high density (HD)/default is 10px but we don’t allow icons size smaller than 12px, hence why here the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
```
joshwooding, line 37:
```suggestion
We have a default icon set that starts from 12px for high and medium density (MD), then 14px for low density (LD), and 16px for touch density (TD). (Since our base size of high density (HD)/default is 10px but we don’t allow icons size smaller than 12px, hence why here the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
```

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -55,7 +55,7 @@ To override the default color of the icon, you can use the `color` css variable.
   />
 </Canvas>
 
-You can supply custom icons as `children` of the `Icon` component by using an `<svg>`.
+You can supply custom icons as 'children' of the `Icon` component by using an `<svg>`.
```
joshwooding, line 58: Just curious about this change. `children` is a special value when dealing with components.

Also @Fercas123 seems like this still needs to be updated

sion-docs: my mistake

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+To override the default color of the icon, you can use the `color` css variable.
+
+<Canvas>
+  <AddDocumentIcon
+    style={{ "--uitkIcon-color": "var(--uitk-status-info-foreground)" }}
+  />
+</Canvas>
+
+You can supply custom icons as `children` of the `Icon` component by using an `<svg>`. It is necessary to pass an aria-label to the Icon whenever adding a custom icon.
```
Fercas123, line 58: @sion-docs fyi. I just added another bit about labels here, if you could check if it makes sense that would be great :)

packages/icons/stories/icon.all.ts
```diff
@@ -543,3 +543,198 @@ export const allIcons = [
   ZipIcon,
   ZipSolidIcon,
 ];
+export const allIconsNames = [
```
el-dav, line 546: Just noticed that this list is missing a few e.g. DarkIcon. Maybe we can update this file to make it a bit easier to manage e.g. not needing to be updated if we generate different icons in the future...

something like:
```tsx
// icon.all.ts
import * as allIconsExport from "../src/components";

export const allIcons = Object.values(allIconsExport);

export const allIconsNames = Object.entries(allIconsExport).map(
  ([name, icon]) => ({
    name,
    icon,
  })
);

```

packages/icons/stories/icon.stories.tsx
```diff
@@ -87,3 +100,32 @@ export const AllIcons: ComponentStory<typeof Icon> = () => {
...
+        {allIconsNames
+          .filter(({ name, icon }) => new RegExp(inputText, "i").test(name))
+          .map(({ name, icon }, i) => {
+            return (
+              <StackLayout style={{ width: "150px" }} align="center">
+                {createElement(icon, {
+                  key: i,
+                  size: 2,
+                })}
+                <p>{name.slice(0, -1)}</p>
```
el-dav, line 124: What's the slice for? It seems to be cutting off the last letter of the name

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+To override the default color of the icon, you can use the `color` css variable.
+
+<Canvas>
+  <AddDocumentIcon
+    style={{ "--uitkIcon-color": "var(--uitk-status-info-foreground)" }}
+  />
+</Canvas>
+
+You can supply custom icons as `children` of the `Icon` component by using an `<svg>`. Whenever you add a custom icon, you need to pass an aria-label to the Icon. 
+
```
sion-docs, line 59: Changed to make the tone more active/2nd person. Should aria-label be `monospace`?

packages/icons/stories/icon.doc.stories.mdx
sion-docs: I have updated the text and made another commit, see below.

packages/icons/stories/icon.stories.tsx
Fercas123: it was removing a comma, but it shouldn't be needed if we change the script

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+  <Story id="icons-icon--toolkit-icon-multiple-sizes" />
+</Canvas>
+
+We provide two types of icons: default and solid. Some special icons may be excluded from it like `plus` and `loader`.
+
+<Canvas>
+  <Story id="icons-icon--toolkit-types" />
+</Canvas>
+
+To override the default color of the icon, you can use the `color` css variable.
```
joshwooding, line 50: I'm thinking we probably want to refer to the whole CSS variable. Thoughts?

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+<Canvas>
+  <AddDocumentIcon
+    style={{ "--uitkIcon-color": "var(--uitk-status-info-foreground)" }}
+  />
+</Canvas>
+
+You can supply custom icons as `children` of the `Icon` component by using an `<svg>`. Whenever you add a custom icon you need to pass an `aria-label` to the Icon.
```
joshwooding, line 58: I still feel like this needs clarification. What does using an <SVG> mean?

Should this be something more like:

```suggestion
You can supply custom icons as `children` of the `Icon` component. `Icon` has the same API as the native `<svg>` element. Whenever you add a custom icon you need to pass an `aria-label` to the Icon.
```

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,103 @@
...
+
+<CSSClassTable of={Icon} />
+
+## Characteristics Used
+
+<CharacteristicUsage of={Icon} />
+
+## Accessibility
+
+To ensure that icon-only buttons are ADA compliant, when icon is inside a button you must display the button’s label in a supporting Tooltip that is revealed on mouse hover or keyboard focus.
```
joshwooding, line 100: I would argue this is not specific to Icon and shouldn't be here. Seems like something that should be on the Button page and potentially linked to

joshwooding (changes requested):
see comments

packages/icons/stories/icon.doc.stories.mdx
Fercas123: I'll update this, @sion-docs / @joshwooding would you say the next paragraph makes sense here or should i move it to accessibility? (or some content in both)

joshwooding: I think it's fine for it to be here. Whether that is achieved by putting it directly here or linking to it and moving it to the accessibility section I don't have a preference

sion-docs: I'm happy to go with Josh's judgment.

packages/icons/stories/icon.doc.stories.mdx
Fercas123: I checked with @wennazhang22 for context and updated this paragraph

packages/icons/stories/icon.doc.stories.mdx
Fercas123: happy to leave it here :)

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+To override the default color of the icon, you can use the `--uitkIcon-color` CSS variable.
+
+<Canvas>
+  <AddDocumentIcon
+    style={{ "--uitkIcon-color": "var(--uitk-status-info-foreground)" }}
+  />
+</Canvas>
+
+You can supply custom icons as `children` of the `Icon` component. `Icon` has the same API as the native `<svg>` element. Whenever you add a custom icon you need to pass an `aria-label` to the Icon.
```
joshwooding, line 58:
```suggestion
If you need a custom SVG icon you can use the `Icon` component. `Icon` has the same API as the native `<svg>` element. Whenever you add a custom icon you need to pass an `aria-label` to the Icon.
```

joshwooding: cc @sion-docs

packages/icons/stories/icon.doc.stories.mdx
pseys: Has this question been resolved now? Sorry for the delay, I must've missed a notification. Its not that they're special, they just don't have solid versions because it's not relevant/required. For example, the plus icon doesn't have a 'fillable' aspect so can't have a solid version.

packages/icons/stories/icon.doc.stories.mdx
```diff
@@ -0,0 +1,98 @@
...
+## Features
+
+We have a default icon set that starts from 12px for high and medium density (MD), then 14px for low density (LD), and 16px for touch density (TD). (Since our base size of high density (HD)/default is 10px but we don’t allow icons size smaller than 12px, hence why here the HD/default will start from 12px, and its 2X is 20px, 3X is 30px etc.)
+You have more size choices by adding multiplies to 2x, 3x, 4x etc.
+
+<Canvas>
+  <Story id="icons-icon--toolkit-icon-multiple-sizes" />
+</Canvas>
+
+We provide two types of icons: default and solid. Most of our icons will have the two versions, except cases in which the icon shape is too narrow to have a solid version like `plus` and `loader`.
```
Fercas123, line 44: @pseys

packages/icons/stories/icon.doc.stories.mdx
sion-docs: Reads fine to me

packages/icons/stories/icon.doc.stories.mdx
pseys: Great, thank you.

packages/icons/stories/icon.doc.stories.mdx
joshwooding: @sion-docs Does the content still need to be changed?

sion-docs: See the most recent comment below, Fernanda has changed it.

packages/icons/stories/icon.doc.stories.mdx
sion-docs: This makes sense.

sion-docs (dismissed):

joshwooding (approved):

PR #572, opened 2022-10-25
Update zindex.stories.mdx

sion-docs:
Small addition for clarity.


packages/theme/stories/foundations/zindex.stories.mdx
```diff
@@ -5,6 +5,9 @@ import { DocGrid } from "docs/components/DocGrid";
 <Meta title="Documentation/Styles and Theming/Foundations/Z-Index" />
 
 # Z-Index
+The z-index property specifies the stack order of a component. A component with a greater stack order is always in front of an element with a lower stack order.
+
+Often used with Shadow.
```
joshwooding, line 10: Maybe we should add a design explanation here?

libertymayc (approved):

packages/theme/stories/foundations/zindex.stories.mdx
libertymayc: do you mean an example?

joshwooding: As in we should answer the question: "Why is it often used with shadow"

joshwooding (approved):

PR #588, opened 2022-10-27
Change -typography-weight tokens to -typography-fontWeight

libertymayc:
Change -typography-weight tokens to -typography-fontWeight to align with -text-fontWeight choice in characteristic tokens

### DIFF
--- a/.changeset/violet-waves-carry.md
+++ b/.changeset/violet-waves-carry.md
@@ -1,0 +1,6 @@
+---
+"@jpmorganchase/uitk-lab": minor
+"@jpmorganchase/uitk-theme": minor
+---
+
+Change -typography-weight tokens to -typography-fontWeight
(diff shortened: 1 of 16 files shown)


joshwooding:
Do we want to update [uitkTheme.ts](https://github.com/jpmorganchase/uitk/blob/main/tooling/theme-editor/src/themes/uitkTheme.ts) too?

libertymayc:
> Do we want to update [uitkTheme.ts](https://github.com/jpmorganchase/uitk/blob/main/tooling/theme-editor/src/themes/uitkTheme.ts) too?

yes, I have the final theme in another branch, I think it may be better to have that in a separate PR rather than each change as its very behind anyway? @joshwooding

joshwooding (approved):

PR #591, opened 2022-10-28
Update introduction.stories.mdx

sion-docs:
Updated based on review comments


packages/theme/stories/introduction.stories.mdx
```diff
@@ -218,12 +218,12 @@ We have already mentioned that components will reference Theme variables, the ex
...
+Note the naming convention for the first variable `--uitkButton-fontSize`. The prefix matches exactly (even in case) the CSS class of the component. That identifies this as a *CSS API variable*. These are not actually assigned any value anywhere in the UI Toolkit code. They are a customisation mechanism intended for use by consumers of Toolkit components. A default (i.e. fallback) value is _always_ provided in this case, and will often reference a theme characteristic variable. If consumers need to override a default style assignment, these API variables provide a simple means to do this. The group of API variables available with each component are published as part of the docs for each component (as are the CSS class names).
```
Fercas123, line 227:
```suggestion
Note the naming convention for the first variable `--uitkButton-fontSize`. The prefix matches exactly (even in case) the CSS class of the component. That identifies this as a *CSS API variable*. These are not actually assigned any value anywhere in the UI Toolkit code. They are a customization mechanism intended for use by consumers of Toolkit components. A default (i.e. fallback) value is _always_ provided in this case, and will often reference a theme characteristic variable. If consumers need to override a default style assignment, these API variables provide a simple means to do this. The group of API variables available with each component are published as part of the docs for each component (as are the CSS class names).
```

libertymayc (approved):

joshwooding (approved):

PR #600, opened 2022-10-31
Add applyClassesTo prop for ToolkitProvider

ImmanuelBaskaran:
Changed applyClassesToChild prop to applyClassesTo prop in ToolkitProvider, if this prop is set to root on a root level Toolkitprovider, then the provider will apply the theme classes to the html element instead of creating a div element. Setting the prop to child will make the ToolkitProvider apply the classes to the child element as before


joshwooding:
I was chatting about this yesterday, I think it would be interesting to explore adding a data attribute to the HTML element

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -94,11 +108,13 @@ interface ToolkitProviderThatInjectsThemeElement {
   children: ReactNode;
   density?: Density;
   theme?: ThemeNameType;
+  applyClassesToBody?: false;
   applyClassesToChild?: false;
   breakpoints?: Breakpoints;
 }
```
alycrys, line 114: These interfaces look repetitive. Anything we could change here?

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -80,10 +84,20 @@ const createThemedChildren = (
   }
 };
 
+interface ToolkitProviderThatAppliesClassesToBody {
+  children: ReactNode;
+  density?: Density;
+  theme?: ThemeNameType;
+  applyClassesToBody?: true;
```
el-dav, line 91: I think I would change these multiple boolean props to be a single enum prop as I don't think it makes sense to have both of these booleans as true e.g.

```
<ToolkitProvider
  applyClassesToBody={true}
  applyClassesToChild={true}
>
```

Instead with an enum you are only allowed one option:

```
<ToolkitProvider
  applyClassesTo="body"
>
```

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -137,9 +154,35 @@ export function ToolkitProvider({
...
+    return () => {
+      if (applyClassesToBody) {
+        // When unmounting/remounting, remove the applied styles from the body
+        document.body.classList.remove(
+          ...themes.map((theme) => `uitk-${theme.name}`),
+          `uitk-density-${density}`
+        );
+      }
+    };
+  }, [applyClassesToBody, density, isRoot, themes]);
```
el-dav, line 184: would you mind checking how often this effect runs? I think it might end up adding and removing the classes from the body every time ToolkitProvider renders due to themes being a new array every render.

We might want to do some optimisation here

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -137,9 +154,35 @@ export function ToolkitProvider({
...
+        console.warn(
+          "\nToolkitProvider can only apply CSS classes to the body if it is the root level ToolkitProvider."
+        );
+      }
+    }
+    return () => {
+      if (applyClassesToBody) {
+        // When unmounting/remounting, remove the applied styles from the body
+        document.body.classList.remove(
+          ...themes.map((theme) => `uitk-${theme.name}`),
```
el-dav, line 179: This map seems to happen in a few places e.g. also in `createThemedChildren`. Maybe we can just do it once and use the result in multiple places?

packages/core/src/__tests__/__e2e__/toolkit-provider/ToolkitProvider.cy.tsx
```diff
@@ -164,6 +163,22 @@ describe("Given a ToolkitProvider", () => {
...
+        <ToolkitProvider density="high" theme="dark" applyClassesToBody>
+          <TestComponent />
+        </ToolkitProvider>
+      );
+
+      cy.get("uitk-theme").should("have.length", 0);
+
+      cy.get("body")
+        .should("exist")
+        .and("have.attr", "class", "uitk-dark uitk-density-high");
```
el-dav, line 178: Is this dependant on the order of classes? Maybe something like:
```
.should('have.class', 'uitk-dark')
.and('have.class', 'uitk-density-high')
```

.changeset/empty-waves-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-core": minor
+---
+
+change applyClassesToChild prop to applyClassesToBody prop in ToolkitProvider, if this prop is set to "root" on a root level Toolkitprovider, then the provider will apply the theme classes to the html element instead of creating a uitk-theme element. Setting the prop to "child" will make the ToolkitProvider apply the classes to the child element as before
```
el-dav, line 5: Changeset needs updating following api change

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -137,16 +146,46 @@ export function ToolkitProvider({
...
+    if (applyClassesTo === "root") {
+      if (isRoot) {
+        // add the styles we want to apply
+        document.documentElement.classList.add(
+          ...themeClassnames,
+          `uitk-density-${density}`
+        );
+      } else {
+        console.warn(
+          "\nToolkitProvider can only apply CSS classes to the body if it is the root level ToolkitProvider."
```
el-dav, line 174: change 'body' to 'root'

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -137,16 +146,46 @@ export function ToolkitProvider({
     (Array.isArray(inheritedThemes) && inheritedThemes.length === 0);
   const density = densityProp ?? inheritedDensity ?? DEFAULT_DENSITY;
   const themeName = getThemeName(themesProp, inheritedThemes);
-  const themes: Theme[] = getTheme(themeName);
+  const themes: Theme[] = useMemo(() => getTheme(themeName), [themeName]);
```
el-dav, line 149: This definitely better.

@joshwooding Do we care that consumers may pass in an inline array of themes which could potentially be a new value each render? e.g. `<ToolkitProvider theme={['light', 'other']} >`

joshwooding: I guess, we should add a comment to say theme should be stable

el-dav: As we said before the alternative is to stringily it to use as a dependency, but that could be overkill

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -118,16 +131,50 @@ export function ToolkitProvider({
...
   const density = densityProp ?? inheritedDensity ?? DEFAULT_DENSITY;
   const themeName = getThemeName(themesProp, inheritedThemes);
-  const themes: Theme[] = getTheme(themeName);
+  // We expect theme to be stable
+  const themNameAsString = themeName.toString();
+  const themes: Theme[] = useMemo(
+    () => getTheme(themeName),
+    // if an array is passed to theme inline, themes would be recalculated each time
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [themNameAsString]
```
steveukx, line 140: You could avoid the lint break, and benefit from the memoising by grouping together more computed variables in the memo:

```typescript
  const theme = useStaticValue(themesProp);
  const {themeClassnames, themes} = useMemo(() => {
    const themeName = getThemeName(theme, inheritedThemes);
    const themes = getTheme(themeName);
    const themeClassnames = themes.map((theme) => `uitk-${theme.name}`);

    return {
      themes,
      themeClassnames
    }
  }, [theme]);
```

Which is using a simple hook to give meaning to the need for a stable reference:

```typescript
function useStaticValue(input: string | string[]): string[] {
  const stringify = JSON.stringify(input);
  return useMemo(() => JSON.parse(stringify), stringify);
}
```

joshwooding (approved):

PR #603, opened 2022-11-01
ToolkitProvider docs

alycrys:


joshwooding (changes requested):
The custom element is no longer used. I believe main has updates in the docs to accommodate this. Maybe they need to be moved here?

alycrys:
Replaced by https://github.com/jpmorganchase/uitk/pull/658

PR #607, opened 2022-11-02
Remove uitk-theme custom element

ImmanuelBaskaran:
Remove uitk-theme custom element and replace it with div element. The custom element may cause confusion during ADA checks and server side rendering.


joshwooding (approved):

packages/core/src/__tests__/__e2e__/toolkit-provider/ToolkitProvider.cy.tsx
```diff
@@ -43,16 +41,16 @@ const TestComponent = ({
...
       cy.mount(
         <ToolkitProvider>
           <TestComponent />
         </ToolkitProvider>
       );
 
-      cy.get("uitk-theme")
+      cy.get("div.uitk-theme")
         .should("have.length", 2)
-        .and("have.attr", "class", "uitk-light uitk-density-medium");
```
joshwooding, line 55: I think we should still check for the light theme here?

joshwooding (changes requested):
Looks good, just left one comment

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -25,31 +23,13 @@ export interface ToolkitContextProps {
...
 
 export const BreakpointContext =
   createContext<Breakpoints>(DEFAULT_BREAKPOINTS);
 
-export const ToolkitContext = createContext<ToolkitContextProps>({
-  density: undefined,
-  themes: [],
-  breakpoints: DEFAULT_BREAKPOINTS,
-});
-
```
ZarkoPernar, line 52: Why are we removing this?

ImmanuelBaskaran: That was an accidental leftover from splitting the ToolkitContext into separate contexts for Density, Breakpoints and theme ( see PR #602  )

ZarkoPernar (approved):

joshwooding (approved):

PR #631, opened 2022-11-07
Remove backwards compatibility and some custom style stories

joshwooding:
This PR is to remove the backward compatibility functionality since we will not deliver this as part of the current work. We can always add it back later using this PR but removing it for new keeps the code clean and prevents us from doing more work the necessary to keep it updated.


packages/lab/stories/metric.stories.tsx
```diff
@@ -247,23 +246,3 @@ export const WithCustomIndicator: ComponentStory<typeof Metric> = () => (
...
-    <Metric direction="up" showIndicator>
-      <MetricHeader subtitle="Total Value" title="Revenue YTD" />
-      <MetricContent value="$801.9B" />
-    </Metric>
-    <Metric direction="down" showIndicator>
-      <MetricHeader subtitle="Total Value" title="Revenue YTD" />
-      <MetricContent value="$801.9B" />
-    </Metric>
-  </div>
-);
```
ZarkoPernar, line 269: This is backwards compat?

ZarkoPernar (approved):
Looking G

packages/lab/stories/metric.stories.tsx
joshwooding: It's a custom override since we're not focusing on for V1 it's something that I think could be removed. Happy to add it back once we want to focus on customisability.

Fercas123 (approved):

PR #638, opened 2022-11-09
Amend site folder structure and add placeholder pages

JoanaMMoreira:
- Create docs subfolders for `components` and `getting-started`
- Add placeholder pages for both
- Amend docusaurus config so docs route points to `/`


EleanorHaproff (commented):
Great to see the urls and folder structures much improved 🎉 

We have three 'Getting Started' docs now - an 'index', 'dev' and 'design' one - is this correct or do we just want the dev and design ones?

JoanaMMoreira:
> Great to see the urls and folder structures much improved 🎉
> 
> We have three 'Getting Started' docs now - an 'index', 'dev' and 'design' one - is this correct or do we just want the dev and design ones?

I suppose it depends on the amount of content we'll have, @james-nash what do you think?

site/docs/components/card.mdx
```diff
@@ -0,0 +1,12 @@
+---
```
james-nash, line 1: Same as for button, I'd remove this page for now.

site/docs/getting-started/design.mdx
```diff
@@ -0,0 +1,11 @@
+---
```
james-nash, line 1: I was tempted to suggest removing this for the same reasons as the component pages. However, let's keep the 3 pages under getting started as it could be useful for our design & content colleagues to see how things behave when we have a docs section like this.

james-nash (changes requested):

james-nash (approved):
Nicely done. Ship it! 🛥️

PR #639, opened 2022-11-09
Fix theme editor build

origami-z:
Another half day wasted because of modular....

- Added additional modular patch and `tsconfig` change, so theme-editor-app could run and build
- Added a build job in GitHub action to ensure we don't break it again without noticing
- Updated `dependencies` in `package.json` to fix modular warning look like below 

> [modular] Package @jpmorganchase/uitk-theme imported in @jpmorganchase/theme-editor-app source but not found in package dependencies or hoisted dependencies - this will prevent you from successfully building, starting or moving esm-views and will cause an error in the next release of modular

### DIFF
--- a/playground/theme-editor-app/package.json
+++ b/playground/theme-editor-app/package.json
@@ -6,9 +6,9 @@
     "type": "app"
   },
   "dependencies": {
+    "@jpmorganchase/theme-editor": "*",
     "@jpmorganchase/uitk-core": "*",
     "@jpmorganchase/uitk-lab": "*",
-    "@jpmorganchase/uitk-theme": "*",
-    "@jpmorganchase/theme-editor": ">=1.0.0"
+    "@jpmorganchase/uitk-theme": "*"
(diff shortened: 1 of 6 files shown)


playground/theme-editor-app/package.json
```diff
@@ -6,6 +6,9 @@
     "type": "app"
   },
   "dependencies": {
-    "@jpmorganchase/theme-editor": ">=1.0.0"
+    "@jpmorganchase/theme-editor": "*",
+    "@jpmorganchase/uitk-core": "*",
+    "@jpmorganchase/uitk-lab": "*",
+    "@jpmorganchase/uitk-theme": "*"
```
libertymayc, line 12: may also need icons?

origami-z: That's not being directly used in the app, probably in the `theme-editor` package

libertymayc (approved):

PR #645, opened 2022-11-10
Implemented Theme modes as data attributes

ImmanuelBaskaran:
Adds the concept of theme modes to the toolkit provider. Instead of having two separate themes such as "uitk-light" and "uitk-dark", The toolkit provider will take one theme and one mode as props. To allow another theme to extend the provided theme "uitk-theme"; the class will always be applied to the toolkit provider in addition to the theme name passed.

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -13,16 +13,16 @@ import { DocsContainer } from "@storybook/addon-docs";
 
 const densities = ["touch", "low", "medium", "high"];
 const DEFAULT_DENSITY = "medium";
-const DEFAULT_THEME = "light";
+const DEFAULT_MODE = "light";
 
 export const globalTypes: GlobalTypes = {
-  theme: {
-    name: "Theme",
+  mode: {
+    name: "Mode",
     description: "Set the color theme",
-    defaultValue: DEFAULT_THEME,
+    defaultValue: DEFAULT_MODE,
     toolbar: {
-      title: "Theme",
-      // show the theme name once selected in the toolbar
+      title: "Mode",
+      // show the mode name once selected in the toolbar
       dynamicTitle: true,
       items: [
         { value: "light", right: "⚪", title: "Light" },
@@ -114,7 +114,7 @@ export const parameters: Parameters = {
     }: ComponentProps<typeof DocsContainer> & { children?: ReactNode }) => (
       // @ts-ignore DocsContainer does not support React18 types
       <DocsContainer context={context}>
-        <ToolkitProvider theme={context.globals?.theme}>
+        <ToolkitProvider mode={context.globals?.mode}>
           {children}
         </ToolkitProvider>
       </DocsContainer>
--- a/cypress/support/commands.tsx
+++ b/cypress/support/commands.tsx
@@ -1,6 +1,6 @@
 import "@testing-library/cypress/add-commands";
 import { mount as cypressMount } from "cypress/react18";
-import type { MountReturn, MountOptions } from "cypress/react";
+import type { MountOptions, MountReturn } from "cypress/react";
 import "cypress-axe";
 import { Options } from "cypress-axe";
 import { PerformanceResult, PerformanceTester } from "./PerformanceTester";
@@ -104,7 +104,7 @@ Cypress.Commands.add("mount", function (children, options) {
   };
 
   return cypressMount(
-    <ToolkitProvider density={this.density} theme={this.theme}>
+    <ToolkitProvider density={this.density} mode={this.theme}>
       {children}
       <AnnouncementListener onAnnouncement={handleAnnouncement} />
     </ToolkitProvider>,
--- a/docs/decorators/withTheme.tsx
+++ b/docs/decorators/withTheme.tsx
@@ -1,8 +1,7 @@
 import { DecoratorFn } from "@storybook/react";
-import { Panel, ToolkitProvider, useTheme } from "@jpmorganchase/uitk-core";
-import { useEffect } from "react";
+import { Panel, ToolkitProvider } from "@jpmorganchase/uitk-core";
 
-const THEMES = ["light", "dark"];
+const MODES = ["light", "dark"];
 
 // Modified from storybook background addon
 // https://github.com/storybookjs/storybook/blob/next/addons/backgrounds/src/helpers/index.ts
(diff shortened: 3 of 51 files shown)


libertymayc (commented):
looking good :)

.storybook/preview.tsx
```diff
@@ -13,16 +13,16 @@ import { DocsContainer } from "@storybook/addon-docs";
...
 const DEFAULT_DENSITY = "medium";
-const DEFAULT_THEME = "light";
+const DEFAULT_MODE = "light";
 
 export const globalTypes: GlobalTypes = {
-  theme: {
-    name: "Theme",
+  mode: {
+    name: "Mode",
     description: "Set the color theme",
```
joshwooding, line 21:
```suggestion
    description: "Set the theme mode",
```

cypress/support/commands.tsx
```diff
@@ -104,7 +104,7 @@ Cypress.Commands.add("mount", function (children, options) {
   };
 
   return cypressMount(
-    <ToolkitProvider density={this.density} theme={this.theme}>
+    <ToolkitProvider density={this.density} mode={this.theme}>
```
joshwooding, line 107:
```suggestion
    <ToolkitProvider density={this.density} mode={this.mode}>
```
See line 65~ That will need to be updated too

packages/theme/css/global.css
```diff
@@ -17,15 +17,15 @@ body {
   font-family: "uitk-sans";
 }
 
-.uitk-theme {
+.uitk-provider {
```
joshwooding, line 20: Thoughts on moving this to ToolkitProvider.css and unifying it with how other components are treated.

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -191,12 +196,16 @@ export function ToolkitProvider({
   }
 }
 
-export const useTheme = (): Theme[] => {
-  return useContext(ThemeContext);
+export const useTheme = (): ThemeNameType => {
```
joshwooding, line 199: Is it worth having separate hooks here? They rely on the same context provider so will cause re-render together.

We could return an object from useTheme with { theme, mode }

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -100,63 +115,52 @@ type ToolkitProviderProps =
...
-};
-
-const getThemeClassName = (themes: Theme[]): string[] =>
-  themes.map((theme) => `uitk-${theme.name}`);
-
 export function ToolkitProvider({
   applyClassesTo,
   children,
   density: densityProp,
   theme: themesProp,
```
alycrys, line 122: as `theme` is a string, shall we use singular for `themesProp`?

joshwooding (changes requested):
Needs a changeset. Otherwise looks good

packages/core/src/theme/Theme.ts
```diff
@@ -28,25 +28,26 @@ export class Theme {
     this.id = id;
     this.name = name;
   }
+}
 
-  getCharacteristicValue(
-    characteristicName: characteristic,
```
alycrys, line 33: Don't we need to pass the mode also here?

packages/core/src/theme/Theme.ts
```diff
@@ -28,25 +28,26 @@ export class Theme {
...
+      return variableValue;
     }
-    return null;
   }
-}
+  return null;
+};
 
 const lightTheme = new Theme("uitk-light", "light");
 const darkTheme = new Theme("uitk-dark", "dark");
```
alycrys, line 53: I believe we should call these modes.
I feel we might need to change few things in here also, as we now have only one theme and two extra modes...??

docs/decorators/withTheme.tsx
```diff
@@ -47,16 +52,16 @@ function SetBackground({ viewMode, id }: { viewMode: string; id: string }) {
...
   return null;
 }
 
 export const withTheme: DecoratorFn = (StoryFn, context) => {
-  const { density, theme } = context.globals;
+  const { density, mode } = context.globals;
 
-  if (theme === "side-by-side" || theme === "stacked") {
-    const isStacked = theme === "stacked";
+  if (mode === "side-by-side" || mode === "stacked") {
```
alycrys, line 63: I thought mode can only be 'light' or 'dark'. Is this change intended?

packages/core/src/theme/Theme.ts
joshwooding: I actually think we should consider removing all the code from here to the end of the file

docs/decorators/withTheme.tsx
joshwooding: This is a storybook only thing. I don't really think the name here actually matters

packages/theme/css/global.css
```diff
@@ -17,15 +17,11 @@ body {
...
-  display: contents;
-}
-
-.uitk-light {
+.uitk-theme[data-mode^="light"] {
   color-scheme: light;
 }
 
-.uitk-dark {
+.uitk-theme[mode="dark"] {
```
alycrys, line 24: typo, missing `data-`

packages/theme/stories/introduction.stories.mdx
```diff
@@ -143,12 +134,12 @@ Consider the example given above:
 In this case, the scope is `:root`. This is equivalent to `html` so variables declared in this way are global. They can be referenced in any CSS that is applied to the page. We constrain the scope of a variable using a CSS selector, exactly as we declare regular CSS rules:
 
 ```css
-.uitk-light {
+.uitk-theme[data-mode^="light"] {
```
alycrys, line 137: Mode should always be 'light' or 'dark' so we don't need the 'starts with '(`^`) attribute selector. Unless I'm missing something

tooling/theme-editor/src/helpers/parseToCss.ts
```diff
@@ -59,7 +59,7 @@ function transformToCSS(patternJsonByScope) {
   patternJsonByScope.forEach((element) => {
     let selector;
     if (element.scope === "mode-all") {
-      selector = `.uitk-light, .uitk-dark`;
+      selector = `.uitk-theme[data-mode^="light"], .uitk-theme[data-mode^="dark"]`;
```
alycrys, line 62: Can this be just `selector = '.uitk-theme'`?

packages/core/src/theme/Theme.ts
ImmanuelBaskaran: The characteristics themselves are not mode dependent.

packages/core/src/toolkit-provider/ToolkitProvider.tsx
```diff
@@ -9,35 +9,44 @@ import React, {
...
+export interface ThemeContextProps {
+  theme: ThemeName;
+  mode: Mode;
+}
 
 export const DensityContext = createContext<Density>(DEFAULT_DENSITY);
 
-export const ThemeContext = createContext<Theme[]>([]);
+export const ThemeContext = createContext<ThemeContextProps>({
+  theme: "",
```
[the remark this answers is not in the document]
ImmanuelBaskaran: The old logic for determining if a toolkit provider was the root provider depended on the fact that theme could be an empty array. Without using an undefined value as the context default, the next best option was an empty string.

joshwooding (approved):

PR #661, opened 2022-11-15
Text component

alycrys:


packages/core/src/text/Display.tsx
```diff
@@ -0,0 +1,51 @@
+import { makePrefixer } from "@jpmorganchase/uitk-core";
```
joshwooding, line 1:
```suggestion
import { makePrefixer } from "../utils";
```

alycrys: Thank you! It was in front of my eyes, I just couldn't see it

packages/lab/stories/link.stories.tsx
```diff
@@ -51,7 +51,10 @@ export const TargetBlankNoIcon: ComponentStory<typeof Link> = () => {
 export const WithTooltip: ComponentStory<typeof Link> = () => {
```
joshwooding, line 51: We should hide this story

packages/theme/css/theme.css
```diff
@@ -35,3 +35,8 @@
 @import url(characteristics/status.css);
 @import url(characteristics/taggable.css);
 @import url(characteristics/text.css);
+
```
joshwooding, line 38: Just curious, why was this put here instead of in global?

joshwooding: Global is included when you import the theme anyway

joshwooding:
Awesome work @alycrys 🎉

packages/theme/css/theme.css
alycrys: There was a discussion about having it as part of theme, but I think the point was more like having (maybe) some global styling per theme. Not sure if that's something we want to do. But either way it should be done in a different PR.

joshwooding (approved):

sion-docs:
The template needs to be applied to this documentation - the following sections/headings are missing:
- Configuring Text
- Accessibility
- Help and support

PR #676, opened 2022-11-17
Add polymorphic behaviour on FlexLayout and FlexItem

Fercas123:


.changeset/stale-mirrors-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@jpmorganchase/uitk-core": minor
+---
+
+Add "as" prop to `FlexLayout` and `FlexItem`.
```
JoanaMMoreira, line 5: Should we add a quick note about how to use the prop? E.g. Add "as" prop to `FlexLayout` and `FlexItem` to allow customisation of the html tag rendered by the component

JoanaMMoreira (approved):

.changeset/stale-mirrors-explode.md
Fercas123: I think we will probably need to add this to the docs of all the layouts with polymorphic behaviour and pull the example in, coming up in a subsequent PR

PR #691, opened 2022-11-21
Update homepage, theme and navbar styles

JoanaMMoreira:


EleanorHaproff (dismissed):
Looks great - the animations are 🚀 🤩

site/src/components/card/Card.module.css
```diff
@@ -0,0 +1,60 @@
+.card {
+  --card-background: #10212b;
+  --ifm-link-decoration: underline;
+}
+
+.card {
+  background-color: var(--card-background);
+  padding: calc(var(--uitk-size-unit) * 2) calc(var(--uitk-size-unit) * 2) 0 calc(var(--uitk-size-unit) * 2);
+  color: var(--uitk-color-white);
```
libertymayc, line 9: are we able to use characteristics here e.g. container-background?

libertymayc (approved):

site/src/components/card/Card.tsx
```diff
@@ -0,0 +1,82 @@
...
+import styles from "./Card.module.css";
+
+type CardIconName = "pencil" | "code" | "arrows";
+
+type CardIconsType = {
+  name: CardIconName;
+  icon: JSX.Element;
+}[];
+
+const cardIcons: CardIconsType = [
```
james-nash, line 18: Not a blocker, but perhaps we should open this up in future PRs. Perhaps we could update the props so that you can pass in any CSS color value (which would include things like `var(--uitk-color-...)`) for the keyline and any React component for the image part.

That might make the card more re-usable in other areas of the website

site/src/pages/index.module.css
```diff
@@ -3,24 +3,62 @@
...
+    flex: 1;
   }
 }
 
-.buttons {
-  display: flex;
-  align-items: center;
-  justify-content: center;
-  gap: var(--uitk-size-unit);
+@media screen and (max-width: 960px) {
```
james-nash, line 59: Is this really meant to be `960px` can this be moved up in the previous `max-width: 996px` media query?

james-nash (dismissed):
Looking great!

I've added a suggestion and question to bear in mind for subsequent updates, but happy to approve this for now.

site/src/pages/index.module.css
JoanaMMoreira: Good spot, I've moved these styles up

feliciackh (approved):
🚀 looking great!

EleanorHaproff (approved):

site/src/components/card/Card.module.css
```diff
@@ -0,0 +1,60 @@
+.card {
+  --card-background: #10212b;
```
james-nash, line 2: If this color value doesn't exist in Odyssey's design tokens, we should maybe raise that with Shey. I'd be keen to understand if there is similar value from the design system we could use (or if perhaps the design system's palette needs to be extended to include this value).

In any case, it's fine for now :-)

james-nash (approved):
Nice work!

Ship it!

site/src/components/card/Card.module.css
JoanaMMoreira: cc @shey-v this is in regards to the content blocks background colour

PR #706, opened 2022-11-23
Change website components density depending on breakpoint

JoanaMMoreira:
The following densities should be applied:

- `xl` and `lg` breakpoints: low density
- `md`, `sm` and `xs` breakpoints: touch density


site/src/theme/Root.tsx
```diff
@@ -1,15 +1,25 @@
...
-const density = "low";
+const DensityProvider = ({ children }) => {
+  const viewport = useCurrentBreakpoint();
+
+  const density = useMemo(
+    () => (viewport === "xl" || viewport === "lg" ? "low" : "touch"),
+    [viewport]
+  );
+
+  return <ToolkitProvider density={density}>{children}</ToolkitProvider>;
```
james-nash, line 16: Could we not move this logic into the `Root` component below and just apply the `density` prop to its `<ToolkitProvider>`? Or is there some kind of benefit to having nested `<ToolkitProvider>`s - one to set the mode and the other to set the density?

JoanaMMoreira: That would be ideal but the reason we can't do that is because we're accessing the ToolkitProvider's context to check the current viewport and we can only do this within its children. This wouldn't be required if we weren't using Odyssey's breakpoints and we wrote our own for the website but I think it would be good to keep everything consistent

james-nash: Ah, understood. Ok, will approve this PR then.

(I wonder if we can feed that back the rest of the team. It would be nice if a future version of the Toolkit provider could let us achieve this without needing to use 2 of them)

JoanaMMoreira: Sure, we briefly discussed this on the dev call yesterday @joshwooding would it make sense to log an issue for this?

joshwooding: Potentially, there's a lot here that can change to make this possible.

james-nash: Off the top of my head, might it be feasible to allow the ToolkitProvider's `density` prop to accept either a value as it does now or a function? That function would then get the current breakpoint as an argument and would return a density value.

EleanorHaproff (approved):

site/src/theme/Root.tsx
joshwooding: That depends if we see this as a valid use-case, so we would need to find out if design approves.

JoanaMMoreira: Makes sense, I'll mark this as resolved for now but let's discuss this separately

PR #710, opened 2022-11-24
Site accordian

EleanorHaproff:
- Added homepage accordion and graphic
- The whole page will be reviewed with @shey-v  once we merge this in


site/src/components/card/Card.module.css
```diff
@@ -1,5 +1,4 @@
 .card {
```
james-nash, line 1: Can we combine this `.card` block with the one immediately below?

site/src/css/custom.css
```diff
@@ -62,6 +63,23 @@
...
+  right: 0;
+}
+
+.uitkAccordionSummary .uitkAccordionSummary-icon:hover {
+  background: var(--accordion-summary-background-hover);
+}
+
+.uitkAccordionDetails .uitkAccordionDetails-content {
+  padding-left: 0;
+}
```
james-nash, line 79: Would it not make more sense to move these rules into `Accordion.module.css`?

Feels a bit risky to globally override them. What happens if we ever want to use a "vanilla" Salt accordion elsewhere on the site?

site/src/pages/index.tsx
```diff
@@ -12,6 +12,7 @@ import { Button } from "@jpmorganchase/uitk-core";
 import styles from "./index.module.css";
 import Features from "../components/features/Features";
 import Card, { CardProps } from "../components/card/Card";
+import Accordion from "../components/accordion/Accordion";
```
james-nash, line 15: Given that our accordion "component" is very specific to the homepage - both in terms of its layout, illustration and the fact the text content is embedded in that component, it might make sense to move those files here under `pages/` so that our `components/` dir is only used for reusable things. That way all the files that make up the content of the homepage are also closer together.

[Docusaurus will ignore files and folders prefixed with an underscore](https://docusaurus.io/docs/creating-pages#routing), so perhaps we could create `site/src/pages/_index/` and move those files into there.

james-nash (changes requested):
It's looking pretty good in the browsers and the content & functionality's all there. However, I think the CSS can be simplified and some minor layout issues need to be fixed too.

site/src/components/accordion/Accordion.module.css
```diff
@@ -0,0 +1,94 @@
...
+}
+
+.accordion .summary {
+  background-color: transparent;
+  padding-left: 0;
+}
+
+.accordion .summary:hover {
+  background: none;
+}
```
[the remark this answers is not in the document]
JoanaMMoreira: This was a design requirement (having the highlight around the chevron only) but I'm happy for us to discuss with Shey

site/src/css/custom.css
JoanaMMoreira: The issue with moving these across to `Accordion.module.css` is that we can't access the icon class for example as it's not defined within that module

site/src/components/accordion/Accordion.module.css
```diff
@@ -0,0 +1,94 @@
...
+  min-height: 70vh;
+}
+
+.accordion {
+  flex: 1;
+  padding: calc(var(--uitk-size-unit) * 2) 0;
+}
+
+.accordion .summary {
+  background-color: transparent;
```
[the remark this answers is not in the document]
JoanaMMoreira: (see comment below)

site/src/components/card/Card.module.css
JoanaMMoreira: We definitely can but I believe the general css variables convention is to declare them in a separate block

site/src/pages/index.tsx
JoanaMMoreira: Makes sense, I've moved both `Accordion` and `Features`

site/src/components/card/Card.module.css
james-nash: Ah Ok. I wasn't aware of that convention. In that case leave it as is.

site/src/css/custom.css
james-nash: This is using CSS Modules, right?

In that case, I think you can [use `:global(...)` to use "global" class names](https://github.com/css-modules/css-modules#exceptions) which CSS modules won't mangle then.

E.g.

```css
/* Inside Accordion.module.css */

.summary :global(.uitkAccordionSummary-icon) {
  /* ... */
}
```

site/src/components/accordion/Accordion.module.css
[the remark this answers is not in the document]
james-nash: Oh, ok. Yeah, let's discuss with Shey.

I think that's a bit misleading as it might create the false impression that only the chevron is clickable. Also, when the chevron rotates as the accordion opens/closes, the highlight does too which looks weird. :-P

site/src/css/custom.css
JoanaMMoreira: Ohh that works :) thank you!

site/src/components/accordion/Accordion.module.css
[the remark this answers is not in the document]
JoanaMMoreira: Agreed, it does look a bit funny!

james-nash (dismissed):
Niiiiice! 😎

james-nash (approved):

PR #715, opened 2022-11-24
Add "scope" option to ToolkitProvider

ImmanuelBaskaran:
Added "scope" as an option for "applyClassesTo" prop, if this is passed into the `ToolkitProvider`, the provider will create a div element and apply the classes to that. The root level toolkit provider now applies the classes to the html element while nested toolkit providers will behave as if "scope" was passed as a prop to "applyClassesTo"


packages/core/stories/toolkit-provider.doc.stories.mdx
```diff
@@ -220,3 +220,32 @@ Then we would see the following HTML.
   </body>
 </html>
 ```
+
+By default, the root level `ToolkitProvider` will apply the theme and density classes and the theme mode data attribute to the `<html>` element and any nested `ToolkitProvider` will create a div with the `uitk-provider` class
+which the classes and data attributes are added,
+if you want the root level `ToolkitProvider` to create a div instead, then pass the value "scope" in the `applyClassesTo`
```
libertymayc, line 226: not a major issue, but to me this is a bit wordy, could split into 2 sentences so it's easier to consume

libertymayc: `which the classes and data attributes are added` probably is a typo

packages/core/src/__tests__/__e2e__/toolkit-provider/ToolkitProvider.cy.tsx
```diff
@@ -146,7 +146,7 @@ describe("Given a ToolkitProvider", () => {
       );
 
       // cy.mount adds a ToolkitProvider
```
libertymayc, line 148: should this comment be deleted

PR #724, opened 2022-11-28
Add default width and height to grid

origami-z:
Current Grid doesn't have any size, so using `<Grid rowData={rowData} />` without any class or style will result in all rows not visible, which is super confusing. Using total width and height variable as default value to accommodate this.


origami-z:
Closing, i don't have time to fix the failing tests because of the change. Will leave with the team to fix the confusing size issue, at least through documentation hopefully.

PR #728, opened 2022-11-28
Homepage refinements

JoanaMMoreira:
- App Header
  - Fix LD key line.
  - Ensure height of orange key line is 3px.
  - App header should be 76px height for TD & 60px for LD.
  - Getting started URL to be changed from ‘dev’ to ‘develop’ (on the Getting started page).

- Footer
  - Cookies policy to be updated to Contact us - links to “Support and contributions” page. 
  - The divider lines need to be in the secondary text color. 

- What to expect block
  - Remove spacing between two paragraphs so the bullets flow
  - Wrap the content / copy so it flows better (i.e. not aligning with the tick but with the copy itself) in smaller viewports.

- Cards
  - Update background color 
  - Remove extra padding 
  - Remove animation on the icons (for now) to be replaced with another. 
  - Change body text in cards to secondary color
  - Keep links white (primary text color) for stand out

- Accordion
  - Remove background on hover completely
  - Update line animation to 5 secs
  - Change key lines (separators to white at 10% opacity). 
  - Amend the accordion chevrons to 24px across LD (base height x2).


james-nash (approved):
Great stuff! 🤘

site/src/components/card/Card.tsx
```diff
@@ -28,26 +29,28 @@ const Card = ({
   url,
   linkText,
   keylineColor,
-  theme = "dark",
+  disableKeylineAnimation = false,
```
EleanorHaproff, line 32: Might be easier to read `keyLineAnimation = 'true'` so you don't have to do `!disableKeylineAnimation` on line 52 (instead just `keylineAnimation`), but not a blocker

EleanorHaproff (approved):
Looking fab 💅

site/src/components/card/Card.tsx
JoanaMMoreira: Makes sense, I'll amend this on my upcoming PR

PR #733, opened 2022-11-28
Getting started overview page

JoanaMMoreira:
Initial implementation of the getting started overview page


site/src/components/card/Card.module.css
```diff
@@ -1,28 +1,38 @@
 .card {
```
libertymayc, line 1: I think on hover there's meant to be a drop shadow

site/src/css/custom.css
```diff
@@ -58,10 +66,108 @@
 }
 
 .navbar__link:hover:not(.navbar__link--active):not(.header-github-link) {
-  background-color: var(--uitk-color-gray-600);
+  background-color: var(--uitk-palette-navigate-background-hover-high);
   border-bottom: 2px solid var(--uitk-navigable-indicator-hover);
 }
```
libertymayc, line 72: --ifm-toc-border-color should use --uitk-separable-tertiary-borderColor (the RHS border of the sidebar)

libertymayc (commented):
LGTM! a few styling comments & the table of contents needs styling but we can do that in a different pr

site/src/css/custom.css
```diff
@@ -58,10 +66,108 @@
...
+  padding: 0;
+}
+
+.theme-doc-sidebar-container .menu__link {
+  border-radius: 0;
+}
+
+.theme-doc-sidebar-container .menu__link--active {
+  font-weight: var(--uitk-navigable-fontWeight-active);
+  border-left: 2px solid var(--uitk-navigable-indicator-active);
```
libertymayc, line 87: i think this is 4px?

libertymayc:
the appheader height or something along those lines needs looking at, seems off and so the indicator overlays the content

Screenshot 2022-11-29 at 09 41 19

site/src/components/card/Card.module.css
JoanaMMoreira: Yes that's true, the reason I didn't add that here is because there is a bit more work that needs to be done (we've discussed having the whole card clickable on bigger screens and not on mobile for example) so I'll raise a separate ticket if that's ok

JoanaMMoreira:
> the appheader height or something along those lines needs looking at, seems off and so the indicator overlays the content
> 
> Screenshot 2022-11-29 at 09 41 19

It will be fixed by another PR :) https://github.com/jpmorganchase/uitk/pull/728

site/src/components/card/Card.module.css
james-nash: Makes sense. Let's get that issue on the backlog and aim to do it once this PR is merged.

Viewing this PR's preview build in the browser has further convinced me that the whole card needs to be clickable! It feels really unintuitive to me to have those enormous cards but I need to click the tiny link text at the bottom to actually go somewhere :-)

JoanaMMoreira: Yes it's definitely more noticeable here! I've raised an issue https://github.com/jpmorganchase/uitk/issues/739

libertymayc (approved):

james-nash (approved):
Lookin' good. Let's get this merged.

(Aside, @JoanaMMoreira and I briefly discussed revisiting how we're doing light/dark portions of the site in the future to see if we can use `<ToolkitProvider>`s instead. However, we agreed that can be a future tidy-up thing that isn't needed right now)

PR #740, opened 2022-11-29
Add issue templates & forms

james-nash:
Add issue templates and forms to make it easier for people to raise good quality bug reports and feature requests.


james-nash:
@joshwooding This is my first stab at creating a Github issue form. As discussed, there's no way to test or view it other than merging it into `main`. Could we do that, so I can see how it looks and make necessary amends?

PR #741, opened 2022-11-29
Use real instead of fake focus on table cells

ZarkoPernar:


packages/grid/src/BaseCell.css
```diff
@@ -54,7 +56,7 @@
   border-color: var(--grid-editableCell-borderColor);
 }
 
-.uitkGridBaseCell-focused {
+.uitkGridBaseCell:focus .uitkGridBaseCell-valueContainer {
```
joshwooding, line 59: You probably want focus-visible here

ZarkoPernar: Good idea

ararus (approved):

ararus (dismissed):

joshwooding (changes requested):
We need to add a changeset

ararus:
> 

Changeset added

ararus (approved):

joshwooding (approved):

PR #750, opened 2022-11-30
fixes YAML schema issues

james-nash:


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,65 @@
+name: Bug report
```
joshwooding, line 1: There's now a bug_report.yml and bug_report.yaml

james-nash: sorry about that. Sorted it now

PR #756, opened 2022-11-30
Make site cards clickable and general styling tweaks

JoanaMMoreira:
- Change cards to be clickable with subtle animations
- General styling tweaks including spacing adjustments on small viewports


site/src/components/card/Card.module.css
```diff
@@ -19,13 +24,27 @@
   flex: 1;
 }
 
+.card:hover {
+  box-shadow: var(--uitk-overlayable-shadow-borderRegion);
```
libertymayc, line 28: a design decision, but I feel --uitk-overlayable-shadow-hover would be more prominent here

libertymayc (approved):

site/src/components/card/Card.module.css
JoanaMMoreira: we've tried that one and decided this one is a bit more subtle but I'll leave it to @shey-v :)

libertymayc (dismissed):

site/src/pages/index.module.css
```diff
@@ -58,5 +62,14 @@
...
     flex-direction: column;
+    padding: calc(var(--uitk-size-unit) * 3) calc(var(--uitk-size-unit) * 2);
+    gap: calc(var(--uitk-size-unit) * 2);
+  }
+}
+
+@media screen and (max-width: 1280px) {
+  .heroTitle {
+    --ifm-heading-line-height: var(--uitk-text-h1-lineHeight);
+    --ifm-h1-font-size: var(--uitk-text-h1-fontSize);
```
EleanorHaproff, line 73: You could move this above the 996px query just for consistency / readability but not a biggie

EleanorHaproff (approved):
Looks great! Great to have the whole card clickable, and it's much more responsive now 🎉

site/src/pages/index.module.css
JoanaMMoreira: I've tried this but I think it may be best to leave it as is because otherwise the hero title looks quite massive when we change to touch density
Screenshot 2022-12-05 at 10 17 42

libertymayc (approved):

PR #761, opened 2022-12-01
Add docs in Button around when to use Link

Fercas123:
add one missing paragraph in the button docs about when to use a link instead


packages/core/stories/button.doc.stories.mdx
```diff
@@ -30,6 +30,8 @@ A Button executes an action when clicked. Button labels should use only one or t
 - **Primary** is the most commonly-used button.
 - **Secondary** performs an action less likely to be triggered.
 
+If the use case requires the user to navigate to a new page or window rather than trigger a function, use a `Link` component instead. The exception to this behaviour is if the navigation is the core action on the page. In this case, a `CTA Button` should be used to trigger the navigation.
```
joshwooding, line 33: I would pass this by @maj-stella too

sion-docs (approved):

packages/core/stories/button.doc.stories.mdx
sion-docs, line 33: This would be best placed under the suggested 'Do not use when...' heading from the template, but that requires a 'Use when...' heading which we don't have here.

Fercas123: @sion-docs i was wondering what to do in this case, given the description of the button is kind of the "Use when" case

Fercas123, line 33:
```suggestion

##### Use when
- You want to indicate actions that the user can take.
- You need to highlight navigation as the core action on the page. In this case, a `CTA Button` should be used to trigger the navigation.

##### Don't use when

The use case requires the user to navigate to a new page or window rather than trigger a function, use a `Link` component instead.
```

Fercas123: @sion-docs @joshwooding does that look better?

joshwooding: Looks good 👍🏽 I would double check "You need to highlight navigation as the core action on the page. In this case, a `CTA Button` should be used to trigger the navigation." with @maj-stella

sion-docs: That's great @Fercas123

Fercas123: @joshwooding I updated it to the previous wording that @maj-stella  had already reviewed and approved.

joshwooding (approved):

PR #767, opened 2022-12-01
Selectable Figma cross check

libertymayc:
Some of the palette tokens dont match Figma but are aligned with updates in #674


libertymayc:
I've put this in draft until https://github.com/jpmorganchase/uitk/pull/674 complete

PR #768, opened 2022-12-01
Getting started design

EleanorHaproff:


site/src/css/custom.css
```diff
@@ -8,6 +8,7 @@
   --ifm-color-primary: var(--uitk-text-primary-foreground);
   --ifm-font-family-base: var(--uitk-typography-fontFamily);
   --card-background: #10212b;
+  --image-border: #c5c9d0;
```
JoanaMMoreira, line 11: This is more of a design question but is there any way we could use a design token here instead of a custom value?

site/docs/getting-started/DesignCards.module.css
```diff
@@ -0,0 +1,40 @@
...
+  flex: 1;
+}
+
+.imgContainer {
+  padding: var(--uitk-size-unit);
+  border: 1px solid var(--image-border);
+}
+
+.imgContainer > img {
+  width: 100%;
```
JoanaMMoreira, line 21: I'm not sure this is required as it makes the images quite large on desktop

site/docs/getting-started/DesignCards.tsx
```diff
@@ -0,0 +1,81 @@
...
+    <Card img={SelectIconImg} />
+  </div>
+);
+
+const Card = ({ img, content }: { img: string; content?: JSX.Element }) => (
+  <div className={styles.card}>
+    <div className={styles.imgContainer}>
+      <img src={img} />
+    </div>
+    <div className={styles.textContainer}>{content}</div>
```
JoanaMMoreira, line 77: Should we just add a `<p>` here since we're always using it?

site/docs/getting-started/IconsTable.module.css
```diff
@@ -0,0 +1,12 @@
+.colored {
+  color: var(--uitk-color-purple-500);
```
JoanaMMoreira, line 2:
```suggestion
  color: var(--site-tertiary-accent-purple);
```

site/docs/getting-started/IconsTable.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </tr>
+    </thead>
+    <tbody>
+      <tr>
+        <td>
+          1X <span className={styles.colored}>(default)</span>
+        </td>
+        <td>
+          <ToolkitProvider density="high">
+            <PrintIcon size={1} />
```
JoanaMMoreira, line 24: Is the icon size default 1? If so we can remove these

JoanaMMoreira:
Looking 🔥 !

site/docs/getting-started/DesignCards.tsx
EleanorHaproff: They already have <p> tags around them in the data. I'd put them as this instead of in the component as the long strings are really hard to read without them in the editor, plus we can handle if we need line breaks etc

site/docs/getting-started/DesignCards.module.css
james-nash: Perhaps using `max-width: 100%` instead would avoid that issue?

site/src/css/custom.css
libertymayc: seems this is container-border-primary in figma

james-nash (approved):

JoanaMMoreira (approved):

PR #787, opened 2022-12-05
Support and contributions

EleanorHaproff:


site/src/components/card/Card.tsx
```diff
@@ -7,27 +7,36 @@ import {
...
 
 export interface CardProps extends ComponentPropsWithoutRef<"div"> {
-  icon: JSX.Element;
+  icon?: JSX.Element;
+  inlineIcon?: JSX.Element;
   title: string;
   description: string;
   url: string;
-  footer: string;
+  footer: FooterProps;
```
JoanaMMoreira, line 28: I think in the future it would be good for this to accept either an object or a string but not a blocker :)

JoanaMMoreira (approved):
Looking great!

libertymayc (approved):

PR #798, opened 2022-12-06
Grid characteristics applied

ararus:
New characteristics applied to grid component


packages/grid/src/Grid.css
```diff
@@ -109,43 +46,48 @@
...
   --grid-headerRowSeparator-gap: var(--uitkGrid-headerRowSeparator-gap, var(--uitk-size-basis-unit));
   --grid-headerRowSeparator-height: var(--uitkGrid-headerRowSeparator-height, 1px);
-  --grid-groupHeaderRowSeparator-color: var(--uitkGrid-groupHeaderRowSeparator-color, var(--grid-groupHeaderRowSeparator-color-default));
+  --grid-groupHeaderRowSeparator-color: var(--uitkGrid-groupHeaderRowSeparator-color, var(--uitk-separable-tertiary-borderColor));
+
+  --grid-groupHeader-color: var(--uitkGrid-groupHeader-color, var(--uitk-text-secondary-foreground));
+  --grid-shadow-color: var(--uitkGrid-shadow-color, var(--uitk-overlayable-shadow-scroll));
+
+  --grid-columnDropTarget-color: var(--uitkGrid-columnDropTarget-color, var(--uitk-target-borderColor-hover));
+  --grid-columnDropTarget-borderWidth: var(--uitkGrid-columnDropTarget-borderWidth, var(--uitk-target-borderWidth-hover));
```
libertymayc, line 76: instead of --uitk-target-borderWidth-hover, can you use 0 var(--uitk-size-border)

packages/theme/css/characteristics/target.css
```diff
@@ -2,6 +2,8 @@
   --uitk-target-background-hover: var(--uitk-palette-interact-background-hover);
 
   --uitk-target-borderColor-hover: var(--uitk-palette-interact-border-hover);
+  --uitk-target-width-hover: 3px;
```
libertymayc, line 5: this token needs to be checked with design as other components need to be considered, i assume it isn't needed as a characteristic though as its component specific?

libertymayc: @pseys @dplsek

PR #809, opened 2022-12-06
Row selection checkbox gets focus

ararus:
Row selection cells show focus border around the interactive element within the cell (radio button or checkbox), not on the cell itself.

### DIFF
--- a/packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
+++ b/packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
@@ -183,18 +183,10 @@ describe("Grid", () => {
     cy.mount(<GridExample />);
 
     const checkCursorPos = (row: number, col: number) => {
-      // Column 0 has interactive elements in cells. We expect the interactive element to have focus.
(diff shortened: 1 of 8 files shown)


packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -183,9 +183,18 @@ describe("Grid", () => {
...
+      if (col === 0) {
+        cy.focused()
+          .parents("td")
+          .should("have.attr", "aria-rowindex", String(row + 1))
+          .should("have.attr", "aria-colindex", String(col + 1));
+      } else {
+        cy.focused()
+          .should("have.attr", "aria-rowindex", String(row + 1))
+          .should("have.attr", "aria-colindex", String(col + 1));
+      }
```
joshwooding, line 197:
```suggestion
        cy.focused()
          .closest("td")
          .should("have.attr", "aria-rowindex", String(row + 1))
          .should("have.attr", "aria-colindex", String(col + 1));
```

ZarkoPernar (approved):

PR #820, opened 2022-12-07
Components page 

JoanaMMoreira:
- Basic components table (sorting will be done on a separate PR)
- We still need to decide if we only want to display status icons on the mobile view but I thought it would be good to get this PR for now so the page is there


Fercas123 (approved):

james-nash (approved):
Great work!

PR #824, opened 2022-12-08
Updating component names 

sion-docs:
Removing use of pascalCase outside of code examples


libertymayc:
Is this still being worked on? 

Partly covered in https://github.com/jpmorganchase/salt-ds/pull/1078

PR #830, opened 2022-12-08
Add Characteristics Used on Text docs

alycrys:
Added CharacteristicUsage and CSSVariableTable to Text docs


libertymayc (dismissed):

packages/core/src/text/Text.css
```diff
@@ -1,7 +1,7 @@
 /* Main css class. Style for body text */
 .uitkText {
-  font-size: var(--uitk-text-fontSize);
-  line-height: var(--uitk-text-lineHeight);
+  font-size: var(--uitkText-fontSize, var(--uitk-text-fontSize));
```
joshwooding, line 3: Do we want public tokens for every variant? I think we should follow [Button](https://github.com/jpmorganchase/uitk/blob/main/packages/core/src/button/Button.css#L86) and have a "global" override that can be scoped with classes vs a per variant solution

packages/core/src/text/Text.css
```diff
@@ -2,6 +2,7 @@
 .uitkText {
   font-size: var(--uitk-text-fontSize);
   line-height: var(--uitk-text-lineHeight);
+  color: var(--text-color);
```
joshwooding, line 5:
```suggestion
  color: var(--uitkText-color, var(--text-color));
```
?

alycrys: Not sure this would be safe. This will potentially break the primary/secondary variants.

packages/core/src/text/Text.css
```diff
@@ -14,10 +15,10 @@
 }
 
 .uitkText-primary {
-  color: var(--uitk-text-primary-foreground);
+  --text-color: var(--uitkText-primary-foreground, var(--uitk-text-primary-foreground));
```
alycrys, line 18: @joshwooding how about this instead?

packages/core/src/text/Text.css
joshwooding: To me, the intention of these variables is to allow consumers to build on Text. They could still scope uitkText-color to the primary or secondary variant classes, but for users who want to just override they shouldn't have to override multiple tokens

libertymayc (approved):

PR #834, opened 2022-12-08
card title icon alignment fix

EleanorHaproff:


site/src/components/card/Card.module.css
```diff
@@ -13,8 +13,9 @@
   --card-icon-background: var(--uitk-color-gray-10);
 }
 
-.cardTitle {
-  --ifm-heading-margin-bottom: var(--uitk-size-unit);
+.cardTitle h2 {
+  margin-top: 0;
```
libertymayc, line 17: i think this aligns better set to var(--uitk-size-unit)?

JoanaMMoreira: I think it's fine given the flex gap we have

JoanaMMoreira (approved):

site/src/components/card/Card.module.css
JoanaMMoreira: Just discussed with @libertymayc , the contributions page cards still need a couple more tweaks - let's discuss on Monday :)

PR #838, opened 2022-12-08
Site table fixes

EleanorHaproff:
#### All tables
- [x] Make table headers sticky
- [x] Enable horizontal scroll (ie for icon / density table)
- [x] Border of header should be using separable / primary / border with shadow / black-L02
- [ ] Row and column dividers should use separable / tertiary / border
- [ ] Header shadow style disappears on mobile view

#### Components Page table
- [x] Hide header and status text on mobile 

#### Getting started pages
- [x] 'Getting started for designers' page - add Figma logo to column header, and remove from individual cells (consistent with Components page). Also ensure the Figma logo is the same as on the Components page.
- [ ] Font-sizes need reducing to make words fit better in columns
- [x] 'Getting started for designers' page  - Change the copy in the table title from “Salt asset libraries” to “Assets and styles” and the copy “Design specifications” to “Salt libraries”.
- [x] 'Getting started for designers' page - Move to one table (instead of two separate ones).

NB The 'responsive fixes' bucket includes reducing page padding on both sides, which should make tables slightly wider and text more readable...


site/src/utils/useIsMobileView.ts
```diff
@@ -0,0 +1,10 @@
+import { useViewport } from "@jpmorganchase/uitk-core";
+
+const useIsMobileView = (): boolean => {
```
Fercas123, line 3: should this be something more reusable? maybe useIsWithinViewport, idk, so you can pass the breakpoint you want to check for?

Fercas123 (approved):
looks really good! just left a tiny comment, but we can change that if there is a use case in the future

PR #840, opened 2022-12-08
Adds a proper favicon to the Salt website

### LINKED ISSUE #678: Favicon design
- [ ] Create Favicon for the site
- [ ] Get sign off from leads
- [ ] Implement in build

james-nash:
* Use different favicon for local dev, PR and prod builds.
* Removes some redundant Docusaurus sample images

closes #678


site/docusaurus.config.js
```diff
@@ -1,6 +1,28 @@
...
+/**
+ * Determines if this is a local dev, PR or production build
+ */
+function getBuildType() {
+  if (
+    process.env.GITHUB_EVENT_NAME === "push" &&
+    process.env.GITHUB_REF_NAME === "main"
+  ) {
+    return "prod";
+  }
```
james-nash, line 13: @joshwooding Do you happen to know if this is the "correct" way to detect when the site is being built off the `main` branch (and will therefore deploy to our production URL)?

I'm fairly confident the conditions for PR builds and local dev are working, as I'm getting the desired favicon in each case. But until this PR gets merged into main, I have no way to test whether or not this condition is the right one.

joshwooding: This should work yeah. Not 100% though

james-nash: Cheers. Ok, I'd say let's try to merge this then. Worst case scenario - the live site has the wrong color favicon :-P

joshwooding (approved):

PR #854, opened 2022-12-09
Add sorting to components table

JoanaMMoreira:


site/docs/components/_index/ComponentsList.tsx
```diff
@@ -63,28 +88,99 @@ const ComponentStatusData = ({
...
   return (
     <div className={styles.componentList}>
       <table>
         <thead>
           <tr>
-            <th>
-              <StorybookLogo />
-              <span>Component</span>
+            <th
+              onClick={() => handleSorting("name")}
```
james-nash, line 148: Wouldn't it be neater to put the `onClick` be on the `<Button>` in `<ComponentHeader>`? That's the actual interactive and focusable element that's being clicked (or potentially activated via keyboard), not the `<th>`.

How you've done it still seems to work (even via keybaord). I suspect that's due to the button's click event bubbling up to the table header. But, I think it make the code easier to understand and maintain if we move the `onClick` to the component that's actually being clicked.

james-nash (dismissed):
Looking pretty sweet!

Just requesting a small tweak to the code though.

site/docs/components/_index/ComponentsList.tsx
JoanaMMoreira: Makes sense, I've added the Button after our chat on Friday so forgot to move the `onClick`

Fercas123 (approved):
LGTM

PR #884, opened 2022-12-13
developer table fixes

EleanorHaproff:


site/docs/getting-started/DevelopFontsAccordion.module.css
```diff
@@ -3,6 +3,10 @@
   --ifm-hr-background-color: var(--uitk-separable-secondary-borderColor);
 }
 
+.accordionWrapper :global(.uitkAccordionDetails) {
+  overflow: visible;
```
libertymayc, line 7: for me it seems this is making the overflow a bit funny when the accordion is 'opening' as it overlaps the following content

EleanorHaproff: good spot! Will fix

PR #885, opened 2022-12-13
Styling tweaks

JoanaMMoreira:
-  Align inline code blocks with text
-  Fix spacing between images on getting started design page
-  Change images captions to use LD/Help token
-  Adjust padding on components table headers to allow a larger clicking target


libertymayc (approved):

EleanorHaproff (approved):
🎉

PR #889, opened 2022-12-13
remove CSSVariableTable from components' docs

alycrys:


libertymayc:
Is this temporary?

alycrys:
> Is this temporary?

Don't think so. Let me remove the component also. If we ever want to bring it back we'll dig into history for it

libertymayc:
hmm, why are we deleting this? it's a pretty good doc table for customisation imo, for familiarisation with the api variables in general - i would get if it were temporary whilst we wait for designers to let us know what should be customisable though

alycrys:
> hmm, why are we deleting this? it's a pretty good doc table for customisation imo, for familiarisation with the api variables in general - i would get if it were temporary whilst we wait for designers to let us know what should be customisable though

I believe the conversation with designers has happened, as this request is coming from Robin

PR #898, opened 2022-12-13
Amend link icon colour and fix table on dev page

JoanaMMoreira:


site/docs/components/_index/ComponentsList.module.css
```diff
@@ -6,6 +6,10 @@
   table-layout: fixed;
 }
 
+.componentList a svg {
+  --uitkIcon-color: var(--uitk-text-primary-foreground);
```
alycrys, line 10: would you need to update the colour for interaction states also?

james-nash: I think it would.

An easy way to achieve that would be to do:

```suggestion
  --uitkIcon-color: currentColor;
```

Then the icon will match the link's `color`, including when that changes on hover and stuff like that

james-nash (changes requested):

PR #904, opened 2022-12-14
Getting started design changes

EleanorHaproff:


shey-v (dismissed):
**Vert Nav.**

- The active left hand nav items currently shift / indent slightly. Same thing happening on live site. Bug fix.

 

**Right-hand Nav.**
- The orange bar doesn’t seem to appear on load, only when scrolling down slightly, is this a bug? Should it not be there on page load?
- Right-hand nav should remain sticky on scroll, currently it disappears…… 

 
**Cards**

- Missing hover state (shadow) please add in to be consistent with all other cards

 
**Padding issue**
- Margin above ‘Swapping from light to dark theme and vice versa’ h4 needs to be updated to 24px - currently it is 32px which is too large

 
**Icon table padding**

- Margin above and below table needs to be amended to 12px - currently 24px which is too large.

 
**Need support call out**

- Rufaro to feedback - update copy.

 
**General copy**
- Overall, copy tweaks are needed on entire page for some sections - Rufaro to feedback.

Fercas123 (approved):
Amazing work! 🎉 LGTM

PR #960, opened 2022-12-20
Open all site links in the same tab

### LINKED ISSUE #857: Site: Make all links open in same tab/window
Docusaurus has an annoying "feature" where links to other websites automatically get `target="_blank"` added to them, which forces them to open in a new browser tab or window. This is a usability issue (it takes away choice from the user) and an accessibility issue (it can disorient some users), so we should work around it if we can.

One option might be to swizzle Docusaurus' `<Link>` component and use our own one instead. When we do this, we should also remove the "new tab" icon beside such links.

It would be good to also raise a bug on Docusaurus so that they fix this in a future release. [Someone already reported it to them in their feedback forum](https://docusaurus.io/feature-requests/p/external-links-opening-in-new-tabs-should-be-optional-a11y), but I don't know if an actual Github issue for this exists. Let's also report this (or at least comment on that existing report). The more people do so, the greater the chance they'll fix it. :-)

### Further reading:

* [W3C WCAG Opening new windows and tabs from a link only when necessary](https://www.w3.org/TR/WCAG20-TECHS/G200.html)
* [CSS Tricks: When to use target=”_blank”](https://css-tricks.com/use-target_blank/)
* [Nielsen Norman Group: Opening Links in New Browser Windows and Tabs](https://www.nngroup.com/articles/new-browser-windows-and-tabs/)
* [Smashing Magazine: Should Links Open In New Windows?](https://www.smashingmagazine.com/2008/07/should-links-open-in-new-windows/)

JoanaMMoreira:
Closes #857


james-nash (approved):
Awesome stuff!

alycrys (approved):

PR #961, opened 2022-12-20
Storybook palette token designs

libertymayc:


ZarkoPernar:
Why don't we use the "swatches" display from the foundations here as well?

libertymayc:
> I find the horizontal display a bit easier to scan quickly rather than scrolling vertically.

that is an option, Paul is currently looking at our own unique designs (this pr is just me playing around atm)

PR #962, opened 2022-12-20
Update Theme 'characteristic' type to only include characteristics

libertymayc:


joshwooding (changes requested):
Can you add a changeset please :)

joshwooding (approved):

PR #967, opened 2022-12-20
Misc HTML, SVG & a11y fixes

james-nash:
Fixing some minor issues flagged by the HTML validator, aXe and other tools. Also optimizing some of our SVG images as part of this.

* Some SVGs were invalid (they were missing the mandatory `offset` attribute on some gradient `<stop>` elements. Presumably Figma is exporting broken SVG syntax 🤦 ), so I've fixed those by hand
* Homepage inlines multiple SVGs and, as part of Docusaurus build, they get optimized which then causes duplicate IDs to be present on the page (the optimization replaces IDs in each SVG file with single-letter names, starting with `a`. So when more than one of those are in the same document, you end up with dupe IDs)
* Some of our decorative SVGs didn't have text alternatives. Since they're just decorative, I set them to `role="presentation"`
* A few instances of incorrect HTML nesting
* Fixed color contrast issues with our customized Prism theme for syntax highlighting

Unsolved aXe issues:

* aXe reports that we have two `<nav>` landmarks on some pages (the main site nav and the vertical section nav), but neither has a header or `aria-label` so they can't be distinguished. It turns out that the [Docusaurus team already have a fix for this in the works](https://github.com/facebook/docusaurus/commit/eecfdebf084facca78b4c9074566df4109dadd5d#diff-947d6ab827b74de49165483648717280cd4ac5e752fef373cb2ad780fb094fcdR45), so once that gets released and we upgrade this issue should go away :-)
* aXe also reports items where it can't determine if there's sufficient color contrast. It's the headings (and intro blurb on the homepage) where we have _image_ backgrounds. I've checked and our images are suitably dark (and we also set a solid background color in case the image fails to load). Therefore, we can safely ignore this issues!


JoanaMMoreira (dismissed):
LGTM :)

JoanaMMoreira (approved):

PR #975, opened 2022-12-22
Use Storybook color component for palette tokens & Opacity block styling

libertymayc:


ZarkoPernar:
I meant how we display the prefix + suffix differently in these two examples

Screenshot 2022-12-22 at 14 06 52
Screenshot 2022-12-22 at 14 06 42

libertymayc:
> I meant how we display the prefix + suffix differently in these two examples
> 
> Screenshot 2022-12-22 at 14 06 52
> 
> Screenshot 2022-12-22 at 14 06 42

I initially intended this, but why I didn't do it this way is 1. because some palettes have lots of tokens, and with the label underneath each swatch it becomes quite unreadable, of course i could just split them by lines (like gray is done in colors), but i found it clearer to split by attribute as it is easier to scan for the type of token you want that way IMO, and 2. because in the figma, it is also separated by attributes, so gives a bit of similarity

ZarkoPernar:
Makes sense @libertymayc 👍

ZarkoPernar (approved):

ZarkoPernar (approved):

PR #970, opened 2022-12-21
Separate content from UI components

JoanaMMoreira:
- Move most of the docs content to MDX files
- Refactor and move docs specific components
- Move homepage accordion content to `index.tsx`
- Add site tsconfig to ESLint
- Move `splitListItems` to utils

PR #989, opened 2023-01-04
Move content component to layouts

JoanaMMoreira:
Most files have already been moved as part of #970 so I believe the `Content` component is the only one left that is misplaced but let me know if you think otherwise


james-nash (approved):
Looks good. Thanks for doing this!

PR #993, opened 2023-01-04
Link is supplying excessive text override

alycrys:

PR #995, opened 2023-01-05
Add changeset for #993

alycrys:


.changeset/thirty-dryers-invent.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Remove styles from base Text component
```
joshwooding, line 5: Could you add some context on why this was needed please

joshwooding (changes requested):
We should add some context on why this was necessary and how it affects the user.

feliciackh (dismissed):

PR #996, opened 2023-01-05
Fix flex layouts polymorphic prop

Fercas123:
Previous change from `extend FlexLayoutProps` to `extend HTMLAttributes` for components built from `FlexLayout` to avoid them inheriting all of `FlexLayout` props meant they no longer had access to its polymorphic `as` property.
This change allows `StackLayout` and `FlowLayout` to use the polymorphic prop.

### DIFF
--- a/packages/core/src/flow-layout/FlowLayout.tsx
+++ b/packages/core/src/flow-layout/FlowLayout.tsx
@@ -2,6 +2,9 @@ import { ElementType, forwardRef, HTMLAttributes } from "react";
 import { FlexLayout, FlexLayoutProps } from "../flex-layout";
 
 export interface FlowLayoutProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * The HTML element used for the root node.
+   */
   as?: FlexLayoutProps<ElementType>["as"];
   /**
    * Defines the default behavior for how flex items are laid out along the cross axis on the current line, default is "stretch".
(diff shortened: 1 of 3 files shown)


ZarkoPernar:
Don't you need to add the whole polymorphic props type to these two components similar to what FlexLayout has?

packages/core/src/flow-layout/FlowLayout.tsx
```diff
@@ -1,22 +1,35 @@
...
+    }
+  >;
+
+type FlowLayoutComponent = <T extends ElementType = "div">(
+  props: FlowLayoutProps<T>
+) => ReactElement | null;
+
+export const FlowLayout: FlowLayoutComponent = forwardRef(
+  <T extends ElementType = "div">(
+    { ...props }: FlexLayoutProps<T>,
```
JoanaMMoreira, line 30: should this be `FlowLayoutProps<T>`?

packages/core/src/stack-layout/StackLayout.tsx
```diff
@@ -1,23 +1,35 @@
...
 
-export const StackLayout = forwardRef<HTMLDivElement, StackLayoutProps>(
-  function StackLayout(props, ref) {
+type StackLayoutComponent = <T extends ElementType = "div">(
+  props: StackLayoutProps<T>
+) => ReactElement | null;
+
+export const StackLayout: StackLayoutComponent = forwardRef(
+  <T extends ElementType = "div">(
+    { ...props }: FlexLayoutProps<T>,
```
JoanaMMoreira, line 30: Same as above

PR #997, opened 2023-01-05
Fix skipped grid stories

ZarkoPernar:
- Row virtualisation was failing on ci but not on local because the ci runs on linux in which the default browser renders the scrollbar, while locally (if on a mac) there are no scrollbars. That resulted in a height difference between the two in the available space for the rows.
- Previously skipped column tests were expecting results from an older data set so just needed some adjustments.


feliciackh (approved):
lgtm

PR #1006, opened 2023-01-11
[ag-grid-theme] Separate generic numeric and editable cell classes, and fix corner tag

ZarkoPernar:
## Separate numeric and editable cell classes from `.editable-numeric-cell` to:
- `.editable-cell`
- `.numeric-cell`

Numeric cells can be editable or not, and we can have non-numeric editable cells. So that's the motivation for two separate 
classes (which can be composed together). Technically, we could remove the numeric class and let the user do that 
themselves but since we already handled text alignment in numeric cells I thought we keep it around

## Fix corner tag
Previously, `overflow: hidden` on the cell prevented us from placing the focused corner tag on top of border but with this change where we replace overflow with `clip-path` and now it is possible.

### Before
Screenshot 2023-01-12 at 15 55 55

### After
Screenshot 2023-01-12 at 15 55 11

### DIFF
--- a/.changeset/fuzzy-foxes-grin.md
+++ b/.changeset/fuzzy-foxes-grin.md
@@ -1,0 +1,14 @@
+---
+"@salt-ds/ag-grid-theme": major
+---
+
+Separate static numeric and editable cell class names in salt ag-grid theme,
+
+Before:
+
+- `.editable-numeric-cell`
+
+After:
+
+- `.numeric-cell`
+- `.editable-cell`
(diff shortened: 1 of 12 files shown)


joshwooding:
What's the motivation behind the change?

ZarkoPernar:
@joshwooding numeric cells can be editable or not, and we can have non-numeric editable cells. So that's the motivation for two separate classes (which can be composed together). Technically, we could remove the numeric class and let the user do that themselves but since we already had it I thought we keep it around.

Fercas123 (approved):

feliciackh (approved):
lgtm

PR #1023, opened 2023-01-16
Fix Link not removing underline on hover or focus

joshwooding:


alycrys (dismissed):

pseys:
Looks good. 

The icon behaviour on the variants (primary and secondary) is not correct on hover, is this just because there's no href in the example? 

https://8d6d7b9d.saltdesignsystem-storybook.pages.dev/?path=/docs/documentation-core-link--external-link

PR #1025, opened 2023-01-17
Tooltip

alycrys:
This PR includes:

**Tooltip refactor** 

- Reason being the fact that is wasn't easy to use by devs. We had the Tooltip and trigger as siblings and we were using the `useTooltip` hook to get the props to be passed to both.
We've refactored it so the trigger needs to be wrapped inside the Tooltip and the `useTooltip` hook is now being used inside the Tooltip component, therefor much cleaner.
- Replaced `floating-ui/react-dom-interactions` (deprecated) with `floating-ui/react`
- Removed `render` and `title` params and added `content` instead

**Changes in all components using Tooltip**

**Move Tooltip to Core** 
- along with `Portal` and `Popper`, as they're being used by Tooltip

**Remove TooltipContext** as it was used only by the deprecated components like list-deprecated and combo-box-deprecated

**Added few more cypress tests**

**Updated documentation**


alycrys:
Replaced by https://github.com/jpmorganchase/salt-ds/pull/1080 and https://github.com/jpmorganchase/salt-ds/pull/1084

PR #1035, opened 2023-01-19
Move split layout to core

Fercas123:


el-dav:
A few initial thoughts on SplitLayout

- It would be nice if we could avoid props like `leftSplitItemProps`/`rightSplititemProps`
- When we set `gap` on `SplitLayout`, do we expect that `gap` to also apply to items in the left/right hand side (if there are multiple items in a side). Currently I don't think this would happen.
- If there are two children on each side, what is the intended wrapping behaviour when we run out of space? Should only the last item in the right hand side wrap, or should the whole right side wrap

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+type SplitLayoutComponent = <T extends ElementType = "div">(
+  props: SplitLayoutProps<T>
+) => ReactElement | null;
+
+export const SplitLayout: SplitLayoutComponent = forwardRef(
+  <T extends ElementType>(
+    { children, wrap = true, ...rest }: SplitLayoutProps<T>,
+    ref?: PolymorphicRef<T>
+  ) => {
+    const warnChildren = Children.toArray(children).length > 2;
```
el-dav, line 39: You can use Children.count for this. Also I'd maybe put this line inside the effect so it doesn't run in prod

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,62 @@
...
+          );
+        }
+      }
+    }, [warnChildren]);
+    return (
+      <FlexLayout
+        direction="row"
+        ref={ref}
+        wrap={wrap}
+        justify="space-between"
```
el-dav, line 55: Because we always have `justify="space-between"` it means we depend on having two children. This means that we couldn't have one item at the end and none at the beginning.

Maybe this is fine, and it would have been true with the old API. But I could see a use case where you conditionally have an item at the beginning and still want an item at the end.

packages/core/src/split-layout/SplitLayout.tsx
Fercas123: i think children.count would ignore conditional children

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+  props: SplitLayoutProps<T>
+) => ReactElement | null;
+
+export const SplitLayout: SplitLayoutComponent = forwardRef(
+  <T extends ElementType = "div">(
+    {
+      direction,
+      endItem,
+      startItem,
+      wrapAtBreakpoint = "sm",
```
el-dav, line 50: I'd have thought by default we would leave `wrapAtBreakpoint` as undefined. What do you think?

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+      ...rest
+    }: SplitLayoutProps<T>,
+    ref?: PolymorphicRef<T>
+  ) => {
+    const justify = endItem && !startItem ? "end" : "space-between";
+    const wrapSplit = useIsViewportLargerThanBreakpoint(wrapAtBreakpoint);
+
+    return (
+      <FlexLayout
```
el-dav, line 59: Do we need to give a SplitLayout className for consistency?

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+        ref={ref}
+        justify={justify}
+        direction={wrapSplit ? "column" : direction}
```
el-dav, line 62: To me the code logic is saying that when the viewport is larger than `wrapAtBreakpoint` then `direction="column"` which would be incorrect (we want the SplitLayout to collapse when the viewport is smaller than the breakpoint).

But the behaviour in Storybook is correct.

Is the `useIsViewportLargerThanBreakpoint(wrapAtBreakpoint)` returning the wrong value?

packages/core/stories/layouts.doc.stories.mdx
```diff
@@ -45,6 +45,14 @@ Defines an equally distributed layout system using columns and rows.
   <Story id="core-layout-gridlayout-griditem--grid-item-wrapper" />
 </Canvas>
 
+## [Split layout](/docs/documentation-core-layout-splitlayout--page)
+
+Defines start and end regions for UI elements within a span, such as a button bar
```
el-dav, line 50: Maybe change `span` to `container` as there may be confusion with the `span` element?

packages/core/stories/split-layout.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+      options: [...FLEX_ALIGNMENT_BASE, "stretch", "baseline"],
+      control: { type: "select" },
+    },
+    as: {
+      type: "string",
+    },
+    gap: {
+      type: "number",
+    },
+    wrap: {
```
el-dav, line 32: change `wrap` to `wrapAtBreakpoint`. Something like:
```
wrapAtBreakpoint: {
      options: Object.keys(DEFAULT_BREAKPOINTS),
      control: { type: "select" },
    },
```

packages/core/stories/split-layout.stories.tsx
```diff
@@ -0,0 +1,130 @@
...
+  InfoIcon,
+  SettingsIcon,
+  TearOutIcon,
+  VisibleIcon,
+} from "@salt-ds/icons";
+
+export default {
+  title: "Core/Layout/SplitLayout",
+  component: SplitLayout,
+  argTypes: {
```
el-dav, line 21: Missing `direction` prop. Would be good to include this as a select e.g:
```
direction: {
    options: ["row", "column"],
    control: { type: "select" },
  },
```

packages/core/src/split-layout/SplitLayout.tsx
```diff
@@ -0,0 +1,70 @@
...
+       */
+      gap?: FlexLayoutProps<ElementType>["gap"];
+      /**
+       * Start component to be rendered.
+       */
+      startItem?: ReactNode;
+      /**
+       * Breakpoint at which the horizontal split wraps.
+       */
+      wrapAtBreakpoint?: keyof Breakpoints;
```
el-dav, line 37: We should make sure this prop name is consistent across components with similar behaviour.

I think for the user this might not feel like 'wrapping' even though that's what the Flex is doing under the hood. What do you think?

packages/core/src/split-layout/SplitLayout.tsx
Fercas123: We went for not giving  their own classNames to layouts that didn't need any styles of their own, like this, `Flow` and `Stack`

packages/core/src/split-layout/SplitLayout.tsx
Fercas123: I added `sm` because I removed wrap all together to simplify the props, would you say we would want to have `wrap` and `wrapAtBreakpoint` ? to me it felt like as a user i could feel like i need both on, i guess `wrapAtBreakpoint` is strictly not really wrapping but changing direction, also this approach doesn't let you not wrap... 
i guess options are: 
1) wrap (boolean) and wrapAtBreakpoint(breakpoint, default undefined)
2) wrapAtBreakpoint(breakpoint, default undefined means no wrap)
3) wrapAtBreakpoint(breakpoint,  add a `default` option that does wrap, and undefined means no wrap)

packages/core/src/split-layout/SplitLayout.tsx
Fercas123: I agree, not very sure about this name, it's not even wrapping, it is changing direction 🤔

packages/core/src/split-layout/SplitLayout.tsx
el-dav: I guess another option is to have neither prop as the user can user a responsive prop with direction?

packages/core/src/split-layout/SplitLayout.tsx
el-dav: See https://github.com/jpmorganchase/salt-ds/pull/1035#discussion_r1097408918. We could remove both wrap and wrapAtBreakpoint?

alycrys (dismissed):

packages/lab/src/utils/useIsViewportLargerThanBreakpoint.ts
```diff
@@ -1,20 +1,16 @@
+import { Breakpoints } from "@salt-ds/core/src/breakpoints";
```
libertymayc, line 1:
```suggestion
import { Breakpoints } from "@salt-ds/core";
```

packages/lab/src/utils/useIsViewportLargerThanBreakpoint.ts
```diff
@@ -1,20 +1,16 @@
...
 import {
-  Breakpoints,
   useCurrentBreakpoint,
   useOrderedBreakpoints,
-} from "@salt-ds/core";
+} from "@salt-ds/core/src/utils/useResponsiveProp";
```
libertymayc, line 5:
```suggestion
} from "@salt-ds/core";
```

feliciackh (approved):

alycrys (approved):

packages/core/stories/layouts.doc.stories.mdx
```diff
@@ -45,6 +45,14 @@ Defines an equally distributed layout system using columns and rows.
   <Story id="core-layout-gridlayout-griditem--grid-item-wrapper" />
 </Canvas>
 
+## [Split layout](/docs/documentation-core-layout-splitlayout--page)
+
+Defines start and end regions for UI elements within a container, such as a button bar
```
sion-docs, line 50: Could this be 'start and end _points_'?

packages/core/stories/split-layout.doc.stories.mdx
```diff
@@ -15,16 +13,16 @@ import HelpAndSupport from "docs/blocks/help-and-support.doc.mdx";
...
 
 <Canvas>
-  <Story id="lab-layout-splitlayout--default-split-layout" />
+  <Story id="core-layout-splitlayout--default-split-layout" />
 </Canvas>
 
 ## Use Split layout when...
 
-There are 2 defined areas that are required to be in the opposite horizontal ends of a container,
+There are 2 defined areas required to be in the opposite horizontal ends of a container,
```
sion-docs, line 25: two rather than 2

PR #1038, opened 2023-01-20
Data Grid cell validation

### LINKED ISSUE #1011: Data Grid: Cell Validation
### Area

UI Components

### The problem

Currently the data grid cannot be configured to show validation states for a cell out of the box. As most things with our grid this can be developed in user land but we want to provide and easy and accessible experience out of the box.

### The solution

Each cell can be passed a validation state, one of:
- info
- warning
- error
- success

### Alternatives and examples

N/A

ZarkoPernar:
### Description
Closes #1011. Documentation is WIP.

This PR adds cell and row validation in grid. The "row validation" is a bit of a lie because there is no row-level api. Instead the user can add validation props to each column in order to achieve row validation. We are also adding no business logic for validation, only the visual styles.

### Changes
This PR adds new props to `GridColumn`:
- `getValidationStatus` this is a function the user provides us that must return one of the allowed validation state names (currently only "warning" and "error" are supported). BaseCell calls this function with all of the row and column data so the user can determine the cell validation based on the value.
- `getValidationMessage` same as the above but here the user returns a custom message we can use in the aria-status region for some a11y support. In the case of error, we also add an `aria-invalid` to the cell itself.
- `validationType` the options here are "strong" and "light", and the default is "light". Because we want the icon to appear on editable cells and for non-editable cells to have only their background affected this 

### Alternative approach
For the validationType, we could completely avoid this by making editable cells have an icon by default and non editable cells have only their background color affected. Ben and Darrin mentioned a lot of different approaches to validation inside the org, so I'm a bit worried about constraining our api so much. 

#### Row validation
localhost_6006_iframe html_args= id=data-grid-data-grid--row-validation viewMode=story

#### Cell validation
localhost_6006_iframe html_args= id=data-grid-data-grid--cell-validation viewMode=story


ZarkoPernar:
@joshwooding what do you think any having a `validationType = withIcon | color` or something like that to the column definition? That would allow the user the flexibility to either have background color + border, or colors + icon. Then we could add guidance in the documentation telling them to use the icon when validating individual cells, and only the colors when they're validating the entire row.

packages/grid/src/GridColumn.tsx
```diff
@@ -96,6 +104,23 @@ export interface GridColumnProps<T = any> {
    * for the given row data item.
    * */
   getValue?: (rowData: T) => any;
+  /**
+   * Cell validation status getter. Should return one of the known validation status names.
```
feliciackh, line 108:
```suggestion
   * Cell validation status getter. Should return one of the known validation status names.
   * Cell validation status getter. Should return one of the known validation status names: "none" | "error" | "warning"
```

packages/grid/stories/grid.doc.stories.mdx
```diff
@@ -200,6 +200,20 @@ To trigger edit mode on cells on your:
 - keyboard: press either the `backspace` _or_ `F2` key, _alternatively_ simply start typing
 - mouse: double click
 
+### Validation
+
+The grid columns have two types of built-in validation, "light" and "strong". The default style is "light". They both only apply the visual styling. You will need to provide the validation logic yourself.
+If you are validating editable cells, we recommend you use a 3rd party library like [yup](https://www.npmjs.com/package/yup) or [zod]https://www.npmjs.com/package/zod). The first validation type "light", applies a status color to the background and border,
```
feliciackh, line 206:
```suggestion
If you are validating editable cells, we recommend you use a 3rd party library like [yup](https://www.npmjs.com/package/yup) or [zod](https://www.npmjs.com/package/zod). The first validation type "light", applies a status color to the background and border,
```

packages/grid/stories/grid.doc.stories.mdx
```diff
@@ -200,6 +200,20 @@ To trigger edit mode on cells on your:
...
+while the second type "strong" also displays a status icon. Light validation is supposed to be used with non-editable cells, and strong with editable cells.
```
feliciackh, line 207:
```suggestion
while the second type "strong" additionally displays a status icon. Light validation is supposed to be used with non-editable cells, and strong with editable cells.
```

feliciackh (dismissed):

packages/grid/src/BaseCell.css
```diff
@@ -27,3 +27,86 @@
...
+.saltGridBaseCell-status-error:focus .saltGridCursor {
+  border-color: var(--salt-status-error-borderColor);
+}
+
+.saltGridBaseCell-status-error:focus .saltGridCornerTag-focusOnly {
+  border-left-color: var(--salt-status-error-borderColor);
+}
+
+.saltGridBaseCell-valueContainer-status-error {
+  background-color: var(--salt-palette-error-background-emphasize);
```
libertymayc, line 41:
```suggestion
  background-color: var(--salt-status-error-background-emphasize);
```

packages/grid/src/BaseCell.css
```diff
@@ -27,3 +27,86 @@
...
+.saltGridBaseCell-status-success:focus .saltGridCursor {
+  border-color: var(--salt-status-success-borderColor);
+}
+
+.saltGridBaseCell-status-success:focus .saltGridCornerTag-focusOnly {
+  border-left-color: var(--salt-status-success-borderColor);
+}
+
+.saltGridBaseCell-valueContainer-status-success {
+  background-color: var(--salt-palette-success-background-emphasize);
```
libertymayc, line 54:
```suggestion
  background-color: var(--salt-status-success-background-emphasize);
```

packages/grid/src/BaseCell.css
```diff
@@ -27,3 +27,86 @@
...
+.saltGridBaseCell-status-warning:focus .saltGridCursor {
+  border-color: var(--salt-status-warning-borderColor);
+}
+
+.saltGridBaseCell-status-warning:focus .saltGridCornerTag-focusOnly {
+  border-left-color: var(--salt-status-warning-borderColor);
+}
+
+.saltGridBaseCell-valueContainer-status-warning {
+  background-color: var(--salt-palette-warning-background-emphasize);
```
libertymayc, line 67:
```suggestion
  background-color: var(--salt-status-warning-background-emphasize);
```

packages/grid/src/BaseCell.css
```diff
@@ -27,3 +27,86 @@
...
+
+.salt-density-touch {
+  --grid-statusIndicator-size: 12px;
+}
+
+.saltGridBaseCell-statusContainer {
+  font-size: var(--saltGrid-statusIndicator-size, var(--grid-statusIndicator-size));
+  position: absolute;
+  bottom: 4px;
+  right: 4px;
```
libertymayc, line 90: will these be affected by density? could use --salt-size-basis-unit here if not

packages/grid/src/GridColumn.tsx
```diff
@@ -15,7 +15,10 @@ import { GridColumnModel, GridRowModel } from "./Grid";
 
 export type GridColumnPin = "left" | "right" | null;
 
-export interface GridCellProps<T> {
+type CellValidationState = "error" | "warning" | "success" | undefined;
```
libertymayc, line 18: do we need undefined if optional?

packages/grid/stories/grid-cellValidation.stories.tsx
```diff
@@ -0,0 +1,334 @@
...
+const discountValues = [1, 0.95, 0.9, 0.8, 0.5];
+
+interface RowExample {
+  id: string;
+  name: string;
+  description: string;
+  amount: number;
+  price: number;
+  discount: string;
+  status?: "error" | "warning" | "success";
```
libertymayc, line 33: i'm wondering if it would be good to export the CellValidationStatus type?

packages/grid/src/BaseCell.tsx
```diff
@@ -12,6 +12,49 @@ export function getCellId<T>(rowKey: string, column: GridColumnModel<T>) {
   return `R${rowKey}C${column.info.props.id}`;
 }
 
+const icons = {
```
libertymayc, line 15: I feel these may be used in other components for validation, may be good separating them out somewhere...? can be done in the future though

packages/grid/src/BaseCell.css
ZarkoPernar: Have to double check in figma

packages/grid/src/BaseCell.tsx
ZarkoPernar: They might be used in FF but not sure. I prefer to keep things co-located until it doesn't make sense anymore. But yeah, as soon as another component needs this we should break them out.

packages/grid/stories/grid-cellValidation.stories.tsx
ZarkoPernar: Yeah, probably

packages/grid/src/GridColumn.tsx
ZarkoPernar: Hmm, I son't remember why I did this 😅

ZarkoPernar: Ah, yeah, I just tried removing it and typescript was complaining. The issue is, that getValidationState is optional but that does not mean the return value is optional as well. You have to make it explicitly so.

package.json
```diff
@@ -161,5 +161,8 @@
       "<rootDir>/node_modules/csstree-validator"
     ]
   },
-  "packageManager": "yarn@3.4.1"
+  "packageManager": "yarn@3.4.1",
+  "devDependencies": {
+    "@faker-js/faker": "^7.6.0"
```
joshwooding, line 166: Can you move this to dependencies so it aligns with everything else?

packages/grid/src/BaseCell.tsx
joshwooding: I wonder if it's still worth putting these in a separate file at least?

packages/grid/src/BaseCell.tsx
```diff
@@ -12,6 +12,49 @@ export function getCellId<T>(rowKey: string, column: GridColumnModel<T>) {
...
+  warning: (
+    <svg
+      aria-hidden
+      // width="1em"
```
joshwooding, line 19: The attributes for the icons seem to be commented out randomly. Was this to fix scaling issues?

joshwooding:
> @joshwooding what do you think any having a `validationType = withIcon | color` or something like that to the column definition? That would allow the user the flexibility to either have background color + border, or colors + icon. Then we could add guidance in the documentation telling them to use the icon when validating individual cells, and only the colors when they're validating the entire row.

I'd have to think more about this. My immediate thoughts are more around with the color option we still want an icon at the end of a row so it might be confusing. Is there a way we can make this map more to the use-case for the row vs cell validation?

packages/grid/src/GridColumn.tsx
libertymayc: would it be better to have getValidationStatus?: (value: GridCellValueProps<T>) => CellValidationState | undefined ?

ZarkoPernar: Probably, good idea 👍

package.json
ZarkoPernar: Sure thing. Just curious, is there a particular reason we don't do deps vs devdeps?

packages/grid/src/BaseCell.css
```diff
@@ -27,3 +27,67 @@
 .saltGridCell-editable:focus .saltGridBaseCell-valueContainer {
   border-color: transparent;
 }
+
+.saltGridBaseCell-status-error {
```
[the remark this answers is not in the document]
ZarkoPernar: @libertymayc do you have an opinion on this?

packages/grid/src/BaseCell.tsx
ZarkoPernar: It was yeah. I'll remove the commented out parts

packages/grid/src/BaseCell.tsx
ZarkoPernar: ✅

packages/grid/src/BaseCell.css
ZarkoPernar: It needs to be hardcoded

packages/grid/src/BaseCell.css
```diff
@@ -60,26 +60,26 @@
...
 
 .salt-density-low {
-  --grid-statusIndicator-size: 10px;
+  --grid-statusIndicator-size: 9px;
 }
 
 .salt-density-touch {
-  --grid-statusIndicator-size: 12px;
+  --grid-statusIndicator-size: 11px;
 }
```
libertymayc, line 76: @bhoppers2008 should the above sizes be in the size foundation? trying to avoid any component specific styles

ZarkoPernar: I think this value will remain specific to this icon in particular. Mainly because Ben was trying to squeeze it in without obscuring overflowing text. We don't necessarily have to put everything in the foundations and characteristics. If it is not repeated in multiple places and we are not confident about reusability, it should remain locally defined. Bad abstractions are worse than no abstractions.

ZarkoPernar:
> tini tiny ones, also not sure if it is related but my local is complaining about line 5 in `ag-grid.stories.tsx` (`import "../salt-ag-theme.css";`) when i build, commenting it out allowed me to review but not sure if that would be an issue to look at.

This is because you first have to build the theme with a separate command in a separate process. Not the most automated system in the world 😅 

Thanks for the review @Fercas123! If you have time, since you now understand this area, can you also review #1237? It builds on top of this one.

packages/grid/stories/grid-cellValidation.stories.tsx
```diff
@@ -0,0 +1,335 @@
...
+        return x;
+      });
+    },
+    [setRows]
+  );
+
+  return { rows, setAmount, setName, setDiscount, setPrice };
+};
+
+export const CellValidation: Story = () => {
```
joshwooding, line 141: Same as the column sorting, could you add an example showing async/server validation?

packages/grid/stories/grid.doc.stories.mdx
```diff
@@ -233,6 +233,21 @@ To trigger edit mode on cells on your:
 - keyboard: press either the `backspace` _or_ `F2` key, _alternatively_ simply start typing
 - mouse: double click
 
+### Cell Validation
+
+To display a cell in one of expected validation states, use the `getValidationState` prop on the `GridColumn` component. This prop is a function that provides row and column data, and expects as return value one of known validation states, "error", "warning", and "success".
```
joshwooding, line 238:
```suggestion
To display a cell in one of the expected validation states, use the `getValidationState` prop on the `GridColumn` component. This prop is a function that provides row and column data and expects a return value of one of the known validation states, "error", "warning", and "success".
```

packages/grid/stories/grid-cellValidation.stories.tsx
ZarkoPernar: @joshwooding ok this has now been updated with an example of async validation using yup which would be the common scenario for real time editing. I could also add a mock server validation example but I think it would be almost an exact copy paste but instead of using yup I would be using a fake fetch. So I don't see a point.

joshwooding (approved):

PR #1040, opened 2023-01-23
Remove display content wrapper in flexlayoyut

Fercas123:
display content wrapper is blocking styles to be passed down to the component's children. removing it and using simple margins instead, since the separators are not going to be used in more than one direction.


joshwooding (approved):

joshwooding (changes requested):
Looks good. Can you add a changeset please :)

Fercas123:
Closing this PR as we will move separators from flex to stack layout.

PR #1041, opened 2023-01-23
Fix sticky behaviour on site docs TOC

JoanaMMoreira:


james-nash (dismissed):
Nicely done! So it wasn't the `display: contents` thing after all, eh?

Maybe Firefox was the one that was buggy.

site/src/css/custom.css
```diff
@@ -346,22 +337,6 @@ ul.table-of-contents li > ul li {
   --ifm-leading: var(--ifm-paragraph-margin-bottom);
 }
 
-main .container > div > div > div {
```
Fercas123, line 349: were we being too explicit and leaving things without this styles? I am just worried that div>div>div is not very descriptive to what we are targeting, and it could be easily broke by someone accidentally adding another wrapper, or removing one of those divs... could it not be done with classes?

JoanaMMoreira:
> Nicely done! So it wasn't the `display: contents` thing after all, eh?
> 
> Maybe Firefox was the one that was buggy.

yeah it's not easy to tell but maybe it was

site/src/css/custom.css
JoanaMMoreira: I agree it would have been better to use classes but I'm not sure if it was possible at the time

libertymayc (changes requested):
this reverts the next/previous buttons height positioning on zoom out (in chrome)

JoanaMMoreira:
> this reverts the next/previous buttons height positioning on zoom out (in chrome)

good spot! In that case it seems those changes were required so I've only amended the toc container height

libertymayc (approved):

PR #1043, opened 2023-01-24
[Data Grid] Resize refactor

ZarkoPernar:


joshwooding (approved):

joshwooding:
@ZarkoPernar Do you think we need a changeset? This is only a refactor but still touches production code.

ZarkoPernar:
Yeah, I was wondering about that. I could add something like "we made  grid dimension measurements more reliable"?

joshwooding:
> Yeah, I was wondering about that. I could add something like "we made grid dimension measurements more reliable"?

I think "Made Grid dimension measurements more reliable" is fine

joshwooding (approved):

PR #1066, opened 2023-01-30
Fix AriaAnnouncer not removing announced text from the DOM

joshwooding:
Updates AriaAnnouncer to remove announced text from the DOM and moves the tests to Cypress.

When investigating migrating the tests, I noticed that the behaviour didn't match what the tests were testing. This rectifies that


Fercas123 (approved):

JoanaMMoreira (approved):

alycrys (dismissed):

ImmanuelBaskaran (approved):

alycrys:
Would we need a changeset here?

alycrys (approved):

PR #1072, opened 2023-01-30
[Data Grid] Add aria-selected to selected rows

### LINKED ISSUE #1067: Add aria-selected to Grid rows

ZarkoPernar:
This used to be on the grid but was accidentally deleted at some point, probably in a conflict resolution. 
Closes #1067

### DIFF
--- a/.changeset/wise-pumas-travel.md
+++ b/.changeset/wise-pumas-travel.md
@@ -2,4 +2,4 @@
 "@salt-ds/data-grid": patch
 ---
 
-Add aria-selected to selected grid rows
+Fixed row selection not being announced by screenreaders
--- a/.changeset/wise-pumas-travel.md
+++ b/.changeset/wise-pumas-travel.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/data-grid": patch
+---
+
+Add aria-selected to selected grid rows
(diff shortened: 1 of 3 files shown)


libertymayc (dismissed):

.changeset/wise-pumas-travel.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/data-grid": patch
+---
+
+Add aria-selected to selected grid rows
```
joshwooding, line 5: This changeset covers what the change was but not really how it affects the user.

We should change it to something like:

```suggestion
Fixed row selection not being announced by screenreaders
```

ZarkoPernar: 👍

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -39,8 +39,8 @@ const clickCell = (row: number, col: number) => {
 
 const checkRowSelected = (row: number, expectedSelected: boolean) => {
   cy.get(`tr[data-row-index="${row}"]`).should(
-    expectedSelected ? "have.class" : "not.have.class",
-    "saltGridTableRow-selected"
+    expectedSelected ? "have.attr" : "not.have.attr",
```
joshwooding, line 42: This test only tests on the attribute's presence but not the value. This would mean that <tr aria-selected="false" /> would be picked up as selected.

joshwooding (changes requested):

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
ZarkoPernar: Indeed

ZarkoPernar: Updated

joshwooding (approved):

PR #1073, opened 2023-01-31
[Data Grid] Improve docs

ZarkoPernar:
### Lazy loading stories
Added LazyStoryCanvas component that uses intersection observer. This was necessary because the docs page for grid was not even loading locally, too many stories loading at once. 

### Story changes
I was trying to make the simplest possible example of Grid to show up as the initial example on the docs page. Also, experimenting a bit with the args in storybook. It is not easy to find a balance between a customisable story example that the reader can play with, and an example that shows up compactly in the "show code" on the docs page. The show code is very brittle and can easily make the example look nonsensical.

### CSS changes
Fixed a bug where the storybook global css was messing up the display of grid. I also wanted to make the stories css representative of what the user will need to do in their app. However, storybooks default display behaviour of centering the flex children makes it hard to do that and not affect other stories. For now we have to set a hardcoded value for the width.

#### Bug screenshot
Screenshot 2023-02-02 at 14 13 52

### withTestIdWrapper
This storybook decorator appears to no longer have any use so I removed it as it was interfering how grid was being displayed in the stories.


feliciackh (approved):

packages/grid/stories/grid.doc.stories.mdx
```diff
@@ -40,13 +42,23 @@ yarn add @salt-ds/data-grid;
...
+To import the grid into your application, use the following import:
 
-```
-import { Grid } from "@salt-ds/data-grid";
+```js
+import { Grid, GridColumn } from "@salt-ds/data-grid";
 ```
 
-Note: ensure the CSS `height` property for `.saltGrid` has been set appropriately to see grid on screen.
+<Banner disableAnnouncer>
```
joshwooding, line 51: We should make sure the team is happy with using Banner here

joshwooding: Just sent a message. If I don't hear back by the end of the day we can merge

libertymayc (approved):

feliciackh (approved):

PR #1084, opened 2023-02-02
Move Tooltip to core

alycrys:
Move Tooltip and useFloatingUI to core


packages/core/src/index.ts
```diff
@@ -11,11 +11,14 @@ export * from "./grid-item";
 export * from "./grid-layout";
 export * from "./link";
 export * from "./panel";
+export * from "./popper";
```
joshwooding, line 14: Before merging this PR we would discuss terminology around the popper/floating situation. This should be agreed between design and dev

joshwooding (approved):

PR #1103, opened 2023-02-07
Fix Panel primary variant background 

libertymayc:
Primary variant was using secondary token


Fercas123 (dismissed):

joshwooding (approved):

joshwooding (changes requested):
I don't think we need a changeset here for Panel to core since it hasn't be released in core yet

joshwooding (approved):

PR #1104, opened 2023-02-07
Only deploy site and storybook to production on changesets publish

EleanorHaproff:


.github/workflows/release.yml
```diff
@@ -35,3 +35,59 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: Use Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16"
+          cache: "yarn"
+
+      - name: Install dependencies
+        run: yarn
```
Fercas123, line 50: Just being curious, how does the release workflow work? (i see there are 3 steps that repeat this setup - install dependencies step... given it's the same repo, how is this not redundant? does something get clear after each step?)
sorry in advance if it is a dumb question, i have no clue of workflows

joshwooding: These steps are redundant since they are in the same job.

EleanorHaproff: Oh yes good spot, will remove

joshwooding (approved):

PR #1111, opened 2023-02-07
Card component doc and styling improvements

libertymayc:
- Fixes sbdocs custom theming overriding Salt styles in SB
- Aligns CSS tokens with convention (i.e. how Button does it)
- Uses Text component in examples

### DIFF
--- a/.storybook/styles.css
+++ b/.storybook/styles.css
@@ -1,14 +1,17 @@
 .salt-theme .sbdocs {
-  background: var(--salt-container-primary-background);
-  color: var(--salt-text-primary-foreground);
   font-family: var(--salt-typography-fontFamily);
   font-size: var(--salt-text-fontSize);
   letter-spacing: var(--salt-text-letterSpacing);
   line-height: var(--salt-text-lineHeight);
 }
 
-.salt-theme .sbdocs h1 {
+.salt-theme .sbdocs.sbdocs-content {
+  background: var(--salt-container-primary-background);
   color: var(--salt-text-primary-foreground);
+  height: fit-content;
+}
+
+.salt-theme .sbdocs h1 {
   font-size: var(--salt-text-h1-fontSize);
   font-weight: var(--salt-text-h1-fontWeight);
   line-height: var(--salt-text-h1-lineHeight);
@@ -16,21 +19,18 @@
 
 .salt-theme .sbdocs h2 {
   border-bottom: none;
-  color: var(--salt-text-primary-foreground);
   font-size: var(--salt-text-h2-fontSize);
   font-weight: var(--salt-text-h2-fontWeight);
   line-height: var(--salt-text-h2-lineHeight);
 }
 
 .salt-theme .sbdocs h3 {
-  color: var(--salt-text-primary-foreground);
   font-size: var(--salt-text-h3-fontSize);
   font-weight: var(--salt-text-h3-fontWeight);
   line-height: var(--salt-text-h3-lineHeight);
 }
 
 .salt-theme .sbdocs h4 {
-  color: var(--salt-text-primary-foreground);
   font-size: var(--salt-text-h4-fontSize);
   font-weight: var(--salt-text-h4-fontWeight);
   line-height: var(--salt-text-h4-lineHeight);
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -15,7 +15,7 @@
 
 /* Styles applied to Card content */
 .saltCard-content {
-  color: var(--saltCard-color, var(--uitk-text-primary-foreground));
+  color: var(--saltCard-color, var(--salt-text-primary-foreground));
   font-size: var(--saltCard-fontSize, inherit);
   letter-spacing: var(--saltCard-letterSpacing, inherit);
   line-height: var(--saltCard-lineHeight, inherit);
@@ -61,7 +61,7 @@ a:focus .saltCard-interactable.saltCard-disabled,
 a:focus .saltCard-interactable.saltCard-disabled {
   --card-borderColor: var(--salt-container-primary-borderColor-disabled);
 
-  color: var(--saltCard-color-disabled, var(--uitk-text-primary-foreground-disabled));
+  color: var(--saltCard-color-disabled, var(--salt-text-primary-foreground-disabled));
   cursor: var(--saltCard-cursor-disabled, var(--salt-selectable-cursor-disabled));
   outline: none;
 }
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -16,10 +16,7 @@
 /* Styles applied to Card content */
 .saltCard-content {
   color: var(--saltCard-color, var(--salt-text-primary-foreground));
-  font-size: var(--saltCard-fontSize, inherit);
-  letter-spacing: var(--saltCard-letterSpacing, inherit);
-  line-height: var(--saltCard-lineHeight, inherit);
-  min-height: var(--saltCard-minHeight, var(--salt-text-minHeight));
+  min-height: var(--salt-text-minHeight);
   padding: var(--saltCard-padding, calc(var(--salt-size-unit) * 3));
 }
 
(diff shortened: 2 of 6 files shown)


.storybook/styles.css
```diff
@@ -7,29 +7,29 @@
   line-height: var(--salt-text-lineHeight);
 }
 
-.salt-theme .sbdocs h1 {
+.salt-theme .sbdocs h1:not(.saltText) {
   color: var(--salt-text-primary-foreground);
   font-size: var(--salt-text-h1-fontSize);
   font-weight: var(--salt-text-h1-fontWeight);
   line-height: var(--salt-text-h1-lineHeight);
```
alycrys, line 14: why not saltText?

.storybook/styles.css
```diff
@@ -7,29 +7,29 @@
...
 }
 
-.salt-theme .sbdocs h2 {
+.salt-theme .sbdocs h2:not(.saltText) {
   border-bottom: none;
```
alycrys, line 18: was there a border-bottom on h2? just curious

packages/core/src/card/Card.css
```diff
@@ -1,22 +1,26 @@
...
 
 /* Styles applied to Card content */
 .saltCard-content {
-  color: inherit;
-  font-size: var(--saltCard-content-fontSize, var(--salt-text-fontSize));
-  letter-spacing: var(--saltCard-content-letterSpacing, normal);
-  line-height: var(--salt-text-lineHeight);
+  --saltText-color: var(--card-color);
+
+  font-size: var(--saltCard-content-fontSize, inherit);
```
alycrys, line 22: don't we need line-height here anymore?

.storybook/styles.css
libertymayc: in Storybook default styling yes

.storybook/styles.css
libertymayc: trying to remember why i did this.. but i don't think its necessary

packages/theme/stories/characteristics/characteristics.stories.mdx
```diff
@@ -57,7 +57,7 @@ theme, --salt-actionable-primary-background resolves to `var(--salt-action-prima
 
 In the following example, we will show how to customize the Card component through characteristics. Currently, Card only has one variant, using tokens such as `---salt-container-primary-background` all of its styling purposes. Here's how it looks by default, shown on a purple Panel for demo purposes:
 
-<Panel style={{ background: "var(--salt-color-purple-10)" }}>
+<Panel style={{ "--saltPanel-background": "var(--salt-color-purple-10)" }}>
```
joshwooding, line 60: We currently don't want to recommend this override behaviour. This change should be discussed with design

joshwooding (changes requested):
Needs further considerations

packages/core/src/card/Card.css
```diff
@@ -1,23 +1,27 @@
 /* Styles applied to the root element */
 .saltCard {
+  --card-borderColor: var(--salt-container-primary-borderColor);
+  --card-color: inherit;
+  --card-shadow: var(--salt-overlayable-shadow);
+  --saltText-color: var(--card-color);
```
joshwooding, line 6: Is this Card's responsibility?

packages/core/stories/card.stories.tsx
```diff
@@ -42,9 +42,9 @@ const Examples = () => (
     <ExampleRow name="Default">
       <Card>
         <div>
-          <h1 style={{ margin: "0", lineHeight: "1.3em" }}>
+          <H1 style={{ margin: "0", lineHeight: "1.3em" }}>
```
joshwooding, line 45: I haven't checked but are these styles still needed?

packages/core/src/card/Card.css
libertymayc: it's intended from design because of the disabled styling, and i did it like this since text would be passed via children - however, very true it should anyway be the responsibility of the consuemr to add the disabled variant to text, since card doesn't contain text necessarily - i just thought this would be an 100% way to ensure it. but i'm going to take it out now that i've reconsidered

packages/core/stories/card.stories.tsx
libertymayc: they were for QA stories, no longer needed (for now anyway)

libertymayc:
drafted - in discussion with Ben/Paul/Alina re nested text styling

libertymayc:
> drafted - in discussion with Ben/Paul/Alina re nested text styling

see #1145

packages/core/stories/card.stories.tsx
```diff
@@ -66,17 +66,17 @@ export const All: ComponentStory<typeof Card> = () => (
...
-      <span>Using Nested DOM Elements</span>
+      <H1>This is Card</H1>
+      <Text>Using Nested DOM Elements</Text>
     </div>
   </Card>
 );
 
 export const Disabled: ComponentStory<typeof Card> = () => (
   <Card data-testid="card-disabled-example" disabled>
     <div>
```
libertymayc, line 77: @alycrys @joshwooding Should we provide an alternative helper class that can be added to surrounding divs to avoid having to add disabled to each...?

packages/core/src/card/Card.css
```diff
@@ -1,23 +1,25 @@
...
 
 /* Styles applied to Card content */
 .saltCard-content {
-  color: inherit;
-  font-size: var(--saltCard-content-fontSize, var(--salt-text-fontSize));
-  letter-spacing: var(--saltCard-content-letterSpacing, normal);
-  line-height: var(--salt-text-lineHeight);
-  min-height: var(--saltCard-content-minHeight, var(--salt-text-minHeight));
+  color: var(--saltCard-color, var(--salt-text-primary-foreground));
+  font-size: var(--saltCard-fontSize, inherit);
```
alycrys, line 19: I'm not sure about making text styles configurable inside Card. I don't quite see the point

packages/core/stories/card.doc.stories.mdx
```diff
@@ -38,6 +38,8 @@ Cards can have interactable styling by setting `interactable={true}`. These card
 
 Cards can have disabled styling applied by setting `disabled={true}`.
 
+Make sure to as well apply `disabled={true}` to any nested components, such as Text, to ensure disabled styling is also applied to the inner content. We discourage the use of links within interactable Cards that can be disabled, since Salt Links cannot be disabled out of the box as they present usability issues, and create misleading interactions within an application.
```
alycrys, line 41: I think we discourage the use of any focusable elements inside the interactable cards regardless of disabled state. Or are we not?

packages/core/stories/card.stories.tsx
```diff
@@ -42,10 +42,8 @@ const Examples = () => (
     <ExampleRow name="Default">
       <Card>
         <div>
-          <h1 style={{ margin: "0", lineHeight: "1.3em" }}>
-            Card with Density
-          </h1>
-          <span>Here is some content</span>
+          <H1>Card with Density</H1>
```
alycrys, line 45: Let's use another Heading inside the card. I know this is just an example, but it'll never be valid in real life. Let's use an H3 styledAs H1 if you wish

packages/core/stories/card.stories.tsx
alycrys: I think it's worth having a discussion about this. Maybe it'll be an enhancement and we keep it as is for now?....

packages/core/stories/card.doc.stories.mdx
libertymayc: yes - this is just a bit design point because of the fact we can disable Text, people may not realise we can't disable Link. I will add to the docs about that too

joshwooding: I actually find it quite confusing reading this.

joshwooding: I would agree with @alycrys the confusion here is that we contradict ourselves by earlier saying focusable elements shouldn't be inside interactable cards and then later say links should not be inside interactable cards.

Regardless of the fact whether links could be disabled or not it is still focusable/interactable therefore shouldn't be inside the card anyway

libertymayc: sure, i copied it from figma. will delete this bit

alycrys (approved):

joshwooding (approved):

PR #1134, opened 2023-02-09
Add a roadmap page (in the Getting Started section for now)

shinytoyrobots:
Getting started seems to be the best interim spot for this page, within the current navigation.


site/docs/getting-started/roadmap.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+This is the collaborative estimate of the design system team for realistic delivery within Q1.
+
+It is intended as a guide to our intended released work, for guidance and information. It should not be considered an absolute commitment.
+
+### System
+
+#### Components
+
+- [Tooltip](https://github.com/jpmorganchase/salt-ds/issues/324)
```
james-nash, line 22: If these are ordered by priority, perhaps it would be better to use a numbered list to indicate that.

Otherwise (if we don't want to convey a priority order within the quarter), I would suggest sorting them alphabetically by component name, so that it's easier to find a given component name.

james-nash (dismissed):
Looking pretty good. Just had a question / suggestion about the ordering of the bullet points in each section

site/docs/getting-started/roadmap.mdx
sion-docs: Numbered lists should be reserved for sequential steps, alphabetically makes sense.

site/docs/getting-started/roadmap.mdx
```diff
@@ -0,0 +1,96 @@
...
+
+# Roadmap
+
+<Content>
+
+## Q1 2023 delivery
+
+This is the collaborative estimate of the design system team for realistic delivery within Q1.
+
+It is intended as a guide to our intended released work, for guidance and information. It should not be considered an absolute commitment.
```
sion-docs, line 16: Suggest "It covers our intended released work for guidance and information...".

site/docs/getting-started/roadmap.mdx
rufcha: I agree, it should be alphabetical.

rufcha:
@shinytoyrobots, I've set up a meeting to walk through my recommendations for this page/section. @james-nash, will make sure you have visibility of any changes.

site/docs/getting-started/roadmap.mdx
shinytoyrobots: It's a good question. It's currently mirroring the component order here - https://github.com/orgs/jpmorganchase/projects/5/views/54. So it is roughly intended to convey prioritization order. But that prioritization order can change.

I will chat to @joshwooding about making our project board (or parts of it) public. Then we can put alphabetically here, and also link to the GitHub section that has more specific priority ordering.

joshwooding: I wouldn't change it to alphabetical since it's been ordered to represent priority. Let me know when you want to chat @shinytoyrobots

james-nash: We know it's priority order, but there's nothing on the page to make that apparent to readers, so it just looks like a jumbled list.

If alphabetical order or using a numbered list aren't options, then I suggest we at least add an introductory sentence along the lines of "Component work has been prioritized into the following order" (I'll let someone else come up with better wording, but hopefully you get the jist of what I'm trying to convey).

shinytoyrobots:
Cypress test is failing on CascadingMenu - which is not part of this PR so I assume separate issue waiting for resolution.

joshwooding (approved):

james-nash (approved):
LGTM :-)  🚀

PR #1140, opened 2023-02-13
Avatar refactor

Fercas123:


joshwooding (changes requested):
Can we split the PR into refactoring and moving, please? It makes it easier to review

packages/lab/stories/avatar.doc.stories.mdx
```diff
@@ -16,44 +16,84 @@ import {
...
 # Avatar
 
-An Avatar is a small circular graphical representation of a person through a personal profile picture or name initials.
+An Avatar is a small circular graphical representation of an entity through a profile picture or name initials.
+It has three variants to account for different types of data:
 
-## Default Avatar
+- Image is used when a photo is available;
+- Initials is used when a photo is unavailable but profile information is;
+- Default is used when neither photo nor profile information are available.
```
el-dav, line 24: Maybe change to 'A default Icon is used'

packages/lab/stories/avatar.doc.stories.mdx
```diff
@@ -16,44 +16,84 @@ import {
...
 
 <Canvas>
-  <Story id="lab-avatar--initials" />
+  <Story id="lab-avatar--with-initials" />
 </Canvas>
 
-# API
+#### Custom SVG
+
+You can pass a custom SVG as a children of the component.
```
el-dav, line 61: You can pass a custom SVG as children to the component to be used an an image.

el-dav (dismissed):

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,74 +1,52 @@
...
 import { clsx } from "clsx";
-import { forwardRef, HTMLAttributes, ImgHTMLAttributes } from "react";
-import { classBase } from "./internal/constants";
-import { DefaultAvatar } from "./internal/DefaultAvatar";
-import { useLoaded } from "./internal/useLoaded";
+import { forwardRef, HTMLAttributes, ReactNode } from "react";
 
 import "./Avatar.css";
 
+const classBase = "saltAvatar";
```
joshwooding, line 8: Can we inline this?

packages/lab/stories/avatar.stories.tsx
```diff
@@ -1,36 +1,83 @@
...
+import { CSSProperties, ReactNode, useEffect, useState } from "react";
+import { Avatar } from "@salt-ds/lab";
 
 export default {
   title: "Lab/Avatar",
   component: Avatar,
 } as ComponentMeta<typeof Avatar>;
+const sizes = [1, 2, 3, 4] as const;
+
+const imgSrc =
```
joshwooding, line 13: This url will need to change

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,74 +1,52 @@
...
-    id: idProp,
-    size = "medium",
-    src,
-    srcSet,
-    title = "user",
-    imgProps,
-    sizes,
+    size = DEFAULT_AVATAR_SIZE,
+    style: styleProp,
+    fallbackIcon = <UserIcon />,
```
joshwooding, line 25: What is the use case for fallbackIcon? I think it would make sense to build an example using it. In my mind, I'm trying to figure out why this wouldn't just be a default to children.

el-dav: I think I agree. The idea was previously that you provide the Avatar all the information (imgSrc, initial data and fallbackIcon) and it would sort out what gets displayed in order of preference. But having moved the responsibility for the image loading to the consumer it can probably all be children.

el-dav: Related question is we want to support both centered Icons e.g. the default, and also svgs which take up the full background circle of the Avatar (primarily for use in the Flag components).

Currently that's controlled by where it has the salt icon class. Are we happy with that logic?

packages/lab/src/avatar/Avatar.css
```diff
@@ -2,65 +2,36 @@
 .saltAvatar {
   --avatar-background: var(--saltAvatar-background, var(--salt-accent-background));
   --avatar-foreground: var(--saltAvatar-foreground, var(--salt-accent-foreground));
```
libertymayc, line 4: I'd say to move these to below directly in the attribute, no need for a local token since it's only being used in 1 place

libertymayc: as in just have `background: var(--saltAvatar-background, var(--salt-accent-background));` below, + same for foreground

packages/lab/stories/avatar/avatar.doc.stories.mdx
```diff
@@ -16,44 +16,90 @@ import {
...
 <Canvas>
-  <Story id="lab-avatar--image" />
+  <Story id="lab-avatar--with-image" />
 </Canvas>
 
-## Initials Avatar
+#### Initials
 
-You can pass the initials as children
+You can pass the initials as a children of the component.
```
libertymayc, line 53:
```suggestion
You can pass the initials as the children of the component.
```

libertymayc: same for above 'a children' maybe?

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,74 +1,88 @@
...
-  sizes?: string;
+  /**
+   * The name that Avatar represents.
+   */
+  name?: string;
+  /**
+   * Defines the function that gets initials. Default is capital first letter of each separate word in name.
+   * If a function is not passed or returns undefined, Avatar will default to Icon.
+   */
+  initialsGetter?: InitialsGetter;
```
joshwooding, line 20: We should think about the name of this

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,74 +1,88 @@
...
 }
 
-const withBaseName = makePrefixer(classBase);
+const withBaseName = makePrefixer("saltAvatar");
+const DEFAULT_AVATAR_SIZE = 2; // medium
+
+const initialsDefaultGetter = (name?: string) =>
+  name
+    ?.split(" ")
+    .slice(0, 2)
```
joshwooding, line 41: Still unsure about this limit

packages/lab/stories/avatar/avatar.stories.tsx
```diff
@@ -1,36 +1,81 @@
...
+    <path d="M12 6A6 6 0 1 1 0 6a6 6 0 0 1 12 0zm-7.207 4.853c-.7-.92-1.199-1.877-1.488-2.853H1.416a5.008 5.008 0 0 0 3.377 2.853zM1.1 7h1.984a6.978 6.978 0 0 1-.002-2H1.099a4.995 4.995 0 0 0 0 2zm.316-3h1.887c.288-.976.787-1.932 1.488-2.853A5.008 5.008 0 0 0 1.417 4zm2.936 0h3.262c-.329-.911-.876-1.819-1.655-2.709C5.197 2.181 4.666 3.09 4.352 4zm4.577 3H10.9a4.995 4.995 0 0 0 0-2H8.903a6.93 6.93 0 0 1 .027 2zm-1.634 3.831A5.01 5.01 0 0 0 10.584 8H8.722c-.273.968-.752 1.917-1.428 2.831zM7.677 8H4.356c.325.942.883 1.882 1.686 2.802C6.827 9.881 7.369 8.941 7.678 8zm.992-4h1.915A5.01 5.01 0 0 0 7.12 1.126C7.845 2.052 8.364 3.015 8.669 4zM4.095 5a5.94 5.94 0 0 0 .001 2h3.822a5.904 5.904 0 0 0-.031-2H4.094z" />
+  </svg>
+);
+
+export const WithCustomSvg = Template.bind({});
+WithCustomSvg.args = {
+  children: CustomSVG,
+};
+
+export const ImageFallback: ComponentStory<typeof Avatar> = () => {
```
joshwooding, line 64: Can this be simplified now?

packages/lab/src/avatar/Avatar.tsx
Fercas123: `initialsReducer` `nameShortener` `initialsExtractor` ? 🤷‍♀️ any suggestions?

packages/lab/stories/avatar/avatar.stories.tsx
Fercas123: there is a simplified one `WithImage` that uses src, but this one has the image as children so you can pass it with props but would still need to handle failure.

packages/lab/src/avatar/Avatar.tsx
pseys: We need to retain the 2 character limit with the new component. Although it would be nice to offer more it adds complexity. If the characters are, for instance WMW, they'd be too wide for the avatar and would either overlap the edge or have to wrap. Alternatively we'd have to look into ways the font size could be reduced to accommodate 3+ characters which would be further work. 

This isn't to say we shouldn't look into it in the future, as a person with a double-barrelled surname I can see them value in doing it, but for now I don't think its a priority.

packages/lab/src/avatar/Avatar.css
```diff
@@ -1,66 +1,36 @@
...
+  --avatar-base-fontSize: var(--saltAvatar-fontSize, var(--salt-accent-fontSize));
+  --avatar-container-size: calc(var(--avatar-base-size) * var(--avatar-size-multiplier));
+  --avatar-fontSize: calc(var(--avatar-base-fontSize) * var(--avatar-size-multiplier));
+  /* Icon styling */
+  --saltIcon-color: var(--avatar-foreground);
+  --salt-size-icon-base: calc(var(--avatar-container-size) / 2);
 }
 
 /* Style applied to the root element */
 .saltAvatar {
```
libertymayc, line 15: I think we should be adding line-height here (--salt-accent-lineHeight), i'm going to tag in figma specs also

sion-docs (approved):

packages/lab/src/avatar/Avatar.css
Fercas123: I tried `--salt-accent-lineHeight`, and it is not noticeable in the other densities, but it seems to be pushing the text up from the centre in high density, specially in size 1

libertymayc: hmm, I'm not noticing a difference myself when I add it in storybook - could you screenshot?
I think even if it's not noticeable it's important in case it's nested inside other components that set the line height or the global line height is changed

el-dav (dismissed):

packages/lab/src/avatar/Avatar.css
Fercas123: it was a tiny move, but I build again and it seems fine now, so i think it was something in my local :) 

image

lineHeight set: 
image

packages/lab/src/avatar/Avatar.css
```diff
@@ -1,66 +1,37 @@
...
-  background: var(--avatar-background);
+  line-height: var(--salt-accent-lineHeight);
```
libertymayc, line 16: if we expose --saltAvatar-fontSize, we should also expose --saltAvatar-lineHeight

libertymayc, line 16:
```suggestion
  line-height: var(--saltAvatar-lineHeight, var(--salt-accent-lineHeight));
```

libertymayc (approved):

pseys:
Thanks for making the changes @Fercas123 all looks good to me.

Issue #1010, opened 2023-01-12
Data Grid: Basic column sorting


ZarkoPernar:
Currently we offer no sorting functionality that exists in every other grid library. As most things with our grid this can be developed in user land with customisable headers but we want to provide and easy and accessible experience out of the box.

Add the ability for a user to sort by a single activating the sort functionality on sortable columns, one at a time. Rows can be sorted based on a column value in ascending and descending order.

Multiple composable sorting columns will not be developed at this time.

```[tasklist]
### Tasks
- [ ] https://github.com/jpmorganchase/salt-ds/issues/1395
```

parsakhan21:
2/13 - dev in progress - should have PR up soon for review

joshwooding:
Addressing current feedback, dev review needed shortly

libertymayc:
> @feliciackh i noticed the sort arrow icon overlaps the label when resizing the column. can we honor the base padding (per density) on either side of the icon, so there's space between the icon and the ellipses whether left or right-aligned?
> 
> I have screenshots but it keeps failing when trying to upload... not sure why... so will send them via symphony

would you be able to add size specs to the characteristics page for grid so @feliciackh knows where to use which :)

joshwooding:
@bhoppers2008 and @dplsek to catch up on the grid size specs. @feliciackh fixing a bug found in design review

dplsek:
@feliciackh @libertymayc I've added the padding to the 'Density examples' where we show padding. The first column of each grid example includes a sort arrow (scaled as necessary). The left set shows with padding visible and the right set with the padding hidden. The same padding rules apply whether the icon is left aligned in the header or right aligned. Let me know if this works for you. Cool.

Light:
[](https://www.figma.com/file/xqqKG90bV9yT8xBJvWayzJ/Data-Grid-V1?node-id=827%3A9905&t=Ym8CdkuugSH0Zbt8-1)

Dark:
[](https://www.figma.com/file/xqqKG90bV9yT8xBJvWayzJ/Data-Grid-V1?node-id=839%3A20398&t=Ym8CdkuugSH0Zbt8-1)

feliciackh:
> @feliciackh @libertymayc I've added the padding to the 'Density examples' where we show padding. The first column of each grid example includes a sort arrow (scaled as necessary). The left set shows with padding visible and the right set with the padding hidden. The same padding rules apply whether the icon is left aligned in the header or right aligned. Let me know if this works for you. Cool.
[...]

PR #1141, opened 2023-02-13
[Data Grid] implement sorting on grid columns

feliciackh:
#1010

## Description

- [x] Add sort functionality to grid columns
- [x] Sort order: `ascending`, `descending` or `none`
- [x] Sort columns using either:
1) values from GridColumn `getValue` prop
2) values from `rowData`
3) custom sorting function
4) server side sorting using `onSortOrderChange `
- [x] Cypress test for sorting
- [x] Update storybook documentation

## Examples
story/data-grid-data-grid--server-side-sort
story/data-grid-data-grid--client-side-sort

## Update
Design approved by Darrin
Content reviewed by Sion
_[WIP] ADA to be reviewed by Stella_

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -447,10 +447,10 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
       return r[sortByColumnId as keyof typeof r];
     });
 
+  const customSortingFn = getColById(sortByColumnId)?.info.props.customSort;
+
   const sortedRowData = useMemo(() => {
     if (!isSortMode || onSortOrderChange) return rowData;
-
-    const customSortingFn = getColById(sortByColumnId)?.info.props.customSort;
 
     if (customSortingFn) {
       return customSortingFn({ rowData, sortOrder });
@@ -468,6 +468,7 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
   }, [
     onSortOrderChange,
     valueGetter,
+    customSortingFn,
     isSortMode,
     rowData,
     sortByColumnId,
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -438,10 +438,16 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
 
   const isSortMode = sortByColumnId && sortOrder !== SortOrder.NONE;
 
-  const sortedRowData = useMemo(() => {
-    const onSortOrderChange =
-      getColById(sortByColumnId)?.info.props.onSortOrderChange;
+  const onSortOrderChange =
+    getColById(sortByColumnId)?.info.props.onSortOrderChange;
 
+  const valueGetter =
+    getColById(sortByColumnId)?.info.props.getValue ||
+    ((r: T) => {
+      return r[sortByColumnId as keyof typeof r];
+    });
+
+  const sortedRowData = useMemo(() => {
     if (!isSortMode || onSortOrderChange) return rowData;
 
     const customSortingFn = getColById(sortByColumnId)?.info.props.customSort;
@@ -449,12 +455,6 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     if (customSortingFn) {
       return customSortingFn({ rowData, sortOrder });
     }
-
-    const valueGetter =
-      getColById(sortByColumnId)?.info.props.getValue ||
-      ((r: T) => {
-        return r[sortByColumnId as keyof typeof r];
-      });
 
     const sortedData = [...rowData].sort((a, b) =>
       valueGetter(a) < valueGetter(b) ? -1 : 1
@@ -465,7 +465,14 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     }
 
     return sortedData;
-  }, [getColById, isSortMode, rowData, sortByColumnId, sortOrder]);
+  }, [
+    onSortOrderChange,
+    valueGetter,
+    isSortMode,
+    rowData,
+    sortByColumnId,
+    sortOrder,
+  ]);
 
   const onClickSortColumn = useCallback(
     (colHeaderId: GridColumnProps["id"]) => {
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -11,7 +11,7 @@ import React, {
   useState,
 } from "react";
 import { makePrefixer } from "@salt-ds/core";
-import { GridColumnInfo } from "./GridColumn";
+import { GridColumnInfo, GridColumnProps } from "./GridColumn";
 import { GridContext } from "./GridContext";
 import { clsx } from "clsx";
 import {
@@ -57,6 +57,7 @@ import { ColumnDragContext } from "./ColumnDragContext";
 import { ColumnGhost } from "./internal/ColumnGhost";
 import { ColumnDropTarget } from "./internal/ColumnDropTarget";
 import { ColumnDataContext } from "./ColumnDataContext";
+import { ColumnSortContext } from "./ColumnSortContext";
 
 const withBaseName = makePrefixer("saltGrid");
 
@@ -65,6 +66,12 @@ export type ColumnGroupRowSeparatorType = "first" | "regular" | "last";
 export type ColumnGroupColumnSeparatorType = "regular" | "none" | "pinned";
 export type GridRowSelectionMode = "single" | "multi" | "none";
 export type GridCellSelectionMode = "range" | "none";
+
+export enum SortOrder {
+  ASC = "asc",
+  DESC = "desc",
+  NONE = "none",
+}
 
 export type RowKeyGetter<T> = (row: T, index: number) => string;
 
@@ -226,6 +233,10 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     headerIsFocusable ? "header" : "body"
   );
 
+  const [sortByColumnId, setSortByColumnId] =
+    useState<GridColumnProps["id"]>("");
+  const [sortOrder, setSortOrder] = useState<SortOrder>(SortOrder.NONE);
+
   const [editMode, setEditMode] = useState<boolean>(false);
   const [initialText, setInitialText] = useState<string | undefined>(undefined);
 
@@ -293,8 +304,8 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
 
   const midGrpByColId = useMemo(() => {
     const m = new Map<string, GridColumnGroupModel>();
-    for (let g of midGroups) {
-      for (let c of g.childrenIds) {
+    for (const g of midGroups) {
+      for (const c of g.childrenIds) {
         m.set(c, g);
       }
     }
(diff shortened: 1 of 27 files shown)


packages/grid/src/HeaderCell.tsx
```diff
@@ -30,25 +38,77 @@ export function HeaderCell<T>(props: HeaderCellProps<T>) {
...
+      <FlexLayout className={className} justify={justify} aria-hidden>
+        {sortOrder === "asc" && <ArrowUpIcon />}
+        {sortOrder === "desc" && <ArrowDownIcon />}
+      </FlexLayout>
+    );
+
+    return icon;
+  };
+
+  const ariaSort =
```
joshwooding, line 62: This would be clearer as a map or a switch e.g.

```ts
const map = {
"asc": "ascending",
"desc": "descending"
}

map[sortOrder] || "none"
```

joshwooding, line 62: This would be clearer as a map or a switch e.g.

```ts
const map = {
"asc": "ascending",
"desc": "descending"
}

map[sortOrder] || "none"
```

packages/grid/src/GridColumn.tsx
```diff
@@ -54,6 +54,20 @@ export interface GridColumnProps<T = any> {
...
+   * To enable column header's keyboard navigation on sort,
+   * users need to set `headerIsFocusable` prop to `true` in Grid component.
+   * */
+  isSortable?: boolean;
+  /**
+   * Custom sorting function.
+   * */
+  customSort?: (args: {
+    rowData: T[];
+    sortByColumnId: string;
```
ZarkoPernar, line 68: I don't think this is necessary. If this is a prop for the GridColumn, then it doesn't make sense because this function will only get called on this same column and therefore always be the same.

packages/grid/src/HeaderCell.css
```diff
@@ -88,3 +88,7 @@
   cursor: col-resize;
   z-index: 2;
 }
+
+[aria-label="sort column"] {
```
ZarkoPernar, line 92: Id target this with a class name instead. The label might change or be customisable.

packages/grid/src/GridColumn.tsx
feliciackh: true that!

packages/grid/src/HeaderCell.tsx
feliciackh: updated to use a map

packages/grid/stories/grid-sortColumns.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+        getValue={(rowData: Investor) => rowData.date}
+        isSortable
+      />
+      <GridColumn
+        name="Score"
+        id="score"
+        getValue={(r) => r.score}
+        onChange={onScoreChange}
+        isSortable
+        customSort={({ rowData, sortOrder }) => {
```
joshwooding, line 56: It would be good to create an example that focuses on data that isn't local and how that interaction would work

feliciackh: @joshwooding good idea - have added that in latest commit. Do let me know if it's similar to what you had in mind

packages/grid/src/Grid.tsx
```diff
@@ -226,6 +228,10 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     headerIsFocusable ? "header" : "body"
   );
 
+  const [sortByColumnId, setSortByColumnId] =
+    useState<GridColumnProps["id"]>("");
```
joshwooding, line 232: Could "" be a valid id?

packages/grid/src/GridColumn.tsx
```diff
@@ -54,6 +54,16 @@ export interface GridColumnProps<T = any> {
    * Unique identifier of the column.
    * */
   id: string; // TODO make optional
+  /**
+   * Enables sorting for the column.
+   * To enable column header's keyboard navigation on sort,
+   * users need to set `headerIsFocusable` prop to `true` in Grid component.
+   * */
+  isSortable?: boolean;
```
joshwooding, line 62: Normally we don't prefix our variable names with `"is"` see `pinned` below.

packages/grid/src/HeaderCell.tsx
```diff
@@ -30,25 +38,82 @@ export function HeaderCell<T>(props: HeaderCellProps<T>) {
...
+  const valueAlignRight = align === "right";
+
+  interface HeaderCellSortingIconProps {
+    justify: FlexContentAlignment;
+  }
+
+  const HeaderCellSortingIcon = ({ justify }: HeaderCellSortingIconProps) => {
+    const className = withBaseName("sortingIcon");
+    const icon = (
+      <FlexLayout className={className} justify={justify} aria-hidden>
```
joshwooding, line 53: Do we need to use FlexLayout here?

packages/grid/src/HeaderCell.tsx
```diff
@@ -30,25 +38,82 @@ export function HeaderCell<T>(props: HeaderCellProps<T>) {
...
+        {sortOrder === "asc" && <ArrowUpIcon />}
+        {sortOrder === "desc" && <ArrowDownIcon />}
+      </FlexLayout>
+    );
+
+    return icon;
+  };
+
+  type AriaSortProps = "none" | "ascending" | "descending";
```
joshwooding, line 62: These should be moved out of the component

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -409,90 +409,40 @@ describe("Grid", () => {
     });
   });
 
-  describe("Row Selection", () => {
-    describe("Uncontrolled & switching selection modes", () => {
-      it("Shows correct columns", () => {
```
joshwooding, line 414: Do we not have to test this anymore?

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -606,4 +556,44 @@ describe("Grid", () => {
     checkRowSelected(2, true);
     checkRowSelected(3, true);
   });
+
+  describe("Column Sorting", () => {
+    it.only("should sort column values when sorting is enabled", () => {
```
joshwooding, line 561:
```suggestion
    it("should sort column values when sorting is enabled", () => {
```

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -606,4 +556,44 @@ describe("Grid", () => {
...
+      cy.mount(<RowSelectionModes />);
+
+      // first click: sort in ascending order
+      cy.findAllByTestId("column-header")
+        .eq(2)
+        .should("have.text", "B")
+        .as("sortableHeader")
+        .click()
+        .should("have.attr", "aria-sort", "ascending")
+        .then(() => {
```
joshwooding, line 571: Was this `then()` necessary?

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
```diff
@@ -606,4 +556,44 @@ describe("Grid", () => {
...
+
+  describe("Column Sorting", () => {
+    it.only("should sort column values when sorting is enabled", () => {
+      cy.mount(<RowSelectionModes />);
+
+      // first click: sort in ascending order
+      cy.findAllByTestId("column-header")
+        .eq(2)
+        .should("have.text", "B")
+        .as("sortableHeader")
```
joshwooding, line 568: What role does this `as()` have? Usually, it would "save" a query for later, but here we don't use it later.

packages/grid/stories/grid-sortColumns.stories.tsx
joshwooding: This example still shows sorting on the client. Even though the initial data is fetched with useSWR, the sort doesn't re-fetch the data from the server sorted but instead works on data it already has locally.

joshwooding (changes requested):

packages/grid/src/HeaderCell.tsx
ZarkoPernar: I assume we could extract this from the standard lib instead of defining it ourselves? Like HTMLTableRowAttributes["aria-sort"] or something like that.

packages/grid/src/Grid.tsx
feliciackh: Hmm not sure how that could be done, as on first sortable column header click, that's when the `colHeaderId` to be sorted gets set, it would be difficult to have an initial `colHeaderId` set as the initial state value - do you have something else in mind?

packages/grid/src/HeaderCell.tsx
feliciackh: I've moved it out of the component instead as was not having much luck on typescript's end using the types `HTMLTableRowElement["ariaSort"]`

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
feliciackh: nice one - could do with just `findCell` directly

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
feliciackh: removed that - tbh I forgot why it was added initially but it makes sense, reading other use cases of `as()` it's not needed in here

packages/grid/src/__tests__/__e2e__/Grid.cy.tsx
feliciackh: We do - these tests that have been removed were simply repeated. You could see them present on lines 
https://github.com/jpmorganchase/salt-ds/pull/1141/files/37d58c0e9cda7534ed9120f85997d6a60993dc68#diff-70083ed0f96d8fa99e5060f265de64108b3cfaa8922bdd71dc2f16a900015665R458
and
https://github.com/jpmorganchase/salt-ds/pull/1141/files/37d58c0e9cda7534ed9120f85997d6a60993dc68#diff-70083ed0f96d8fa99e5060f265de64108b3cfaa8922bdd71dc2f16a900015665R413

packages/grid/src/HeaderCell.tsx
```diff
@@ -30,25 +39,84 @@ export function HeaderCell<T>(props: HeaderCellProps<T>) {
...
+  interface HeaderCellSortingIconProps {
+    justify: FlexContentAlignment;
+  }
+
+  const HeaderCellSortingIcon = ({ justify }: HeaderCellSortingIconProps) => {
+    const className = withBaseName("sortingIcon");
+    const icon = (
+      <div
+        className={className}
+        style={{ display: "flex", justifyContent: `${justify}` }}
```
joshwooding, line 56: We try to avoid inline styles. Can this be done via CSS?

feliciackh: noted and yes!

packages/grid/stories/grid-sortColumns.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+    dummyInvestors[rowIndex].score = value;
+  };
+
+  return (
+    <div
+      style={{
+        position: "relative",
+        // @ts-ignore
+        "--salt-overlayable-background":
+          "var(--salt-color-gray-20-fade-background-readonly)",
```
libertymayc, line 78: better to use palette token `--salt-palette-neutral-secondary-background-readonly`

feliciackh: as discussed with Darrin, we'll use for each mode these values for the time being until secondary scrims are worked on
light mode: `--salt-color-gray-60-fade-background`
dark mode: `--salt-color-gray-300-fade-background`

joshwooding (dismissed):

joshwooding (dismissed):

joshwooding (approved):

PR #1145, opened 2023-02-14
Text disabled prop

libertymayc:
Adds a disabled prop to Text so that we have a clean consistent way to add disabled text styling when Text is nested inside other disabled components.


joshwooding (approved):

joshwooding:
Nice work!

PR #1155, opened 2023-02-15
Component documentation template

james-nash:
Adds MDX files that outline the sections that need to be completed for each component's docs page.

### DIFF
--- a/site/docs/components/_content-template/index.mdx
+++ b/site/docs/components/_content-template/index.mdx
@@ -1,0 +1,7 @@
+# Component Name
+
+## Usage
+
+## Exmaples
+
+...


james-nash (commented):
Comments from review call with @shey-v & @bhoppers2008

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+  Add a bulleted list of any other design-related guidance the user may need to
+  consider.
+-->
+
+### Additional materials
+
+<!--
+  Add a bulleted list of links to additional, component-specific materials:
+
+  * [Full component specification (in Figma)](https://figma.com/path/to/spec)
```
shey-v, line 183: Worth noting, these component specs will only be available to internal JPMC users.

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+<!--
+  Add a bulleted list of any other design-related guidance the user may need to
+  consider.
+-->
+
+### Additional materials
+
+<!--
+  Add a bulleted list of links to additional, component-specific materials:
```
shey-v, line 181: May have to differentiate between internal vs external user and link each consumer group out to relevant docs?

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+
+<!--
+  Show the typical import statement used to include this component in your
+  code.
+-->
+
+```js
+import { ComponentName } from "@some-org/some-package";
+```
+
```
shey-v, line 200: After import section, could we list out a bunch of component specific features here? before the props table, in a bulleted list?

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+- ...
+
+## Designing
+
+### Theming considerations
+
+<!--
+  Describe how Salt characteristics and theming impact the component, and any
+  modifications that can be made.
+-->
```
james-nash, line 169: @bhoppers2008 & @shey-v I wonder if we should omit this "Theming considerations" section as well, for similar reasons to the anatomy and spacing spec ones.

site/docs/components/_content-template/index.mdx
james-nash: Good spot. In that case, I'll remove that as a suggestion for now.

(Out of curiosity: Is there any reason those specs are internal only? What's stopping us making them public?)

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+  section. Include links out to useful articles.
+-->
+
+- [Relevant industry standard](https://example-standard.org/)
+- [Useful article](https://some-site.com/article/bla)
+- ...
+
+## Designing
+
+### Theming considerations
```
shey-v, line 164: I'm not sure if this needs to be moved to the developing section, what specifically are we going to talk about here? If an example could be provided that would be great. Happy to leave as is for now, but still unclear on what content will live here.

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,218 @@
...
+```js
+import { ComponentName } from "@some-org/some-package";
+```
+
+### Props
+
+<!--
+  A table of the props for this component in the following format:
+   
+| Prop  | Description | Data type.          | Default  | Mandatory (?)  |
```
shey-v, line 206: In these table headings, do we need 'Data type' and 'Mandatory' ?

site/docs/components/_content-template/index.mdx
james-nash: We're quite a way off having the technical means to add JPMC-only content on the site, so I'd suggest that this needs to be limited to only public links for now (but I can add a note in the guidance to that effect).

shey-v: One for @bhoppers2008 to go into more detail on, I believe we don't share this info out in the open as it is more a working file for us as designers to re-design and re-build our components whilst also producing a spec for handoff to devs and it may contain sensitive data that's only relevant for JPMC designers.

site/docs/components/_content-template/index.mdx
james-nash: Yes. Those are important things a developer needs to know in order to use the API.

However, given that this table (including whatever columns it has) is going to be auto-generated, maybe we can remove that manual example altogether.

Then the guidance for this would simply be: "Auto-generated props table goes here" :-)

site/docs/components/_content-template/index.mdx
shey-v: One for @bhoppers2008 to go into more detail on, however, I believe we don't share this info out in the open as it is more a working file for us as designers to re-think, re-design and re-build our Figma components whilst also producing a spec for developer handoff. It also may contain sensitive data that's only relevant for JPMC designers.

site/docs/components/_content-template/index.mdx
james-nash: As per my earlier comment below, I think we can remove this section altogether.

site/docs/components/_content-template/index.mdx
shey-v: I do think we could keep in the characteristics table for now, to show what tokens have been used. I've currently got it in my wireframes and it works quite nicely, however let's discuss this one in the call we have with Liberty and Paul next week around how useful this is for designers/developer partners. Thoughts?

site/docs/components/_content-template/index.mdx
james-nash: I'm not sure what you mean. Can you provide an example of what you have in mind? Also, how would this differ from what's being conveyed in the "Configurations" section?

shey-v: Sure, have tagged you in Figma. Actually now to think about it, it wont differ from the 'configurations' section, it just means when the time comes to rolling out the new template, I move that info somewhere else, instead of having a description and live example for each feature :)

site/docs/components/_content-template/index.mdx
james-nash: I'm leaning towards review this for now. Partly because that meeting is next week and we'd like to get this PR merged this week.

I think there's a very good chance some kind of styles / design token information will be added in the future, but I think it'll be a while before we have a good handle on what consumers need to know and how best to give that to them. Also, similar to the props tables, I'm fairly confident this will be something that can and should be auto-generated from code and/or Figma. In that case, there might not be anything a component team needs to provide at all as we can just generate this section "for free".

TL;DR: I still want to remove this section. Any objections?

sion-docs: No objections, we definitely need guidance that will hopefully come next week

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,213 @@
...
+<!--
+  Describe how Salt characteristics and theming impact the component, and any
+  modifications that can be made.
+-->
+
+### Design guidance
+
+<!--
+  Add a bulleted list of any other design-related guidance the user may need to
+  consider.
```
james-nash, line 175: Would be nice if we can provide a short example of the kind of guidance that would included here. Bear in mind we already explain the component's purpose, dos and don'ts further up the page - so what _else_ is it that would go here?

@shey-v, @bhoppers2008 if you have suggestions we could include here, that would be awesome!

Also, might this section include Figma component usage instructions (e.g. if there are non-intuitive quirks designers need to be aware of in order to successfully use this component in Figma)? Or would that kind of thing deserve a new (optional) section of its own?

shey-v: I'm keen to move do's and don'ts (when to use and when not to use) down here as it is specific to design usage, patterns and the component behaviour. 

i.e. https://spectrum.adobe.com/page/button/#Usage-guidelines
i.e. https://atlassian.design/components/button/usage#best-practices

This kind of content would be relevant here. 

From what I've seen, each component in the Community space and internally, comes with Figma component guides which tell the user how to configure the component to the variants and states they need etc. I think perhaps this could go into it's own section or just be mentioned in the copy, i.e. refer to the component guides to view Fgima component usage instructions?

sion-docs:
Should we mandate/suggest that they add some internal navigation/anchor links? Docusaurus gives us this out of the box but these are potentially long complex pages without internal nav.

site/docs/components/_content-template/index.mdx
james-nash: Cool. We'll leave this as is for now then.

site/docs/components/_content-template/index.mdx
james-nash: I'd prefer not to move this stuff under a section labelled "Design(ing)" as it's relevant information for product & dev people too. My worry is that people in those roles wouldn't think to look at that section/tab since they don't indentify as designers.

I therefore propose we create a "Usage" (or "Guidelines" or something else to that effect) section and move the Dos & Don'ts and also this references section in there. Hopefully that would be perceived as more inclusive and so not only designers, but others will look at this too.

The Atlassian example you linked to does this. Others like IBM Carbon & Orbit Kiwi also do this and I really like it:

screenshot of a Carbon component page featuring a "Usage" tab

screenshot of a Kiwi component page featuring a "Guidelines" tab

james-nash:
> Should we mandate/suggest that they add some internal navigation/anchor links? Docusaurus gives us this out of the box but these are potentially long complex pages without internal nav.

@sion-docs I'd prefer to keep that sort of thing out of the template. My thinking is that it should just contain the "raw" content that we need. How we present that, including in-page navigation aids like TOC links should be automatic.

As an interim solution, while this stuff is still published on Storybook, we could perhaps look into adding TOC links. I found an addon that might be worth a go: https://storybook.js.org/addons/storybook-docs-toc

site/docs/components/_content-template/index.mdx
shey-v: Fair enough, I'll have to change my designs.

sion-docs: I do think this info is particularly useful for users who aren't designers. I'm happy to have another heading but I'm not fond of the word 'Usage' as it's a bit vague. The whole page is about Usage really.

james-nash: OK. I've pushed an update that does what I suggested (I went with "Usage" as the title for now, but happy to change that if others have better suggestions).

@sion-docs & @shey-v please take a look and let me know what you think. Cheers!

shey-v: I was thinking about seperating this info out into 'when to use' and 'when not to use'. I feel like perhaps we need an 'Overview' tab, where this info lives, followed by 'Developing' then 'Designing' maybe?

sion-docs:
> > Should we mandate/suggest that they add some internal navigation/anchor links? Docusaurus gives us this out of the box but these are potentially long complex pages without internal nav.
> 
> 
> 
> @sion-docs I'd prefer to keep that sort of thing out of the template. My thinking is that it should just contain the "raw" content that we need. How we present that, including in-page navigation aids like TOC links is the job of the page template we'll design and build (and should be as automated as possible).
> 
> 
> 
> As an interim solution, while this stuff is still published on Storybook, we could perhaps look into adding auto-generated TOC links. I found an addon that might be worth a go: https://storybook.js.org/addons/storybook-docs-toc

Looks good, I would love to add this to our storybook docs - especially with the new structure

james-nash:
Given that we've done a number of localised changes, now might be a good time to take step back and see if the current heading outline makes sense. This preview's probably the most convenient way to see it: https://github.com/jpmorganchase/salt-ds/blob/component-doc-template/site/docs/components/_content-template/index.mdx (ignore the gibberish above the "Component Name" H1, that's just Github's markdown preview struggling with our metadata).

Questions for you to ponder:

* Are the heading levels correct?
    * E.g. do we all agree that "Content" and "Accessibility" should be at the same level (H2) as "Usage", "Designing" & "Developing"?
* Is the order correct?
    * E.g. do we all agree that "Developing" should come last of all?
    * Also consider that some sections (like "Content" and "Accessibility") are optional

sion-docs:
> 

Content and Accessibility should be under Usage.

site/docs/components/_content-template/index.mdx
shey-v: @sion-docs  @james-nash  In this section Designers would like to see:

* Best practises and high level guidance in terms of design principles.

* An anatomy diagram. This would be helpful for designers to clearly communicate what a component is made up of - particularly for components that are made up of a set of different components. I.e. a Form Field. + making it clear that an input is inside of a FormField and within that is an icon… etc. 

* A link to a sticker sheet. This would be helpful for designers to see ALL variations and permutations that are possible with the components, i.e. our current FF documentation page doesn’t show you all examples and possible alignments - label left for example 

* Sizing and spacing spec. This would be useful when designers who aren’t familiar with our densities and base X values… and considering there will be a new size of ultra hd in the release of the data grid component.. it would be a nice to have.

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,208 @@
...
+};
+
+# Component Name
+
+<!--
+  Example screenshot or live component demo here of the default/most common
+  usage.
+
+  For screenshots use Markdown:
+
```
bhoppers2008, line 61: I'm wondering if we should be more opinionated here and ask for live component, rather than image?

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,208 @@
...
+  use for the label text, recommended maximum character counts, etc.
+-->
+
+## Accessibility
+
+<!--
+  OPTIONAL: Remove this section if it is not applicable.
+
+  Describe any considerations the users may need to take into account when using
+  the component.
```
bhoppers2008, line 161: Maybe we could add some guidance about what is worth documenting, assuming we don't want every minute detail documented. These things come to mind as of value:

- Accessibility decisions that deviate from industry standard. (e.g., we've introduced an arrow key navigation, where usually is only tab).
- Accessibility requirements that drive design (e.g., how multi-column forms should wrap).
- Best practices regarding accessible content (e.g., error message announcements must not rely on descriptions that require the user to visually or audibly perceive content ).
- Best practices regarding implementation (e.g., to make sure this component is accessible be sure to provide a relevant alt text).

site/docs/components/_content-template/index.mdx
bhoppers2008: Does it actually help us to create silo'd views of these component pages. I'm beginning to think it makes more sense to separate content by intent, not subject. Using, Styling, Coding, etc. What we think might be of interest to a designer, might also be useful to dev, and vice versa. Everything Shey has linked to designing, could also fit under a 'Style' section, or spill into 'Usage', or even touch on 'code'. This overlaps with the conversation we had about design tokens yesterday... that's something yet to be defined, but would be good to agree the structure now, that would easily accommodate something related to 'tokens', when made 'visible'.

site/docs/components/_content-template/index.mdx
james-nash: Yeah, I agree. Will update the guidance accordingly.

site/docs/components/_content-template/index.mdx
james-nash: Thanks. I think I can add your suggestions almost verbatim!

site/docs/components/_content-template/index.mdx
james-nash: 💯 on using intent rather than subject or role. I'll have a stab at renaming / rejigging things in moment.

site/docs/components/_content-template/index.mdx
bhoppers2008: Does make me wonder about where accessibility should live though... @shinytoyrobots mentioned it shouldn't have its own section/tab. I agree with that perspective (since it touches on different aspects) _but_ then i find myself thinking about this as a template. Interspersing accessibility content (or call outs) seems like an opportunity for inconsistency to creep in, whilst a single section or tab, would be straightforward. Thoughts?

site/docs/components/_content-template/index.mdx
james-nash: Discussed with @shey-v on a call and we've agreed to reintroduce the Anatomy section as all research so far indicates there's a strong demand for that from designers.

Leaving away the sizing/spacing section for now though, since we need to better understand _why_ people want that. If we get sufficient clarity, we can add that in a future revision of this template.

site/docs/components/_content-template/index.mdx
```diff
@@ -0,0 +1,234 @@
...
+<!--
+  Add a bulleted list of best practices and princples that are relevant to this
+  component. Where other pages on the site explain those principles in more
+  detail, consider linking to them rather than repeating the explanation here.
+-->
+
+- ...
+- ...
+
+### Anatomy
```
james-nash, line 182: @shey-v & @sion-docs Should the Anatomy be moved up into the "Overview" section perhaps? I think it could be useful for content folks too, since it could also visualise areas where text is displayed. So, when the "Content" section says something like "limit Button labels to 99 characters", they could refer to the anatomy diagram to see exactly which part of the component we're referring to.

sion-docs: I like the idea of putting it there, are we dispensing with Designing and Developing as mentioned previously?

james-nash: Hmm.. maybe there needs to be a bigger re-jig.

How about this:

* **Overview** (provide a high-level overview of the component's purpose and structure
    * Live example
    * Brief description
    * **Anatomy**
* **Usage** (more detail to help people understand when and how to use the component)
    * **Dos**
    * **Dont's**
    * **Best practices** (can this be merged with or replace "References" below?)
    * **Content**
    * **Accessibility**
    * **References**
* **Configurations** (collection of live examples showcasing key features or config options of the component)
    * **_[Config example 1]_**
    * **_[Config example 2]_**
    * ...

* **Figma** (info and resources specific to the Figma version of this component)
    * Link to stickersheet
* **React** (props table and resources specific to the React version of this component)
    * **Import**
    * **Props**

sion-docs: I'm happy to merge Best practices and References under Best practices, I'm not sure how many components will have content for References.

bhoppers2008: I think that's a tough call to make without seeing the mock-ups. If the overview section is the page header, above tabs, i wouldn't put the anatomy in there. It's not high enough priority. 

What's the difference between Dos/Don'ts and Best practices? Aren't they the same thing? Suggest we merge those, rather than with References.

Can we lead with configurations? Feels more compelling if the examples are the first thing that you see on the page, rather than blocks of usage text. 

So i'd throw in a proposal:

- Overview (eg and description)
- Configs
- Usage
  - Dos/Dont's
  - Content
  - Accessibility
  - References
- Styling
  - Anatomy (serves as a key to the below styling attributes)
  - Spacing/Size (TBD)
  - Tokens (TBD)
  - +other design content shey had outlined.
- Code
  - Importing
  - Props props props

james-nash: Your proposal looks nice, @bhoppers2008. I'm happy to go with that for now.

My only slight question mark is: If design resources like a Figma stickersheet live under a section/tab/whatever called "Styling", is that where someone would go looking for it?

However, we can speculate here until the cows come home :-) So, let's go with this for now. Once we begin filling in this template and adjusting our page designs accordingly, we'll be able to begin testing it with users.

It may well be fine as is. And if it's not, I suspect it's just a case of moving a section somewhere else or renaming a title. So, worst case scenario we need to do some find-and-replace on a bunch of `.mdx` files at some point. I can live with that :-)

bhoppers2008: Totally agree.

A designer made a comment the other day that we don't do a good job of supporting their working practices... so maybe things like the figma link, repo link, site link, could live somewhere in the header, so they're front and centre and don't need any navigation. Like you say... something to consider.

james-nash: Just pushed `Outline v97.final.FINAL-for-real-this-time.pdf` :-)

If you're all happy please "approve" this PR. Thanks!

bhoppers2008 (approved):

sion-docs (approved):

shey-v (approved):

PR #1158, opened 2023-02-15
[RFC] Responsive props

el-dav:
See [Storybook examples](https://aaa0656c.saltdesignsystem-storybook.pages.dev/?path=/story/lab-responsive-props--make-responsive-example)

The idea here is to allow consumers to add responsive behaviour to their components without baking the logic into every component in our design system. This leaves room for an alternative/similar system based off container queries which could be more useful in certain circumstances e.g. micro-frontends.

Consumers would need to add a `<BreakpointProvider>` at the root of their application.

They can then use various tools to add responsive behaviour
- `makeResponsive` is a higher-order-component which will turn all the props of a given component into responsive props e.g:

```jsx
<Component color="white" />
````

could then be used as 

````jsx
<ResponsiveComponent color={{ default: "white", lg: "black" }} />
````

- `useResponsiveValue` is a hook which will return a resolved value based on the viewport e.g:

`const value = useResponsiveValue({ default: 'Small', lg: 'Large' })`

value will be 'Small' on small viewports and 'Large' on large viewports

- `useResponsiveProps` is similar to `useResponsive value` except you pass an object to the hook, and each key/value pair in that object will be turned into it's own responsive prop. E.g:

```
const value = useResponsiveProps({ 
    text: {
        default: 'Small',
        lg: 'Large'
    },
    color: {
        default: 'red',
        md: 'green'
    }
})
```

On small screens text is 'Small' and color is 'red'. On medium screens text is still 'Small' and but color is 'green'. On large screens text is 'Large' and color is 'green'.

- ResponsiveChildren is a component which allows you to pass different ReactNodes to be used at different viewport sizes E.g.
```jsx
<ResponsiveChildren md={<h1>Medium</h1>)>
    <button>Default</button>
</ResponsiveChildren>
```
will show a button by default but on screens above medium it will show an h1.

Other things to note
- All components/hooks are typed with the appropriate breakpoints
- a responsive configuration must pass a 'default' value which will be used as a fallback if none of the provided breakpoint values are matched
- You don't have to use a responsive object for every prop of a component wrapped with `makeResponsive`. You can just provide a value as usual if you don't want that prop to have responsive behaviour.
- The implementation uses `matchMedia` under the hood. Components only re-render on breakpoint changes
- If a consumer is using different viewports then they can create their own responsive system using `createResponsiveSystem` e.g:

```
 const { 
    BreakpointsProvider,
    useResponsiveProps,
    useResponsiveValue,
    makeResponsive,
    ResponsiveChildren
} = createResponsiveSystem({ tiny: 0, kindabig: 800, mahoosive: 1000 });
```

All the returned components/hooks/HOC will then be typed appropriately.

We should also think about how we could make this available. There are currently no dependancies on the design system itself, and the responsive utilities can be used on any component (including primitives). So it could be a stand-alone package.

### DIFF
--- a/packages/lab/src/breakpoints/BreakpointsProvider.tsx
+++ b/packages/lab/src/breakpoints/BreakpointsProvider.tsx
@@ -1,15 +1,12 @@
 import { ReactNode, useState, useEffect, Context, useMemo } from "react";
 
 import {
-  // Breakpoints,
   BpType,
   Breakpoint,
   OrderedBreakpoints,
   BreakpointsContextType,
   BreakpointsOrder,
 } from "./types";
-
-// import { BreakpointsContext } from "./BreakpointsContext";
 
 const getBreakpointsOrder = <BP extends BpType>(
   orderedBreakpoints: OrderedBreakpoints<BP>
@@ -47,6 +44,7 @@ type GetBreakpointMatchersConfig<BP extends BpType> = {
   breakpoints: BP;
 };
 
+// Generated the MediaQueryList matchers which will be used to detect the current breakpoint
 const getBreakpointMatchers = <BP extends BpType>({
   orderedBreakpoints,
   breakpoints,
@@ -102,6 +100,7 @@ export const createBreakpointsProvider = <BP extends BpType>(
   breakpoints: BP,
   BreakpointsContext: Context<BreakpointsContextType<BP>>
 ) => {
+  // Generate breakpoint metadata and matchers in outer scope to maintain referential integrity for react hooks
   const orderedBreakpoints = getOrderedBreakpoints(breakpoints);
   const breakpointOrder = getBreakpointsOrder(orderedBreakpoints);
   const breakpointMatchers = getBreakpointMatchers({
@@ -114,6 +113,7 @@ export const createBreakpointsProvider = <BP extends BpType>(
       getCurrentBreakpoint({ breakpointMatchers, orderedBreakpoints })
     );
 
+    // setup the event handlers for detecting breakpoint changes
     useEffect(() => {
       const matchersCleanerupFns: (() => void)[] = [];
 
--- a/packages/lab/src/breakpoints/BreakpointsProvider.tsx
+++ b/packages/lab/src/breakpoints/BreakpointsProvider.tsx
@@ -1,0 +1,159 @@
+import { ReactNode, useState, useEffect, Context, useMemo } from "react";
+
+import {
+  // Breakpoints,
+  BpType,
+  Breakpoint,
+  OrderedBreakpoints,
+  BreakpointsContextType,
+  BreakpointsOrder,
+} from "./types";
+
+// import { BreakpointsContext } from "./BreakpointsContext";
+
+const getBreakpointsOrder = <BP extends BpType>(
+  orderedBreakpoints: OrderedBreakpoints<BP>
+): BreakpointsOrder<BP> => {
+  const breakpointsOrder = Object.fromEntries(
+    orderedBreakpoints.map((breakpoint, order) => [breakpoint, order] as const)
+  );
+
+  return breakpointsOrder as BreakpointsOrder<BP>;
+};
+
+const getOrderedBreakpoints = <T extends Record<string, number>>(
+  breakpoints: T
+): (keyof T)[] => {
+  return Object.entries(breakpoints)
+    .sort(([bpA, bpAValue], [bpB, bpBValue]) => {
+      return bpAValue - bpBValue;
+    })
+    .map(([bp], order) => {
+      return bp;
+    });
+};
+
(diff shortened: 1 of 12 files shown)


packages/lab/stories/breakpoints.stories.tsx
```diff
@@ -0,0 +1,145 @@
...
+    default: "blue",
+    xs: "red",
+    sm: "green",
+    md: "tomato",
+    lg: "goldenrod",
+    xl: "cyan",
+  });
+
+  return (
+    <BreakpointsProvider>
```
Fercas123, line 67: could the component reuse the `BreakpointContext` from `SaltProvider`?

el-dav: I think it would replace the BreakpointContext if we're happy with this solution.

libertymayc:
LGTM!! this is really great stuff :)

packages/lab/src/breakpoints/BreakpointsProvider.tsx
```diff
@@ -0,0 +1,159 @@
...
+  if (!currentBreakpoint) {
+    throw new Error(
+      "Current viewport matches none of the specified breakpoints"
+    );
+  }
+
+  return currentBreakpoint;
+};
+
+export const createBreakpointsProvider = <BP extends BpType>(
```
joshwooding, line 99: Would the BreakpointsProvider replace the ViewportProvider?

Also I would change this name to match what it exports

packages/lab/stories/breakpoints.stories.tsx
joshwooding: Lol just seen this after my comment

packages/lab/src/breakpoints/ResponsiveChildren.tsx
```diff
@@ -0,0 +1,37 @@
...
+};
+/**
+ * ResponsiveChildren is a component which allows you to change the rendered component based on the viewport E.g.:
+ * <ResponsiveChildren
+ *    md={<h1>Medium Viewport Child</h1>}
+ * >
+ *    <h1>Default Child</h1>
+ * </ResponsiveChildren>`
+ */
+export const createResponsiveChildren = <BP extends BpType>(
```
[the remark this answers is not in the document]
el-dav: The main reason was you can't pass the responsive configuration object directly to a component e.g.
`<Component children={{ default: <div>Thing</div>, sm: <div>Thing2</div> }}` and `<Component>{{ default: <div>Thing</div>, sm: <div>Thing2</div> }}</Component> are both invalid in react, so this provides a convenient method.

el-dav:
> Nice work, I wonder if we want to offer the HOC. What benefits do you think it provides?

Yeah I think it's really valuable, similar to something like the `styled` HOC for styled components. You can make your sub-components responsive outside the main body of your component and then inline the responsive props making it easier to. rationalise the behaviour.

In theory you could also use it to wrap all of your components where they are defined similar to `memo`, or `forwardRef` to make them all responsive. Not saying we would do this on our side, but consumers might want to.

DavieReid:
Can you link to the Storybook examples please?  The link in the description takes me to a markdown cheatsheet for some reason...

Were any other solutions looked at before deciding on this API?  There are a few I can think of below:

1. Stiches - https://stitches.dev/docs/responsive-styles
2. Facepaint - https://github.com/emotion-js/facepaint
3. Vanilla Extract - https://seek-oss.github.io/braid-design-system/css/responsiveStyle/


One potential downside is that brreakpoints are triggered by screen *size*.  For touch devices it is usually better to use other methods e.g.

### OK for mobile
```
@media (max-width: 64rem) {
    // triggered when screen size is <= 64 rem
}
```

### Better for mobile
```
@media(hover:none) and (pointer: coarse) {
    // triggered when you can't hover and using a pointing device of limited accuracy (a finger)
}
```

el-dav:
> 

Thanks for the comment! The link is fixed now!

With regards to existing solutions the approach was definitely inspired by existing solutions e.g. Braid or MaterialUIs `sx` prop. But there were a couple of reasons why existing solutions weren't quite right: 

- Salt is built using vanilla css and css-variables, so a css-in-js solution isn't really appropriate
- Ideally we want to be able to change prop values as well as simply styling (e.g. conditionally render a component) so we need something which react can hook into and affects rendering

Having said that we recognise that depending on the application this might not be correct for the consumer project. This is why we avoided baking the logic into the components themselves, instead providing this as an optional tool which consumers can use if it's helpful.

i definitely get your point in terms of screen-size vs device. I think both are valid. Sometimes we want functionality to change based on screen size, even if the device type is the same. But changing behaviour based on device type is also interesting.

joshwooding:
Closing for now since we can use it as a reference from the issue. We can reopen once we're ready

PR #1159, opened 2023-02-15
guard controls that could cause storybook to error

Fercas123:
We have a few controls in storybook that cause an error if interacted with. 
- limiting to string when possible
- disabling controls when the option would be `set option`


alycrys (dismissed):

packages/core/stories/border-item.stories.tsx
```diff
@@ -8,12 +8,10 @@ export default {
     position: {
       control: { type: "select" },
     },
-    horizontalAlignment: {
-      control: { type: "select" },
-    },
-    verticalAlignment: {
-      control: { type: "select" },
+    as: {
```
joshwooding, line 11: Would it work to put these argTypes globally instead of in every component?

Fercas123: I guess it wouldn't harm, specially `as` and `ref` that will be in a bunch of places i would imagine

PR #1177, opened 2023-02-21
Checkbox

ImmanuelBaskaran:
Checkbox

Removed `CheckboxBase` and replaced with `Checkbox`
Added `error` prop for error state styling.

CheckboxGroup

Removed `legend` and `LegendProps` prop; will be implemented by FormField.
Replaced `row` prop with `direction` prop.
Added `wrap` prop.

CheckboxIcon

Added `error` prop for error state styling.
Added `disabled` prop for disabled state styling.


packages/lab/src/checkbox/Checkbox.tsx
```diff
@@ -1,77 +1,114 @@
...
-
 export const Checkbox = forwardRef<HTMLDivElement, CheckboxProps>(
-  function Checkbox(props, ref) {
-    const {
+  function Checkbox(
+    {
       checked: checkedProp,
-      className,
-      defaultChecked: defaultCheckedProp,
+      className: classNameProp,
```
libertymayc, line 40: I think there's no need to rename this

libertymayc (approved):
lgtm as initial refactor

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,46 @@
...
-.salt-density-high {
-  --checkbox-fontSize: 11px;
-}
 
 .saltCheckbox {
   --checkbox-label-marginTop: calc(var(--salt-size-unit) * var(--checkbox-top-multiplier));
   --checkbox-space-multiplier: var(--saltCheckbox-space-multiplier, 3);
   --checkbox-top-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
+  --salt-text-label-fontSize: var(--salt-text-fontSize);
+  --salt-text-label-lineHeight: var(--salt-text-lineHeight);
```
alycrys, line 8: you don't need these. the Label component will apply the text styles

ImmanuelBaskaran: I added those overrides because the label styles didn't match the previous snapshot, but I'll remove it for now.

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,44 @@
...
-}
-.salt-density-high {
-  --checkbox-fontSize: 11px;
-}
 
 .saltCheckbox {
-  --checkbox-label-marginTop: calc(var(--salt-size-unit) * var(--checkbox-top-multiplier));
   --checkbox-space-multiplier: var(--saltCheckbox-space-multiplier, 3);
   --checkbox-top-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
+  --checkbox-label-marginTop: calc(var(--salt-size-unit) * var(--checkbox-top-multiplier));
```
alycrys, line 6: I believe the checkbox should not have any margins

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,44 @@
...
+.saltCheckbox-input {
+  cursor: inherit;
+  height: 10px;
+  left: 0;
+  margin: 0;
+  opacity: 0;
+  padding: 0;
+  position: absolute;
+  top: 0;
+  width: 10px;
```
alycrys, line 42: don't think you need to set width and height on input

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,29 +1,26 @@
...
-  --checkbox-size: 16px;
   --checkbox-density-bar-y: 6px;
   --checkbox-density-bar-height: 2px;
 }
 
 .salt-density-touch {
-  --checkbox-size: 18px;
   --checkbox-density-bar-y: 6px;
   --checkbox-density-bar-height: 2px;
 }
```
alycrys, line 20: I think we might be able to get rid of all of the above. what do you think?

ImmanuelBaskaran: I'll have to find a characteristic that gives the same values

libertymayc: the density bar ones are so random, probably incorrect. the design spec should be updated with the right sizes for everything soon

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,39 @@
...
-}
-.salt-density-high {
-  --checkbox-fontSize: 11px;
-}
 
 .saltCheckbox {
-  --checkbox-label-marginTop: calc(var(--salt-size-unit) * var(--checkbox-top-multiplier));
   --checkbox-space-multiplier: var(--saltCheckbox-space-multiplier, 3);
   --checkbox-top-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
 }
```
libertymayc, line 6: i don't think the above two tokens are used

packages/lab/src/checkbox/CheckboxGroup.css
```diff
@@ -1,16 +1,21 @@
...
-  flex-direction: column;
-  vertical-align: top;
 }
 
-/* Styles applied to root component if `row={true}` */
+/* Styles applied to root component if `direction={"horizontal"}` */
 .saltCheckboxGroup-horizontal {
-  --saltControlLabel-right-marginRight: calc(var(--salt-size-unit) * 3);
+  display: flex;
+  gap: var(--salt-size-unit);
```
libertymayc, line 11:
```suggestion
  gap: calc(var(--salt-size-unit)*0.75)
```

libertymayc: spec updated

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,38 @@
...
-  --checkbox-fontSize: 14px;
-}
-.salt-density-high {
-  --checkbox-fontSize: 11px;
-}
 
 .saltCheckbox {
-  --checkbox-label-marginTop: calc(var(--salt-size-unit) * var(--checkbox-top-multiplier));
-  --checkbox-space-multiplier: var(--saltCheckbox-space-multiplier, 3);
   --checkbox-top-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
```
libertymayc, line 4: i think this token is a bit unnecessary?

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,38 @@
...
-  --checkbox-fontSize: 16px;
-}
-.salt-density-low {
-  --checkbox-fontSize: 14px;
-}
-.salt-density-high {
-  --checkbox-fontSize: 11px;
-}
 
 .saltCheckbox {
```
libertymayc, line 3:
```suggestion
.saltCheckbox {
align-items: center;
```

libertymayc: checkbox and label are misaligned on zoom in

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,38 @@
...
-.saltCheckbox .saltControlLabel-label {
-  vertical-align: baseline;
-  min-height: var(--saltCheckbox-label-minHeight, var(--salt-text-minHeight));
-}
-
-.saltCheckbox-disabled {
-  --saltControlLabel-color: var(--salt-text-primary-foreground-disabled);
+/* Styles applied to input element */
+.saltCheckbox-input {
+  cursor: inherit;
```
libertymayc, line 30: causes cursor on label - desired?

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,29 +1,26 @@
...
 
 .saltCheckboxIcon {
+  --checkbox-size: var(--salt-size-selectable);
   --checkbox-borderWidth: var(--salt-size-border);
   --checkbox-box-offset: calc(var(--checkbox-borderWidth) / 2);
   --checkbox-height: calc(var(--checkbox-viewbox) - var(--checkbox-borderWidth));
```
libertymayc, line 26: I don't think we need --checkbox-viewbox (below - see spinner for example on not needing that and it scales by density) , or this height token, or the borderwidth token (on checked, i think it should just change color, not sure though so we can wait for specs)

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -89,6 +100,17 @@
   --checkbox-icon-tick-fill: var(--saltCheckbox-icon-indeterminate-bar-color-disabled, var(--salt-selectable-foreground-partialDisabled));
 }
 
+/* Styles applied if `indeterminate={true}` and `error={true}` */
+.saltCheckboxIcon-error.saltCheckboxIcon-indeterminate {
+  --checkbox-icon-tick-fill: var(--saltCheckbox-icon-stroke-error, var(--salt-status-error-borderColor));
+}
+
+/* Styles applied to radio button icon when disabled and with error */
```
libertymayc, line 108: need to check this is a use case with designer

packages/lab/src/checkbox/Checkbox.tsx
```diff
@@ -1,79 +1,169 @@
...
+    };
+
     const handleChange = createChainedFunction(
-      onChange,
+      handleInternalChange,
       groupContext?.onChange
     );
 
-    let checked = checkedProp;
+    let isCheckchecked = checkedProp;
```
alycrys, line 118:
```suggestion
    let isChecked = checkedProp;
```

packages/lab/src/checkbox/Checkbox.tsx
```diff
@@ -1,79 +1,169 @@
...
-          className={`${classBase}-label`}
+        <input
+          aria-checked={indeterminate ? "mixed" : checked}
+          name={name}
+          value={value}
+          {...inputProps}
+          checked={checked}
+          className={withBaseName("input")}
+          data-indeterminate={indeterminate}
+          defaultChecked={defaultChecked}
```
alycrys, line 152: I believe we can only pass `checked` attribute to checkbox

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,80 +1,88 @@
...
+}
+
+.saltCheckbox-error:hover .saltCheckboxIcon,
+.saltCheckboxIcon-error:hover {
+  background: var(--salt-palette-error-background-emphasize);
+}
+
+/* Styles applied if `error={true}` */
+.saltCheckbox-error.saltCheckbox-disabled .saltCheckboxIcon,
+.saltCheckbox-error.saltCheckbox-disabled:hover .saltCheckboxIcon {
```
alycrys, line 54: Do we have a disabled:hover state?

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,80 +1,88 @@
...
 /* Styles applied if `disabled={true}` on hover */
 .saltCheckboxIcon.saltCheckboxIcon-disabled:hover,
 .saltCheckbox.saltCheckbox-disabled:hover .saltCheckboxIcon {
   stroke: var(--saltCheckbox-icon-stroke-disabled, var(--salt-selectable-borderColor-disabled));
-  stroke-width: var(--saltCheckbox-icon-strokeWidth-disabled, var(--salt-size-border));
+  background: var(--salt-palette-interact-background-disabled);
+}
+
+/* Styles applied if `error={true}` */
+.saltCheckbox-error .saltCheckboxIcon,
```
alycrys, line 42: there should be a `.saltCheckboxIcon-error` class you could use instead

packages/lab/src/checkbox/Checkbox.tsx
ImmanuelBaskaran: It's a react specific attribute, when it's rendered, the correct "checked" props is applied

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,80 +1,83 @@
...
-.saltCheckboxIcon:hover {
-  stroke: var(--saltCheckbox-icon-stroke, var(--salt-selectable-borderColor));
-  stroke-width: var(--saltCheckbox-icon-strokeWidth-hover, var(--salt-size-border));
+.saltCheckbox:hover .saltCheckboxIcon {
+  stroke: var(--saltCheckbox-icon-stroke, var(--salt-selectable-borderColor-hover));
+  background: var(--salt-palette-interact-background-hover);
 }
 
 /* Styles applied if `disabled={true}` */
+.saltCheckbox.saltCheckbox-disabled:hover .saltCheckboxIcon,
```
alycrys, line 30: .saltCheckbox not needed

alycrys: there is no disabled:hover spec, so you can remove this actually

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,80 +1,83 @@
...
+}
+
+/* Styles applied if `error={true}` on hover */
+.saltCheckbox-error:hover .saltCheckboxIcon {
+  background: var(--salt-palette-error-background-emphasize);
+}
+
+/* Styles applied if `error={true}` and `disabled={true}` */
+.saltCheckbox-error.saltCheckbox-disabled .saltCheckboxIcon,
+.saltCheckbox-error.saltCheckbox-disabled:hover .saltCheckboxIcon {
```
alycrys, line 49: same here, disabled:hover can be removed

libertymayc:
status tokens ready to be consumed!

Fercas123 (dismissed):
nice one!

packages/lab/src/checkbox/Checkbox.css
```diff
@@ -1,43 +1,39 @@
...
-  --checkbox-top-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
+  --checkbox-gap-multiplier: var(--saltCheckbox-top-multiplier, 0.75);
 }
 
 .saltCheckbox {
-  line-height: var(--saltCheckbox-lineHeight, 1);
+  cursor: var(--salt-selectable-cursor-hover);
+  background: var(--saltCheckbox-background, none);
+  display: flex;
+  gap: calc(var(--salt-size-unit) * var(--checkbox-gap-multiplier));
```
libertymayc, line 11:
```suggestion
  gap: var(--saltCheckbox-gap, calc(var(--salt-size-unit) * 0.75));
```

the multiplier token seems a bit unnecessary

alycrys (dismissed):

packages/lab/src/checkbox/CheckboxGroup.css
```diff
@@ -1,16 +1,29 @@
...
-  flex-direction: column;
-  vertical-align: top;
 }
 
-/* Styles applied to root component if `row={true}` */
+/* Styles applied to root component if `direction={"horizontal"}` */
 .saltCheckboxGroup-horizontal {
-  --saltControlLabel-right-marginRight: calc(var(--salt-size-unit) * 3);
+  display: flex;
+  gap: calc(var(--salt-size-unit) * 0.75);
```
libertymayc, line 11: gap should maybe have css api if the vertical one does

packages/lab/src/checkbox/CheckboxIcon.css
```diff
@@ -1,80 +1,82 @@
...
   --checkbox-density-bar-height: 2px;
-}
 
-.saltCheckboxIcon {
   --checkbox-borderWidth: var(--salt-size-border);
+  --checkbox-size: var(--salt-size-selectable);
   --checkbox-box-offset: calc(var(--checkbox-borderWidth) / 2);
-  --checkbox-height: calc(var(--checkbox-viewbox) - var(--checkbox-borderWidth));
+  /* 14px is used to match the svg's viewbox */
+  --checkbox-box-size: calc(14px - var(--checkbox-borderWidth));
```
libertymayc, line 11: i think this is fragile, i see the comment but i think this value should be density aware not fixed at 14

libertymayc (approved):

PR #1182, opened 2023-02-21
Avatar refactor - all props exploration

Fercas123:


packages/lab/src/avatar/Avatar.css
```diff
@@ -24,6 +27,7 @@
   overflow: hidden;
   user-select: none;
   border-radius: 50%;
+  outline: auto;
 }
 
 /* Style applied to the image/svg element. We specifically allow SVGs as the Avatar image to enable SVGs such as Flags to take up the full Avatar space */
```
Fercas123, line 33: @el-dav @joshwooding do you know what are lines 38 to 41 for? i'm not sure what are they meant to do, hide text in svgs?

joshwooding: I think it's the hide the broken image icon. It really only needs to be applied to img but worth checking

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,50 +1,88 @@
...
 const withBaseName = makePrefixer("saltAvatar");
 const DEFAULT_AVATAR_SIZE = 2; // medium
 
+const initialsDefaultGetter = (name: string | undefined) =>
+  name &&
+  name
+    .split(" ")
+    .map((n) => n[0])
+    .join("")
+    .toUpperCase();
```
libertymayc, line 44: should we limit the amount of initials? or add an ellipsis, dash etc on overflow? (e.g. 10 word name shows the middle initials only when it overflows)

packages/lab/src/avatar/Avatar.css
```diff
@@ -24,15 +27,12 @@
   overflow: hidden;
   user-select: none;
   border-radius: 50%;
+  outline: auto;
```
libertymayc, line 30: this gives a white outline in dark mode, which I don't think is intended?

Fercas123: this was to avoid a blue outline otherwise, might have to look into it a bit more

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,50 +1,89 @@
-import { FlexLayout, makePrefixer } from "@salt-ds/core";
+import { makePrefixer } from "@salt-ds/core";
 import { UserIcon } from "@salt-ds/icons";
 import { clsx } from "clsx";
 import { forwardRef, HTMLAttributes, ReactNode } from "react";
+import { useLoaded } from "./internal/useLoaded";
 
 import "./Avatar.css";
 
+export type InitialsGetter = (name: string) => string | undefined;
```
el-dav, line 9: should name be optional?

packages/lab/src/avatar/Avatar.tsx
```diff
@@ -1,50 +1,89 @@
...
+  /**
+   * Icon to be used as a default item. Defaults to `UserIcon`
+   */
   fallbackIcon?: ReactNode;
 }
 
 const withBaseName = makePrefixer("saltAvatar");
 const DEFAULT_AVATAR_SIZE = 2; // medium
 
+const initialsDefaultGetter = (name: string | undefined) =>
```
el-dav, line 38: should it be (name?: string) rather than (name: string | undefined).

Or if you make name optional in the InitialsGetter type you can just typer as InitialsGetter

packages/lab/src/avatar/internal/useLoaded.ts
```diff
@@ -0,0 +1,34 @@
...
+      setLoaded("loaded");
+    };
+    image.onerror = () => {
+      if (!active) {
+        return;
+      }
+      image.style.display = "none";
+      setLoaded("error");
+    };
+    return () => {
```
el-dav, line 28: should probably clean up the onload and onerror events here.

I think you can do it by unsetting the properties, but might be better to use `image.addEventListener('load', ...)` so you can do `image.removeEventListener('load', ...)`

el-dav:
Do we want to keep the stories with the `` as children now that you can pass src?

el-dav (approved):

PR #1189, opened 2023-02-23
Fix TS error in Spinner

libertymayc:
Fixes TS error in Spinner


alycrys:
Is it possible to get rid of the lint warning on line 121?

alycrys (approved):

PR #1215, opened 2023-02-28
Fix ag-grid uitk theme floating filter input height in High density

pedroBlazquez:
### Changes in this PR

The `list-item-height` variable was set to 60px (same height as "touch" density) for HD.
This was making the dropdown element inside the floating filter taller than it should be in HD.

### Before
image

### After
image


ZarkoPernar (commented):
Hey @pedroBlazquez, thanks for contributing! Can you run `yarn changeset` in order to add the changelog. This could qualify as "patch".

PR #1235, opened 2023-03-01
Change size-selection to size-selectable

libertymayc:
Renamed `--salt-size-selection` to `--salt-size-selectable`


Fercas123 (approved):

joshwooding:
This is a breaking change and should not have been merged.

PR #1240, opened 2023-03-02
Add color name to color

origami-z:
Currently when choosing color using `ColorChooser`, `Color` object doesn't contain color name (e.g. orange 700) even if picking color from the swatches.  Adding `colorName` so at least there is a way to get the information when needed.

Also fixed `isTransparent` so any hex with `00` alpha value will return true.

### DIFF
--- a/packages/lab/src/color-chooser/ColorHelpers.ts
+++ b/packages/lab/src/color-chooser/ColorHelpers.ts
@@ -5,7 +5,9 @@ import { isTransparent } from "./color-utils";
 export function getColorNameByHexValue(
   hexValue: string | undefined,
   disableAlpha = false,
-  saltColorOverrides?: Record<string, string>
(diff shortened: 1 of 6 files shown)


packages/lab/src/color-chooser/ColorHelpers.ts
```diff
@@ -5,7 +5,9 @@ import { isTransparent } from "./color-utils";
 export function getColorNameByHexValue(
   hexValue: string | undefined,
   disableAlpha = false,
-  saltColorOverrides?: Record<string, string>
+  saltColorOverrides?: Record<string, string>,
+  /** Whe disabled, color names not recognized will be undefined instead of hex values */
```
libertymayc, line 9:
```suggestion
  /** When disabled, color names not recognized will be undefined instead of hex values */
```

PR #1244, opened 2023-03-03
Move avatar to core

Fercas123:


site/docs/components/_index/components-list.tsx
```diff
@@ -150,8 +150,8 @@ export const componentDetails: ComponentDetails[] = [
     name: "Avatar",
     description:
       "Represents a person. It has three variants to account for different types of data: photo, initials and a default version when neither is available.",
-    devStatus: ComponentStatus.IN_BACKLOG,
-    designStatus: ComponentStatus.IN_BACKLOG,
+    devStatus: ComponentStatus.IN_PROGRESS,
```
libertymayc, line 153: @Fercas123 i'm not sure if this is ready now or in progress, feel free to change :)

alycrys (approved):

alycrys (dismissed):

pseys:
Can we change the opening text from: 

An Avatar is a small circular graphical representation of a user entity through a profile picture or name initials. It has three variants to account for different types of data:
- Image is used when a photo is available;
- Initials is used when a photo is unavailable but profile information is;
- A default Icon is used when neither photo nor profile information are available.

to

The Avatar component is a graphical element which can be used to represent a person, group, entity, or communication channel.  It has three variants to account for different types of data: 
- Image is used when a photo is available; 
- Initials can be used when a photo is unavailable but profile information is; 
- Icon can be used when neither a photo nor profile information is available.


Can we use 'user-solid' as the default icon. It's closer to the original and spec'd in Figma. 

There's a background colour on the image variant, which shouldn't be there. I can cause the appearance of a slight halo effect. Its a bug that's been flagged in UITK but not corrected yet. 

Is it possible not to refer to Icon as default? Based on the logic we've discussed before its not actually a default, as if a image is present the default would arguably be image. Can we refer to the variants as: Image, Initials, Icon

el-dav:
Could we change the fallback text in the ImageFallback example to be something with one or two characters?

https://735f992a.saltdesignsystem-storybook.pages.dev/?path=/story/core-avatar--image-fallback

We probably shouldn't have an example where the text overflows the Avatar. I think any initials will do.

el-dav (dismissed):

el-dav (approved):

PR #1247, opened 2023-03-06
Fix typing and build  errors

JoanaMMoreira:
The following changes will allow us to run `yarn build` within `mosaic-site`:
- Extend monorepo's tsconfig in `mosaic-site`
- Fix typings across site
- Pin version of lab packages used in the site to avoid conflict (should be fixed in the next Mosaic release)


mosaic-site/src/components/card/Card.tsx
```diff
@@ -82,7 +84,7 @@ export const InlineCard = ({
   footerText,
   keylineColor,
 }: CardProps): JSX.Element => {
-  const ref = useRef<HTMLDivElement>();
+  const ref = useRef<HTMLDivElement>() as MutableRefObject<HTMLDivElement>;
```
libertymayc, line 87:
```suggestion
  const ref = useRef<HTMLDivElement>(null);
```

mosaic-site/src/_index/accordion/Accordion.tsx
```diff
@@ -15,7 +15,7 @@ type HomepageAccordionProps = {
 const HomepageAccordion = ({
   homePageInfo,
 }: HomepageAccordionProps): JSX.Element => {
-  const ref = useRef<HTMLDivElement>();
+  const ref = useRef<HTMLDivElement>() as MutableRefObject<HTMLDivElement>;
```
libertymayc, line 18:
```suggestion
  const ref = useRef<HTMLDivElement>(null);
```

mosaic-site/src/components/card/Card.tsx
```diff
@@ -34,7 +36,7 @@ export const Card = ({
   keylineColor,
   keyLineAnimation = true,
 }: CardProps): JSX.Element => {
-  const ref = useRef<HTMLDivElement>();
+  const ref = useRef<HTMLDivElement>() as MutableRefObject<HTMLDivElement>;
```
libertymayc, line 39: would suggest same as below but useOnScreen needs non null, maybe worth seeing if we can change the types in that hook in mosaic

libertymayc (approved):

Issue #1236, opened 2023-03-02
Migrating docusaurus components


JoanaMMoreira:
- [x] Navbar
- [x] Vertical navigation
- [x] ToC
- [x] Hamburger menu
- [x] Footer
- [x] Docs pagination
- [x] Code blocks

parsakhan21:
March 2 - in progress

PR #1253, opened 2023-03-07
mosaic chores (#1236)

mark-tate:
- added missing env vars
- updated next config to add i18n locales
- added back the api routes (may remove again during deployment PR)
- pipped Mosaic to beta.24 providing search


mosaic-site/src/pages/api/snapshots/[...route].tsx
```diff
@@ -0,0 +1,37 @@
+import * as path from 'path';
```
Comment (DavieReid) [inline mosaic-site/src/pages/api/snapshots/[...route].tsx]: why do we have this?  withContent middleware should be doing this....

JoanaMMoreira (approved):

PR #1255, opened 2023-03-08
Card refactor

libertymayc:
Considerations and further improvements after design spec created:

- InteractableCard component created; extends Card component  (deprecated `interactive` from Card)
- Interactable redesign; new `accentPosition` prop
- Disabled can only be used for interactable cards (deprecated `disabled` from Card)
- CSS clean up and corrections
- Showcased more variations within stories


libertymayc:
still in progress - moved out of draft for storybook links

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,33 +26,41 @@ This default example is intended to show the way the Card component’s content
...
 
-Cards can have interactable styling by setting `interactable={true}`. These cards could be wrapped in Link, for example, for interactivity. We discourage the use of any focusable elements within interactable Cards.
+Cards can have actions by making use of `InteractableCard`. We discourage the use of any focusable elements within interactable Cards.
 
 <Canvas>
   <Story id="core-card--interactable" />
 </Canvas>
 
-## Disabled
+## Disabled Cards
```
alycrys, line 37:
```suggestion
## Disabled Card
```

packages/core/src/card/InteractableCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+      className={clsx(
+        withBaseName(),
+        withBaseName("interactable"),
+        {
+          [withBaseName("disabled")]: disabled,
+        },
+        className
+      )}
+      onClick={onClick}
+      ref={ref}
```
alycrys, line 32: should we add a tabIndex if there's an onClick?

packages/core/src/card/InteractableCard.tsx
```diff
@@ -0,0 +1,38 @@
...
+   * If `true`, the card will be disabled.
+   */
+  disabled?: boolean;
+}
+
+export const InteractableCard = forwardRef<
+  HTMLDivElement,
+  InteractableCardProps
+>(function Card(props, ref) {
+  const { className, disabled, onClick, children, ...rest } = props;
```
alycrys, line 19: should onClick be mandatory? given that we separated this?

alycrys: Maybe not, but just feels like something is missing. It looks like interactable, but might not actually be interactable.

libertymayc: I considered this initially too but saw in Button that onClick isn't mandatory...

alycrys: I think the button is different. We always expect a button to have an action on it, but the card is split in two, one with action and one without. 
Something to consider, I guess.

alycrys: tbh I don't quite see the value in separating them

libertymayc: it is split in Figma, so for consistency - and it was swayed as it gets confusing having a disabled prop when the component is not necessarily interactive

.changeset/stupid-elephants-trade.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/core": minor
+---
+
+Removed `interactable` and `disabled` props from Card
```
joshwooding, line 5: This is a breaking change and won't be able to be made. We should deprecate these props instead

joshwooding (changes requested):

.changeset/stupid-elephants-trade.md
joshwooding: Another way of doing this is to still have InteractableCard and when interactable is present to conditionally render it. Would make things more complicated though

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -16,7 +16,7 @@ import {
 
 # Card
 
-A Card is a sheet of material that serves as an entry point to more detailed information. It is a convenient means of displaying content composed of different elements whose size or supported actions vary.
+Cards contain content and actions about a single subject.
```
libertymayc, line 19: TBU

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,33 +26,53 @@ This default example is intended to show the way the Card component’s content
...
 
 # API
 
 ```
-import { Card } from "@salt-ds/lab";
+import { Card, InteractableCard } from "@salt-ds/core";
 ```
 
-## Props
+## Interactable Card Props
```
joshwooding, line 73: This removes Card's props?

.changeset/stupid-elephants-trade.md
joshwooding, line 5: This will need to be updated

packages/core/src/card/InteractableCard.tsx
```diff
@@ -0,0 +1,44 @@
...
+   * If `true`, the card will be disabled.
+   */
+  disabled?: boolean;
+  /**
+   * Action called on click.
+   */
+  onClick: React.MouseEventHandler<HTMLDivElement>;
+}
+
+export const InteractableCard = forwardRef<
```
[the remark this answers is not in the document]
joshwooding: The reason why I ask is that if we narrowed the use-case of this component we can be more opinionated in how it works.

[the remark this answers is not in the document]
libertymayc: cc @bhoppers2008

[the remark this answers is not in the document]
bhoppers2008: Depends how you think about this component... i tend to think of it as a base component which can be used to facilitate various patterns. So it's an interactive version of card, which could be used for navigation, action, selection... just depends what app teams want to do with it.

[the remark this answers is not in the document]
joshwooding: If you view it as a base component, it could work. I think it would make sense to show other use cases possible on the docs. The issue with something being interactable like this is that semantically they are quite different.

If the card was performing an action, then it would be something like a button and therefore would have button semantics, but if it's navigating to a page then it would have link semantics. Selection would get more confusing because it depends if the card is alone or with other cards, then it could have radio or checkbox semantics.

Having a component that could provide all these possible interactions could be confusing, this might not be a problem if the relevant context is given to the component e.g. either behaviour being consistent within an app or the card's content providing context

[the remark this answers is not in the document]
libertymayc: > Selection would get more confusing because it depends if the card is alone or with other cards, then it could have radio or checkbox semantics.

from @bhoppers2008 and I's recent discussion, I think this wouldn't apply - that would be from a SelectableCard component which branched off of InteractableCard (in the future). If we just have the first 2 you mentioned, or similar, interactive actions then it may be easier, but definitely we need to state clearly in the spec what the limit of interactions should be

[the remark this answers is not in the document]
bhoppers2008: Josh raises a good point /:-| I'm going to contradict myself here. 

I think it's fine that the card doesn't inherit radio/ check/ button semantics, since first and foremost, it's a container and it should be styled as such. We follow the same principle elsewhere as well. Now for the contradiction... we're applying selectable characteristics to the card to create the states atm... so we are already assuming a semantic. 

I believe the styles are correct regardless, since they come from the 'Interact' palette. So i think this gives us two options... we leave it referencing 'Selectable' (maybe a bit hacky?) or we add hover and active tokens to 'Container', which reference 'Interact'.

...or there is another more radical option. We redesign the interactable card 8\ There have been suggestions that the 'Interactable card' should take on a similar aesthetic to what we have on the site. So it would have an accent bar running across the bottom (which would reference accent, but could also be overridden to another colour from the categorical palette). In this case it wouldn't reference selectable at all, and might be [something like this](https://www.figma.com/file/jHTj8hPMa2IL638Mu4kYnH/Untitled?node-id=1%3A25&t=6WUoMDgQUdMid9aS-1).

@pseys @dplsek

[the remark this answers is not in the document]
libertymayc: call scheduled for monday :D

packages/core/src/card/Card.css
```diff
@@ -27,43 +25,51 @@
...
 
-  cursor: var(--saltCard-cursor-hover, var(--salt-selectable-cursor-hover));
+  cursor: var(--salt-selectable-cursor-hover);
   position: relative;
 }
 
 /* Styles applied on active state if `interactable={true}` */
-.saltCard-interactable:active {
+.saltCard-interactable:active,
+.saltCard-interactable:focus {
```
alycrys, line 39: .saltCard-interactable:focus is overridden here. Is this intended?

packages/core/src/button/Button.tsx
```diff
@@ -1,9 +1,9 @@
-import { ComponentPropsWithoutRef, forwardRef, ReactElement } from "react";
 import { clsx } from "clsx";
+import { ComponentPropsWithoutRef, forwardRef, ReactElement } from "react";
+import { useButton } from "./useButton";
```
alycrys, line 3: I believe this change shouldn't be part of this PR

packages/core/src/card/Card.css
```diff
@@ -9,61 +9,61 @@
...
-  --card-borderColor: var(--salt-selectable-borderColor-hover);
+a:focus-visible .saltCard-interactable,
+.saltCard-interactable:focus, 
+.saltCard-interactable:focus-visible {
+  --card-cursor: var(--salt-selectable-cursor-hover);
   --card-shadow: var(--salt-overlayable-shadow-hover);
 
-  cursor: var(--saltCard-cursor-hover, var(--salt-selectable-cursor-hover));
-  position: relative;
+  color: var(--salt-text-primary-foreground);
```
alycrys, line 39: should this also be customisable?

packages/core/src/card/Card.css
```diff
@@ -9,61 +9,61 @@
...
-/* Styles applied if `disabled={true}` */
-.saltCard-disabled,
-.saltCard-disabled.saltCard-interactable,
-a:focus .saltCard-interactable.saltCard-disabled {
+.saltCard-interactable.saltCard-disabled:active{
   --card-borderColor: var(--salt-container-primary-borderColor-disabled);
+  --card-cursor: var(--salt-selectable-cursor-disabled);
+  --card-shadow: var(--salt-overlayable-shadow);
 
+  background: var(--saltCard-background, var(--salt-container-primary-background-disabled));
```
alycrys, line 61: should `--saltCard-background` be `--saltCard-disabled-background`?

packages/core/src/card/Card.tsx
```diff
@@ -4,34 +4,33 @@ import { makePrefixer } from "../utils";
...
 ) {
-  const { className, disabled, interactable, children, ...rest } = props;
-
+  const { className, children, disabled, interactable, ...rest } = props;
   return (
     <div
       className={clsx(
         withBaseName(),
         {
+          /* **Deprecated:** Will be removed soon */
```
alycrys, line 33: can we replace `soon` with something else?

packages/core/src/card/Card.tsx
```diff
@@ -4,34 +4,33 @@ import { makePrefixer } from "../utils";
 import "./Card.css";
 
 const withBaseName = makePrefixer("saltCard");
-
 export interface CardProps extends ComponentPropsWithoutRef<"div"> {
   /**
+   * **Deprecated:** Use the InteractableCard instead
```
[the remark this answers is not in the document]
libertymayc: good idea, thank you

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,38 +26,70 @@ This default example is intended to show the way the Card component’s content
   <Story id="core-card--default" />
 </Canvas>
 
-## Interactable
+## Interactable Card
```
joshwooding, line 29: We should have content review this to make sure it's not confusing having two Interatable Card headers

packages/core/stories/card/card.stories.tsx
```diff
@@ -70,26 +76,127 @@ export const Default: ComponentStory<typeof Card> = () => (
...
+      <InteractableCard style={{ maxWidth: "500px" }}>
+        <div style={{ textAlign: "center" }}>
+          <div>Visit Salt</div>
+        </div>
+      </InteractableCard>
+    </a>
+  );
+};
+
+// export const InteractableAsAction: ComponentStory<typeof Card> = () => {
```
joshwooding: Can we delete this?

joshwooding (changes requested):

alycrys (approved):

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,38 +26,70 @@ This default example is intended to show the way the Card component’s content
...
+
+<CSSClassTable of={InteractableCard} />
+
+## Characteristics Used in Card
+
+<CharacteristicUsage of={Card} />
+
+## Characteristics Used in Interactable Card
 
 <CharacteristicUsage of={Card} />
```
feliciackh, line 95: QQ - is this meant to be  <CharacteristicUsage of={InteractableCard} /> instead?

libertymayc: yes, thank you!

packages/core/stories/card/card.stories.tsx
```diff
@@ -41,17 +50,24 @@ const Examples = () => (
...
-          <Text>Here is some content</Text>
-        </div>
+        <Text>Content</Text>
       </Card>
     </ExampleRow>
   </>
 );
 
+const InteractableExamples = () => (
+  <>
```
feliciackh, line 60: Not a blocker - these fragments probably aren't needed (same for above Examples component)

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,38 +26,62 @@ This default example is intended to show the way the Card component’s content
...
-  <Story id="core-card--interactable" />
+  <Story id="core-card--interactable-as-link" />
 </Canvas>
 
-## Disabled
+## Disabled Card
+
+**Deprecated:** In previous versions, Cards could have disabled styling applied by setting `disabled={true}`.
+
+**Updated:** Interactable cards can curently have disabled styling applied by setting `disabled={true}`.
```
feliciackh, line 43:
```suggestion
**Updated:** Interactable cards can currently have disabled styling applied by setting `disabled={true}`.
```

packages/core/src/card/InteractableCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const withBaseName = makePrefixer("saltInteractableCard");
+
+// TODO: Remove omissions when Card props deprecated
+export interface InteractableCardProps
+  extends Omit<CardProps, "disabled" | "interactable"> {
+  /**
+   * Accent border position: defaults to "bottom"
+   *
+   * Must be one of  "bottom" | "top" | "left"
```
joshwooding, line 17:
```suggestion
```
This info is a duplication of the type signature below

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,38 +26,62 @@ This default example is intended to show the way the Card component’s content
   <Story id="core-card--default" />
 </Canvas>
 
-## Interactable
+## Interactable Card
 
-Cards can have interactable styling by setting `interactable={true}`. These cards could be wrapped in Link, for example, for interactivity. We discourage the use of any focusable elements within interactable Cards.
+**Deprecated:** In previous versions, Cards could have interactable styling by setting `interactable={true}`. These cards could be wrapped in Link, for example, for interactivity. We discourage the use of any focusable elements within interactable Cards.
```
joshwooding, line 31: The **Deprecated** and **Updated** sections in this document seem weird to me and feel like they lack context.
Is it even worth highlighting these as deprecated in the docs? We should improve the changeset by adding rich diffs of how a user would migrate from using the deprecated Card props to the new InteractableCard

joshwooding (changes requested):

alycrys (approved):

packages/core/src/card/Card.css
```diff
@@ -1,69 +1,64 @@
 /* Styles applied to the root element */
 .saltCard {
-  --card-borderColor: var(--salt-container-primary-borderColor);
```
origami-z, line 3: What if `--card-borderColor` was used by consumer?

packages/core/src/card/Card.css
```diff
@@ -1,69 +1,64 @@
...
+  min-height: var(--saltCard-minHeight, 100%);
   overflow: hidden;
-  transition: box-shadow 0.3s cubic-bezier(0.4, 0, 0.2, 1); /* TODO: this should possibly be moved to animation.css? */
+  text-decoration: none;
 }
 
 /* Styles applied to Card content */
 .saltCard-content {
-  color: var(--saltCard-color, var(--salt-text-primary-foreground));
-  min-height: var(--salt-text-minHeight);
```
origami-z, line 19: Need to document these changes?

packages/core/src/utils/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./capitalize";
```
origami-z, line 1: Better to be moved independently ?

origami-z (commented):
Deprecation means anything worked before should still work.  There are a few changes feel disrupting. Whether getting a new Interactive Card component within v1 really that importatnt?

.changeset/stupid-elephants-trade.md
```diff
@@ -0,0 +1,33 @@
...
++   const MyInteractableCard = () => (
++       <InteractableCard />
++   );
++
++   const MyInteractableDisabledCard = () => (
++       <InteractableCard disabled />
++   );
+```
+
+Moved `capitalize` from lab to core
```
libertymayc, line 33:
```suggestion
```

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -16,7 +16,7 @@ import {
 
 # Card
 
-A Card is a sheet of material that serves as an entry point to more detailed information. It is a convenient means of displaying content composed of different elements whose size or supported actions vary.
+A Card is a visually distinct container that displays a snapshot of information about a single subject, and acts as an entry point to more detailed information.
```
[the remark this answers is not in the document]
sion-docs: Yes. @libertymayc, can you advise on use cases?

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,38 +26,58 @@ This default example is intended to show the way the Card component’s content
...
 <Canvas>
-  <Story id="core-card--interactable" />
+  <Story id="core-card--interactable-as-link" />
 </Canvas>
 
-## Disabled
+## Disabled Interactable Card
 
-Cards can have disabled styling applied by setting `disabled={true}`.
+Interactable cards can currently have disabled styling applied by setting `disabled={true}`.
```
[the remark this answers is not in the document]
sion-docs: Agreed. @libertymayc can you advise on default disabled styling?

sion-docs (dismissed):

packages/core/stories/card/card.stories.tsx
```diff
@@ -70,26 +81,144 @@ export const Default: ComponentStory<typeof Card> = () => (
...
+    <SaltProvider mode="dark">
+      <InteractableExamples />
+    </SaltProvider>
+  </div>
+);
+
+export const InteractableAsLink: ComponentStory<typeof Card> = () => {
+  const handleClick = () => {
+    console.log("Clicked");
+    window.open("https://saltdesignsystem.com/");
```
joshwooding, line 120: This won't be accessible. We should figure out a better example

.changeset/stupid-elephants-trade.md
```diff
@@ -0,0 +1,31 @@
+---
+"@salt-ds/core": minor
+"@salt-ds/lab": minor
```
joshwooding, line 3:
```suggestion
```

joshwooding (approved):

PR #1264, opened 2023-03-09
[WIP] Add a check for the class order in clsx

Fercas123:


Fercas123:
won't implement for now

PR #1265, opened 2023-03-09
Font family on text component

origami-z:
1. `--salt-text-fontFamily` was not applied on text components at all, so add it to `.saltText` with test
2. Adds new variables like `--salt-text-h1-fontFamily` which bridges to default text one. This is useful for code generator tool which generates CSS automatically from design softwares like Figma, or design tokens typography definitions (e.g. [draft format#typography](https://design-tokens.github.io/community-group/format/#typography)) where font family is defined together with other text properties like font size, font weight, line height.

Open for alternative suggestion on 2nd point

### DIFF
--- a/.changeset/real-numbers-shake.md
+++ b/.changeset/real-numbers-shake.md
@@ -1,0 +1,6 @@
+---
+"@salt-ds/core": patch
+---
+
+- Fix Text components doesn't have `--salt-text-fontFamily` applied.
+- Apply new font family variables from `@salt-ds/theme` to Text components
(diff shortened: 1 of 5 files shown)


libertymayc:
if we do 2., we should probably keep the standard and refer to the foundation value `--salt-typography-fontFamily` rather than characteristic token referencing another characteristic token, & looking at the link in the description, will it be useful then to also add letterSpacing token to the variants?

origami-z:
> if we do 2., we should probably keep the standard and refer to the foundation value `--salt-typography-fontFamily` rather than characteristic token referencing another characteristic token

Updated


> looking at the link in the description, will it be useful then to also add letterSpacing token to the variants

There are a few more can be added but optional, where i think font family is something compulsory for defining a type style

alycrys (approved):

origami-z:
Closing in favor of #1354

PR #1285, opened 2023-03-13
[Tooltip] Add strategy and portal props to Tooltip

ZarkoPernar:
### Why
When we are using the Tooltip in a container that is clipped, we need a portal in order to escape the bounding container that is clipping the tooltip and render the tooltip at the root of the html document. While `strategy="fixed"` is useful in most cases because the tooltip is still nested within the dom hierarchy, it doesn't work in the case of css `clip-path` property.

### Changes
- add `portal` prop to Tooltip
- add `strategy` prop to Tooltip

### DIFF
--- a/packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
@@ -156,4 +156,38 @@ describe("GIVEN a Tooltip", () => {
       cy.get("li").should("be.visible");
     });
   });
+
+  describe("WHEN portal prop is true", () => {
+    it("should render the tooltip at the root of the body element", () => {
+      cy.mount(
+        <div id="container">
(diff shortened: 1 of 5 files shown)


packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
```diff
@@ -156,4 +156,38 @@ describe("GIVEN a Tooltip", () => {
...
+      );
+
+      cy.get("#container .saltTooltip").should("not.exist");
+
+      cy.get("body > [data-floating-ui-portal] > .saltTooltip").should("exist");
+    });
+  });
+
+  describe("By default the strategy prop is 'absolute'", () => {
+    it("should render the tooltip with css position fixed", () => {
```
feliciackh, line 175:
```suggestion
    it("should render the tooltip with css position absolute", () => {
```

PR #1293, opened 2023-03-15
Country symbol

el-dav:
PR Summary:

New package called 'countries'
A base CountrySymbol component (similar to Icon) which is used as the wrapper for all the CountrySymbol components
Imported all the country symbol SVGs from figma
Created a script to:
Format the names of the CountrySymbols (remove special characters, strip country codes, convert to us-eng)
Generate the CountrySymbol components from the SVGs
Optimise the svgs of the resulting CountrySymbol components
Create an index file to export all the CountrySymbol components
Create a countryCodes array and countryMetaMap to be exported for consumers to use
Create a countrySymbolMap for consumers to use to map from countryCode to CountrySymbolComponent
create a lazyMap for use in LazyCountrySymbol component
Created LazyCountrySymbol component which allows consumers to change the symbol using the code prop and uses code-splitting to (hopefully) reduce bundles size
Created relevant stories
Added tests
Added documentation


packages/countries/scripts/generateCountrySymbol.mjs
```diff
@@ -0,0 +1,404 @@
...
+            };
+          },
+        },
+      ],
+    });
+
+    const fileContents = Mustache.render(template, {
+      svgElements: svgPaths.data,
+      componentName,
+      ariaLabel: baseCountryName.toLowerCase(),
```
Fercas123, line 183: could this not change the reader's emphasis?
```suggestion
      ariaLabel: baseCountryName,
```

el-dav: Yeah potentially, I based it on what we were doing with the original icon script so thought there might be a reason

el-dav:
PR Summary:
- New package called 'countries'
- A base CountrySymbol component (similar to Icon) which is used as the wrapper for all the CountrySymbol components
- Imported all the country symbol SVGs from figma
- Created a script to:
  - Format the names of the CountrySymbols (remove special characters, strip country codes, convert to us-eng)
  - Generate the CountrySymbol components from the SVGs
  - Optimise the svgs of the resulting CountrySymbol components
  - Create an index file to export all the CountrySymbol components
  - Create a countryCodes array and countryMetaMap to be exported for consumers to use
  - Create a countrySymbolMap for consumers to use to map from countryCode to CountrySymbolComponent
  - create a lazyMap for use in LazyCountrySymbol component
-  Created LazyCountrySymbol component which allows consumers to change the symbol using the `code` prop and uses code-splitting to (hopefully) reduce bundles size
- Created relevant stories

pseys:
This is looking great. On review I've noticed several of the SVGs have broken on export. I'm trying to find availability for a designer to take a look at this and fix the issues. A quick look through I've spotted at the very least the following needs fixing: 
- UN
- Tuvalu
- Virgin Islands (British)
- Turks and Caicos Islands
- Slovakia
- Slovenia
- Pitcairn
- Peru
- Niue
- Montserrat
- Cayman Islands
- Kenya

packages/countries/src/lazy-country-symbol/LazyCountrySymbol.tsx
```diff
@@ -0,0 +1,36 @@
+import { Suspense, useEffect, useState, useTransition } from "react";
```
joshwooding, line 1: Just to add what we've discussed here:

We should be able to simplify this by moving the Suspense component to be on the user side.
We can't use setTransition but we should be able to just use Line 23 directly with the prop

packages/countries/src/countrySymbolMap.ts
joshwooding, line 1: We should probably not ship this map since it will pull in all the Symbols and recommend the lazy component instead.

origami-z:
I've submitted a review request to change 4 flags in Figma
- GD
- PT
- UG
- UN

pseys (commented):
The following paragraph is inaccurate in relation to Country Symbol (it relates to icon). The whole paragraph can be removed.

"Our base size for high density (HD) elements is 10px. However, for accessibility and visual design reasons, we don't recommend icon sizes that are smaller than 12px. Therefore, our default HD Country Symbol icons start at 12px but they increase size in multiples of 10px—so that 2x size is 20px, 4x is 40px, etc."

el-dav:
> 

@joshwooding Should this be a major or will that bump it to v2?

el-dav:
> 

Thanks, this had already been removed but I think the latest hadn't been deployed to storybook when you reviewed. You can view the latest here: 
https://0d329424.saltdesignsystem-storybook.pages.dev/?path=/story/documentation-country-symbols-country-symbol--page

pseys (approved):
I've reviewed the storybook doc and it all looks good to me.

PR #1308, opened 2023-03-15
Customise docs pagination

JoanaMMoreira:
Add styling overrides to pagination component so it matches the current site


mosaic-site/src/layouts/DetailTechnical/DetailTechnical.module.css
```diff
@@ -70,3 +70,73 @@
     display: none;
   }
 }
+
+.docPaginator {
+  background-color: #10212b;
```
libertymayc, line 75: whats this hex?

libertymayc:
just seen text selection bg is wrong, we could add in this one --salt-text-background-selected to src/css/base.css like the old site, can be added in any pr though as not coupled to this :) 
otherwise lgtm! just the hex value

JoanaMMoreira:
> just seen text selection bg is wrong, we could add in this one --salt-text-background-selected to src/css/base.css like the old site, can be added in any pr though as not coupled to this :) otherwise lgtm! just the hex value

I've added this in :)

Fercas123 (approved):

joshwooding (approved):

PR #1319, opened 2023-03-17
Update custom-property-no-foundations lint rule

libertymayc:
Uses updated characteristics/foundations list, and usages are corrected


alycrys:
Shall we wait for feedback from design or can this be merged? Looks like there are a lot of TODOs

libertymayc:
> Shall we wait for feedback from design or can this be merged? Looks like there are a lot of TODOs

The specs need to be done for the components with the TODOs, which probably won't come for a while and will need maybe new characteristics (e.g. white focus outline) or component redesign to fit characteristics - we should merge this as it would be too much design work to wait and the usages commented as todo's are wrong anyway

alycrys (approved):

PR #1337, opened 2023-03-22
Add `basis` to FlexItem

origami-z:
Without the basis, need to use `style` to give flex basis for simple 2 column layout like below

image

For the 2nd row, i'm can't get input / form field to be exactly half of the space

```
<FlexLayout gap={1}>
  {/* without flexBasis input will be bigger than the empty space */}
  <FlexItem grow={1} shrink={1} style={{ flexBasis: 1 }}>
    <FormField label="Decimal">
      <Input />
    </FormField>
  </FlexItem>
  <FlexItem grow={1} shrink={1}></FlexItem>
</FlexLayout>
```

### DIFF
--- a/.changeset/poor-lions-hear.md
+++ b/.changeset/poor-lions-hear.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added new `basis` prop to FlexItem to control flex-basis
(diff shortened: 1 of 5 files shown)


Fercas123 (approved):
LGTM

PR #1351, opened 2023-03-23
Portal

alycrys:


packages/lab/src/portal/Portal.tsx
```diff
@@ -95,12 +62,9 @@ export const Portal = forwardRef<HTMLElement, PortalProps>(function Portal(
...
-  if (mounted && portalRef.current && children) {
-    return createPortal(
-      <SaltProvider>{children}</SaltProvider>,
-      portalRef.current
-    );
-  }
-
-  return null;
+  return (
+    <FloatingPortal root={container as HTMLElement} id={id}>
```
mark-tate, line 66: What is FloatingPortal has more props added ?
There is no way to change defaults

[preserveTabOrder](https://floating-ui.com/docs/FloatingPortal#preservetaborder)`

packages/lab/src/portal/Portal.tsx
```diff
@@ -67,25 +53,6 @@ export const Portal = forwardRef<HTMLElement, PortalProps>(function Portal(
 
   const container = getContainer(containerProp) ?? document.body;
```
mark-tate, line 54: This will break if container is not defined and SSR is used

packages/lab/src/portal/Portal.tsx
```diff
@@ -67,25 +53,6 @@ export const Portal = forwardRef<HTMLElement, PortalProps>(function Portal(
...
 
-  useIsomorphicLayoutEffect(() => {
```
mark-tate, line 70: How will we approach the Desktop use-case with Portal ?

mark-tate: What are our thoughts on controlling the placement of content?
I see floating-window has middleware, I guess we create a way to use this?

Most folks get a default but there is a way that it can be overridden for the Desktop use-case etc ?
Can you point us at any code which might exist already?

alycrys: Still trying to figure it out

Issue #1137, opened 2023-02-10
Desktop Support


joshwooding:
Teams need to have multi-window support with components. This would include features like opening a tooltip in a new window and opening each Cascading Menu list in new windows.

```[tasklist]
### Tasks
- [ ]  Discuss the current abstractions
```

Requirements:

- Injection of window
- Window positioning
- Styling injection

mark-tate:
How do we see this work's priorities dovetailing with List creation ?

PR #1353, opened 2023-03-23
Desktop support poc

joshwooding:
This PR is a POC implementation for #1137 

It tackles runtime style injection as well as window injection. It uses the concept of a component injection too similar to how window in labs works.

Window and styles are currently new packages to ensure we don't introduce any circular dependencies


packages/styles/src/use-style-injection/useStyleInjection.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    return () => {
+      const sheetsMap = windowSheetsMap.get(targetWindow);
+      const style = sheetsMap?.get(id);
+      if (style) {
+        targetWindow.document.head.removeChild(style);
+        sheetsMap?.delete(id);
+      }
+    };
+  });
```
el-dav, line 50: Should we have a dependency array? I don't think we need to do this every render

shinytoyrobots:
Is there a reason why we haven't progressed this PR?

joshwooding:
> Is there a reason why we haven't progressed this PR?

This work has been handed to @el-dav and @ImmanuelBaskaran. They are currently investigating the steps forward and setting up dev environments.

PR #1355, opened 2023-03-23
Add colorName to Color

origami-z:
Currently when choosing color using `ColorChooser`, `Color` object doesn't contain color name (e.g. orange 700) even if picking color from the swatches.  Adding `colorName` so at least there is a way to get the information when needed.

Also fixed `isTransparent` so any hex with `00` alpha value will return true.

Redo of #1240

### DIFF
--- a/.changeset/happy-hats-argue.md
+++ b/.changeset/happy-hats-argue.md
@@ -1,0 +1,6 @@
+---
+"@salt-ds/lab": minor
+---
+
+- Add `colorName` to `Color` which represents name of Salt color palette when match, e.g. Orange700
+- Fix color with alpha value of 0 not correctly recognized by `isTransparent`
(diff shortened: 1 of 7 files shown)


libertymayc (dismissed):

.changeset/happy-hats-argue.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/lab": minor
+---
+
+- Add `colorName` to `Color` which represents name of Salt color palette when match, e.g. Orange700
+- Fix color with alpha value of 0 not correctly recognized by `isTransparent`
```
joshwooding, line 6:
```suggestion
- Fix colors with an alpha value of 0 not being correctly recognized by `isTransparent`
```

joshwooding (approved):

PR #1358, opened 2023-03-23
Links focus-visible styling

libertymayc:
Replaced browser default focus ring with our own


joshwooding:
I'm not sure about this. What is the intention behind this change?

PR #1367, opened 2023-03-24
Fix broken links on the site

JoanaMMoreira:
Fix a few broken links we've noticed after the site was deployed

### DIFF
--- a/mosaic-site/docs/getting-started/developing.mdx
+++ b/mosaic-site/docs/getting-started/developing.mdx
@@ -149,7 +149,7 @@ function App() {
 export default App;
 ```
 
-Head over to our available [components](../components/) to view their respective APIs.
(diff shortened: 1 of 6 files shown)


mosaic-site/docs/getting-started/developing.mdx
```diff
@@ -149,7 +149,7 @@ function App() {
 export default App;
 ```
 
-Head over to our available [components](../components/) to view their respective APIs.
+Head over to our available [components](../components/index) to view their respective APIs.
```
james-nash, line 152: I expect the URLs of our site's pages not to change as part of the Mosaic migration.

Can Mosaic not be configured so that the path `/components/` points to the page with the status table, just as it does on the current Docusaurus site?

Then these links wouldn't need changing as well.

james-nash (changes requested):

PR #1368, opened 2023-03-24
Fix Avatar fallback

joshwooding:


packages/core/src/avatar/internal/useImage.ts
```diff
@@ -0,0 +1,35 @@
...
+  }, [src]);
+
+  useIsomorphicLayoutEffect(() => {
+    if (!src) {
+      return;
+    }
+
+    let active = true;
+    const image = new Image();
+    image.src = src;
```
joshwooding, line 20: img.src was never previously set

packages/core/stories/avatar/avatar.stories.tsx
```diff
@@ -62,21 +62,5 @@ WithCustomSvg.args = {
...
-    img.onerror = () => {
-      setChildren(initials);
-    };
-    img.onload = () => {
-      setChildren(<img src={src} alt={alt} />);
-    };
-  }, [src, alt]);
-
-  return <Avatar size={3}>{children}</Avatar>;
+  return <Avatar size={3} src="bad_url" name="Peter Piper" />;
```
joshwooding, line 65: Since the fallback is handled within Avatar now it's simpler to show this demo

packages/core/src/avatar/internal/useImage.ts
```diff
@@ -0,0 +1,35 @@
...
+export function useImage({ src }: ImgHTMLAttributes<HTMLImageElement>) {
+  const [status, setStatus] = useState<
+    "pending" | "loading" | "loaded" | "error"
+  >("loading");
+
+  useEffect(() => {
+    setStatus(src ? "loading" : "pending");
+  }, [src]);
+
+  useIsomorphicLayoutEffect(() => {
```
joshwooding, line 13: Changed this to useLayout so the status check happens quicker and the content doesn't flash as bad

joshwooding (commented):
Need to add tests to this feature won't break in the future

packages/core/stories/avatar/avatar.stories.tsx
Fercas123: what happens if the user wants to pass props to the image? I think this example might be redundant for an image as src, and we should have one that shows how to pass it as children when you want to add other props to your image, but it would still need to be handled outside wouldn't it?

joshwooding: We should provide another example showing how img can be a child

Fercas123 (approved):

packages/core/src/avatar/Avatar.tsx
```diff
@@ -64,15 +64,23 @@ export const Avatar = forwardRef<HTMLDivElement, AvatarProps>(function Avatar(
     "--saltAvatar-size-multiplier": `${size}`,
   };
 
-  const hasImgNotFailing = useLoaded({ src }) !== "error" && src;
-
+  const status = useAvatarImage({ src });
+  const hasImgNotFailing = status === "loaded";
   if (hasImgNotFailing) {
-    children = <img alt={name} src={src} />;
+    children = <img alt="" src={src} />;
```
Fercas123, line 70:
```suggestion
    children = <img alt={name} src={src} />;
```

Fercas123 (dismissed):

Fercas123 (approved):

PR #1378, opened 2023-03-27
Remove salt prefix from site url

JoanaMMoreira:


mark-tate:
This PR was created to convert 'local' namespace references to 'salt'.

When we intially created the salt site, we did so, just using the 'local' namespace.

The 'local' namespace is just a sandbox for local development. 
Think of it like a scratchpad for demos or pocs.

Using the 'local source plugin' we can map local files to the 'salt' namespace, so we can use the 'salt' namespace for all docs pulled from both local and remote sources.

PR #1399, opened 2023-03-28
Update tokens on Radio button

alycrys:
Apply updated design tokens to radio button icon


packages/lab/src/radio-button/RadioButtonIcon.css
```diff
@@ -1,78 +1,86 @@
...
-.saltRadioButtonIcon-disabled .saltRadioButtonIcon-border,
-.saltRadioButton:hover .saltRadioButtonIcon-disabled .saltRadioButtonIcon-border {
-  fill: var(--saltRadioButton-icon-fill-disabled, var(--radioButton-icon-fill));
-  stroke: var(--saltRadioButton-icon-strokeColor-disabled, var(--salt-selectable-borderColor-disabled));
+/* Styles applied to radio button icon if checked and disabled */
+.saltRadioButtonIcon-disabled.saltRadioButtonIcon-checked .saltRadioButtonIcon-circle,
+.saltRadioButton:hover .saltRadioButtonIcon-disabled.saltRadioButtonIcon-checked .saltRadioButtonIcon-circle {
+  fill: var(--saltRadioButton-icon-fill-checked-disabled, var(--salt-container-primary-background-disabled));
+  stroke: var(--saltRadioButton-icon-strokeColor-checked-disabled, var(--salt-selectable-borderColor-selectedDisabled));
+  stroke-width: calc(var(--salt-size-border) * 6.5);
```
libertymayc, line 43: JW, where the 6.5 comes from

ImmanuelBaskaran (dismissed):

libertymayc:
Tokens seem fine to me, but I wonder whether we should be setting the fill like our tokens for button, i.e. making a local token --radio-button-fill and then setting it only in one place as a base and adjusting the value based on the class instead of the current impl

packages/lab/src/radio-button/RadioButtonIcon.tsx
```diff
@@ -29,10 +29,10 @@ export const RadioButtonIcon = ({
       viewBox="0 0 14 14"
       width="14"
     >
-      <circle className={withBaseName("border")} cx="7" cy="7" r="6.5" />
+      <circle className={withBaseName("circle")} cx="7" cy="7" r="6.5" />
```
alycrys, line 32: @libertymayc this is the 6.5 value referenced in css

alycrys:
> 
Sure, I can do that

alycrys:
> Tokens seem fine to me, but I wonder whether we should be setting the fill like our tokens for button, i.e. making a local token --radio-button-fill and then setting it only in one place as a base and adjusting the value based on the class instead of the current impl

Sure, I can do that

packages/lab/src/radio-button/RadioButtonIcon.css
```diff
@@ -1,78 +1,86 @@
...
   margin-top: var(--salt-size-border);
 }
 
-/* Styles applied to RadioButtonIcon border */
-.saltRadioButtonIcon-border {
-  fill: var(--saltRadioButton-icon-fill, var(--radioButton-icon-fill));
+/* Styles applied to radio button icon circle */
+.saltRadioButtonIcon-circle {
+  fill: var(--saltRadioButton-icon-fill, var(--salt-container-primary-background));
   stroke: var(--saltRadioButton-icon-strokeColor, var(--salt-selectable-borderColor));
```
libertymayc, line 13: just spoke with @navkaur76 and i think this one is supposed to be selectable-foreground, and same for wherever border is used (i noted in the specs)
she needs to clarify with @pseys since naturally i would assume border is the correct token here too

alycrys: the actual colours don't seem to match the design though

packages/lab/src/radio-button/RadioButtonIcon.tsx
libertymayc: can this radius be documented in the size specs @navkaur76 if it makes sense to? if it relates more strongly to the svg though maybe its unnecessary

libertymayc:
> > Tokens seem fine to me, but I wonder whether we should be setting the fill like our tokens for button, i.e. making a local token --radio-button-fill and then setting it only in one place as a base and adjusting the value based on the class instead of the current impl
> 
> Sure, I can do that

sorry, I think I take it back if you are able to just revert back to the previous commit. Button has variants which is why we have it there but the local tokens kind of just bloat it up for radio button. it would make sense i think to do it for only the tokens which change on error state, sort of like line 114 in formfield.css

packages/lab/src/radio-button/RadioButtonIcon.css
libertymayc: agreed, when i inspected it does use border which is why i msged nav

libertymayc (approved):

alycrys:
> > > Tokens seem fine to me, but I wonder whether we should be setting the fill like our tokens for button, i.e. making a local token --radio-button-fill and then setting it only in one place as a base and adjusting the value based on the class instead of the current impl
> > 
> > 
> > Sure, I can do that
> 
> sorry, I think I take it back if you are able to just revert back to the previous commit. Button has variants which is why we have it there but the local tokens kind of just bloat it up for radio button. it would make sense i think to do it for only the tokens which change on error state, sort of like line 114 in formfield.css

I've added the ones that would repeat. Is it ok?

libertymayc:
> > > > Tokens seem fine to me, but I wonder whether we should be setting the fill like our tokens for button, i.e. making a local token --radio-button-fill and then setting it only in one place as a base and adjusting the value based on the class instead of the current impl
> > > 
> > > 
> > > Sure, I can do that
> > 
> > 
> > sorry, I think I take it back if you are able to just revert back to the previous commit. Button has variants which is why we have it there but the local tokens kind of just bloat it up for radio button. it would make sense i think to do it for only the tokens which change on error state, sort of like line 114 in formfield.css
> 
> I've added the ones that would repeat. Is it ok?

I think I would do something like

.saltRadioButton {
  --radioButton-strokeColor: selectable token
  --radioButton-strokeColor-hover: selectable hover token
}

.saltRadioButton.saltRadioButton-error {
 the above local tokens, set to the status token
}

.saltRadioButton .saltRadioButtonIcon-circle {
  stroke: var(--saltRadioButton-strokeColor, var(--radioButton-strokeColor))
}

.saltRadioButton:hover .saltRadioButtonIcon-circle {
  stroke: var(--saltRadioButton-strokeColor-hover, var(--radioButton-strokeColor-hover))
}

then you don't need to be having the specific .saltRadioButton:hover .saltRadioButton-error .saltRadioButtonIcon-circle selector for example, and there are only 2 local tokens, which get changed on the error status, and only 2 css api tokens needed too

packages/lab/src/radio-button/RadioButtonIcon.css
```diff
@@ -1,78 +1,89 @@
...
 }
 
-/* Styles applied to RadioButtonIcon border */
-.saltRadioButtonIcon-border {
-  fill: var(--saltRadioButton-icon-fill, var(--radioButton-icon-fill));
-  stroke: var(--saltRadioButton-icon-strokeColor, var(--salt-selectable-borderColor));
+/* Styles applied to radio button icon circle */
+.saltRadioButtonIcon-circle {
+  fill: var(--saltRadioButton-fill, var(--salt-container-primary-background));
+  stroke: var(--saltRadioButton-strokeColor, var(--salt-selectable-borderColor));
```
libertymayc, line 16: @navkaur76 has just confirmed the border tokens should be selectable foreground

alycrys: updated ;)

libertymayc (approved):

PR #1420, opened 2023-03-31
Accent foreground disabled

libertymayc:


packages/theme/css/foundations/palette.css
```diff
@@ -20,6 +20,7 @@
   --salt-palette-accent-border: var(--salt-color-blue-500);
   --salt-palette-accent-border-disabled: var(--salt-color-blue-500-fade-border);
   --salt-palette-accent-foreground: var(--salt-color-white);
+  --salt-palette-accent-foreground-disabled: var(--salt-color-white-fade-foreground);
```
alycrys, line 23: is this correct?

libertymayc: yes, accent is used on a blue background so white with opacity is fine - same as what we do with button (actionable foreground cta)

PR #1434, opened 2023-04-02
removing setTimeout and adding onBlur event on the cell grid

AmeyDagaria007:
Cell value changed to new value entered, unless used "Esc" key
now this behaviour is working fine 
Please Review @origami-z

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -453,7 +453,6 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     rowIdx: number,
     colIdx: number
   ) => {
-    setTimeout(() => {
       const selector =
         part === "body"
           ? `td[data-row-index="${rowIdx}"][data-column-index="${colIdx}"]`
@@ -464,7 +463,6 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
       } else {
         console.warn(`focusCellElement can't find the element`);
       }
-    }, 0);
   };
 
   const startEditMode = (text?: string) => {
--- a/packages/grid/src/NumericColumn.tsx
+++ b/packages/grid/src/NumericColumn.tsx
@@ -4,6 +4,7 @@ import { GridColumnModel, GridRowModel } from "./Grid";
 import {
   ChangeEventHandler,
   KeyboardEventHandler,
+  FocusEventHandler,
   ReactNode,
   useEffect,
   useRef,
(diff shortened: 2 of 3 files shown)


packages/grid/src/Grid.tsx
```diff
@@ -453,7 +453,6 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
     rowIdx: number,
     colIdx: number
   ) => {
-    setTimeout(() => {
```
ZarkoPernar, line 456: What is the reason you are removing this setTimeout? It was originally put here for a reason, to counter a race condition. I forgot the exact scenario but will need to do some manual QA locally to see if anything is broken.

packages/grid/src/NumericColumn.tsx
```diff
@@ -50,6 +50,10 @@ export function NumericCellEditor<T>(props: NumericEditorProps<T>) {
   const onChange: ChangeEventHandler<HTMLInputElement> = (e) => {
     setEditorText(e.target.value);
   };
+  
+  const onBlur: ChangeEventHandler<HTMLInputElement> = (e) => {
```
ZarkoPernar, line 54: This should be `FocusEventHandler<HTMLInputElement>`

ZarkoPernar (changes requested):
Thanks for the contribution Amey! I need to talk to our designer in case this functionality was intentional. Even if it was, we can still change it if enough people complain. Out of curiosity, are you using the salt data grid in production?

packages/grid/src/Grid.tsx
ZarkoPernar: Ah, turns out the test caught the scenario that has a race condition.

feliciackh: More context: removing `setTimeOut` results in one of the tests for keyboard navigation failing, we believe `setTimeOut` was initially added to counter virtualization lag, i.e. when we focus on a cell that hasn't been rendered yet

AmeyDagaria007:
> Thanks for the contribution Amey! I need to talk to our designer in case this functionality was intentional. Even if it was, we can still change it if enough people complain. Out of curiosity, are you using the salt data grid in production? (Feel free to reach out to me directly on Symphony)

Sorry is there any open slack or is symphoney an internal tool ?

packages/grid/src/NumericColumn.tsx
AmeyDagaria007: Yeah Sure will Update it Thanks a lot

packages/grid/src/Grid.tsx
AmeyDagaria007: @ZarkoPernar @feliciackh  Got it . The Reason to remove this was as it pushes it in event queue so it dosent run in sync and due to that onBlur function is not gettin called as it focus after a while .

AmeyDagaria007:
Thanks @ZarkoPernar Sure . if possible can you please give the link for cla ?

origami-z:
@AmeyDagaria007 See contribution guide [here](https://github.com/jpmorganchase/.github/blob/main/CONTRIBUTING.md).

PR #1437, opened 2023-04-03
Rename grid folder to data-grid

joshwooding:
Renames packages/grid to packages/data-grid.

This is so that all our package folder names match the package name.


feliciackh (dismissed):

joshwooding:
Will do this again after the grid PRs are merged

PR #1410, opened 2023-03-30
Countries

el-dav:
PR Summary:

New package called 'countries'

A base CountrySymbol component (similar to Icon) which is used as the wrapper for all the CountrySymbol components
Imported all the country symbol SVGs from figma

Created a script to:
- Generate the CountrySymbol components from the SVGs
- Optimise the svgs of the resulting CountrySymbol components
- Create an index file to export all the CountrySymbol components
- create a lazyMap for use in LazyCountrySymbol component
- Make the ID for each <mask> in an svg unique

Ran generation script

Created LazyCountrySymbol component which allows consumers to change the symbol using the code prop and uses code-splitting to (hopefully) reduce bundles size

Created relevant stories

Added tests

Added documentation

PR #1438, opened 2023-04-03
Align icons script with countries script

joshwooding:
Aligns the icons script with #1410


packages/icons/scripts/generateIcons.mjs
```diff
@@ -0,0 +1,230 @@
...
+
+    await fs.writeFile(newFilePath, formattedResult, {
+      encoding: "utf8",
+    });
+
+    return componentName;
+  }));
+};
+
+// Generate the index file to export the CountrySymbol components
```
el-dav, line 162:
```suggestion
// Generate the index file to export the Icon components
```

el-dav:
Could you double check if you need to run yarn after adding rimraf? I get a change when I run yarn locally after checking out the branch

joshwooding:
> Could you double check if you need to run yarn after adding rimraf? I get a change when I run yarn locally after checking out the branch

Thanks, done!

PR #1451, opened 2023-04-03
Add `sitemap.xml` and `robots.txt` to mosaic site

JoanaMMoreira:
- Add `sitemap.xml` and `robots.txt` to mosaic site 
- Add script to copy over `sitemap.xml` everytime we generate a snapshot


Fercas123 (dismissed):

mosaic-site/public/sitemap.xml
joshwooding, line 1: This site map is generated every time but it’s only being copied statically?

Do you think it’s worth adding a script to copy this over instead?

JoanaMMoreira: Yeah that makes sense, should the script run when we generate the snapshots?

joshwooding: Makes sense :)

PR #1461, opened 2023-04-04
Make easier to hide icon in external links

libertymayc:
Make easier to hide icon in external links by changing types

### DIFF
--- a/.changeset/violet-walls-act.md
+++ b/.changeset/violet-walls-act.md
@@ -3,3 +3,9 @@
 ---
 
 Make hiding icons in external links easier
+
+```tsx
+<Link IconComponent={null} href="#root" target="_blank">
+  This has no icon
+</Link>
+```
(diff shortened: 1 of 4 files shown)


joshwooding (approved):

joshwooding (approved):
Looks good

ImmanuelBaskaran (dismissed):

alycrys:
When we refactored Link it was agreed by design (along with Stella) that the icon on external links should always be visible. I'm curious what's the reason for this change

libertymayc:
> When we refactored Link it was agreed by design (along with Stella) that the icon on external links should always be visible. I'm curious what's the reason for this change

its for wrapping components like Card in Link

joshwooding:
@alycrys @libertymayc I think we should have this functionality. We should either not document it on the Link page or add a warning that this is only needed when you wrap a component in the Link

ImmanuelBaskaran (approved):

alycrys (approved):

PR #1462, opened 2023-04-04
Data Grid / Click outside confirms value

### LINKED ISSUE #1372: Data Grid Editable Cell not updated after blur
### Package name(s)

Data Grid v1.0.0 (@salt-ds/data-grid)

### Description

After editing a cell value, clicking away will revert cell value instead of accepting it

### Steps to reproduce

1. Go to data grid [storybook](https://storybook.saltdesignsystem.com/?path=/story/documentation-data-grid-data-grid--page)
2. Found "Editable cells" example
3. Double click any of the editable cell to enter edit mode
4. Make some changes to the value
5. Click away to other cells
6. Cell value reverted to original value

### Expected behavior

Cell value changed to new value entered, unless used "Esc" key

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

ZarkoPernar:
### Description
**The work in this PR is split logically by commit.** Closes #1372 

### Why the refactor?
Currently we store the editor input value in the editor component. This means if we wanted to confirm the editor value when we click outside that editor we have to somehow react to an event outside the component and report the current value to the Grid. Initially, I attempted to do this with listening on `onBlur` on the input, however this didn't work because there is a tricky race condition where we destroy the component before the onBlur is called. It could have been solved potentially, but it led down a rabbit hole of handling race conditions and various event handlers being refactored.

### Changes
- Pulled out the editor value from the editor component to the top of the Grid. This comes with some performance penalty because now typing triggers a re-render in the Grid. This should be fine for now as we don't have any noticeable issues with other state changes in Grid like cell navigation or selection.
- With this refactor the fix for clicking outside and confirming value is easier since we don't have any race conditions. We listen to the global click, or a click on a different cell, and end the editing with the final value. There is a slight conflict with onMouseDown listener on the cell and the global click listener so we check if the click hasn't originated inside the body of the grid.
- Fixed a bug where you can't start typing with a letter "a"
- Added some missing tests


packages/grid/src/Grid.tsx
```diff
@@ -702,6 +713,9 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
       if (part === "header" && !headerIsFocusable) {
         return;
       }
+      if (editMode) {
```
origami-z, line 716: What if click on header or other intractable elements outside of the grid?

ZarkoPernar: Yeah, I was thinking of setting up a global click listener but then realised it will have conflicts with existing listeners. Also, Im pretty sure calling `.stopPropagation()` on a react event will not stop it from bubbling to the actual dom listeners? Maybe we handle this in next iteration.

ZarkoPernar: But to answer your question, nothing 😅

ZarkoPernar: Actually, I think I found a solution. Stay tuned.

ZarkoPernar: @origami-z Ok, I have now updated the PR with a solution for this. Check out the [listen to click on the page and on the header](https://github.com/jpmorganchase/salt-ds/pull/1462/commits/b6e6bd7c42965576f237e5ab45acfa00597c733c) commit.

Fercas123 (dismissed):

packages/grid/src/EditorContext.ts
```diff
@@ -1,10 +1,11 @@
 import { createContext, useContext } from "react";
 
 export interface EditorContext {
-  initialText?: string;
+  editorText: string;
```
mark-tate, line 4: this would be considered a breaking change  in future ?

How will we handle breaking changes, once the grid gets adoption ?
Do we know if anyone is using this already ?

PR #1481, opened 2023-04-05
Add site redirects and remove remote source

JoanaMMoreira:
- Add site redirects so users can access pages without using the salt prefix e.g. `https://next.saltdesignsystem.com/getting-started`, `https://next.saltdesignsystem.com/components`, etc
- Remove remote source as it's not required
- Update snapshots


Fercas123 (approved):

mark-tate (approved):

james-nash:
The behaviour I'm seeing when I run this locally is:

* Open `http://localhost:3000/`, and my browser is redirected (via HTTP 308 permanent redirect) to `http://localhost:3000/salt/index`
* The nav links in the header all point to URL paths prefixed with `salt/`
* Entering any "old" URLs, e.g. `http://localhost:3000/getting-started/`, does an HTTP redirect to the corresponding `salt/getting-started/index` URL

If this is the expected behaviour of these changes, then I'm happy to approve this PR.

However, longer term, I still want us to get back to the current, "cleaner" URL structure without a `salt/` prefix and `index` suffix. Happy to raise separate issue for Mosaic and/or our site to get that on our backlog(s) to be resolved in the future.

JoanaMMoreira:
> The behaviour I'm seeing when I run this locally is:
> 
> * Open `http://localhost:3000/`, and my browser is redirected (via HTTP 308 permanent redirect) to `http://localhost:3000/salt/index`
> * The nav links in the header all point to URL paths prefixed with `salt/`
> * Entering any "old" URLs, e.g. `http://localhost:3000/getting-started/`, does an HTTP redirect to the corresponding `salt/getting-started/index` URL
> 
> If this is the expected behaviour of these changes, then I'm happy to approve this PR.
> 
> However, longer term, I still want us to get back to the current, "cleaner" URL structure without a `salt/` prefix and `index` suffix. Happy to raise separate issues for Mosaic and/or our site to get that on our backlog(s) to be resolved in the future.

Yeah this is the only workaround for now, there will need to be some work on the Mosaic side to allow a default source so it doesn't have to be explicit in the url like it is now

james-nash (approved):

PR #1501, opened 2023-04-11
Fix a bug where the Data Grid cell editor value is dismissed when the user clicks away

joshwooding:
An investigation into an alternative fix to #1462.

This PR reverts https://github.com/jpmorganchase/salt-ds/pull/1462 and tries to solve it differently it avoids API changes and focuses on fixing the bug.

It will be easier to look at specific commits to review it.

The three commits here are:
The revert of the original commit
The commit that adds the fix
A commit that cherrypicks the changeset and tests

### DIFF
--- a/packages/grid/src/Grid.tsx
+++ b/packages/grid/src/Grid.tsx
@@ -227,7 +227,7 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
   );
 
   const [editMode, setEditMode] = useState<boolean>(false);
-  const [editorText, setEditorText] = useState<string>("");
+  const [initialText, setInitialText] = useState<string | undefined>(undefined);
 
   const resizeClient: ScrollableProps<T>["resizeClient"] = useCallback(
     (dimensions) => {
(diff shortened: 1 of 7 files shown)


feliciackh (dismissed):

packages/grid/src/TextCellEditor.tsx
```diff
@@ -50,15 +53,26 @@ export function TextCellEditor<T>(props: TextCellEditorProps<T>) {
...
+
+  useEffect(() => {
+    const input = inputRef.current;
+    const focusOut = (event: FocusEvent) => {
+      if (!input?.contains(event.target as Node)) {
+        endEditMode(editorText);
+      }
+    };
+
+    document?.addEventListener("mousedown", focusOut, true);
```
origami-z, line 70: Do we need some comment around capture v.s. race condition?

origami-z (commented):
I don't think numeric editor works? Was testing on 2nd column below
https://316a45cd.saltdesignsystem-storybook.pages.dev/?path=/story/data-grid-data-grid--editable-cells

packages/grid/src/TextCellEditor.tsx
joshwooding: Added :)

joshwooding:
> I don't think numeric editor works? Was testing on 2nd column below https://316a45cd.saltdesignsystem-storybook.pages.dev/?path=/story/data-grid-data-grid--editable-cells

Ah yes, I think it wasn't fixed in the original PR. Just added a commit that fixes this there too. I tested it with Dropdown and it works

packages/grid/src/Grid.tsx
```diff
@@ -471,45 +471,39 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
...
+  const endEditMode = (value: string) => {
+    if (!editMode) {
+      return;
+    }
+    if (cursorColIdx == undefined) {
+      console.error(`endEditMode: cursorColIdx is undefined in edit mode`);
+      return;
+    }
+    const c = cols[cursorColIdx];
+    const handler = c.info.props.onChange;
```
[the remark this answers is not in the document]
joshwooding: Those changes are due to a revert and not a manual change. I've made it easier to review the PR by keeping things in separate commits and the first should be ignored.

PR #1502, opened 2023-04-11
When to use Panel guidance; color bug fix

libertymayc:
When to use Panel guidance added to docs


ImmanuelBaskaran (approved):

packages/core/src/panel/Panel.css
```diff
@@ -11,7 +11,7 @@
 /* Styles applied to the root element */
 .saltPanel {
   background: var(--saltPanel-background, var(--panel-background));
-  color: var(--saltPanel-color, inherit);
+  color: var(--saltPanel-color, initial);
```
joshwooding, line 14: Is this change missing a changeset? cc @libertymayc @ImmanuelBaskaran

PR #1504, opened 2023-04-11
Site design tweaks

JoanaMMoreira:
- hide scrollbar on docs side navigation
- remove hover background colour on navbar items
- add base root text styles to match Docusaurus
- amend features list icon colour
- remove hover background colour on hamburger menu button
- increase spacing between GitHub logo and label on mobile menu
- remove hover background colour on mobile menu close button
- amend mobile menu items font size
- change background image size on docs to make all pages consistent
- remove duplicate utils function
- update code blocks font family


james-nash (approved):
Nicely done! 👏

joshwooding (approved):

mosaic-site/src/_index/features/Features.module.css
```diff
@@ -1,3 +1,7 @@
+.listItem > svg {
+  --saltIcon-color: var(--salt-actionable-primary-background-active);
```
alycrys, line 2: Just checking, shouldn't this be 'foreground'?
CC @libertymayc

mosaic-site/src/css/global/code.css
```diff
@@ -10,7 +10,7 @@ pre {
   --pre-color: #393a34;
   --code-font-size: 90%;
   --pre-line-height: 1.45;
-  --font-family-monospace: SFMono-Regular, Menlo;
+  --font-family-monospace: var(--salt-text-code-fontFamily);
```
alycrys, line 13: this var is only being used in one place. Do you think it makes sense to remove it and use `--salt-text-code-fontFamily` directly?

JoanaMMoreira: I think this is just to help simply the code on line 72 ` font: var(--code-font-size) / var(--pre-line-height) var(--font-family-monospace);`

mosaic-site/src/_index/features/Features.module.css
alycrys: could be the right choice, was just checking. If I remember well, @pseys mentioned we should use foreground for SVGs, but he might be referring only to Salt components

mosaic-site/src/css/global/code.css
alycrys: oh, I thought those vars were created because they had custom values.

alycrys (approved):

PR #1508, opened 2023-04-11
Fix ag-grid-theme icon mapping

joshwooding:


origami-z:
indeterminate may need a font regeneration, it's quite noticeably misaligned to the left 

Could add `columnGroupShow` use case to "Column Group" example to test icon mapping for expand/collapse column grouping

joshwooding:
> indeterminate may need a font regeneration, it's quite noticeably misaligned to the left
> 
> Could add `columnGroupShow` use case to "Column Group" example to test icon mapping for expand/collapse column grouping

We should raise the font regeneration as a separate issue but I've updated the Column Group example :)

Issue #431, opened 2022-09-21
Form Field v1


ZarkoPernar:
Figma Link: https://www.figma.com/file/KSpYg21r69mHjXaQwWgAu3/Form-Field
___

```[tasklist]
### Released in
- [ ] Design
- [x] Dev
```

```[tasklist]
### Tasks
- [ ] https://github.com/jpmorganchase/salt-ds/issues/1030
- [ ] https://github.com/jpmorganchase/salt-ds/issues/1742
```

All versions assume support for:
- All densities
- All modes
- Accessibility considerations

 
### Version 1

1. Form label
  a. Placement: Top, Right (on left)
  b. Width: 100%
  c. Overflow: Wrap
2. States: Default, Hover, Selected, Read-only, Disabled
  a. Support for custom states (Markets’ ‘Non-default’)
3. Help text
  a. Width: 100%
  b. Overflow: Wrap
4. Focus ring support
5. Validation text: Error, Warning, Success
  a. Width: 100%
  b. Overflow: Wrap
6. Integrations: Input
  a. Activation line
  b. Variants: Primary, Secondary
  c. Validation states (inc. fill & icon Fixed): Error, Warning
7. Control child validation state

parsakhan21:
Ben to work on up front design work. Start on Monday.

james-nash:
11/24: Meeting today to review proposals

parsakhan21:
1/18 - no progress yet on this; focusing on datagrid - wrap up by tomorrow

JohannaJoergensen:
@bhoppers2008 - any updates on this?

joshwooding:
Moved to blocked due to lack of current capacity

joshwooding:
@joshwooding and @alycrys to review PR

libertymayc:
@bhoppers2008 documenting CSS API for explicit use cases:

--saltFormFieldNext-width (default is 100%)
--saltInputNext-paddingLeft (should be used when embedded)
--saltInputNext-textAlign (default is left)

for custom states (e.g. Markets non default)
--saltInputNext-background
--saltInputNext-background-active
--saltInputNext-background-hover
--saltInputNext-borderColor
--saltInputNext-borderColor-active
--saltInputNext-borderColor-hover
--saltInputNext-borderColor-focused
--saltInputNext-outlineColor


*note these will change to --saltInput prefix once we deprecate the current Input and switch over to the new one :-)

origami-z:
Figma library released

PR #1510, opened 2023-04-12
Form Field Next

libertymayc:
@bhoppers2008 #431 

Included in this PR:
- General design changes
- Focus ring support
- Form label and helper text
- Basic API decisions [RFC]
- InputNext basic component
- Disabled state for FormField and InputNext
- Read only state for FormField and InputNext
- InputNext - activation line (new design - always on Input)

Not included:
- Static adornments, can come in separate PR to avoid this becoming too large
- Validation states, can come in separate PR to avoid this becoming too large
- Specific support for controls (out of v1 scope, just demonstrated here)
- Specific support for multiple inputs (out of v1 scope, just demonstrated here)
- Documentation - will be added when decisions are more solidified


el-dav:
Design question: 
With the design changes does the variant prop now have any affect on a FormField which wraps Radio/Buttons or checkboxes? E.g. in [this story](https://3ff4a91b.saltdesignsystem-storybook.pages.dev/?path=/story/lab-form-field-next--with-controls&args=variant:secondary).

If not then maybe variant is actually a prop of input type components rather than the FormField

libertymayc:
> Design question: With the design changes does the variant prop now have any affect on a FormField which wraps Radio/Buttons or checkboxes? E.g. in [this story](https://3ff4a91b.saltdesignsystem-storybook.pages.dev/?path=/story/lab-form-field-next--with-controls&args=variant:secondary).
> 
> If not then maybe variant is actually a prop of input type components rather than the FormField?
> 
> @bhoppers2008

yeah, variants will come under Input now - one thing happening here though is Dropdown uses Button and the variants for Button (i.e. secondary specifically) doesn't match up with the secondary variant here - although, that could just be a Dropdown variant I guess, that overrides the Button styling within Dropdown, which would be ok, if that's appropriate for Dropdown when used outside of FormField too

el-dav:
@joshwooding 
One thing that we discussed and should maybe explore is to use a more composition-based API for FormField e.g.

```jsx
<FormField>
  <Label>My Label</Label>
  <Input />
  <HelperText>My HelperText</HelperText>
</FormField>
```

This would be a pretty big change and so needs more discussion/confirmation. It has a little more setup for the consumer in the simple use case but has certain advantages e.g.
- Can pass a ref/prop to each component individually
- More flexibility on how to arrange the components
- Could work nicely with our Layout components

libertymayc:
> @joshwooding One thing that we discussed and should maybe explore is to use a more composition-based API for FormField e.g.
> 
> ```js
> <FormField>
>   <Label>My Label</Label>
>   <Input />
>   <HelperText>My HelperText</HelperText>
> </FormField>
> ```
> 
> This would be a pretty big change and so needs more discussion/confirmation. It has a little more setup for the consumer in the simple use case but has certain advantages e.g.
> 
> * Can pass a ref/prop to each component individually
> * More flexibility on how to arrange the components
> * Could work nicely with our Layout components

I like this a lot, @bhoppers2008 and I were thinking about this yesterday using Grid Layout for alignment of the labels when placed on the left, this would make that possible

packages/lab/src/form-field-next/FormFieldNext.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { FormFieldContext } from "../form-field-context";
+import { FormFieldLabel, FormFieldLabelProps } from "./FormFieldLabel";
+import { FormHelperText } from "./FormHelperText";
+
+import "./FormFieldNext.css";
+
+export type FormFieldLabelPlacement = "top" | "left";
+
+export interface FormFieldProps
+  extends HTMLAttributes<HTMLDivElement>,
```
el-dav, line 19: API is already feeling cleaner :)

joshwooding:
I would prefer to use the compositional API that is being discussed. @libertymayc I agree with you here from the looks of the requirements, a component to wrap the controls seems necessary

packages/lab/src/dropdown-next/DropdownBase.tsx
```diff
@@ -0,0 +1,182 @@
...
+        popupWidth,
+        rootRef,
+        width,
+      }
+    );
+    const [maxPopupHeight, setMaxPopupHeight] = useState<number | undefined>(
+      undefined
+    );
+
+    const middleware = isDesktop
```
ImmanuelBaskaran, line 86: the isDesktop boolean will likely be removed in favor of a cleaner solution.

Perhaps we could remove it early.

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+/* TODO: These issues (in helper text) need consideration */
+
+export const MultiInput: ComponentStory<typeof FormFieldNext> = (props) => {
+  return (
+    <div style={{ width: "366px" }}>
+      <div style={{ height: 16 }} />
+      <FormFieldNext
+        label="Paired fields"
+        helperText="*User entry in either field will automatically poluate the corresponding field with the correct value"
```
ImmanuelBaskaran, line 125:
```suggestion
        helperText="*User entry in either field will automatically populate the corresponding field with the correct value"
```

libertymayc:
Q for devs: should we change disabled and readonly props to one 'state' prop that's an enum, since both can't/shouldn't be applied at once?

libertymayc:
> > @joshwooding One thing that we discussed and should maybe explore is to use a more composition-based API for FormField e.g.
> > ```js
> > <FormField>
> >   <Label>My Label</Label>
> >   <Input />
> >   <HelperText>My HelperText</HelperText>
> > </FormField>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This would be a pretty big change and so needs more discussion/confirmation. It has a little more setup for the consumer in the simple use case but has certain advantages e.g.
> > 
> > * Can pass a ref/prop to each component individually
> > * More flexibility on how to arrange the components
> > * Could work nicely with our Layout components
> 
> I like this a lot, @bhoppers2008 and I were thinking about this yesterday using Grid Layout for alignment of the labels when placed on the left, this would make that possible It does though add more room for error and slightly more work for user (e.g., ensuring states are given to the control and helper text consistently)
> 
> we would possibly also need to consider providing a wrapper i.e.
> 
> ```js
> <FormField>
>   <Label>My Label</Label>
>   <FormFieldControls>
>       <Input />
>       <Input />
>    </FormFieldControls>
>   <HelperText>My HelperText</HelperText>
> </FormField>
> ```
> 
> unless we want to give the responsibility of multi inputs layout being handled by the user - or otherwise recommend which Layout to use for controls in this case

**This exploration will come in next PR.**

origami-z:
If we're aiming to merge this with three next components, then will need to document the plan of what to do with the existing ones. There are various teams using them now, so migration / adoption should not be an afterthought. 

e.g. will next be the ones eventually promoted to core? what's the difference between current input / dropdown and new ones?

packages/lab/src/form-field-next/FormFieldLabel.tsx
```diff
@@ -0,0 +1,33 @@
...
+  label?: string;
+}
+
+export const FormFieldLabel = ({
+  className,
+  disabled,
+  label,
+  ...restProps
+}: FormFieldLabelProps) => (
+  <Label
```
alycrys, line 25: why not use variant and disabled props?

packages/lab/src/form-field/FormHelperText.css
```diff
@@ -33,7 +33,7 @@
 }
 
 /* Class applied to the helper text container */
-.saltFormFieldHelperText {
```
alycrys, line 36: i think we shouldn't touch this component at all

packages/lab/src/form-field-next/FormFieldLabel.tsx
libertymayc: variant is always secondary - have improved and used disabled prop

packages/core/src/text/Label.tsx
```diff
@@ -1,9 +1,11 @@
 import { forwardRef } from "react";
 import { Text, TextProps } from "./Text";
 
+export type LabelProps = Omit<TextProps<"label">, "as">;
+
 export const Label = forwardRef<
   HTMLLabelElement,
-  Omit<TextProps<"label">, "as">
+  LabelProps
```
alycrys, line 8: Is this necessary? I would avoid mixing this in

packages/lab/src/form-field-next/FormFieldHelperText.tsx
```diff
@@ -0,0 +1,31 @@
+import clsx from "clsx";
+import { Label, LabelProps, makePrefixer } from "@salt-ds/core";
+import { FormFieldProps } from "./FormFieldNext";
+
+import "./FormFieldHelperText.css";
+
+const withBaseName = makePrefixer("saltFormFieldHelperText");
+
+export interface FormFieldHelperTextProps
+  extends Omit<LabelProps, "variant" | "styleAs"> {
```
alycrys, line 10: You can use TextProps instead. No need to add the LabelProps

packages/core/src/text/Label.tsx
libertymayc: makes sense, done

packages/lab/src/form-field-next/FormFieldNext.css
```diff
@@ -0,0 +1,62 @@
...
+  vertical-align: top;
+  width: var(--saltFormFieldNext-width, 100%);
+}
+
+.saltFormFieldNext-labelLeft {
+  align-self: center;
+  grid-template-columns: auto 1fr;
+}
+
+.saltFormFieldNext-labelTop {
```
Fercas123, line 19: do we need this class?

packages/lab/src/form-field-next/FormFieldHelperText.tsx
```diff
@@ -0,0 +1,31 @@
...
+}
+
+export const FormFieldHelperText = ({
+  className,
+  disabled,
+  helperText,
+  ...restProps
+}: FormFieldHelperTextProps) => {
+  return (
+    <Label
```
alycrys, line 22: I'm thinking that help text shouldn't return a label. Having two label elements is confusing. Shall we use a 'div' instead? 
Text returns a div element by default. Should be something like:
```
<Text styleAs="label" />
```

Fercas123 (approved):
Looks good! just a non blocking small one

packages/lab/src/form-field-next/FormFieldNext.css
libertymayc: I'm going to keep empty for now as the grid css might change and want to remind us this is how we're doing it

packages/lab/src/form-field-next/FormFieldHelperText.tsx
libertymayc: Zarko suggested the same, have changed :)

packages/lab/src/form-field/FormHelperText.css
libertymayc: I've kept this because the name of the file is FormHelperText and so keeping patterns the class name should match the file name
It is also making storybook confused as these styles are overriding the form field next helper text padding due to having the same class name (idk why that's happening as that css file is not imported, i guess it's all bundled together but ideally that shouldnt happen?)

packages/lab/src/form-field-next/FormFieldNext.tsx
```diff
@@ -0,0 +1,147 @@
...
+  ) => {
+    const rootRef = useRef<HTMLDivElement>(null);
+
+    const handleTriggerRef = useForkRef(rootRef, ref);
+
+    const labelId = useId(LabelProps?.id);
+
+    const a11yValue = useA11yValue({
+      labelId,
+      helperText,
```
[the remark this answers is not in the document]
joshwooding: @alycrys is right, aria-described by should take an id rather than the actual text

packages/lab/src/form-field-next/FormFieldLabel.tsx
```diff
@@ -0,0 +1,33 @@
...
```
joshwooding, line 25: This label is not currently hooked up the control inside the FormField

packages/lab/src/form-field/FormHelperText.tsx
```diff
@@ -17,7 +17,7 @@ export const FormHelperText = <E extends React.ElementType = "p">({
   if (helperText) {
     if (helperTextPlacement === "bottom") {
       return (
-        <p className={`saltFormFieldHelperText`} {...restProps}>
+        <p className={`saltFormHelperText`} {...restProps}>
```
joshwooding, line 20:
```suggestion
        <p className="saltFormHelperText" {...restProps}>
```

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export default {
+  title: "Lab/Form Field Next",
+  component: FormFieldNext,
+} as ComponentMeta<typeof FormFieldNext>;
+
+export const Skeleton: ComponentStory<typeof FormFieldNext> = (props) => {
+  return (
+    <div style={{ width: "366px" }}>
+      <div style={{ height: 16 }} />
```
joshwooding, line 13: Nit: Can we use the layout components here instead

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -0,0 +1,217 @@
...
+  return (
+    <>
+      <FormFieldNext
+        helperText="Default helper text"
+        label="Default Form Field label"
+        {...props}
+      >
+        <InputNext defaultValue="Value" />
+      </FormFieldNext>
+      <div style={{ height: 40 }} />
```
joshwooding, line 40: Same here too

packages/lab/src/form-field-next/FormFieldLabel.tsx
libertymayc: I think this is fixed, but may be missing something ?

joshwooding: It's fixed now

packages/lab/src/form-field-next/FormFieldLabel.css
```diff
@@ -0,0 +1,9 @@
+.saltFormFieldLabel.saltText {
+  align-self: center;
+  font-weight: var(--saltFormField-label-fontWeight, var(--salt-text-fontWeight-strong));
+  width: 100%;
+}
+
+.saltFormFieldLabel:hover + .saltFormFieldNext-controls * {
+  --inputNext-borderColor: var(--salt-editable-borderColor-hover);
```
alycrys, line 8: is this needed?

packages/lab/src/form-field-next/FormFieldNext.css
```diff
@@ -0,0 +1,62 @@
+.saltFormFieldNext {
+  border: 0;
+  display: inline-grid;
+  font-size: var(--saltFormFieldNext-fontSize, var(--salt-text-fontSize));
```
alycrys, line 4: not sure we need to set the font-size. The Text components already have this in

packages/lab/src/form-field-next/FormFieldNext.css
```diff
@@ -0,0 +1,62 @@
...
+  min-width: 0;
+  position: relative;
+  padding: 0;
+  column-gap: var(--salt-size-adornmentGap);
+  row-gap: var(--salt-size-adornmentGap);
+  vertical-align: top;
+  width: var(--saltFormFieldNext-width, 100%);
+}
```
alycrys, line 12: pretty sure we don't need the `border`, `min-width` and `padding` here

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -0,0 +1,222 @@
...
+   */
+  defaultValue?: HTMLInputElement["defaultValue"];
+  /**
+   * If `true`, the component is disabled.
+   */
+  disabled?: HTMLInputElement["disabled"];
+  /**
+   * The HTML element to render the Input, e.g. 'input', a custom React component.
+   */
+  inputComponent?: ElementType;
```
alycrys, line 39: i think we should remove `inputComponent`. @joshwooding do you have any info about why this was added initially?

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -0,0 +1,222 @@
...
+  /**
+   * Callbacks for events.
+   */
+  onChange?: (event: ChangeEvent<HTMLInputElement>, value: string) => void;
+  onFocus?: FocusEventHandler<HTMLInputElement>;
+  onKeyDown?: KeyboardEventHandler<HTMLInputElement>;
+  onKeyUp?: KeyboardEventHandler<HTMLInputElement>;
+  onMouseUp?: MouseEventHandler<HTMLInputElement>;
+  onMouseMove?: MouseEventHandler<HTMLInputElement>;
+  onMouseDown?: MouseEventHandler<HTMLInputElement>;
```
alycrys, line 54: I feel most of these can be passed inside `inputProps` and no need to add them as props. Worth discussing

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -0,0 +1,222 @@
...
+      disabled: a11yDisabled,
+      readOnly: a11yReadOnly,
+      ...restA11y
+    } = {},
+  } = useFormFieldPropsNext();
+
+  const isDisabled = disabled || a11yDisabled;
+  const isReadOnly = readOnlyProp || a11yReadOnly;
+
+  const [focused, setFocused] = useState(false);
```
alycrys, line 135: don't think we need `focused`

libertymayc: this is used for the styling since tab focuses on the inner input rather than the outer div. the CSS is a bit funky if we want to do it the other way?

packages/lab/src/input-next/InputNext.tsx
alycrys: not really. The only thing we do with these handlers is pass them down to the input, so i think that if someone needs them, they can add them to inputProps themselves

packages/lab/src/input-next/InputNext.tsx
alycrys: this should work

```
-.saltInputNext-focused {
+.saltInputNext:has(:focus) {
```

libertymayc: are we allowed to use has yet?

libertymayc:
seen some changes in chromatic, currently fixing

alycrys (approved):

PR #1522, opened 2023-04-14
Strict lint rule for token camel case usage

libertymayc:
Lint rule only allows for defined CSS attributes and merged states (e.g. activeDisabled) to be allowed in tokens


packages/theme/stories/foundations/size.stories.mdx
```diff
@@ -12,8 +12,10 @@ All size values are built on the basis of 4px, `--salt-size-basis-unit`. We enco
 
 <DocGrid>
   <SpacingBlock spacingVar="--salt-size-accent" />
+  <SpacingBlock spacingVar="--salt-size-adornment-gap" />
```
libertymayc, line 15: adornmentGap wasn't documented, so I don't feel the need to document the deprecation

joshwooding: I think the non-documentation was a bug. We should document it as deprecated :)

joshwooding (commented):
Where did this change request come from? Was it noticed elsewhere? I thought that the style lint setup should catch these sort of issues?

feliciackh (approved):

libertymayc:
> Where did this change request come from? Was it noticed elsewhere? I thought that the style lint setup should catch these sort of issues?

it wasn't that strict; but I've changed this PR to make it be

libertymayc:
@joshwooding 
Added lint rule, which is fine for core - but if you look at the failing list I think we may not be able to do it like this, depending on how strict we want to go...

Tokens like `--calendar-day-outOfRange-text-color`, I think outOfRange is better than out-of-range, in this scenario (personal preference), anything that stands as one piece should be camel case - but things like `-adornmentGap` should be `-adornment-gap`, because it would make sense to have `-adornment-color`, `-adornment-size`... etc - which can only be determined by us if it makes sense or not, not a lint rule

Thoughts?

.changeset/famous-queens-dream.md
```diff
@@ -0,0 +1,13 @@
+---
+"@salt-ds/core": minor
+"@salt-ds/lab": minor
+"@salt-ds/theme": minor
+---
+
+Lint rule to strictly only allow for CSS attributes and merged states (`activeDisabled`, `blurSelected`, `partialDisabled`, `selectedDisabled`) to be in camel case within tokens
+
+Deprecated `--size-salt-adornmentGap`; replaced with `--salt-size-adornment-gap`
```
origami-z, line 9: This just got [released yesterday](https://github.com/jpmorganchase/salt-ds/releases/tag/%40salt-ds%2Ftheme%401.3.0) and now we're asking the users to change again. 

I think it's time we take a pause and think about how to deal with theme token additional / change. 

@joshwooding @pseys

joshwooding: Totally agree, I've already raised this.

PR #1543, opened 2023-04-17
Re-add lang attribute on site

joshwooding:
My previous PR that fixed font removed the lang attribute on the HTML element. This add it back

### DIFF
--- a/site/src/pages/_document.tsx
+++ b/site/src/pages/_document.tsx
@@ -1,0 +1,13 @@
+import { Html, Head, Main, NextScript } from "next/document";
+
+export default function Document() {
+  return (
+    <Html lang="en">
+      <Head />
+      <body>
+        <Main />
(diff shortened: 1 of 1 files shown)


feliciackh (approved):

origami-z:
Is there a need for an inline comment like "this is exactly the same as mosaic template, but additional lang"?

Or from another angle, shall we add this as default to mosaic?

joshwooding:
> Is there a need for an inline comment like "this is exactly the same as ~mosaic~ next template, but additional lang"?
> 
> ~Or from another angle, shall we add this as default to mosaic?~

Not really sure, but adding it wouldn't hurt

feliciackh (approved):

JoanaMMoreira (approved):

alycrys (approved):

PR #1553, opened 2023-04-19
Improve accessibility of Card

libertymayc:
ADA feedback addressed


packages/core/stories/card/card.stories.tsx
```diff
@@ -10,12 +10,14 @@ import {
   Text,
   FlexLayout,
   GridLayout,
-  Label,
   Link,
+  Display1,
```
alycrys, line 14: doesn't seem to be used

alycrys (approved):

alycrys (approved):

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,12 +26,26 @@ This default example is intended to show the way the Card component’s content
...
+<Canvas>
+  <Story id="core-card--default-with-button-and-image" />
+</Canvas>
+
 ## Interactable Card
 
-To apply actions to Cards use the `InteractableCard` component. We discourage the use of any focusable elements within interactable Cards.
+Use the `InteractableCard` component to make the entire Card interactable. Do **not** use any focusable elements within Interactable Card.
+
+To use as a link, wrap Card in a [Link](?path=/docs/documentation-core-link--page) component.
```
libertymayc, line 45: Is this ok? @joshwooding and should we add something like 'Keep content minimal as all content will be read out by screen reader' 
cc @bhoppers2008 @ivan-calderon

packages/core/stories/card/card.doc.stories.mdx
```diff
@@ -26,12 +26,26 @@ This default example is intended to show the way the Card component’s content
   <Story id="core-card--default" />
 </Canvas>
 
+## Adding actions
+
+If you intend to add any links or buttons within your Card, only add them into a default Card, not an interactable Card, to ensure ADA compliancy.
```
libertymayc, line 31: Should we add advice/a story on how to add link styling to a "link" in interactable cards, like on the current site? @joshwooding @ivan-calderon

packages/core/stories/card/card.doc.stories.mdx
joshwooding: Yeah, I think we should add that guidance as well as shorten the text in the example.

packages/core/stories/card/card.doc.stories.mdx
joshwooding: It depends if it's a valid pattern. I'll let @ivan-calderon answer

packages/core/stories/card/card.doc.stories.mdx
bhoppers2008: Maybe content (Delme or Mitch) might have some guidance that we can add/link to for this as well? Came across this article, which suggests there's no special consideration for screen readers regarding content... it should just be well written: https://accessibility.blog.gov.uk/2017/02/08/advice-for-creating-content-that-works-well-with-screen-readers/

Makes sense to reduce the text in the example, but we should recommend against multiple lines regardless of a screen reader. It should be easily scannable by all users. Given the speed of most screen readers, it's probably more of an issue for sighted than non-sighted users :) Maybe we can tweak the guidance to reflect more general advice, with screen readers being one of the supporting reasons.

packages/core/stories/card/card.doc.stories.mdx
libertymayc, line 33:
```suggestion
If you intend to add any links or buttons within your Card, only add them into a default Card, not an Interactable Card. This will ensure ADA compliance.
```

joshwooding (changes requested):
@libertymayc looks like the card tests are failing

joshwooding (approved):

Issue #1044, opened 2023-01-25
Migrate Component documentation from Storybook to the Site


joshwooding:
https://github.com/jpmorganchase/salt-ds/issues/1044#tasklist-block-d06f224f-b3fd-42ac-80e1-178877d3c3fc

james-nash:
Moved this to Q3, since we'll be using the first week of that quarter to tackle this

shinytoyrobots:
Is the intent to add all of the individual component documentation issues linked in this epic?

PR #1558, opened 2023-04-20
Add component page layout and button example

JoanaMMoreira:
- Add `Button` example pages
- Add new `DetailsBase` layout to be shared across getting started docs and component pages
- Add new `DetailComponent` layout specifically for component pages
- Update snapshots
- Page preview at https://saltdesignsystem-git-component-docs-spike-joshwooding.vercel.app/salt/components/button


site/public/sitemap.xml
```diff
@@ -34,5 +34,25 @@
 		<changefreq>weekly</changefreq>
 		<priority>0.5</priority>
 	</url>
+	<url>
+		<loc>/salt/components/button/accessibility.mdx</loc>
```
james-nash, line 38: Just noticed that this whole `sitemap.xml` file seems to be broken. The contents of `<loc>` should be the the full URL to the page (see https://www.sitemaps.org/protocol.html), but instead seems to be the path to the source file.

I realise that this bug has nothing to do with this PR, so I've raised a separate issue for it: #1559

james-nash (approved):
Nice work!

site/docs/components/button/accessibility.mdx
```diff
@@ -0,0 +1,27 @@
...
+
+#### When a Button has an icon only and no label
+
+- Pass an `aria-label` to Button that describes its purpose.
+
+### Keyboard interactions
+
+| Key                                  | Function                                                                                                                                                                                                                                                                |
+| ------------------------------------ | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| <kbd>Tab</kbd>                       | - If focus is above the button, Tab key press moves focus onto the button.<br/>- If the button is disabled and is set to receive focus, receives focus.<br/>- If the button has focus, Tab moves focus out of the button, to next focusable component in the tab order. |
```
libertymayc, line 25:
```suggestion
| <kbd>Tab</kbd>                       | - If focus is above the Button, Tab key press moves focus onto the Button.<br/>- If the Button is disabled and is set to receive focus, receives focus.<br/>- If the Button has focus, Tab moves focus out of the Button, to next focusable component in the tab order. |
```

site/docs/components/button/accessibility.mdx
```diff
@@ -0,0 +1,27 @@
...
+| <kbd>Tab</kbd> and <kbd>Shift</kbd>  | - If focused, Tab+Shift combo moves focus out of the component to the previous component in the tab order.                                                                                                                                                              |
+| <kbd>Enter</kbd> or <kbd>Space</kbd> | - If the button has focus, Enter or Space key press activates the button.<br/>- If the button is disabled and can receive focus, does nothing.                                                                                                                          |
```
libertymayc, line 27: I think we use capitals for component name (other comment)
There's also a difference here of 'if focused' and 'if the button has focus' between tab and shift and enter/space - I think the wording should be aligned for clarity

site/docs/components/button/accessibility.mdx
JoanaMMoreira: I'm happy to make these changes but this page for now is just an example that I've copied from Figma so we could see the layouts in action. I think we'll need a proper copy review as we add the component docs into the site. What do you reckon @james-nash @shey-v ?

libertymayc: awesome, I guess it will be checked anyway by content :)

james-nash: Agreed. This PR is just about the technical mechanism of rendering MDX content split across different files into the tabs of the component page layout.

Migrating existing content into the page layout (once it's all done) will be a separate exercise - see #1044

libertymayc (dismissed):

ZarkoPernar:
Im seeing some wild flickering going on when I switch between different tabs. Do you know what is causing this @JoanaMMoreira?

site/docs/components/button/accessibility.mdx
```diff
@@ -0,0 +1,27 @@
+---
+title: Button
+layout: DetailComponent
+sidebar:
+  label: Accessibility
+data:
+  descriptionRef:
```
mark-tate, line 7: NIT: I would not suffix it as ref, it's just description

site/src/layouts/index.ts
```diff
@@ -1,3 +1,4 @@
+export * from "./DetailComponent";
```
mark-tate, line 1: I would rename the Layouts and remove this Detail prefix (needs Mosaic change)

JoanaMMoreira:
> Im seeing some wild flickering going on when I switch between different tabs. Do you know what is causing this @JoanaMMoreira?

As discussed yesterday, this happens across the site when changing pages so it will need to be addressed separately

site/src/layouts/index.ts
JoanaMMoreira: I'll make the change after we change it in Mosaic :)

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,67 @@
...
+    // update tab when the route changes
+    const currentTab = tabs.find(({ name }) => route?.includes(name));
+
+    const defaultTabAndRoute = () => {
+      setActiveTabIndex(0);
+      push(`${newRoute}${tabs[0].name}`);
+    };
+
+    currentTab ? setActiveTabIndex(currentTab.id) : defaultTabAndRoute();
+  }, [route]);
```
ZarkoPernar, line 51: Like Josh mentioned yesterday, this useEffect is also unnecessary. What you are doing here is basically derived state. https://react.dev/learn/you-might-not-need-an-effect You can simply use the router url as state. No need for useState at all.

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  const SecondarySidebar = <TableOfContents />; // TODO: replace with custom component pages sidebar
+
+  return (
+    <DetailBase sidebar={<Sidebar sticky>{SecondarySidebar}</Sidebar>}>
+      <p>{description}</p>
+      <Tabs activeTabIndex={activeTabIndex} onActiveChange={onActiveChange}>
```
ZarkoPernar, line 58: Then it would look something like
```tsx
      <Tabs activeTabIndex={figureOutIndexFromUrl(url)} onActiveChange={(index) => push(figureOutUrlFromIndex(index)}>
```

ZarkoPernar (commented):
Would be good to remove that useEffect but other than that it looks good 👍

site/src/layouts/DetailComponent/DetailComponent.tsx
JoanaMMoreira: Thank you! I've made some tweaks, let me know how that looks

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const { push } = useRouter();
+  const { route } = useRoute();
+
+  const newRoute = route?.substring(0, route.lastIndexOf("/") + 1);
+
+  const useData = useStore((state: CustomSiteState) => state.data);
+
+  const { description } = useData || {};
+
+  const updateRouteWhenTabChanges = useCallback((index: number) => {
```
joshwooding, line 30: useCallback should be removed unless you are sure it is needed.

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const currentTab = tabs.find(({ id }) => index === id);
+
+    currentTab
+      ? push(`${newRoute}${currentTab.name}`)
+      : push(`${newRoute}${tabs[0].name}`);
+  }, []);
+
+  const updateTabWhenRouteChanges = useMemo(() => {
```
joshwooding, line 38: This useMemo is not needed and the variable name should be updated so it doesn't seem like this is a function

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const currentTab = tabs.find(({ name }) => route?.includes(name));
+
+    const defaultTabAndRoute = () => {
```
joshwooding, line 41: Do you know if we need to handle this at all? Either way, it will need to be in a useEffect

site/src/layouts/DetailComponent/DetailComponent.tsx
```diff
@@ -0,0 +1,66 @@
...
+      push(`${newRoute}${tabs[0].name}`);
+      return 0;
+    };
+
+    return currentTab ? currentTab.id : defaultTabAndRoute();
+  }, [route]);
+
+  const SecondarySidebar = <TableOfContents />; // TODO: replace with custom component pages sidebar
+
```
joshwooding, line 50: I would have expected the code to be something like:

```tsx
const currentTab = tabs.find(({ name }) => route?.includes(name));
const currentTabIndex = currentTab?.id ?? 0
```

site/src/layouts/DetailComponent/DetailComponent.tsx
JoanaMMoreira: Yeah we need to handle the scenario where the user clicks or navigates to the Button root page (in which case we want to display the first page "examples" and select the first tab). Do you mean only this specific scenario should be in a useEffect?

site/src/layouts/DetailComponent/DetailComponent.tsx
JoanaMMoreira: This would be fine but we also need to update the url when setting the tab to 0 hence the `defaultTabAndRoute` function

joshwooding: I'm not sure if we need to change the URL. I think it's better to keep the root and the explicit separate

JoanaMMoreira: Okay makes sense, I'll update this to only default the tab to 0 but maintain the url

JoanaMMoreira: @joshwooding as discussed I've moved this redirect into a `useEffect`

joshwooding (approved):

PR #1580, opened 2023-04-25
Input Next validation

libertymayc:
Included in this PR:

Added `validationStatus` to `InputNext`, `FormFieldNext`, `FormFieldContextNext`

Added `StatusAdornment`, `ErrorAdornment`, `SuccessAdornment`, `WarningAdornment` components

Removed `disabled` prop from `FormFieldHelperText`, `FormFieldLabel`

Renamed `useA11yValueValue` to `u11yValueAriaProps`
Removed `disabled` and `readOnly` out of `u11yValueAriaProps` and separated out `disabled`, `readOnly` within `FormFieldContextNext`
Removed `useA11yValue` hook

Removed `type`, `onChange` prop from InputNext

cc @bhoppers2008


libertymayc:
One problem with this currently is read only styles in FormField override the background of an Input with status background
Design POV - error or warning Inputs shouldn't be allowed to be disabled/readonly
I think ideally we would have a :not:has(... input with status children) on the readonly styles... but for now, what's the best way to solve?

cc @alycrys

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -162,8 +168,10 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
         withBaseName(),
         {
           [withBaseName("focused")]: !isDisabled && focused,
-          [withBaseName("disabled")]: isDisabled,
-          [withBaseName("readOnly")]: isReadOnly,
+          [withBaseName("disabled")]: !validationStatus && isDisabled,
+          [withBaseName("readOnly")]: !validationStatus && isReadOnly,
+          [withBaseName("warning")]: validationStatus === "warning",
+          [withBaseName("error")]: validationStatus === "error",
```
alycrys, line 174: Suggested change:
```
[withBaseName(validationStatus)]: validationStatus,
```

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -206,3 +206,38 @@ export const LongLabelsWithWrap: ComponentStory<typeof FormFieldNext> = (
...
+      <InputNext validationStatus="error" defaultValue="Input value" />
+    </FormFieldNext>
+    <FormFieldNext
+      validationStatus="warning"
+      label="Form Field label"
+      helperText="Helper text"
+      {...props}
+    >
+      <InputNext validationStatus="warning" defaultValue="Input value" />
+    </FormFieldNext>
```
libertymayc, line 230: @alycrys problem with current implementation: having to provide prop to both FF and Input
Would it be best to use CSS to add styling to Input if prop is given in FF? This would apply to checkbox etc too, so would be a lot of styling responsibility in FF

libertymayc:
> Is `variant` being disregarded when `validationSatus` is applied?

there will always be background & border in both variants now, so yes

packages/lab/stories/form-field-next/form-field-next.stories.tsx
alycrys: I think adding it to the FF context will keep things consistent...until we get rid of the context (if/when).

libertymayc:
> check label hover when validation applied

fixed

alycrys:
I'm not too sure about having a StatusAdornment component. It will not be used by its own. Not sure what the guidance would be here. @joshwooding any thoughts on this?

packages/lab/src/form-field-next/FormFieldHelperText.css
```diff
@@ -1,4 +1,16 @@
...
 }
+
+.saltFormFieldHelperText .saltText {
+  font-style: var(--saltFormField-helperText-fontStyle, var(--salt-editable-help-fontStyle));
+}
+
+.saltFormFieldHelperText-withValidation {
+  column-gap: var(--salt-size-adornmentGap);
+}
+.saltFormFieldNext:has(.saltInputNext-disabled, .saltInputNext-readOnly) .saltStatusIndicator {
```
libertymayc, line 14: @alycrys any ideas of how else to handle this for firefox?

packages/lab/src/form-field-next/FormFieldNext.tsx
```diff
@@ -91,13 +95,16 @@ export const FormField = forwardRef(
...
       id: idProp,
+      validationStatus: validationStatusProp,
       ...restProps
     }: FormFieldProps,
     ref: ForwardedRef<HTMLDivElement>
   ) => {
     const labelId = useId();
     const helperTextId = useId();
 
+    const validationStatus = readOnly || disabled ? undefined : validationStatusProp;
```
libertymayc, line 106: Changing the prop to an enum of `disabled | readonly | warning | error` would make this cleaner.

libertymayc: cc @bhoppers2008 assuming disabled and readonly would take precedence over the statuses here, if we are to have separate props as is

libertymayc: after conversation with @alycrys, decision stands as using guidance to say that using validation plus a disabled or readonly prop would be wrong. it is possible, but discouraged and doing so would make for broken behaviour which we won't be reponsible for
cc @joshwooding

.changeset/green-icons-jog.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/lab": minor
+---
+
+Added `validationStatus` to InputNext
+Added `StatusAdornment`, `ErrorAdornment` and `WarningAdornment` components
```
libertymayc, line 6: do we want these in a lab/internal folder? @joshwooding

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -206,3 +206,63 @@ export const LongLabelsWithWrap: ComponentStory<typeof FormFieldNext> = (
...
+      </FormFieldNext>
+      Warning disabled Input
+      <FormFieldNext
+        validationStatus="warning" 
+        label="Form Field label"
+        helperText="Helper text"
+        {...props}
+      >
+        <InputNext disabled defaultValue="Input value" />
+      </FormFieldNext> */}
```
libertymayc, line 265: above examples show broken behaviour through bad code

packages/lab/src/form-field-next/FormFieldNext.tsx
alycrys: I don't think we should add conditions like this in the code. I believe these should be part of guidance in the documentation. 
I believe we should never hide the error status.

packages/lab/src/form-field-next/FormFieldHelperText.css
alycrys: As discussed, I believe we can avoid using `:has` selector here

libertymayc: i think you may have reviewed outdated code :)

packages/lab/src/form-field-next/FormFieldNext.tsx
```diff
@@ -120,17 +125,15 @@ export const FormField = forwardRef(
         )}
         {...restProps}
       >
-        <FormFieldContextNext.Provider value={{ a11yProps: a11yValue }}>
+        <FormFieldContextNext.Provider
+          value={{ a11yProps: a11yValue, validationStatus: validationStatus }}
+        >
           {label && (
             <FormFieldLabel id={labelId} disabled={disabled} label={label} />
```
alycrys, line 132: no need to pass `disabled` to label

.changeset/green-icons-jog.md
alycrys: changelog needs to be updated

libertymayc: @alycrys updated

.changeset/green-icons-jog.md
```diff
@@ -0,0 +1,12 @@
+---
+"@salt-ds/lab": minor
+---
+
+Added `validationStatus` to `InputNext` and `FormFieldNext`
+
+Added `StatusAdornment`, `ErrorAdornment` and `WarningAdornment` components
```
alycrys, line 7: + SuccessAdornment

joshwooding:
> I'm not too sure about having a StatusAdornment component. It will not be used by its own. Not sure what the guidance would be here. @joshwooding any thoughts on this?

StatusAdornment will be used by Input, Dropdown, ComboBox, DatePicker, Grid and more so it needed to be consumable by them all. I think maybe in the long term it could be added to StatusIndicator but for now I thought it would be easier to create a quick component

packages/lab/src/form-field-next/FormFieldHelperText.tsx
```diff
@@ -9,24 +10,32 @@ const withBaseName = makePrefixer("saltFormFieldHelperText");
...
-  disabled?: boolean;
 }
 
 export const FormFieldHelperText = ({
   className,
-  disabled,
   helperText,
   ...restProps
 }: FormFieldHelperTextProps) => {
+  const { disabled, readOnly, validationStatus, ...restFormFieldProps } = useFormFieldPropsNext();
```
alycrys, line 20: I don't think you need `readOnly` and `...restFormFieldProps`

.changeset/green-icons-jog.md
joshwooding: > do we want these in a lab/src/internal folder? @joshwooding

For now I'll say no, consumers might want to use them in their custom components

packages/lab/src/form-field-next/FormFieldNext.tsx
joshwooding: It's an interesting point. I feel like I'm between the two decisions with good points on either side.

Even though it's going to be recommended that you do not use validation plus disabled/readonly I'm not sure whether it's worth deviating from the norm and restricting it. You technically could still have an API which was more aligned with HTML but have the disabled/readonly states always override the styles of the validation states

libertymayc: > You technically could still have an API which was more aligned with HTML but have the disabled/readonly states always override the styles of the validation states

could you explain more what this would look like?

alycrys (approved):

PR #1581, opened 2023-04-25
Form Field composition based API

libertymayc:
cc @bhoppers2008 

Explores a composition based API for FormField component
Adds a `FormFieldControlWrapper` component to contain multi input controls

> "This would be a pretty big change and so needs more discussion/confirmation. It has a little more setup for the consumer in the simple use case but has certain advantages e.g.
> 
> Can pass a ref/prop to each component individually
> More flexibility on how to arrange the components
> Could work nicely with our Layout components"
> @el-dav


packages/lab/src/form-field-next/FormFieldControlWrapper.css
```diff
@@ -0,0 +1,7 @@
+.saltFormFieldControlWrapper {
```
el-dav, line 1: I'm unsure about whether this component is necessary since all it is doing is giving a default layout when we have multiple inputs in a FormField, which you could also achieve using a horizontal Stack.

How certain are we that this is the only viable layout for these type of controls? If we're pretty sure then I could see this as being a convenient default. But it could be confusing if consumers will occasionally want a different layout.

libertymayc: good point, could use Layout components if we aren't too strict about it

packages/lab/src/form-field-next/FormFieldHelperText.tsx
```diff
@@ -1,21 +1,19 @@
...
 
 const withBaseName = makePrefixer("saltFormFieldHelperText");
 
 export interface FormFieldHelperTextProps
   extends Omit<TextProps<"label">, "variant" | "styleAs"> {
-  helperText: FormFieldProps["helperText"];
   disabled?: boolean;
 }
 
 export const FormFieldHelperText = ({
```
el-dav, line 13: I think now we need to pass `disabled` down via context from the FormField, otherwise it looks like consumers will need to set it individually on FormField, HelperText and Label to keep it in sync

libertymayc: true ... do we keep the prop on helper text as well, in case the component is used alone?

el-dav: I was thinking these components would only be used in FF is there another use-case?

I think we would only keep the prop if there was a case for having a different `disabled` state on the helpertext/label to the rest of the FormField, and then it could be used to override the context. But I can't imagine why that would be necessary

packages/lab/src/form-field-next/FormFieldControlWrapper.css
bhoppers2008: We talked about this as a solution to a left-aligned left label, where the field with the longest label in a form drives the width of all other field labels. I don't mind how we achieve that though :)

libertymayc: this wrapper is actually the container for multi controls

bhoppers2008: Apparently i misunderstood :) There's no requirement for vertical stacking of multiple controls (haven't ever seen anything other than Check/Radio groups... but that would be a single group component, rather than a stack in a form field). Just horizontal grouping.

packages/lab/src/form-field-next/FormFieldNext.tsx
```diff
@@ -17,15 +15,7 @@ export interface FormFieldProps
    */
   disabled?: boolean;
   /**
-   * The helper text content
-   */
-  helperText?: string;
```
libertymayc, line 22: would I need to deprecate these given how recent they were added? @joshwooding

packages/lab/src/form-field-next/FormFieldControlWrapper.css
libertymayc: Q for devs is: we either use FlexLayout (as Ben has suggested, until InlineLayout is available which is in backlog) or we provide this Control Wrapper component?

packages/lab/src/form-field-next/FormFieldNext.tsx
joshwooding: Removing these is fine

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -18,60 +21,48 @@ export default {
 export const Skeleton: ComponentStory<typeof FormFieldNext> = (props) => {
```
alycrys, line 21: What's the idea behind the Skeleton? Do we need to keep it?

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -18,60 +21,48 @@ export default {
...
+      </FormFieldNext>
+      <FormFieldNext labelPlacement="left" {...props}>
+        <Label>Default Form Field label</Label>
+        <HelperText>Default helper text</HelperText>
+      </FormFieldNext>
     </FlowLayout>
   );
 };
 
 export const Default: ComponentStory<typeof FormFieldNext> = (props) => {
```
alycrys, line 36: I suggest splitting these examples into Default and LabelPlacement.

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -118,21 +106,26 @@ export const WithControls: ComponentStory<typeof FormFieldNext> = (props) => {
 export const MultiInput: ComponentStory<typeof FormFieldNext> = (props) => {
```
alycrys, line 106: Could we merge somehow the WithControls and MultiInput? Seems to me they're quite similar.

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -189,19 +170,20 @@ export const LongLabelsWithWrap: ComponentStory<typeof FormFieldNext> = (
...
+        </Label>
         <InputNext defaultValue="Primary Input value" />
       </FormFieldNext>
-      <FormFieldNext
-        readOnly
-        label="Form Field label"
-        helperText="Helper text that's very long. Additional text explaining that this is a readonly Form Field."
-        {...props}
-      >
+      <FormFieldNext readOnly {...props}>
```
alycrys, line 180: Do we need to add this readonly example here? I'd replace it with a left aligned label, or even move into a LabelPlacement story

packages/lab/stories/form-field-next/form-field-next.stories.tsx
libertymayc: it's to show the 'bare bones' of Form Field without any controls inside, and to show how the height is controlled by the controls so it 'collapses' - I think it may be a good idea to keep to understand that concept of having to provide controls

alycrys (approved):

origami-z:
Would prop like `labelPlacement` be on `Label` component in this kind of set up?

PR #1592, opened 2023-04-26
Input Static adornments

libertymayc:
V1 of FormFieldNext includes **static** InputNext adornments.

### DIFF
--- a/packages/lab/src/input-next/InputNext.tsx
+++ b/packages/lab/src/input-next/InputNext.tsx
@@ -30,7 +30,7 @@ export interface InputProps
    */
   disabled?: HTMLInputElement["disabled"];
   /**
-   * End adornment components
+   * End adornment component
    */
   endAdornment?: ReactNode;
   /**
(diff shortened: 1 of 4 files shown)


packages/lab/src/input-next/InputNext.tsx
```diff
@@ -28,6 +29,10 @@ export interface InputProps
    * If `true`, the component is disabled.
    */
   disabled?: HTMLInputElement["disabled"];
+  /**
+   * End adornment components
```
alycrys, line 33:
```suggestion
   * End adornment component
```

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -36,6 +41,10 @@ export interface InputProps
    * If `true`, the component is read only.
    */
   readOnly?: boolean;
+  /**
+   * Start adornment components
```
alycrys, line 45:
```suggestion
   * Start adornment component
```

origami-z:
Do we call `startAdornment` prop composition as well? (v.s. strictly `children`)

PR #1593, opened 2023-04-27
Add component examples live preview

JoanaMMoreira:


site/docs/components/button/examples.mdx
```diff
@@ -8,12 +8,50 @@ data:
...
-#### Guidance
+</LivePreview>
+
+<LivePreview componentName="button" exampleName="Primary" >
+
+### Primary
+
+Use the primary button for routine, non-urgent actions.
+
+</LivePreview>
```
alycrys, line 25: Do we need to have the heading and description contained in the LivePreview? I believe it would make more sense to keep things separate.

Like we do in the docs stories:
```
## Features

Button contains three variants `cta`, `primary` and `secondary`. If not specified, the default value is `primary`.

<Canvas>
  <Story id="core-button--all" />
</Canvas>
```

joshwooding: I believe the current layout isn't the final layout. The next iteration should only display the title and description when a specific example is shown so it makes sense for them to be together

JoanaMMoreira: That's correct, the next iteration will include a different view with a list where users can select individual examples so it will be good to group the examples + related content

PR #1606, opened 2023-05-02
Add controlled Input example story

libertymayc:
Add controlled Input example story to InputNext

### DIFF
--- a/packages/lab/stories/input-next/input-next.stories.tsx
+++ b/packages/lab/stories/input-next/input-next.stories.tsx
@@ -28,13 +28,13 @@ export const Controlled: ComponentStory<typeof InputNext> = (args) => {
   const [value, setValue] = useState("Value");
 
   return (
-      <InputNext
-        {...args}
-        value={value}
-        onChange={(event) => {
-          setValue(event.target.value);
-        }}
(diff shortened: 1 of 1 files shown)


alycrys (approved):

packages/lab/stories/input-next/input-next.stories.tsx
```diff
@@ -23,6 +24,23 @@ export const Secondary: ComponentStory<typeof InputNext> = (args) => {
...
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const value = event.target.value;
+    setValue(value);
+  };
+
+  return (
+    <InputNext
+      {...args}
+      value={value}
+      onChange={handleChange}
```
libertymayc, line 39: `onChange={(event) => { setValue(event.target.value); }}`
@alycrys this would cause typecheck errors for us due to us using the default onChange method - is this going to be a potential problem for users too? (can be handled with the current implementation)

alycrys: I believe this is the correct approach

alycrys (approved):

PR #1617, opened 2023-05-03
Fixed error in run local server instructions

shinytoyrobots:
The current instructions direct to a local "page not found", as the url should read http://localhost:3000/salt/index


joshwooding (approved):

joshwooding (changes requested):
Will DM

PR #1622, opened 2023-05-04
Remove the site snapshots from the repo

joshwooding:
This PR removes the snapshots from the repo.

I've added the `gen:snapshot` to the deploy script so contributors no longer need to keep updating the snapshots.


JoanaMMoreira (approved):
Perfect, I had this on my to do list 😄

site/.gitignore
```diff
@@ -8,3 +8,4 @@
 # Generated files
 /.tmp/
 /.next/
+/snapshots/
```
JoanaMMoreira, line 11: I guess we can also remove this from `.prettierignore`

joshwooding: It would still make sense to have it in prettierignore file for local dx improvements

PR #1625, opened 2023-05-04
Add ability to pass insertion point to useComponentCssInjection

ImmanuelBaskaran:


packages/styles/src/use-style-injection/InsertionPointProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  return (
+    <insertionPointContext.Provider value={value}>
+      {children}
+    </insertionPointContext.Provider>
+  );
+}
+
+export function useInsertionPoint() {
+  const value = useContext(insertionPointContext);
```
ImmanuelBaskaran, line 31: Return value may be null here, but insertBefore can take null as the second parameter and as a result will just append the child to the end of the node's children

packages/styles/src/use-style-injection/InsertionPointProvider.tsx
```diff
@@ -0,0 +1,33 @@
+import { createContext, ReactNode, useMemo, useContext } from "react";
+
+export interface InsertionPointContextType {
+  insertionPoint: ChildNode | null;
+}
+
+const insertionPointContext = createContext<ChildNode | null>(
+  window.document.head.firstChild
```
el-dav, line 8: is there a potential danger here if `window.document.head.firstChild` changes or is removed?

I think it might be better to have null as the default and in `useStyleInjection` something like:

```
targetWindow.document.head.insertBefore(
      styleMap.styleElement,
      insertionPoint || targetWindow.document.head.firstChild
    );
```

el-dav (approved):

PR #1632, opened 2023-05-09
Add check for targetWindow to UseComponentCssInjection

ImmanuelBaskaran:
This adds a check to  UseComponentCssInjection if the targetWindow is null or if it is equal to the main window.

We do this because of the use of style-inject, which already handles style injection into the main window.


joshwooding:
< We do this because of the use of style-inject, which already handles style injection into the main window.

Why wouldn't we remove style-inject?

ImmanuelBaskaran:
Closing this as #1657 was merged as a result of the decision to remove style-inject

PR #1638, opened 2023-05-10
List v1

Fercas123:
List Next V1 includes: 
- declarative single select list
- _[moved to V2]_ deselectable list

`/list-next` contains 3 components. 
- _[moved to V2]_ `Highlighter` comes from the original list, and it will be useful for whenever we need to highlight text. 
- `ListNext` renders the list component, and uses `useList` to control items
- `ListItemNext` is the item wrapper.

it also contains the `useList`, the hook that will control the list component. 

for reviewers, probably the easiest place to start is `ListItemNext`


packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,284 @@
...
+    return;
+  });
+
+  // Effects
+  useEffect(() => {
+    // sets list in first render
+    list = listRef.current;
+  }, []);
+
+  useEffect(() => {
```
Fercas123, line 243: event listeners and available options state are set in this effect, that updates everytime the list ref changes

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,284 @@
...
+  };
+
+  const handleBlur = useEventCallback(() => {
+    setFocusedIndex(null);
+    if (onBlur) {
+      onBlur();
+    }
+  });
+
+  const handleMouseDown = useEventCallback(() => {
```
Fercas123, line 166: this handler is needed to know if on focus the event comes from a click or the keyboard, if it doesn't come from the keyboard it doesn't add the visual focus ring.

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,284 @@
...
+  const findPreviousOption = (
+    currentOption: Element,
+    moves: number
+  ): Element => {
+    // Return the previous option if it exists; otherwise, returns first option
+    const currentOptionIndex = activeOptions.indexOf(currentOption);
+    return activeOptions[currentOptionIndex - moves] || activeOptions[0];
+  };
+
+  const updateScroll = (currentTarget: Element) => {
```
Fercas123, line 135: because we are using active descendant, we need to update the scroll ourselves

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,71 @@
...
+  cursor: var(--salt-selectable-cursor-disabled);
+}
+.saltListItemNext-disabled .saltText {
+  color: var(--list-item-text-color);
+}
+
+/* disabled list item when selected */
+.saltListItemNext-disabled[aria-selected="true"] {
+  border-block: var(--salt-size-border) var(--salt-container-borderStyle) var(--salt-selectable-borderColor-selectedDisabled);
+  --list-item-background: var(--salt-selectable-background-selectedDisabled);
```
libertymayc, line 63: we usually take Button.css as the standard, so having e.g. --list-item-background and --list-item-background-disabled in the root then setting disabled: var(--list-item-background-disabled) here

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,85 @@
...
+      id,
+      value,
+      onClick,
+      ...props
+    },
+    ref
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-hightligher",
```
libertymayc, line 44:
```suggestion
      testId: "salt-highlighter",
```

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,34 @@
+/* Default variables applied to the root element */
+.saltList {
+  --list-background: var(--salt-container-primary-background);
+  --list-borderStyle: var(--salt-container-borderStyle);
+  --list-borderWidth: var(--salt-size-border);
+  --list-height: calc(var(--salt-size-stackable) * var(--list-displayedItemCount));
+}
+
+/* Style applied to the root element */
+.saltList {
```
libertymayc, line 10: I'd call this ListNext to ensure it doesn't get confused until the renames are done

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,34 @@
...
+  user-select: none;
+  padding-inline-start: 0;
+  box-sizing: content-box;
+}
+
+/* Style applied to the borderless variant */
+.saltList-borderless {
+  --list-borderStyle: none;
+}
+
```
libertymayc, line 29: focused selected styles seem wrong (the offset)

packages/theme/css/characteristics/selectable.css
```diff
@@ -20,13 +20,12 @@
...
   --salt-selectable-foreground-selected: var(--salt-palette-interact-foreground-active);
   --salt-selectable-foreground-selectedDisabled: var(--salt-palette-interact-foreground-activeDisabled);
-
   --salt-selectable-background: var(--salt-palette-interact-background);
   --salt-selectable-background-hover: var(--salt-palette-interact-background-hover);
   --salt-selectable-background-selected: var(--salt-palette-interact-background-active);
   --salt-selectable-background-blurSelected: var(--salt-palette-interact-background-blurSelected);
   --salt-selectable-background-disabled: var(--salt-palette-interact-background-disabled);
-
+  --salt-selectable-background-selectedDisabled: var(--salt-palette-interact-background-selectedDisabled);
```
libertymayc, line 28: needs changeset for the tokens added

packages/theme/css/foundations/fade.css
```diff
@@ -33,6 +33,7 @@
   --salt-color-gray-90-fade-border-readonly: rgba(132, 135, 142, var(--salt-palette-opacity-border-readonly));
   --salt-color-gray-200-fade-border-readonly: rgba(97, 101, 110, var(--salt-palette-opacity-border-readonly));
 
+  --salt-color-blue-30-fade-background: rgba(164, 213, 244, var(--salt-palette-opacity-background));
```
libertymayc, line 36: tokens added need to be added to theme docs

packages/lab/stories/list-next/list.doc.stories.mdx
```diff
@@ -0,0 +1,68 @@
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { ListItemNext, ListNext } from "@salt-ds/lab";
+import {
+  CharacteristicUsage,
+  CSSClassTable,
+} from "css-variable-docgen-components";
+
+<Meta title="Documentation/Lab/List next" component={ListNext} />
```
libertymayc, line 8:
```suggestion
<Meta title="Documentation/Lab/List Next" component={ListNext} />
```

packages/lab/stories/list-next/list.doc.stories.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+<CSSClassTable of={ListNext} />
+
+<CSSClassTable of={ListItemNext} />
+
+## Characteristics used
+
+<CharacteristicUsage of={ListNext} />
+
+<CharacteristicUsage of={ListItemNext} />
```
libertymayc, line 68: perhaps worth a h3 heading for listnext vs listitemnext on the above

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,184 @@
...
+        className={clsx(
+          withBaseName(),
+          {
+            [withBaseName("borderless")]: borderless,
+            [withBaseName("focusable")]: !childrenCount,
+          },
+          className
+        )}
+        role="listbox"
+        tabIndex={disabled ? undefined : 0}
```
libertymayc, line 172: wondering if it should be -1 or undefined on disabled

packages/lab/src/list-next/Highlighter.tsx
```diff
@@ -0,0 +1,50 @@
...
+import highligherCss from "./Highlighter.css";
+
+const withBaseName = makePrefixer("saltHighlighter");
+
+export interface HighlighterProps {
+  matchPattern?: RegExp | string;
+  text?: string;
+}
+
+export const Highlighter = ({
```
libertymayc, line 15: re the PR description; will this be used in other components?

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,71 @@
...
+.saltListItemNext {
+  /* Color */
+  --list-item-text-color: var(--salt-text-primary-foreground);
+  --list-item-height: var(--salt-size-stackable);
+  --list-item-background: none;
+}
+
+/* Style applied to the root element */
+.saltListItemNext {
+  color: var(--list-item-text-color);
```
libertymayc, line 11:
```suggestion
  color: var(--list-item-next-text-color);
```

until renamed

packages/lab/src/list-next/Highlighter.tsx
Fercas123: I don't think there are any others that are using it outside list, but we could move it to a more general place if we get a case for it to be reused :)

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,85 @@
...
+        value={value}
+        onClick={onClick}
+      >
+        <Text as="p" className={withBaseName("textWrapper")}>
+          {itemTextHighlightPattern === null ? (
+            content
+          ) : (
+            <Highlighter
+              matchPattern={itemTextHighlightPattern}
+              text={label || (children as string)}
```
ZarkoPernar, line 78: It might be worth typing this on the props level. Something like:
```tsx
type ListItemNextProps = ListItemNextBaseProps & { children: string; } | { label: string; }
```

ZarkoPernar: I assume the user needs to provide either a label string or children as a string.

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,187 @@
...
+   */
+  borderless?: boolean;
+  /**
+   * If `true`, items in list will be deselectable.
+   */
+  deselectable?: boolean;
+  /**
+   * Optional id prop
+   *
+   * Used as suffix of List id: `list-{id}`
```
ZarkoPernar, line 53: I assume the user would want to provide their own id so they could select it with `getElementById` or some similar use-case. Changing the user provided id seems odd.

ZarkoPernar: Just checked and you are not using it like the comment suggests :D

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+    if (onKeyDown) {
+      onKeyDown(currentItem);
+    }
+    return;
+  });
+
+  // Effects
+  useEffect(() => {
+    // sets list in first render
+    list = listRef.current;
```
ZarkoPernar, line 245: Why do we do this? Might need to add a comment explaining the scenario we are handling here because it is not obvious to me.

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+const withBaseName = makePrefixer("saltListItemNext");
+
+export interface ListItemNextProps extends HTMLAttributes<HTMLLIElement> {
+  itemTextHighlightPattern?: RegExp | string;
+  label?: string;
+  disabled?: boolean;
+  focused?: boolean;
+  selected?: boolean;
+  showCheckbox?: boolean;
```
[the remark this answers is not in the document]
feliciackh: ahh `showCheckbox` will be used only in multiselect, currently not part of V1. Will have it removed now, but it'll be back in next phase. We'll also consider perhaps a different selection type then :)

packages/lab/src/list-next/ListNext.tsx
feliciackh: Updated the docs

packages/lab/src/list-next/ListItemNext.tsx
Fercas123: > I assume the user needs to provide either a label string or children as a string.

yep, it is either children or label, we could also decide to just go with children to be fair, the problem adding it as types is that interfeces can't extend them

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,72 @@
+/* Default variables applied to the root element */
+.saltListItemNext {
+  /* Color */
+  --list-item-next-text-color: var(--salt-text-primary-foreground);
```
alycrys, line 4: i believe these vars should be `--listItemNext-text-color` format

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,72 @@
...
+.saltListItemNext > * {
+  pointer-events: none;
+}
+
+/* Style applied to presentation items (headers and empty message) */
+.saltListItemNext[role="presentation"] {
+  cursor: default;
+}
+
+/* Pseudo-class applied to the root element on focus */
```
alycrys, line 33: this is not a pseudo-class, it's just a class

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,40 @@
+/* Default variables applied to the root element */
+.saltListNext {
+  --list-displayedItemCount: 4;
```
alycrys, line 3: i think the default value should be set in js

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,40 @@
...
+  --list-background: var(--salt-container-primary-background);
+  --list-borderStyle: var(--salt-container-borderStyle);
+  --list-borderWidth: var(--salt-size-border);
+  --list-height: calc(var(--salt-size-stackable) * var(--list-displayedItemCount));
```
alycrys, line 7: we don't need these vars. can use the values directly

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,162 @@
...
+      onBlur,
+      onMouseDown,
+    });
+
+    const forkedRef = useForkRef(ref, listRef);
+
+    const renderContent = () => {
+      return Children.map(children, (listItem, index) => {
+        if (!isValidElement(listItem)) return;
+        const { disabled: propDisabled, ...restListItemProps } = listItem.props;
```
alycrys, line 122:
```suggestion
        const { disabled: propDisabled, ...restListItemProps } = listItem.props as ListNextProps;
```

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,162 @@
...
+    useComponentCssInjection({
+      testId: "salt-list-next",
+      css: listNextCss,
+      window: targetWindow,
+    });
+
+    const childrenCount = Children.count(children);
+    const emptyList = childrenCount === 0;
+
+    const disabled = disabledListProp;
```
alycrys, line 87: is this necessary?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+        evt.preventDefault();
+        toggleSelectItem(nextItem);
+        break;
+      default:
+        break;
+    }
+    if (onKeyDown) {
+      onKeyDown(currentItem);
+    }
+    return;
```
alycrys, line 237: what is this return supposed to do?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+  });
+
+  // Effects
+  useEffect(() => {
```
alycrys, line 241: I suggest using `useIsomorphicLayoutEffect`

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+
+export const useList = ({
+  children,
+  displayedItemCount,
+  onFocus,
+  onKeyDown,
+  onBlur,
+  onMouseDown,
+}: UseListProps) => {
+  const listRef = useRef<HTMLUListElement | null>(null);
```
alycrys, line 55: maybe do this in the actual ListNext file and passing it to the hook?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+  });
+
+  // Effects
+  useEffect(() => {
+    // sets list in first render
+    list = listRef.current;
+  }, []);
+
+  useEffect(() => {
+    const prepare = (list: HTMLUListElement) => {
```
alycrys, line 247: `addListers`?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,289 @@
...
+      list.addEventListener("keydown", handleKeyDown);
+      list.addEventListener("focus", handleFocus);
+      list.addEventListener("blur", handleBlur);
+      list.addEventListener("mousedown", handleMouseDown);
+    };
+
+    const tearDown = (list: HTMLUListElement | null): void => {
```
alycrys, line 254: `removeListeners`?

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,72 @@
...
+  outline: var(--salt-focused-outline);
+  outline-offset: calc(var(--salt-focused-outlineWidth) * -1);
+}
+
+/* active list item when focused */
+.saltListItemNext:hover:not([role="presentation"]):not(.saltListItemNext-disabled),
+.saltListItemNext-focused {
+  --list-item-next-background: var(--list-item-next-background-hover);
+  cursor: var(--salt-selectable-cursor-hover);
+}
```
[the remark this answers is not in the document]
feliciackh: just checked - `.saltListItemNext-focused` applies only to List Items on focus using keyboard nav, `role=presentation` items by default aren't focusable (logic within `useList`) so we could skip adding this check

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,84 @@
...
+import listItemNextCss from "./ListItemNext.css";
+
+const withBaseName = makePrefixer("saltListItemNext");
+
+export interface ListItemNextProps extends HTMLAttributes<HTMLLIElement> {
+  itemTextHighlightPattern?: RegExp | string;
+  label?: string;
+  disabled?: boolean;
+  focused?: boolean;
+  selected?: boolean;
```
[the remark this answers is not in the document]
feliciackh: `focused` is used explicitly to determine list items that are being navigated and focused on using keyboard, we needed to separate `focused/selected` to keep track of and sync both mouse and keyboard events
`selected` is set to true whenever a list item is selected either using mouse `onClick` or keyboard nav on `Enter/Space`

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.saltListNext:empty:before {
+  height: 100%;
+  padding-inline: var(--salt-size-unit);
+  display: flex;
+  position: relative;
+  align-items: center;
+  content: var(--list-emptyMessage);
+}
```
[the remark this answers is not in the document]
feliciackh: seems like this is still needed, updated the name to `content: var(--listNext-emptyMessage);`

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,72 @@
...
+
+/* Pseudo-class applied to the root element on focus */
+.saltListItemNext-focused {
+  outline: var(--salt-focused-outline);
+  outline-offset: calc(var(--salt-focused-outlineWidth) * -1);
+}
+
+/* active list item when focused */
+.saltListItemNext:hover:not([role="presentation"]):not(.saltListItemNext-disabled),
+.saltListItemNext-focused {
```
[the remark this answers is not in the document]
feliciackh: we will need this to ensure styles for list item on focus using keyboard navigation display the same as list item on hover using mouse. Updated to include comment on what each of the classes cover

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,287 @@
...
+  const listRef = useRef<HTMLUListElement | null>(null);
+  let list = listRef.current;
+
+  const [activeDescendant, setActiveDescendant] = useState<string>("");
+  const [focusedIndex, setFocusedIndex] = useState<number | null>(null);
+  const [selectedIndexes, setSelectedIndexes] = useState<number[]>(
+    getSelected(children)
+  );
+  const [allOptions, setAllOptions] = useState<Element[]>([]);
+  const [activeOptions, setActiveOptions] = useState<Element[]>([]);
```
libertymayc, line 64:
```suggestion
  const activeOptions = allOptions.filter((option) => option.getAttribute("aria-disabled") !== "true");
```

ZarkoPernar:
@Fercas123 @feliciackh the only change from the call that I would recommend is to refactor the useEffect not to set the event listeners, and instead return them so they can be applied to the element. However, since only List is using useList this can also be changed later.

ZarkoPernar (approved):

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,376 @@
...
+    //   );
+    //   cy.findByRole("tooltip").should("exist");
+    // });
+  });
+
+  describe("list height", () => {
+    it("should render list with 3 list items", () => {
+      cy.mount(SingleSelectList({}));
+
+      cy.findByRole("listbox").should("have.css", "height", "108px"); // 36px * 3
```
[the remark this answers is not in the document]
feliciackh: @ZarkoPernar looks like you were reviewing an older version of test, here's the latest one: https://github.com/jpmorganchase/salt-ds/pull/1638/files#diff-02b552591bf60a02c32e17c1e8a6bfcb46ae70a984b29e2ae132f3a41256139dR28-R35

`toFixed()` was added recently bc tests were coming back with `107.98217239px` instead of `108px`

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,84 @@
...
+        onClick={onClick}
+      >
+        <Text as="p" className={withBaseName("textWrapper")}>
+          {itemTextHighlightPattern === null ? (
+            content
+          ) : (
+            <Highlighter
+              matchPattern={itemTextHighlightPattern}
+              text={content}
+            />
```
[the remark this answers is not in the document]
feliciackh: not at the moment - we'll add one in for list v2

packages/lab/src/list-next/useList.ts
feliciackh: adding this code block here so it's easier for future reference

currently we've been doing `setActiveOptions` in the useEffect: 
```
setActiveOptions(
        Array.from(list.children)
          .filter((child) => child.getAttribute("role") === "option")
          .filter((option) => option.getAttribute("aria-disabled") !== "true")
);
```
`useEffect` will be refactored, will review the recommended approach when we work on the feedback on Monday (fyi @libertymayc)

libertymayc: commented this one state is derived from the other, so I dont think it needs to be in state too

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,84 @@
...
+      className: classNameProp,
+      disabled,
+      focused,
+      itemTextHighlightPattern,
+      label,
+      role = "option",
+      selected,
+      id,
+      value,
+      onClick,
```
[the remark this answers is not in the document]
alycrys: that's right, but we're not changing it, so we can just leave it to be passed within `...props`

packages/lab/src/list-next/ListNext.css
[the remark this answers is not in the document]
alycrys: As per the discussions we had, we're not implementing the empty list functionality anymore. So I don't think this is needed

.changeset/silent-dogs-brush.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/lab": minor
+"@salt-ds/theme": minor
+---
+
+add List V1 component to lab
```
alycrys, line 6: I'd add a few more details here

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+  describe("List height", () => {
+    it("should render list with 3 list items", () => {
+      cy.mount(<SingleSelectList />);
+
+      cy.findByRole("listbox").then(($el) => {
+        expect($el.height()?.toFixed(0)).to.equal("108"); // 36px * 3
+      });
+    });
+
+    it("should render list with 2 list items, with displayedItemCount lower than childrenCount", () => {
```
alycrys, line 37: i believe the list will actually render all 3 items, but only 2 will be visible and there will be a scrollbar. I think the description should be amended

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+      });
+    });
+
+    it("should render empty list", () => {
+      cy.mount(<SingleSelectList children={<>Empty list</>} />);
+
+      cy.findByRole("listbox").then(($el) => {
+        expect($el.height()?.toFixed(0)).to.equal("36"); // 36px
+      });
+    });
```
alycrys, line 59: this will need to be changed. When there are no items there should be height = 0

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+    cy.findByRole("option", { name: ITEMS[1].label })
+      .click()
+      .should("have.attr", "aria-selected", "true");
+    // click to select the same list item
+    cy.findByRole("option", { name: ITEMS[1].label })
+      .click()
+      .should("have.attr", "aria-selected", "true");
+  });
+
+  it("should deselect previous list item when a new list one is selected, for non deselectable", () => {
```
alycrys, line 90: no need to mention `deselectable`

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+type ItemType = { label: string; value: string };
+
+const ITEMS: ItemType[] = [
+  { label: "list item 1", value: "item 1" },
+  { label: "list item 2", value: "item 2" },
+  { label: "list item 3", value: "item 3" },
+];
+
+// Uncontrolled single select list
+describe("An uncontrolled single select list", () => {
```
alycrys, line 12: you should follow the GIVEN, WHEN, THEN, SHOULD format. see tests in core

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+    cy.findByRole("option", { name: ITEMS[1].label }).should(
+      "not.have.attr",
+      "aria-selected",
+      "true"
+    );
+  });
+
+  // Disabled list
+  describe("A disabled single select list", () => {
+    it("should not allow any items to be selected", () => {
```
alycrys, line 109: "all items should be disabled" maybe?

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+          }
+        />
+      );
+
+      cy.findByRole("option", { name: ITEMS[1].label }).should(
+        "have.attr",
+        "aria-disabled",
+        "true"
+      );
+    });
```
alycrys, line 139: I think you should also test with click and check 'aria-selected'

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+                {ITEMS[2].label}
+              </ListItemNext>
+            </>
+          }
+        />
+      );
+
+      cy.findByRole("option", { name: ITEMS[1].label })
+        .realHover()
+        .should("not.have.class", "saltListItemNext-focused");
```
alycrys, line 162: i'm a little confused. Do we expect that while hovering the item, it'll have the focused class applied?

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+  describe("Keyboard navigation and interactions", () => {
+    it("should select list item on Space or Enter key", () => {
+      cy.mount(<SingleSelectList />);
+
+      cy.findByRole("listbox").focus();
+      cy.realPress("ArrowDown");
+      cy.realPress("Space"); // select second list item
+      cy.findByRole("option", { name: ITEMS[1].label }).should(
+        "have.class",
+        "saltListItemNext-focused"
```
alycrys, line 176: in your description you're testing for selection, but you're testing for focused on the first item and that the second item is not selected. I think you need to stick with the description

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+      cy.findByRole("listbox").focus();
+      cy.findByRole("option", { name: ITEMS[0].label }).should(
+        "have.attr",
+        "aria-selected",
+        "true"
+      );
+      cy.findByRole("option", { name: ITEMS[0].label }).should(
+        "have.class",
+        "saltListItemNext-focused"
+      );
```
alycrys, line 196: like here 👍

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,317 @@
...
+      cy.findByRole("listbox").blur(); // remove focus
+      cy.findByRole("listbox").focus(); // focus again
+      // focus should be back on the previously focused item
+      cy.findByRole("option", { name: ITEMS[1].label }).should(
+        "have.class",
+        "saltListItemNext-focused"
+      );
+    });
+
+    it("should re-focus on previously selected item when an item was previously selected, unless focus has moved", () => {
```
alycrys, line 217: i don't think we need this test. we only check for the focus index, so selected has no impact

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,170 @@
...
+      handleKeyDown(evt);
+      onKeyDown && onKeyDown(activeElement);
+    };
+
+    const blurHandler = (evt: FocusEvent) => {
+      handleBlur();
+      onBlur && onBlur(evt);
+    };
+
+    const mousedownHandler = (evt: KeyboardEvent) => {
```
libertymayc, line 141:
```suggestion
    const mouseDownHandler = (evt: KeyboardEvent) => {
```
not major, but think it's good to align the casing, and same for keyDownHandler

packages/theme/stories/palettes/interact.stories.mdx
```diff
@@ -26,6 +26,8 @@ Colors representing actions such as editing or selecting an item.
           "var(--salt-palette-interact-background)",
         "--salt-palette-interact-background-disabled":
           "var(--salt-palette-interact-background-disabled)",
+        "--salt-palette-interact-background-activeDisabled":
+          "var(--salt-palette-interact-background-activeDisabled",
```
libertymayc, line 30:
```suggestion
          "var(--salt-palette-interact-background-activeDisabled)",
```

packages/theme/css/characteristics/selectable.css
```diff
@@ -20,13 +20,12 @@
...
   --salt-selectable-foreground-selected: var(--salt-palette-interact-foreground-active);
   --salt-selectable-foreground-selectedDisabled: var(--salt-palette-interact-foreground-activeDisabled);
-
   --salt-selectable-background: var(--salt-palette-interact-background);
   --salt-selectable-background-hover: var(--salt-palette-interact-background-hover);
   --salt-selectable-background-selected: var(--salt-palette-interact-background-active);
   --salt-selectable-background-blurSelected: var(--salt-palette-interact-background-blurSelected);
   --salt-selectable-background-disabled: var(--salt-palette-interact-background-disabled);
-
+  --salt-selectable-background-selectedDisabled: var(--salt-palette-interact-background-activeDisabled);
```
libertymayc, line 28: needs documenting in selectable.stories.mdx

.changeset/silent-dogs-brush.md
libertymayc, line 6: added CSS variables

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,30 @@
+/* Style applied to the root element */
+.saltListNext {
+  margin: 0;
+  background: var(--salt-container-primary-background);
+  border-color: var(--salt-container-primary-borderColor);
+  border-style: var(--salt-container-borderStyle);
+  border-width: var(--salt-size-border);
+  height: calc(var(--salt-size-stackable) * var(--listNext-displayedItemCount));
```
libertymayc, line 8: what happens if --listNext-displayedItemCount is undefined?

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,30 @@
...
+  display: flex;
+  position: relative;
+  align-items: center;
+  content: var(--listNext-emptyMessage);
+}
+
+/* Pseudo-class applied to the root element on focus */
+.saltListNext-focusable:focus-visible {
+  outline: var(--salt-focused-outline);
+  outline-offset: calc(0 - var(--salt-focused-outlineWidth));
```
libertymayc, line 29:
```suggestion
  outline-offset: calc(-1 * var(--salt-focused-outlineWidth));
```
easier to understand

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,235 @@
...
+    []
+  );
+
+export const useList = ({ children, displayedItemCount }: UseListProps) => {
+  const listRef = useRef<HTMLUListElement | null>(null);
+  let list = listRef.current;
+
+  const [activeDescendant, setActiveDescendant] = useState<string>("");
+  const [focusedIndex, setFocusedIndex] = useState<number | null>(null);
+  const [selectedIndexes, setSelectedIndexes] = useState<number[]>(
```
libertymayc, line 44:
```suggestion
  const [selectedIndices, setSelectedIndices] = useState<number[]>(
```

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+              "saltListItemNext-focused"
+            );
+          });
+        });
+
+        describe("WHEN on Home key press", () => {
+          it("THEN focus on first list item", () => {
+            cy.mount(<SingleSelectList />);
+
+            cy.findByRole("listbox").focus();
```
alycrys, line 298: maybe select another item first and then press Home. This test might still pass if there is no handler on Home. Focusing the list will focus the first item anyway, right?

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+            cy.realPress("End");
+            cy.findByRole("option", { name: ITEMS[2].label }).should(
+              "have.class",
+              "saltListItemNext-focused"
+            );
+          });
+        });
+
+        describe("WHEN an item was previously interacted with mouse", () => {
```
alycrys, line 320: I don't see any mouse interaction in this test

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+                (3 * itemHeight).toString()
+              ); // 36px * 3
+            });
+          });
+        });
+      });
+
+      describe("WHEN disabled", () => {
+        describe("WHEN entire list is disabled", () => {
+          it("THEN does not allow any items to be selected", () => {
```
alycrys, line 116:
```suggestion
          it("THEN all items are disabled", () => {
```

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+                }
+              />
+            );
+
+            cy.findByRole("option", { name: ITEMS[1].label }).should(
+              "have.attr",
+              "aria-disabled",
+              "true"
+            );
+          });
```
alycrys, line 152:
```suggestion
        describe("WHEN list item is disabled", () => {
          it("THEN it's not selectable", () => {
            cy.mount(
              <SingleSelectList
                children={
                  <>
                    <ListItemNext value={ITEMS[0].value}>
                      {ITEMS[0].label}
                    </ListItemNext>
                    <ListItemNext value={ITEMS[1].value} disabled>
                      {ITEMS[1].label}
                    </ListItemNext>
                    <ListItemNext value={ITEMS[2].value}>
                      {ITEMS[2].label}
                    </ListItemNext>
                  </>
                }
              />
            );
           cy.findByRole("option", { name: ITEMS[1].label })
               .click()
            cy.findByRole("option", { name: ITEMS[1].label }).should(
              "not.have.attr",
              "aria-selected"
            );
          });
```

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+                    </ListItemNext>
+                  </>
+                }
+              />
+            );
+
+            cy.findByRole("option", { name: ITEMS[1].label })
+              .realHover()
+              .should("not.have.class", "saltListItemNext-focused");
+          });
```
alycrys, line 176: Hover should never Focus. I don't think this test is necessary

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+        it("THEN select list item on Space or Enter key", () => {
+          cy.mount(<SingleSelectList />);
+
+          cy.findByRole("listbox").focus();
+          cy.realPress("ArrowDown");
+          cy.realPress("Space"); // select second list item
+          cy.findByRole("option", { name: ITEMS[1].label }).should(
+            "have.class",
+            "saltListItemNext-focused"
+          );
```
alycrys, line 190: Should this item also be selected? That's what the description says.
Would the second item be focused after ArrowDown?

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,346 @@
...
+              "true"
+            );
+            cy.findByRole("option", { name: ITEMS[0].label }).should(
+              "have.class",
+              "saltListItemNext-focused"
+            );
+          });
+        });
+
+        describe("WHEN an item was previously focused", () => {
```
alycrys, line 214: I'd split this into a "WHEN list re-focus" section a level up maybe

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/* Class applied to list item on focus using keyboard navigation only */
+.saltListItemNext-focused {
+  outline: var(--salt-focused-outline);
+  outline-offset: calc(var(--salt-focused-outlineWidth) * -1);
+}
+
+/* Active list item on hover using mouse */
+.saltListItemNext:hover:not([role="presentation"]):not(
```
alycrys, line 33: do we still have role=presentation in the code? We only cared for it when there used to be an empty message. Now that we removed it, I don't think this is necessary

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,170 @@
...
+    const listStyles = {
+      ...style,
+      "--listNext-displayedItemCount": displayedItemCount,
+    };
+
+    const activeElement = document.getElementById(activeDescendant);
+
+    const focusHandler = () => {
+      handleFocus();
+      onFocus && onFocus(activeElement);
```
alycrys, line 128:
```suggestion
      onFocus?.(activeElement);
```

alycrys: same for the others ;)

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  const activeOptions = allOptions.filter(
+    (option) => option.getAttribute("aria-disabled") !== "true"
+  );
+
+  const [mouseDown, setMouseDown] = useState<boolean>(false);
+
+  const getListItemIndex = (item: Element): number => {
+    const optionIndex = allOptions.indexOf(item);
+    return optionIndex !== -1 ? optionIndex : -1;
```
alycrys, line 57: this doesn't really make sense

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,235 @@
...
+  const toggleSelectItem = (element: Element) => {
+    const elementIndex = getListItemIndex(element);
+    setSelectedIndexes([elementIndex]);
+  };
+
+  const justFocusItem = (element: Element) => {
+    setActiveDescendant(element.id);
+    setFocusedIndex(getListItemIndex(element));
+    updateScroll(element);
+  };
```
alycrys, line 92:
```suggestion
  const focusAndSelect = (element: Element) => {
    selectItem(element)
    focusItem(element)
  };

  const focusFirstItem = () => {
    // Find first active item
    const firstItem = activeOptions[0];
    if (firstItem) {
      focusAndSelect(firstItem);
    }
  };

  const focusLastItem = () => {
    // Find last active item
    const lastItem = activeOptions[activeOptions.length - 1];
    if (lastItem) {
      focusAndSelect(lastItem);
    }
  };

  const selectItem = (element: Element) => {
    const index = allOptions.indexOf(element);
    setSelectedIndexes([index]);
  };

  const focusItem = (element: Element) => {
    const index = allOptions.indexOf(element);
    
    setActiveDescendant(element.id);
    setFocusedIndex(index);
    updateScroll(element);
  };
```

alycrys: just an idea ;)

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,235 @@
...
+  // Handlers
+  const handleClick = ({ currentTarget }: MouseEvent<HTMLUListElement>) => {
+    const nonClickableTarget = activeOptions.indexOf(currentTarget) === -1;
+    if (nonClickableTarget) {
+      return;
+    }
+    toggleSelectItem(currentTarget);
+    setActiveDescendant(currentTarget.id);
+    setFocusedIndex(null);
+    updateScroll(currentTarget);
```
alycrys, line 137:
```suggestion
focusAndSelect(currentTarget)
```

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
feliciackh: Ah good spot - that was one of the tests written in earlier stages, no longer applicable now will remove

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,297 @@
...
+type ItemType = { label: string; value: string };
+
+const ITEMS: ItemType[] = [
+  { label: "list item 1", value: "item 1" },
+  { label: "list item 2", value: "item 2" },
+  { label: "list item 3", value: "item 3" },
+];
+
+describe("GIVEN a list", () => {
+  describe("GIVEN an uncontrolled single select list", () => {
```
alycrys, line 12: if there are no tests for controlled list, then no need to mention uncontrolled

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,297 @@
...
+        "aria-selected",
+        "true"
+      );
+    });
+
+    describe("GIVEN displayedItemCount", () => {
+      const itemHeight = 36;
+
+      describe("WHEN undefined", () => {
+        it("THEN list height should be 3 times itemHeight", () => {
```
alycrys, line 77:
```suggestion
        it("THEN list height should be 3 times item height", () => {
```

alycrys: same for the others

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
```diff
@@ -0,0 +1,297 @@
...
+          cy.findByRole("listbox").then(($el) => {
+            expect($el.height()?.toFixed(0)).to.equal(
+              (3 * itemHeight).toString()
+            );
+          });
+        });
+      });
+
+      describe("WHEN lower than childrenCount", () => {
+        it("THEN list height should be displayedItemCount times itemHeight", () => {
```
alycrys, line 89:
```suggestion
      describe("WHEN displayedItemCount=2", () => {
        it("THEN list height should be 2 times item height", () => {
```

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,61 @@
+/* Default variables applied to the root element */
+.saltListItemNext {
+  /* Color */
```
alycrys, line 3: don't think we need this comment

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,61 @@
...
+  --listItemNext-text-color: var(--salt-text-primary-foreground);
+  --listItemNext-background: none;
+}
+
+/* Style applied to the root element */
```
alycrys, line 8: no need to comment twice on the same class, only one will be applied anyway

alycrys:
couple more small comments. other than that it looks great

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
feliciackh: yeah that makes sense - focusing the list will focus first item yes. Updated the test now

libertymayc (dismissed):

alycrys (dismissed):

origami-z (commented):
I mentioned this to @Fercas123 this morning, a bit of summary below. 

From API level, I'm not convinced by only supporting list item being `children` as a v1 offering. If having a choice to add support later, I'd rather vote for only supporting `source` (like current `List`) not `children` as a start, which should be easier to write and still support use cases we identify so far.

Purely from composition angle, presentational components (e.g. layout component, banner, contact detail) can have their children elements literally being *anything* passed in and stay untouched. But List needs to control children (e.g. selection, highlight), so we need things like `React.Children`/`React.cloneElement` (in this PR and relevant pitfall mentioned in [React doc](https://react.dev/reference/react/Children) ) or other approaches like old descendent approach.

From consumers perspective, any non-hard coded items will need to `map` once in their own code to become children items, then `map` again by `List` code, which doesn't feel ideal. I also can't work out how to effectively send back full list item information in event handlers (i.e. current List's 2nd argument in `onSelectionChange`). Is it reading some required `string` and ask users to map back to original object they have?

For work coming up next, how are we going to deal with `Dropdown` / `Combobox`? If continuation of `children` is needed, what to do with button / input v.s. list? What about virtualized list with `children`?

Some may argue about supporting item groups looks nicer via declarative API, however whether that creates more complex internal code for us? e.g. determining group or not. On top of this, what about custom components? If we're going to restrict what can be put inside a component, then being `children` v.s. old `ListItem` way of doing things makes no difference?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,227 @@
...
+  Children,
+  isValidElement,
+  MouseEvent,
+  KeyboardEvent,
+  ReactNode,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
+import {
```
joshwooding, line 11: This seems a bit redundant and could be replaced with strings directly.

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,59 @@
...
+  display: flex;
+  gap: var(--salt-size-unit);
+  position: relative;
+  align-items: center;
+  white-space: nowrap;
+  transition: background var(--salt-duration-perceptible) ease-in-out;
+}
+
+.saltListItemNext > * {
+  pointer-events: none;
```
joshwooding, line 21: I'm going to guess, but this is probably not what you would want. What was the use-case here?

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,59 @@
...
+}
+
+.saltListItemNext[aria-selected="true"] {
+  border-block: var(--salt-size-border) var(--salt-container-borderStyle) var(--salt-selectable-borderColor-selected);
+  --listItemNext-background: var(--salt-selectable-background-selected);
+  --saltIcon-color: var(--salt-selectable-foreground-selected);
+}
+
+/* Style applied to disabled items */
+.saltListItemNext-disabled {
```
joshwooding, line 47: I'm wondering if we just use ["aria-disabled"] here

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,75 @@
...
+    ref
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-item-next",
+      css: listItemNextCss,
+      window: targetWindow,
+    });
+
+    const className = clsx(
```
joshwooding, line 48: We should move this to be inline so we're consistent across the codebase. It also means we don't need to rename className to classNameProp

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,75 @@
...
+      withBaseName(),
+      {
+        [withBaseName("disabled")]: disabled,
+        [withBaseName("focused")]: focused,
+      },
+      classNameProp
+    );
+
+    const content = label || (children as string);
```
joshwooding, line 57: Why is this casting needed?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,169 @@
...
+   *The id for the list component
+   */
+  id?: string;
+  /**
+   * The number of items displayed in the visible area.
+   * Note that this determines the max height of the list.
+   */
+  displayedItemCount?: number;
+  // ListNextControlProps
+  onBlur?: FocusEventHandler;
```
joshwooding, line 47: These shouldn't need to be declared as they should already be provided by: HTMLAttributes<HTMLUListElement>

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-next",
+      css: listNextCss,
+      window: targetWindow,
+    });
+
+    const childrenCount = Children.count(children);
+
+    const listId = useId(idProp) || "listNext"; // TODO: check why useId needs to return undefined
```
joshwooding, line 81: It needs to return undefined in the initial render due to the polyfill and ssr support.

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,169 @@
...
+      handleBlur,
+      handleMouseDown,
+    } = useList({
+      children,
+      displayedItemCount,
+    });
+
+    const forkedRef = useForkRef(ref, listRef);
+
+    const renderContent = () => {
```
joshwooding, line 102: I don't think this should be in the function and move inline

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,169 @@
...
+}
+
+export const ListNext = forwardRef<HTMLUListElement, ListNextProps>(
+  function ListNext(
+    {
+      children,
+      className,
+      disabled: listDisabled,
+      displayedItemCount: displayedItemCountProp = 4,
+      ListItem = DefaultListItem,
```
joshwooding, line 60: I don't think this is used?

joshwooding (changes requested):

packages/lab/src/list-next/ListItemNext.tsx
Fercas123: it was wrapped in text, but no longer needed. should we keep label as a prop?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,257 @@
...
+    const currentItem =
+      (activeDescendant && document.getElementById(activeDescendant)) ||
+      activeOptions[0];
+    let nextItem = currentItem;
+    if (!currentItem) {
+      return;
+    }
+    setShowFocusRing(true);
+    switch (key) {
+      // TODO: since we are getting rid of displayItemCount. should we do this with height?
```
Fercas123, line 195: @joshwooding thoughts on this one?

packages/lab/src/list-next/ListNextContext.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext } from "@salt-ds/core";
+import { SyntheticEvent, useContext } from "react";
+
+export interface ListItemNextContextValue {
+  disabled: boolean;
```
libertymayc, line 5: I assume this should be optional
```suggestion
  disabled?: boolean;
```

packages/lab/src/list-next/ListNextContext.tsx
```diff
@@ -0,0 +1,18 @@
...
+  id?: string;
+  select: (event: SyntheticEvent<HTMLLIElement>) => void;
+  isSelected: (id: string) => boolean;
+  isFocused: (id: string) => boolean;
+}
+
+export const ListItemNextContext = createContext<
```
libertymayc, line 12: should the file name be ListItemNextContext ?

libertymayc: or I feel like rather, this should be ListemNextContext, given it's used in ListNext?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,124 @@
...
+      handleKeyDown,
+      handleBlur,
+      handleMouseDown,
+      activeDescendant,
+      contextValue,
+    } = useList({
+      disabled: listDisabled,
+      selected: selectedProp,
+      defaultSelected,
+      // onChange,
```
libertymayc, line 79: what's the plan for onChange here?

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -0,0 +1,53 @@
+/* Default variables applied to the root element */
+.saltListItemNext {
+  --listItemNext-text-color: var(--salt-text-primary-foreground);
+  --listItemNext-background: none;
+}
+
+.saltListItemNext {
+  color: var(--listItemNext-text-color);
+  background: var(--listItemNext-background);
+  height: var(--salt-size-stackable, auto);
```
libertymayc, line 10:
```suggestion
  height: calc(var(--salt-size-base) + var(--salt-spacing-100));
```
this spacing addition might supposed to be padding, would check with designer

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-item-next",
+      css: listItemNextCss,
+      window: targetWindow,
+    });
+
+    const listContext = useListItems();
+    const id = useId(idProp) || value;
+    const itemId = `list-${listContext?.id || "listNext"}--list-item--${id}`;
```
libertymayc, line 47: think we could have something other than "listNext" here (maybe another default id? the one above?) because otherwise when it changes to list from list next, it would be `list-list--...`

alycrys: I agree. I also believe we don't need to add any other words if there is already an id provided. So something like `${listContext?.id || "listNext"}-${id}` makes more sense to me

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,82 @@
...
+    ref
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-item-next",
+      css: listItemNextCss,
+      window: targetWindow,
+    });
+
+    const listContext = useListItems();
```
alycrys, line 45: I normally prefer to destructure values here. Makes it easier for me to see what it being used

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,120 @@
...
+      children,
+      className,
+      disabled: listDisabled,
+      id,
+      onSelect,
+      onFocus,
+      onBlur,
+      onKeyDown,
+      onMouseDown,
+      selected: selectedProp,
```
alycrys, line 48: same here

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,120 @@
...
+      contextValue,
+    } = useList({
+      disabled: listDisabled,
+      selected: selectedProp,
+      defaultSelected,
+      id: listId,
+      ref: listRef,
+    });
+
+    const focusHandler = (event: FocusEvent<HTMLUListElement>) => {
```
alycrys, line 80: We normally use `handleFocus`. It appears you've switched the words. Was it intentional?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  // takes care of keydown when using keyboard navigation
+  const handleKeyDown = (event: KeyboardEvent<HTMLUListElement>) => {
+    const { key } = event;
+    const currentItem = getActiveItem();
+    let nextItem = currentItem;
+    if (!currentItem) {
+      return;
+    }
+    setShowFocusRing(true);
```
alycrys, line 198: shouldn't this happen only when focus happened?

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,250 @@
...
+  const selectItem = useCallback(
+    (element: HTMLElement) => {
+      const newValue = element?.dataset.value;
+      if (newValue) {
+        setSelectedItem(newValue);
+        updateActiveDescendant(element);
+      }
+    },
+    [setSelectedItem, updateActiveDescendant]
+  );
```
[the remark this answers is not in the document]
Fercas123: it is, otherwise they would make the dependencies of the context callback (select) change on every render.

packages/lab/src/list-next/ListNext.tsx
Fercas123: because handleFocus has already been declared in the useListHook.

alycrys: Maybe rename the other one as `handleFocusProp`? Would be more inline with the current pattern

packages/lab/src/list-next/ListItemNext.tsx
Fercas123: given `id` and `disabled` will be declared below and would have to be renamed, and we need to check that the context is not coming with the default undefined, readability wise it ends up being a bit of the same imo, but happy to destructure it.

packages/lab/src/list-next/ListItemNext.css
Fercas123: changed and i will double check with Ivan

packages/lab/src/list-next/useList.ts
Fercas123: nice catch! yeah, then i can default to false, rely on focus and forget about this ⭐ ta

.changeset/silent-dogs-brush.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/lab": minor
+"@salt-ds/theme": minor
+---
+
+Add single select List Next to lab
```
joshwooding, line 6: Could we add a rich codeblock that shows the default usage

packages/lab/src/__tests__/__e2e__/list-next/ListNext.cy.tsx
joshwooding, line 1: Rather than checking for the focused class I would check the active descendant on the list element

packages/lab/src/common-hooks/keyUtils.ts
```diff
@@ -14,13 +14,15 @@ export const ArrowUp = "ArrowUp";
 export const ArrowDown = "ArrowDown";
 export const ArrowLeft = "ArrowLeft";
 export const ArrowRight = "ArrowRight";
+export const Delete = "Delete";
```
joshwooding, line 17: I think I mentioned this before but can re remove these changes?

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,91 @@
...
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-item-next",
+      css: listItemNextCss,
+      window: targetWindow,
+    });
+    const id = useId(idProp) || value;
+
+    const listContext = useListItems();
```
joshwooding, line 46:
```suggestion
    const listContext = useListItem();
```
I would prefer this was singular since it's in one list item

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,91 @@
...
+    },
+    ref
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-list-item-next",
+      css: listItemNextCss,
+      window: targetWindow,
+    });
+    const id = useId(idProp) || value;
```
joshwooding, line 44: I'm not sure about this. If you had two lists there would be a clash of ids if the values were the same

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const {
+      id: contextId,
+      disabled: contextDisabled,
+      select,
+      isSelected,
+      isFocused,
+    } = listContext;
+
+    const itemId = `${contextId || "listNext"}--${id}`;
+    const content = label || children;
```
joshwooding, line 58: The usage of label or children seems confusing, is there a reason to support both?

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,16 @@
...
+.saltListNext {
+  margin: 0;
+  background: var(--salt-container-primary-background);
+  border-color: var(--salt-container-primary-borderColor);
+  border-style: var(--salt-container-borderStyle);
+  border-width: var(--salt-size-border);
+  height: 100%;
+  outline: none;
+  overflow-y: auto;
+  overflow-x: hidden;
```
joshwooding, line 11: Do we need to prevent x overflow here?

packages/lab/src/list-next/ListNext.css
```diff
@@ -0,0 +1,16 @@
...
+  position: relative;
+  user-select: none;
```
joshwooding, line 13: Why are we preventing text selection?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * If true, all items in list will be disabled.
+   */
+  disabled?: boolean;
+  /**
+   * Optional id prop
+   *
+   *The id for the list component
+   */
+  id?: string;
```
joshwooding, line 29: Do we need to redeclare this when it's provided by HTMLAttributes?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -0,0 +1,120 @@
...
+    };
+
+    const handleMouseDown = (event: MouseEvent<HTMLUListElement>) => {
+      mouseDownHandler();
+      onMouseDown?.(event);
+    };
+
+    return (
+      <ListNextContext.Provider value={contextValue}>
+        <ul
```
joshwooding, line 102: I believe the ul needs aria-disabled set to true when the list is disabled

packages/lab/src/list-next/ListItemNext.tsx
Fercas123: legacy copy paste. can i get rid of label? if so and it is a standard, probably one for the dev practices file?

packages/lab/src/list-next/ListNext.css
Fercas123: I think so, if we have overflow auto in one side (and really, it is the intent for that to happen) it shouldn't scroll on the other side

packages/lab/src/list-next/useList.ts
```diff
@@ -0,0 +1,248 @@
...
+      isFocused,
+    }),
+    [disabled, id, select, isSelected, isFocused]
+  );
+
+  return {
+    focusHandler,
+    keyDownHandler,
+    blurHandler,
+    mouseDownHandler,
```
ZarkoPernar, line 244: Any reason we don't want to name these onKeyDown, onBlur... etc?

packages/lab/src/list-next/ListNext.tsx
Fercas123: @ZarkoPernar

packages/lab/src/list-next/useList.ts
joshwooding: We should probably discuss a standard, it could be something like:
on* for props or parameters and handle* for the functions

.changeset/silent-dogs-brush.md
```diff
@@ -0,0 +1,15 @@
+---
+"@salt-ds/lab": minor
+"@salt-ds/theme": minor
+---
+
+Add new list components: ListNext, ListItemNext
+
+```
+<ListNext>
+  <ListItemNext value={Alaska}>
```
joshwooding, line 10:
```suggestion
  <ListItemNext value="Alaska">
```

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -0,0 +1,88 @@
...
+      id: contextId,
+      disabled: contextDisabled,
+      select,
+      isSelected,
+      isFocused,
+    } = listContext;
+
+    const itemId = `${contextId || "listNext"}--${id}`;
+    const disabled = disabledProp || contextDisabled;
+    const selected = selectedProp || isSelected(value);
```
alycrys, line 57: i have a feeling this might prone to errors. but we'll deal with it if it happens

alycrys (approved):

PR #1645, opened 2023-05-10
Nav Item

joshwooding:


james-nash:
Nice work!

I like the API approach you've taken!

packages/lab/src/nav-item/NavItem.css
```diff
@@ -0,0 +1,81 @@
...
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+  position: relative;
+  background: var(--salt-navigable-primary-background);
+  gap: var(--salt-spacing-75);
+}
+
+.saltNavItem-horizontal {
+  min-height: calc(var(--salt-size-stackable) + var(--salt-size-accent) + var(--salt-spacing-100));
```
libertymayc, line 11:
```suggestion
  min-height: calc(var(--salt-size-stackable) + var(--salt-size-bar) + var(--salt-spacing-100));
```

libertymayc: new token update + stackable will be deprecated, i think it should be size-base and then padding above and below of spacing-50, or whatever this is calculated to be

joshwooding: Not sure if it's worth tackling this atm, versus when we have design guidance

libertymayc: np

libertymayc (approved):

packages/lab/stories/nav-item/nav-item.stories.tsx
```diff
@@ -0,0 +1,279 @@
...
+import { Card, H2, Link } from "@salt-ds/core";
+
+import "./nav-item.stories.css";
+
+export default {
+  title: "Lab/Nav Item",
+  component: NavItem,
+};
+
+const Template: Story<NavItemProps> = (args) => <NavItem {...args} />;
```
alycrys, line 13: I think Template is not necessary here

alycrys (approved):

PR #1660, opened 2023-05-12
Add examples list view

JoanaMMoreira:
- Add new `ExamplesListView` component (that displays either a list on larger screens or a dropdown on smaller screens) to allow users to select from a list of all available examples
- Update `LivePreview` component so it accepts an optional list prop
- Preview at https://saltdesignsystem-git-add-examples-list-view-joshwooding.vercel.app/salt/components/button/examples


site/src/components/components/LivePreview.module.css
```diff
@@ -1,13 +1,16 @@
 .container {
   border: 1px var(--salt-container-borderStyle) var(--salt-container-primary-borderColor);
```
libertymayc, line 2:
```suggestion
  border: var(--salt-size-border) var(--salt-container-borderStyle) var(--salt-container-primary-borderColor);
```

libertymayc (approved):

feliciackh (approved):

joshwooding:
This will need to be checked by Stella before we roll these changes out fully. I'm seeing some interesting focus order issues

JoanaMMoreira:
> This will need to be checked by Stella before we roll these changes out fully. I'm seeing some interesting focus order issues

Yeah it would be good to do an ADA review. Something to keep in mind is that most of these components are in labs so I'm not sure it's worth fixing these issues now or once we start working on them

PR #1667, opened 2023-05-15
Text alignment prop on InputNext

libertymayc:
Added `textAlignment` prop to `InputNext` with possible configurations: "left" (default), "right", "center"


packages/lab/src/input-next/InputNext.tsx
```diff
@@ -173,6 +178,8 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
           [withBaseName("focused")]: !isDisabled && focused,
           [withBaseName("disabled")]: isDisabled,
           [withBaseName("readOnly")]: isReadOnly,
+          [withBaseName(`textAlign${capitalize(textAlignment)}`)]:
```
joshwooding, line 181: I think this is better done via a different method.

Instead of relying on classes, you can use the CSS variable in styles e.g. https://github.com/jpmorganchase/salt-ds/blob/bf0a790fb8121fb50ee592af965ac0312e814249/packages/core/src/flex-layout/FlexLayout.tsx#L93-L99

joshwooding (changes requested):

packages/lab/stories/input-next/input-next.stories.tsx
```diff
@@ -91,13 +89,18 @@ export const ReadonlyEmptyMarker: ComponentStory<typeof InputNext> = (args) => {
...
-      <div className="alignCenter">
-        <InputNext defaultValue={args.defaultValue ?? "Value"} {...args} />
-      </div>
-      <div className="alignRight">
-        <InputNext defaultValue={args.defaultValue ?? "Value"} {...args} />
-      </div>
+    <FlowLayout style={{ maxWidth: "266px" }}>
+      <InputNext defaultValue={args.defaultValue ?? "Value"} {...args} />
+      <InputNext
+        textAlignment="center"
```
alycrys, line 95: rename here also

alycrys (approved):

joshwooding (approved):

PR #1666, opened 2023-05-15
Fix InputNext CSS bugs

libertymayc:
Fixes InputNext CSS 
- activation width
- cursor
- background (see secondary disabled example)
- uses grid-template-areas

PR #1671, opened 2023-05-16
Tweak FormField stories for documentation

libertymayc:
Stories improved/updated in alignment with what we're exposing in V1 and documentation of usage guidance

### DIFF
--- a/.changeset/popular-mayflies-study.md
+++ b/.changeset/popular-mayflies-study.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": minor
+---
+
+Added `textAlign` prop to `InputNext` with possible configurations: "left" (default), "right", "center"
--- a/packages/lab/src/input-next/InputNext.css
+++ b/packages/lab/src/input-next/InputNext.css
@@ -117,7 +117,7 @@
   min-width: 0;
   overflow: hidden;
   padding: 0;
-  text-align: var(--saltInputNext-textAlign, left);
+  text-align: var(--inputNext-textAlign);
   width: 100%;
 }
 
--- a/packages/lab/src/input-next/InputNext.tsx
+++ b/packages/lab/src/input-next/InputNext.tsx
@@ -51,6 +51,10 @@ export interface InputProps
    */
   startAdornment?: ReactNode;
   /**
+   * Alignment of text within container. Defaults to "left"
+   */
+  textAlign?: "left" | "center" | "right";
+  /**
    * Validation status.
    */
   validationStatus?: "error" | "warning" | "success";
@@ -102,6 +106,7 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
     role,
     startAdornment,
     style,
+    textAlign = "left",
     value: valueProp,
     // If we leave both value and defaultValue undefined, we will get a React warning on first edit
     // (uncontrolled to controlled warning) from the React input
@@ -165,6 +170,11 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
     setFocused(true);
   };
 
+  const inputStyle = {
+    "--inputNext-textAlign": textAlign,
+    ...style,
+  };
+
   return (
     <div
       className={clsx(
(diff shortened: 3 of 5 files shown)


packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -50,154 +50,166 @@ export const Disabled: ComponentStory<typeof FormFieldNext> = (props) => {
...
+        <FormLabel>
           Form Field label that's extra long. Showing that labels wrap around to
           the next line.
-        </Label>
+        </FormLabel>
         <InputNext defaultValue="Primary Input value" />
       </FormFieldNext>
       <FormFieldNext {...props}>
-        <Label>Default Form Field label</Label>
+        <FormLabel>Form Field label</FormLabel>
```
alycrys, line 114: this example seems to be more of a long helper text example. is it necessary?

alycrys: I'd combine Label and Long Label. what do you think?

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -50,154 +50,166 @@ export const Disabled: ComponentStory<typeof FormFieldNext> = (props) => {
...
-        <Label>Disabled Form Field</Label>
+        <FormLabel>Disabled Form Field</FormLabel>
         <InputNext defaultValue="Primary Input value" />
-        <HelperText>This field has been disabled</HelperText>
+        <FormHelperText>This field has been disabled</FormHelperText>
       </FormFieldNext>
       <FormFieldNext disabled {...props}>
-        <Label>Disabled Form Field</Label>
+        <FormLabel>Disabled Form Field</FormLabel>
         <InputNext variant="secondary" defaultValue="Secondary Input value" />
```
alycrys, line 59: I'd say we should stick to only primary version for input. It gets confusing having so many variations of input in FF stories.

packages/lab/stories/form-field-next/form-field-next.stories.tsx
bhoppers2008: I think this makes sense. We're not actually using this example on site, since we don't want to recommend long labels. It'll just be an example in storybook so we know how it behaves if a consumer does do it.

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -50,154 +50,166 @@ export const Disabled: ComponentStory<typeof FormFieldNext> = (props) => {
...
+//         </Controls>
+//         <FormHelperText>
+//           *User must enter all values in the string to complete the input
+//         </FormHelperText>
+//       </FormFieldNext>
+//     </FlowLayout>
+//   );
+// };
 
 export const Readonly: ComponentStory<typeof FormFieldNext> = (props) => {
```
bhoppers2008, line 153: @libertymayc not sure if it's just storybook but i can make out a faint background in the secondary field when it should be transparent... can you see the same?

libertymayc: fixed in #1666

packages/lab/stories/form-field-next/form-field-next.stories.tsx
```diff
@@ -38,166 +38,160 @@ export const Default: ComponentStory<typeof FormFieldNext> = (props) => {
...
   return (
     <FlowLayout style={{ width: "366px" }}>
       <FormFieldNext readOnly {...props}>
-        <Label>Readonly Form Field</Label>
+        <FormLabel>Readonly Form Field</FormLabel>
         <InputNext defaultValue="Primary Input value" />
-        <HelperText>This Form Field is readonly</HelperText>
+        <FormHelperText>This Form Field is readonly</FormHelperText>
       </FormFieldNext>
       <FormFieldNext readOnly {...props}>
```
alycrys, line 156: don't think this second example is necessary

alycrys (approved):

alycrys:
needs changeset

PR #1677, opened 2023-05-17
Support Tabstrip  `activeTabIndex` to be null

### LINKED ISSUE #1676: Tabstrip throws error when passing `null` to `activeTabIndex`
### Package name(s)

Lab v1.0.0-alpha.4 (@salt-ds/lab)

### Description

I want to find a way to not set any active tab, which i understand is not a standard use case. However, when attempting to pass `null` to `Tabstrip` component (following similar convention to `List` for example), there is error being thrown.

### Steps to reproduce

```
<Tabstrip activeTabIndex={null}>
```

Stackblitz reproduce: https://stackblitz.com/edit/salt-template-feca2y?file=App.tsx

### Expected behavior

The code should not throw any error.

Even better, showing no tab being active state.

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

origami-z:
Fixes #1676

### DIFF
--- a/.changeset/brown-sloths-change.md
+++ b/.changeset/brown-sloths-change.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": minor
+---
+
+Fix Tabstrip crashes when setting `activeTabIndex` to null
--- a/.changeset/brown-sloths-change.md
+++ b/.changeset/brown-sloths-change.md
@@ -2,4 +2,4 @@
 "@salt-ds/lab": minor
 ---
 
-Fix Tabstrip crashes when setting `activeTabIndex` to null
+Fix Tabstrip crashing when `activeTabIndex` is set to null
--- a/packages/lab/stories/tabstrip/tabstrip.stories.tsx
+++ b/packages/lab/stories/tabstrip/tabstrip.stories.tsx
@@ -1,5 +1,5 @@
 import { useCallback, useMemo, useRef, useState } from "react";
-import { Link, SaltProvider, Text } from "@salt-ds/core";
+import { Button, FlexLayout, Link, SaltProvider, Text } from "@salt-ds/core";
 import {
   EditableLabel,
   Tab,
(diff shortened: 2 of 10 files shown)


packages/lab/stories/tabstrip/tabstrip.stories.tsx
```diff
@@ -1039,3 +1039,23 @@ export const DraggableTabsWithOverflow = () => {
     </div>
   );
 };
+
+export const NullTabIndex = () => {
```
origami-z, line 1043: Use this story to test

origami-z:
Tested by MCP team, to address clicking active tab not firing 'fake' inactive tab before having the PR change

joshwooding (commented):
I'm not too sure about this... I understand that some teams might want to do this but we've previously spoken about this being a hard requirement of tabs

origami-z:
> I'm not too sure about this... I understand that some teams might want to do this but we've previously spoken about this being a hard requirement of tabs

We have to do something for MCP right now, given no other option will be available.

Either we officially support in TS, or something like #1683

.changeset/brown-sloths-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": minor
+---
+
+Fix Tabstrip crashes when setting `activeTabIndex` to null
```
joshwooding, line 5:
```suggestion
Fix Tabstrip crashing when `activeTabIndex` is set to null
```

ZarkoPernar:
> I'm not too sure about this... I understand that some teams might want to do this but we've previously spoken about this being a hard requirement of tabs

Why a hard requirement? It will obviously be used with no selection in some scenarios.

joshwooding:
> > I'm not too sure about this... I understand that some teams might want to do this but we've previously spoken about this being a hard requirement of tabs
> 
> Why a hard requirement? It will obviously be used with no selection in some scenarios.

Technically we don't support those scenarios. A requirement for Tabs is that it always has an active item. I think it's worth merging this anyway and relying on documentation to convey this.

origami-z:
> Why a hard requirement? It will obviously be used with no selection in some scenarios.

When Homepage (being not a part of tab) is active, none of the view tabs should be active. If they override css pure visually, then `onActiveChange` won't fire when clicking back.

joshwooding (approved):

joshwooding (approved):

PR #1679, opened 2023-05-17
Update grid stories import to be relative

origami-z:
Import using package name breaks modular typecheck locally...


alycrys (approved):

joshwooding (changes requested):
All of the other files use the package import. Is this issue only in this file or do you see it everywhere?

origami-z:
> All of the other files use the package import. Is this issue only in this file or do you see it everywhere?

Only this file has problem locally.

All grid stories reference Grid components using relative imports, referencing core/lab using package.

PR #1684, opened 2023-05-18
Add TabstripNext and TabNext

ZarkoPernar:
## Usage
```tsx
<TabstripNext defaultSelected="Home">
  <TabNext value="home">Home</TabNext>
  <TabNext value="transactions">Transactions</TabNext>
  <TabNext value="loans">Loans</TabNext>
</TabstripNext>
```


ZarkoPernar:
@origami-z that is a lot of bugs with the old tab active indicator that I am reusing. I have seen a lot of bugs as well. I will remove it for now with a static bottom border. The animation alone is just not worth hassle.

ZarkoPernar:
> > activeIndex = null
> > controlled and uncontrolled
> > overflow menu
> > truncation with maxWith on tab
> > centered tabs
> 
> Some observation trying out the storybook preview
> 
> 1. Select overflowed item, does not make it visible?
>    
>    * "Loans" being activated instead of "Liquidity"
>    * image

Yes, that is the current expected behavior of uncontrolled tabstrip. I want us to not support reordering of the non-controlled version because I cannot see how I can reliably implement that without some really weird useMemo code. I saw Steve had all kinds of checks, just in order to support that feature.

> 2. Overflow position is the end or next to tabs?
>    
>    * image
>    * Would be good to have a story with a lot more tabs to be captured in Chromatic

Yeah, I have tried the css order hack that the previous tabstrip used in order to position the button "inline" with the tabs and it was very buggy with the resize observer approach I have taken. Might be possible to fix in future versions but I think it is good enough tradeoff for code simplicity.

ZarkoPernar:
> 5. Using keyboard on overflow button, space keypress to open (no highlighted item) then space again, focus move back to previous item. Check w/ design what's expected behaviour around keyboard access?

Good catch. I think the most obvious behaviour is to close the menu and return focus to the overflow button. Current behaviour is almost as we are selecting something, even tho nothing changes. So it seems buggy.

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,150 @@
...
+  justify-content: center;
+  background: var(--saltTabs-tab-background, var(--tabs-tab-background));
+  border: none;
+  border-radius: 0;
+  color: var(--salt-text-primary-foreground);
+  cursor: var(--saltTabs-tab-cursor, var(--tabs-tab-cursor));
+  display: var(--tabs-tabstrip-display);
+  height: var(--saltTabs-tab-height, var(--tabs-tabstrip-height));
+  letter-spacing: var(--saltTabs-tab-letterSpacing, var(--tabs-tab-letterSpacing, 0));
+  min-width: var(--saltTabs-tab-minWidth, 40px);
```
libertymayc, line 26: should this 40px come from spacing?

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,150 @@
+/* Class applied to root Tab element */
+.saltTabNext {
+  --saltEditableLabel-padding: 0;
+  --saltEditableLabel-height: var(--tabs-tabstrip-height);
+  --saltInput-minWidth: 4em;
+
+  --tabs-tab-background: var(--salt-navigable-primary-background);
+  --tabs-tab-background-hover: rgba(0, 0, 0, 0.08);
```
libertymayc, line 8: whats this color?

packages/lab/src/tabs-next/TabstripNext.css
```diff
@@ -0,0 +1,138 @@
+/* Component class applied to the root element */
+.saltTabstripNext {
+  --tabs-tabstrip-height: var(--saltTabs-tabstrip-height, var(--salt-size-stackable));
+  --tabs-tabstrip-dragging-display: none;
+  --tabs-tabstrip-display: inline-flex;
+  --tabs-tabstrip-background: transparent;
+
+  --saltTabs-activationIndicator-height: 1px;
+  --tabs-activationIndicator-background: var(--salt-container-primary-borderColor);
+  --tabs-activationIndicator-thumb-height: 2px;
```
libertymayc, line 10:
```suggestion
  --tabs-activationIndicator-thumb-height: var(--salt-editable-borderWidth-active);
```

libertymayc: not sure if this makes sense but i assume it is and the other 1px above can be salt-size-border

packages/lab/src/tabs-next/TabstripNext.css
```diff
@@ -0,0 +1,138 @@
...
+  --tabs-tab-spacing: var(--salt-spacing-100);
+
+  align-items: flex-start;
+  align-self: var(--saltTabs-tabstrip-alignSelf, stretch);
+  font-size: var(--salt-text-fontSize);
+  font-weight: var(--salt-text-fontWeight);
+  position: relative;
+  overflow: hidden;
+  display: flex;
+  min-width: 28px;
```
libertymayc, line 20: token?

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,208 @@
...
+    };
+  }, [tabs.length]);
+
+  return (
+    <div
+      role="tablist"
+      className={clsx(
+        withBaseName(),
+        withBaseName("horizontal"),
+        [withBaseName(`variant-${variant}`)],
```
libertymayc, line 122:
```suggestion
        withBaseName(variant),
```

packages/lab/stories/tabs-next/tabs-next.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import { TabPanelNext, TabsNext } from "@salt-ds/lab";
+import { Text } from "@salt-ds/core";
+
+export default {
+  title: "Lab/TabsNext/TabsNext",
```
libertymayc, line 5:
```suggestion
  title: "Lab/Tabs Next/Tabs Next",
```

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
```diff
@@ -0,0 +1,234 @@
+import { useState } from "react";
+import { Story } from "@storybook/react";
+import { Button, FlexLayout } from "@salt-ds/core";
+import { TabstripNextProps, TabstripNext, TabNext } from "@salt-ds/lab";
+import { AddIcon } from "@salt-ds/icons";
+import "./tabstrip-next.stories.css";
+
+export default {
+  title: "Lab/TabsNext/TabstripNext",
```
libertymayc, line 9:
```suggestion
  title: "Lab/Tabs Next/Tabstrip Next",
```

packages/lab/src/tabs-next/TabNext.css
ZarkoPernar: This is the new opacity that doesn't exist as a characteristic yet (0.08). I suggested we hard-code this as it is a potential one off, and also we cannot use opacities in rgba() constructor anyways.

packages/lab/src/tabs-next/TabstripNext.css
```diff
@@ -0,0 +1,138 @@
+/* Component class applied to the root element */
+.saltTabstripNext {
+  --tabs-tabstrip-height: var(--saltTabs-tabstrip-height, var(--salt-size-stackable));
```
libertymayc, line 3: size-stackable is now size-base + spacing-100 where spacing-100 addition is coming from the padding top and bottom. would be good to use the new tokens as stackable is being deprecated eventually - not sure if new sizes are in tabs spec yet though so can be done later

ZarkoPernar: I will check with Nav

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,293 @@
...
+    });
+    it("THEN overflow menu can be closed with ArrowLeft and focus is returned to last visible tab", () => {
+      cy.mount(<SimpleTabstrip width={120} />);
+      cy.findByRole("combobox").click();
+      cy.findByRole("listbox").should("exist");
+      cy.focused().realPress("ArrowLeft");
+      cy.findByRole("listbox").should("not.exist");
+      cy.focused().should("have.attr", "role", "tab");
+    });
+    it("THEN overflow menu item can be selected with Enter and focus is returned to the overflow button", () => {
```
origami-z, line 285: Is this the intended behaviour? Should active one receive focus?

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,159 @@
...
+  }: TabProps,
+  ref: ForwardedRef<HTMLDivElement>
+): ReactElement<TabProps> {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-tab-next",
+    css: tabCss,
+    window: targetWindow,
+  });
+  if (index === undefined || onClick === undefined || onKeyDown === undefined) {
```
feliciackh, line 73:
```suggestion
  if (index === undefined || onClick === undefined || onKeyDown === undefined || onKeyUp === undefined ) {
```

error msg seems to indicate `onKeyUp` is required

origami-z (commented):
Feels `id`,`label` and `index` needs some alignment on when to use.

Don't think selecting from overflow should go back to the button.

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,293 @@
...
+        cy.findByRole("tablist").should("have.class", "saltTabstripNext");
+        cy.get(".saltTabstripNext-inner")
+          .findAllByRole("tab")
+          .should("have.length", 5)
+          .eq(4)
+          .should("be.visible");
+      });
+      it("THEN no overflow indicator will be present", () => {
+        cy.mount(<SimpleTabstrip width={400} />);
+        cy.findByRole("tablist").findByRole("button").should("not.exist");
```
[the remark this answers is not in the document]
ZarkoPernar: Good spot

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
ZarkoPernar: I agree, will change.

packages/lab/src/tabs-next/OverflowMenu.tsx
```diff
@@ -0,0 +1,220 @@
...
+    (context) => context.itemVisibility
+  );
+
+  const tabList = tabs
+    .map((tab) => {
+      const label = tab.props.label;
+
+      return {
+        label,
+        id: getTabId(label),
```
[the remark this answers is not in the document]
ZarkoPernar: Yeah, I am currently refactoring the use of ids, so the user can have multiple identical labels, and pass ids. This needs to be documented. Otherwise Overflow will break.

packages/lab/src/tabs-next/OverflowMenu.tsx
```diff
@@ -0,0 +1,220 @@
...
+  const { isOverflowing } = useOverflowMenu<HTMLButtonElement>();
+
+  if (!isOverflowing) {
+    return null;
+  }
+
+  return <OverflowMenuImpl {...props} />;
+}
+
+function OverflowMenuImpl({
```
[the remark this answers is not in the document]
ZarkoPernar: I wanted a tight control over the popup behavior. Wasn't sure if Dropdown was the best level of abstraction for this use-case. It should definitely use ListNext in next iteration.

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,173 @@
...
+                    onKeyUp: noop,
+                    onKeyDown: (e) => {
+                      const focusableElements = Array.from(
+                        outerRef.current?.querySelectorAll<HTMLDivElement>(
+                          `[data-overflow-item]:not([data-overflowing]) [role="tab"], [data-overflow-menu] button`
+                        ) ?? []
+                      );
+                      const focusableIndex =
+                        focusableElements.findIndex((tab) => {
+                          return label === tab.dataset.label;
```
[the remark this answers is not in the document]
ZarkoPernar: Dataset gets automatically created as an object from all `data-*` attributes. I will refactor this to use ids.

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,173 @@
...
+    label: string;
+    id: string;
+  }) => {
+    if (item) {
+      const index = tabs.findIndex((tab) => {
+        const label =
+          typeof tab.props.children === "string"
+            ? tab.props.children
+            : tab.props.label;
+        return label === item.label;
```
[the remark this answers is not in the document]
ZarkoPernar: Another one for the TODO in the id refactor. 😅

packages/lab/src/tabs-next/OverflowMenu.tsx
```diff
@@ -0,0 +1,216 @@
...
+  const { ref, overflowCount } = useOverflowMenu<HTMLDivElement>();
+  const itemVisibility = useOverflowContext(
+    (context) => context.itemVisibility
+  );
+
+  const tabList = tabs.filter(({ id }) => !itemVisibility[id]);
+
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-tabstrip-list",
```
libertymayc, line 64: should this be salt-overflow-menu to follow convention?

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,159 @@
...
+} from "react";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { TabProps } from "./TabsNextTypes";
+
+import tabCss from "./TabNext.css";
+
+const withBaseName = makePrefixer("saltTabNext");
+
+//TODO not ideal - duplicating the Icon then hiding one in css based on density - is there a nicer way ?
```
libertymayc, line 23: would a density provider make sense here?

ZarkoPernar: How does density provider help here? Not sure I understand 🤔

libertymayc: show or hide based on density?

libertymayc:
close tab tabstrip hover colors seem wrong on hover over button vs tab

origami-z:
I'm seeing Tab's activation indicator being much thicker than expected..?

origami-z:
On keyboard interaction with overflow menu, should there be a focus ring around either menu button or list item when menu is open?

packages/lab/src/tabs-next/TabNext.tsx
ZarkoPernar: Ok, so do a conditional render in javascript? I don't see how that is a better solution than a css conditional display.

ZarkoPernar:
> On keyboard interaction with overflow menu, should there be a focus ring around either menu button or list item when menu is open?

Just checked with nav, the current behaviour is correct.

origami-z:
> > On keyboard interaction with overflow menu, should there be a focus ring around either menu button or list item when menu is open?
> 
> Just checked with nav, the current behaviour is correct.

This means it's a design change from existing Tabs overflow as well as dropdown focus ring behaviour?

ZarkoPernar:
> close tab tabstrip hover colors seem wrong on hover over button vs tab

@libertymayc wrong in what way?

libertymayc:
overflow not seeming to be present correctly on story - will send you screenshot

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+    describe("WHEN a activeTabIndex is provided", () => {
+      it("THEN the activeTabIndex is selected", () => {
+        cy.mount(<DefaultTabs activeTabIndex={1} width={400} />);
+        cy.findAllByRole("tab")
+          .eq(1)
+          .should("have.attr", "aria-selected", "true");
+        cy.findByText("Content for Transactions tab");
+      });
+    });
+    describe("WHEN a onActiveChange is provided", () => {
```
alycrys, line 41:
```suggestion
    describe("WHEN an onActiveChange is provided", () => {
```

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+          .eq(1)
+          .should("have.attr", "aria-selected", "true");
+      });
+    });
+    describe("WHEN a defaultActiveTabIndex is null", () => {
+      it("THEN no tab is selected", () => {
+        cy.mount(<DefaultTabstrip defaultActiveTabIndex={null} width={400} />);
+        cy.findAllByRole("tab")
+          .eq(0)
+          .should("have.attr", "aria-selected", "false");
```
alycrys, line 87: either the description is wrong or the test. I'd go with the test 😄

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+describe("Navigation, Given a Tabstrip", () => {
+  describe("WHEN initial size is sufficient to display all contents", () => {
+    describe("WHEN the tabstrip is first rendered", () => {
+      describe("WHEN the tabstrip receives keyboard focus", () => {
+        it("THEN focus will be transfered to the first tab", () => {
+          cy.mount(
+            <StackLayout>
+              <button data-testid="tabstop-1" />
+              <SimpleTabstrip width={400} />
+              <button data-testid="tabstop-2" />
```
alycrys, line 215: are the additional buttons necessary?

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
ZarkoPernar: Because I am only checking the first tab? I should probably loop through all tabs and check none are selected. I naively checked only the first since that is what usually gets auto-selected 😅

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
alycrys: i think we can do without the buttons in some of these tests. In this one you're only testing tabbing in, so there's no need for them. We don't do this in the other components, we Tab straight in. I don't mind the technique, I'm only mentioning this so we can have consistency through all our tests

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
alycrys, line 1: I think it would make sense to split this into TabsNext.cy.tsx and TabstripNext.cy.tsx

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+    describe("WHEN the selected Tab is in the overflow menu", () => {
+      it("THEN the active tab will be moved from the overflow menu to the end of visible tabs", () => {
+        cy.mount(<SimpleTabstrip width={220} />);
+        cy.get(".saltTabstripNext-inner [role='tab']:first-child").should(
+          "have.ariaSelected"
+        );
+
+        cy.findByRole("combobox").click();
+        cy.findByRole("listbox");
+        cy.focused().realPress("ArrowDown").realPress("Enter");
```
alycrys, line 183:
```suggestion
        cy.findByRole("combobox").click();
        cy.realPress("ArrowDown").realPress("Enter");
```

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+    describe("WHEN the selected Tab is moved to visible", () => {
+      describe("WHEN the component is resized", () => {
+        it("THEN the active tab will be moved from the overflow menu to the end of visible tabs", () => {
+          cy.mount(<SimpleTabstrip width={320} />);
+          cy.get(".saltTabstripNext-inner [role='tab']:first-child").should(
+            "have.ariaSelected"
+          );
+          cy.findByRole("combobox").click();
+          cy.findByRole("listbox");
+          cy.focused().realPress("ArrowDown").realPress("Enter");
```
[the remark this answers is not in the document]
ZarkoPernar: Same here

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,417 @@
...
+            cy.realPress("ArrowRight");
+            cy.wait(50);
+            cy.realPress("Tab");
+            cy.wait(50);
+            cy.realPress(["Shift", "Tab"]);
+            cy.wait(50);
+            cy.findAllByRole("tab")
+              .eq(0)
+              .should("be.focused")
+              .should("have.ariaSelected");
```
alycrys, line 312: how do we know that this was the selected tab? the test only shows that this is. Not totally sure, but I think the second tab should be selected after the ArrowRight (or maybe I'm missing something)

packages/lab/src/tabs-next/OverflowMenu.tsx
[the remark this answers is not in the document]
alycrys: add a TODO?

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,178 @@
...
+  return isValidElement(child) && child.type === TabNext;
+}
+
+export type TabstripNextProps = PropsWithChildren<{
+  /* Value for the uncontrolled version. Set to null in order to have no tabs be selected. */
+  activeTabIndex?: number | null;
+  /* Callback for the controlled version. */
+  onActiveChange?: (index?: number) => void;
+  /* Initial value for the uncontrolled version. Set to null in order to have no tabs be selected by default. */
+  defaultActiveTabIndex?: number;
```
alycrys, line 34: i think we'd also need `null` here

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,120 @@
+/* Class applied to root Tab element */
+.saltTabNext {
+  --tabs-tab-background: var(--salt-navigable-primary-background);
+  --tabs-tab-background-hover: rgba(0, 0, 0);
+  --tabs-tab-backgroundOpacity-hover: 0.08;
```
alycrys, line 5: are we waiting for tokens from design?

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,120 @@
...
+.saltTabNext {
+  flex-shrink: 0;
+  align-items: center;
+  align-self: stretch;
+  justify-content: center;
+  background: var(--saltTabs-tab-background, var(--tabs-tab-background));
+  border: none;
+  border-radius: 0;
+  color: var(--salt-text-primary-foreground);
+  cursor: var(--saltTabs-tab-cursor, var(--tabs-tab-cursor));
```
alycrys, line 23: I would imagine `color` would be more customisable than `cursor` :)

packages/lab/src/tabs-next/TabNext.css
libertymayc: the whole opacity scale is awaiting restructure, although not sure when this change will be made cc @pseys

packages/lab/src/tabs-next/TabPanelNext.tsx
```diff
@@ -0,0 +1,38 @@
...
+    window: targetWindow,
+  });
+
+  return (
+    <div
+      {...htmlAttributes}
+      className={clsx(withBaseName(), className)}
+      ref={forwardedRef}
+      role="tabpanel"
+      tabIndex={0}
```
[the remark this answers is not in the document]
ZarkoPernar: It should be focusable unless there is a focusable element within it. So I need to document this and allow the user to manually control the tabIndex https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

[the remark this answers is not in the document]
ZarkoPernar: Specifically https://www.w3.org/WAI/ARIA/apg/patterns/tabs/#keyboardinteraction

packages/lab/src/tabs-next/OverflowMenu.tsx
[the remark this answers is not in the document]
ZarkoPernar: I did on line 131

packages/lab/src/tabs-next/TabNext.css
ZarkoPernar: I added an override as part of the css api

ZarkoPernar: We can continue this convo on teams

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
```diff
@@ -0,0 +1,68 @@
...
+        cy.mount(<DefaultTabs defaultActiveTabIndex={1} width={400} />);
+        cy.findAllByRole("tab")
+          .eq(1)
+          .should("have.attr", "aria-selected", "true");
+        cy.findByText("Content for Transactions tab");
+      });
+    });
+    describe("WHEN defaultActiveTabIndex is null", () => {
+      it("THEN first tab is selected", () => {
+        // @ts-expect-error
```
alycrys, line 28: i believe there is no need for this anymore

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,87 @@
...
+    css: tabCss,
+    window: targetWindow,
+  });
+  if (index === undefined || onClick === undefined || onKeyDown === undefined) {
+    throw Error(
+      "index, onClick, onKeyDown are required props, they would normally be injected by Tabstrip, are you creating a Tab outside of a Tabstrip"
+    );
+  }
+  const root = useRef<HTMLDivElement>(null);
+  const setForkRef = useForkRef(ref, root);
```
alycrys, line 49: is there a reason for this?

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,87 @@
...
+      e.preventDefault();
+      onClick(e, index);
+    },
+    [index, onClick]
+  );
+
+  return (
+    <div
+      {...props}
+      className={withBaseName()}
```
alycrys, line 61: I think we also need to pass `className`?

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,177 @@
...
+    [uniqueId]
+  );
+  const getTabId = getTabIdProp || _getTabId;
+  const tabs = Children.toArray(children)
+    .filter(isTab)
+    .map((tab) => {
+      const { label } = tab.props;
+      const id = tab.props.id || getTabId(label);
+      return { tab, id };
+    });
```
alycrys, line 74: there's a lot of hard work to create these tab `id`s (also in Tabs). could it be possible to create the tab id in Tab?

packages/lab/stories/tabs-next/tabs-next.doc.stories.mdx
```diff
@@ -0,0 +1,154 @@
...
+import { ArgsTable, Meta, Story, Canvas } from "@storybook/addon-docs";
+import { Link, SaltProvider, Text } from "@salt-ds/core";
+import { TabPanelNext, TabsNext, TabNext, TabstripNext } from "@salt-ds/lab";
+import {
+  CharacteristicUsage,
+  CSSClassTable,
+} from "css-variable-docgen-components";
+import HelpAndSupport from "docs/blocks/help-and-support.doc.mdx";
+
+<Meta title="Documentation/Lab/TabsNext" />
```
libertymayc, line 11:
```suggestion
<Meta title="Documentation/Lab/Tabs Next" />
```

packages/lab/stories/tabs-next/tabs-next.doc.stories.mdx
```diff
@@ -0,0 +1,154 @@
...
+##### Don't use when
+
+If your content isn't logically grouped into separate categories or if there are too many categories. In this case, an alternative navigation method may be more appropriate.
+
+## Features
+
+Tabs can be aligned to the left or centered within the Tabstrip. If not specified, the default alignment is left.
+
+<Canvas>
+  <Story id="lab-tabsnext-tabsnext--default" />
```
libertymayc, line 55:
```suggestion
  <Story id="lab-tabs-next-tabs-next--default" />
```

+ all below

packages/lab/stories/tabs-next/tabs-next.doc.stories.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+`import { TabsNext, TabPanelNext, TabstripNext } from "@salt-ds/lab";`
+
+### Props
+
+<ArgsTable of={TabsNext} />
+
+## Related components
+
+The Tab component is part of the [Tabstrip](?path=/docs/documentation-tabstrip-next-tabstripnext--page) component.
```
libertymayc, line 123: link doesnt exist

packages/lab/src/tabs-next/TabNext.css
libertymayc: not yet since the 'restructure' (similar to spacing work) is on hold for now, but Paul is getting round to it soon cc @pseys :)

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,87 @@
+// TODO close button needs to be a button. Hence tab needs to include 2 buttons
```
alycrys, line 1: what close button?

packages/lab/src/__tests__/__e2e__/tabs-next/TabsNext.cy.tsx
ZarkoPernar: We do, because this is testing an "incorrect" use case

packages/lab/src/tabs-next/TabNext.tsx
ZarkoPernar: I gues not

ZarkoPernar: I guess not

packages/lab/src/tabs-next/TabNext.tsx
ZarkoPernar: ✅

packages/lab/src/tabs-next/TabstripNext.tsx
ZarkoPernar: No, because we need these ids in Tabstrip to pass them to overflow menu and overflow library

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,177 @@
...
+                    label,
+                    tabIndex: isActive ? 0 : -1,
+                    selected: isActive,
+                    index,
+                    onClick: () => {
+                      setActiveTabIndex(index);
+                      onActiveChange?.(index);
+                    },
+                    onKeyUp: noop,
+                    onKeyDown: (e) => {
```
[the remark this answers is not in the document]
ZarkoPernar: It will just create more indirection because I have to do something like `createHandleKeyDown(isActive, index)` which is then a factory function etc.

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,177 @@
...
+                      const focusableElements = Array.from(
+                        outerRef.current?.querySelectorAll<HTMLDivElement>(
+                          `[data-overflow-item]:not([data-overflowing]) [role="tab"], [data-overflow-menu] button`
```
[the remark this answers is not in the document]
ZarkoPernar: Sounds good

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,177 @@
...
+  /* Value for the uncontrolled version. Set to null in order to have no tabs be selected. */
+  activeTabIndex?: number | null;
+  /* Callback for the controlled version. */
+  onActiveChange?: (index?: number) => void;
+  /* Initial value for the uncontrolled version. Set to null in order to have no tabs be selected by default. */
+  defaultActiveTabIndex?: number | null;
+  /* Align the tabs to the center. Left aligned by default. */
+  align?: "center";
+  /* Set a tab max-width in order to enable tab truncation */
+  tabMaxWidth?: number;
```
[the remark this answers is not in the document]
ZarkoPernar: Tabstrip can be used without Tabs, if you are making your own tabbed interface.

packages/lab/src/tabs-next/TabNext.tsx
ZarkoPernar: removed, old comment

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,87 @@
...
+  {
+    ariaControls,
+    children,
+    className,
+    index,
+    label,
+    onClick,
+    onFocus: onFocusProp,
+    onKeyDown,
+    onKeyUp,
```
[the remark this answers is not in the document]
ZarkoPernar: It will get overriden by Tabstrip anyways

packages/lab/src/tabs-next/TabsNext.tsx
```diff
@@ -0,0 +1,84 @@
...
+    controlled: activeTabIndexProp,
+    default: defaultActiveTabIndex ?? 0,
+    name: "useTabs",
+    state: "activeTabIndex",
+  });
+
+  const uniqueId = useId();
+  const getTabId = useCallback(
+    (label: string) => {
+      return `tab-${uniqueId ?? "unknown"}-${label}`;
```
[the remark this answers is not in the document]
ZarkoPernar: It doesn't really matter. No one can do much with it anyways since it is unique

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,122 @@
...
+  color: inherit;
+  cursor: inherit;
+  align-items: center;
+  font-family: inherit;
+  font-size: inherit;
+  font-weight: inherit;
+  height: var(--salt-size-base);
+  margin-top: var(--salt-spacing-50);
+  outline: none;
+  padding: 0 var(--tabs-tab-spacing);
```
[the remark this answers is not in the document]
ZarkoPernar: Why? The tabstrip controls the spacing

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,122 @@
...
+  flex-shrink: 0;
+  align-items: center;
+  align-self: stretch;
+  justify-content: center;
+  background: var(--saltTabs-tab-background, var(--tabs-tab-background));
+  border: none;
+  border-radius: 0;
+  color: var(--saltTabs-tab-color, var(--tabs-tab-color));
+  cursor: var(--saltTabs-tab-cursor, var(--tabs-tab-cursor));
+  display: var(--tabs-tabstrip-display);
```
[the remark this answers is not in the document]
ZarkoPernar: legacy code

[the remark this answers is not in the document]
ZarkoPernar: I will remove it

packages/lab/src/tabs-next/TabPanelNext.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { makePrefixer } from "@salt-ds/core";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+
+import tabPanelCss from "./TabPanelNext.css";
+
+const withBaseName = makePrefixer("saltTabPanelNext");
+
+export interface TabPanelNextProps extends HTMLAttributes<HTMLDivElement> {
+  label: string;
```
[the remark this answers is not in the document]
ZarkoPernar: We use it to create tabs inside tabstrip. It is not used directly on the tabpanel children

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
```diff
@@ -0,0 +1,252 @@
...
+          <TabNext label={label} key={label}>
+            {label}
+          </TabNext>
+        ))}
+      </TabstripNext>
+    </div>
+  );
+};
+
+export const TruncatedTabs: TabstripStory = ({
```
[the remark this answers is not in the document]
ZarkoPernar: Removed, we are not doing truncation in v1

packages/lab/src/tabs-next/TabNext.css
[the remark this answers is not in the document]
alycrys: between the tabs and inside the tab?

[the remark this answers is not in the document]
libertymayc: since it doesn't get redefined to any other value can just use the spacing token directly

packages/lab/src/tabs-next/TabstripNext.tsx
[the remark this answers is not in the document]
alycrys: but can Tabs also use these features? Don't see why not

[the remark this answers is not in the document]
ZarkoPernar: Tabs and Tabstrip both have this feature. Tabs extends most of Tabstrip prop types

packages/lab/src/tabs-next/TabsNextTypes.ts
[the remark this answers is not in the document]
ZarkoPernar: Will refactor the types in v2

packages/lab/src/tabs-next/TabNext.css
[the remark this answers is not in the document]
ZarkoPernar: It will be easier to change in the future if I don't have to hunt down the use of spacing tokens and instead can refactor the top level tabs variables only.

packages/lab/src/tabs-next/TabstripNext.css
```diff
@@ -0,0 +1,68 @@
+/* Component class applied to the root element */
+.saltTabstripNext {
+  --tabs-tabstrip-display: var(--saltTabs-tabstrip-display, inline-flex);
+  --tabs-tabstrip-background: var(--saltTabs-tabstrip-background, transparent);
+
+  --tabs-activationIndicator-thumb-height: var(--saltTabs-activationIndicator-thumb-height, var(--salt-size-bar));
+  --tabs-activationIndicator-height: var(--saltTabs-activationIndicator-height, var(--salt-size-border));
+  --tabs-activationIndicator-background: var(--saltTabs-activationIndicator-background, var(--salt-container-primary-borderColor));
```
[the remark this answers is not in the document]
ZarkoPernar: It doesn't make sense as a border either, since there are two borders anyways, one for tabs and one for tab. Id rather keep it this way. It doesn't necessarily need to refer to a component.

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -28,7 +29,7 @@
   min-width: var(--saltTabs-tab-minWidth, var(--salt-spacing-400));
   outline: none;
   position: var(--saltTabs-tab-position, var(--tabs-tab-position));
-  border-bottom-width: var(--tabs-activationIndicator-thumb-height);
+  border-bottom-width: var(--tabs-tab-border-width);
```
libertymayc, line 32:
```suggestion
  border-bottom-width: var(--saltTabs-tab-border-width, var(--salt-size-bar));
```

local token unnecessary

packages/lab/src/tabs-next/TabstripNext.css
```diff
@@ -23,8 +22,8 @@
   content: "";
   position: absolute;
   inset: auto 0 0 0;
-  height: var(--tabs-activationIndicator-height);
-  background: var(--tabs-activationIndicator-background);
+  height: var(--tabs-border-width);
+  background: var(--tabs-border-color);
```
libertymayc, line 26:
```suggestion
  height: var(--saltTabs-border-width, var(--salt-size-border));
  background: var(--saltTabs-border-color, var(--salt-container-primary-borderColor));
```

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
```diff
@@ -0,0 +1,223 @@
+import { useState } from "react";
+import { Story } from "@storybook/react";
+import { Button, FlexLayout } from "@salt-ds/core";
+import { TabstripNextProps, TabstripNext, TabNext } from "@salt-ds/lab";
+import { AddIcon } from "@salt-ds/icons";
+import "./tabstrip-next.stories.css";
+
+export default {
+  title: "Lab/Tabs Next/Tabstrip Next",
```
feliciackh, line 9:
```suggestion
  title: "Lab/Tabstrip Next/Tabstrip Next",
```

packages/lab/src/tabs-next/TabstripNext.css
libertymayc: there's nothing wrong with the original, just unnecessary, this is cleaner

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
ZarkoPernar: That was intentional since I want that story nested under tabs

packages/lab/stories/tabs-next/tabs-next.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+          <TabPanelNext label={label} key={label}>
+            <Text>Content for {label} tab</Text>
+          </TabPanelNext>
+        ))}
+      </TabsNext>
+    </div>
+  );
+};
+
+export const Controlled: TabsStory = ({ width = 600, ...props }) => {
```
Fercas123, line 123: would it makes sense to have something external controlling the tabs? or something else that happens with the state change so the difference between cotrolled/uncontrolled examples is visually more evident?

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+          <TabNext label={label} key={label}>
+            {label}
+          </TabNext>
+        ))}
+      </TabstripNext>
+    </div>
+  );
+};
+
+export const ActiveIndexNull: TabstripStory = ({
```
Fercas123, line 180: what is the usecase of having active index set to null?

ZarkoPernar: Sometimes, when using the tabstrip alone you want no tabs selected by default. You do that with null, since undefined will offload the state to useControlled.

packages/lab/stories/tabs-next/tabs-next.stories.tsx
ZarkoPernar: Good point

packages/lab/stories/tabstrip-next/tabstrip-next.stories.tsx
Fercas123: I see, makes sense... i think currently that is making the tabs inaccessibly via keyboard in the example, could it be missing a tabindex or something like that?

ZarkoPernar: Ah, good catch! I think the first one should be focusable in that case 🤔

Fercas123 (approved):

.changeset/orange-ducks-wash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": minor
```
origami-z, line 2: Good to have a short-summary of the API changes / usage guideline, so consumers can read the changelog to adapt if needed

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,122 @@
...
+  --tabs-tab-background: var(--salt-navigable-primary-background);
+  --tabs-tab-background-hover: rgba(0, 0, 0);
+  --tabs-tab-backgroundOpacity-hover: 0.08;
+  --tabs-tab-color: var(--salt-text-primary-foreground);
+  --tabs-tab-letterSpacing: 0;
+  --tabs-tab-cursor: pointer;
+  --tabs-tab-position: relative;
+}
+
+.salt-theme[data-mode="dark"] .saltTabNext {
```
origami-z, line 12: Should there a mode dependent CSS in components?

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,78 @@
...
+  ref: ForwardedRef<HTMLDivElement>
+): ReactElement<TabProps> {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-tab-next",
+    css: tabCss,
+    window: targetWindow,
+  });
+  if (index === undefined || onClick === undefined || onKeyDown === undefined) {
+    throw Error(
```
origami-z, line 42: I thought it's `throw new Error`? Add a test maybe?

packages/lab/src/tabs-next/TabsNext.tsx
```diff
@@ -0,0 +1,74 @@
+import { Children, cloneElement, isValidElement, ReactNode } from "react";
+import { StackLayout, useControlled, useId } from "@salt-ds/core";
+import { TabstripNext, TabstripNextProps } from "./TabstripNext";
+import { TabNext } from "./TabNext";
+import { TabPanelNext } from "./TabPanelNext";
+import { TabElement } from "./TabsNextTypes";
+
+export type TabsNextProps = Omit<TabstripNextProps, "defaultActiveTabIndex"> & {
```
origami-z, line 8: So props are different? `defaultActiveTabIndex` vs `defaultSelectedTab`

packages/lab/src/tabs-next/TabsNext.tsx
```diff
@@ -0,0 +1,74 @@
...
+}: TabsNextProps) => {
+  const tabs = Children.toArray(children).filter(isTabPanel);
+  const [selectedTab, setSelectedTab] = useControlled({
+    controlled: selectedTabProp,
+    default: defaultSelectedTab ?? tabs[0]?.props.value,
+    name: "useTabs",
+    state: "selectedTab",
+  });
+
+  const uniqueId = useId();
```
origami-z, line 31: Should this honour id passed in?

packages/lab/src/tabs-next/TabstripNext.tsx
```diff
@@ -0,0 +1,178 @@
...
+  selectedTab?: string | null;
+  /* Callback for the controlled version. */
+  onSelectTab?: (
+    e: KeyboardEvent<HTMLElement> | MouseEvent<HTMLElement>,
+    value: string
+  ) => void;
+  /* Initial value for the uncontrolled version. Set to null in order to have no tabs be selected by default. */
+  defaultSelectedTab?: string | null;
+  /* Align the tabs to the center. Left aligned by default. */
+  align?: "center";
```
origami-z, line 39: Type should be 2?

packages/lab/stories/tabstrip-next/tabstrip-next.stories.css
```diff
@@ -0,0 +1,3 @@
+#root {
```
origami-z, line 1: Why is this needed? If this impacts story, then the user likely will bump into the same problem and we need document this

yarn.lock
```diff
@@ -3614,6 +3623,92 @@ __metadata:
...
+"@fluentui/keyboard-keys@npm:^9.0.3":
+  version: 9.0.3
+  resolution: "@fluentui/keyboard-keys@npm:9.0.3"
+  dependencies:
+    "@swc/helpers": ^0.4.14
+  checksum: ab723fb9580d71f83861189cc3447d8f1500b2d76f7f71c915ab8f23591bd47d9b8626ae98dcb22f1f8e815f6c350a93553158fa465e2d40a5febd5b722a88ea
+  languageName: node
+  linkType: hard
+
+"@fluentui/priority-overflow@npm:^9.0.3":
```
origami-z, line 3635: It's worth to check the bundle size impact of these transient dependencies

.changeset/orange-ducks-wash.md
ZarkoPernar: Yep, I was going to do that after we finalize and all agree on the api

packages/lab/src/tabs-next/TabNext.css
ZarkoPernar: Unfortunately, I don't have these characteristics defined elsewhere yet. I can also add as a todo and an issue?

packages/lab/src/tabs-next/TabsNext.tsx
ZarkoPernar: Whoops, that is a leftover from the previous version. Will remove

packages/lab/src/tabs-next/TabsNext.tsx
```diff
@@ -0,0 +1,74 @@
...
+  selectedTab: selectedTabProp,
+  defaultSelectedTab,
+  onSelectTab: onSelectTabProp,
+  ...props
+}: TabsNextProps) => {
+  const tabs = Children.toArray(children).filter(isTabPanel);
+  const [selectedTab, setSelectedTab] = useControlled({
+    controlled: selectedTabProp,
+    default: defaultSelectedTab ?? tabs[0]?.props.value,
+    name: "useTabs",
```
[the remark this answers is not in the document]
ZarkoPernar: Yeah, I think I copied from the usage from old tabs but I guess there it was inside a hook. Will rename.

packages/lab/src/tabs-next/TabsNextTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+  HTMLAttributes<HTMLElement>,
+  "onClick" | "onKeyUp"
+> & {
+  /* Value prop is mandatory and must be unique in order for overflow to work. */
+  value: string;
+  /* Label prop is mandatory for accessibility. */
+  label: string;
+  id?: string;
+  disabled?: boolean;
+  ariaControls?: AriaAttributes["aria-controls"];
```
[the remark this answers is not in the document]
ZarkoPernar: Probably, yes. It is a leftover from original tabs.

packages/lab/src/tabs-next/TabstripNext.tsx
ZarkoPernar: Yeah, it should probably be left and center. Instead of undefined and center.

yarn.lock
ZarkoPernar: This is necessary only due to how our storybook setup wraps the contents by default, which is to center everything. I don't think we need to document this. This is basic flex layout.

ZarkoPernar: Josh checked when he did the initial experiment. He said it was the size of two of our buttons. I think our code would likely be of somewhat similar size if we decided to maintain our own version.

packages/lab/src/tabs-next/TabNext.tsx
ZarkoPernar: `throw Error` works as well.

packages/lab/src/tabs-next/TabsNext.tsx
origami-z: Yes, this makes sure consumer could apply an ID of choice in the application so they can write logic or test against it.
Thus why useId accepts a default.

yarn.lock
origami-z: Have you confirmed? If so, add a line to PR description for future reference of current decision making.

packages/lab/src/tabs-next/TabsNext.tsx
ZarkoPernar: Sounds good

yarn.lock
ZarkoPernar: Adds 5kb minified and gzipped. https://bundlephobia.com/package/@fluentui/react-overflow@9.0.19

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,122 @@
...
+.saltTabNext:hover:not([aria-selected="true"]):not(.saltTabNext-disabled) {
+  border-bottom-color: var(--salt-navigable-indicator-hover);
+}
+
+.saltTabNext:hover:not(.saltTabNext-disabled)::before,
+.saltTabNext:focus-visible::before {
+  content: "";
+  position: absolute;
+  inset: 0;
+  background: var(--saltTabs-tab-background-hover, var(--tabs-tab-background-hover));
```
libertymayc, line 109: related to the other comment, I'm not really sure this token is doing anything?

Fercas123:
image

im not very sure how i got here, but if you tab into the button, open with enter, exit, move around and do that a few times, at some point enter/space opens the dropdown but sends the focus to the first item in tabstrip

joshwooding:
> image
> im not very sure how i got here, but if you tab into the button, open with enter, exit, move around and do that a few times, at some point enter/space opens the dropdown but sends the focus to the first item in tabstrip.
> 
> i think it happens when you open and arrow down too quickly

Do you have a clear reproduction? If not I don't think this is an actual issue we can fix

Fercas123:
> > image
> > im not very sure how i got here, but if you tab into the button, open with enter, exit, move around and do that a few times, at some point enter/space opens the dropdown but sends the focus to the first item in tabstrip.
> > i think it happens when you open and arrow down too quickly
> 
> Do you have a clear reproduction? If not I don't think this is an actual issue we can fix

seems to happen consistently if you press enter/space + arrow down on the button

Fercas123 (approved):

PR #1686, opened 2023-05-18
[WIP] Add props table component to site docs

JoanaMMoreira:


mark-tate:
This could be a mosaic source rather than a script and any components could be mosaic components, so can be re-used across other sites and use-cases?

Would be good to consider Salt site complete and move all further feature changes to mosaic codebase. This will  push us towards the goal of salt site updates being content only.

Future Salt spokes would be generated from a generator of mosaic features and theme. This will help us converge on a single theme/components/layout for all Mosaic/Salt sites. 

Worth discussing with @joshwooding @DavieReid  @lilyvc  and sharing your own thoughts as well

JoanaMMoreira:
> This could be a mosaic source rather than a script and any components could be mosaic components, so can be re-used across other sites and use-cases?
> 
> Would be good to consider Salt site complete and move all further feature changes to mosaic codebase. This will push us towards the goal of salt site updates being content only.
> 
> Future Salt spokes would be generated from a generator of mosaic features and theme. This will help us converge on a single theme/components/layout for all Mosaic/Salt sites.
> 
> Worth discussing with @joshwooding @DavieReid @lilyvc and sharing your own thoughts as well

That sounds really good, it would be good to discuss how the mosaic source would work in practice. @joshwooding @DavieReid @mark-tate @lilyvc should I setup a call?

PR #1698, opened 2023-05-22
rename style injection id to testId as it is optional

el-dav:


packages/core/src/avatar/Avatar.tsx
```diff
@@ -61,8 +61,8 @@ export const Avatar = forwardRef<HTMLDivElement, AvatarProps>(function Avatar(
 ) {
   const { window: targetWindow } = useWindow();
   useComponentCssInjection({
-    id: "salt-avatar",
     css: avatarCss,
+    testId: "salt-avatar",
```
ImmanuelBaskaran, line 65: Nitpick, testId should be above css. Not a blocker.

ImmanuelBaskaran (approved):

PR #1701, opened 2023-05-23
Input v1 cleanup

libertymayc:
Cleanup of InputNext for v1


packages/lab/src/input-next/InputNext.tsx
```diff
@@ -131,17 +102,6 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
 
   const [focused, setFocused] = useState(false);
 
-  const misplacedAriaProps = {
-    "aria-activedescendant": ariaActiveDescendant,
-    "aria-expanded": ariaExpanded,
-    "aria-owns": ariaOwns,
-    role,
-  };
-  const inputProps = mergeA11yProps(
```
joshwooding, line 140: This used to merge the aria-labelledby props but this no longer happens?

libertymayc: I moved that to line 156 as thought it was a bit clunky just to merge that, what do you think?

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -17,10 +15,8 @@ import "./InputNext.css";
 
 const withBaseName = makePrefixer("saltInputNext");
 
-// TODO: Double confirm whether this should be extending DivElement given root is `<div>`.
-// And forwarded ref is not assigned to the root like other components.
 export interface InputProps
-  extends Omit<HTMLAttributes<HTMLDivElement>, "defaultValue"> {
+  extends Omit<InputHTMLAttributes<HTMLDivElement>, "defaultValue"> {
```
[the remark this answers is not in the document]
libertymayc: mirrors checkbox etc, not too sure myself what the difference is here?

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -188,20 +148,25 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
...
       <input
+        aria-label={clsx(
+          a11yProps?.["aria-labelledby"],
+          inputProps["aria-labelledby"]
+        )}
         id={id}
         className={clsx(withBaseName("input"), inputProps?.className)}
         disabled={isDisabled}
         readOnly={isReadOnly}
-        ref={ref}
```
alycrys, line 199: I think we might want to add an `inputRef` in case there is a need to ref the actual input. I believe input is a little different from checkbox and radio button. Not sure. Maybe we should add it as an enhancement later if there will be a need for it

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
origami-z: > mirrors checkbox etc, not too sure myself what the difference is here?

This adds more valid attribute for TypeScript. Check `@types/react` for detail

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -188,20 +148,25 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
         },
         classNameProp
       )}
+      ref={ref}
       style={inputStyle}
       {...other}
     >
       <input
+        aria-label={clsx(
+          a11yProps?.["aria-labelledby"],
```
origami-z, line 157: `aria-labelledby` is an `id`? `aria-label` is plain text?

packages/lab/src/input-next/InputNext.tsx
libertymayc: cc @joshwooding thoughts?

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
joshwooding: I guess with this, there's no guarantee that we forward the props though. So this could lead to inconsistency?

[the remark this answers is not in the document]
origami-z: My suggestion would be looking at those additional ones, and see whether they will be applicable / working correctly when added as input props.

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -131,17 +98,6 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
 
   const [focused, setFocused] = useState(false);
 
-  const misplacedAriaProps = {
```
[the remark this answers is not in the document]
joshwooding: Not sure, it probably existed before we moved Input to this repo. To make a guess it was probably props that weren't being explicitly forwarded to the input element

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
joshwooding: Is it better to just use:

```ts
ComponentPropsWithoutRef<"div">, Pick<ComponentPropsWithoutRef<"input">, ...values>
```

packages/lab/src/input-next/InputNext.tsx
joshwooding: I think inputRef might be needed

packages/lab/src/input-next/InputNext.tsx
```diff
@@ -196,13 +153,17 @@ export const Input = forwardRef<HTMLInputElement, InputProps>(function Input(
...
-        ref={ref}
         value={value}
         tabIndex={isReadOnly || isDisabled ? -1 : 0}
         onBlur={handleBlur}
         onChange={handleChange}
         onFocus={!isDisabled ? handleFocus : undefined}
+        {...a11yProps}
         {...restInputProps}
+        aria-labelledby={clsx(
+          a11yProps?.["aria-labelledby"],
```
origami-z, line 164: Spreading `a11yProps` first and then picking one key out to override may lead to code confusion. e.g. we destructure things we need to modify from `props` and spread rest

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
libertymayc: yes that makes sense - I guess it was for e.g. dropdown.. may need it in that case but wondering if until dropdown is working on we should leave that complexity out? or should i bring them back explicitly?

.changeset/three-kids-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": minor
+---
+
+`InputProps` interface changed to extend `InputHTMLAttributes<HTMLDivElement>` , `mergeA11yProps` removed, `ref` moved to container div
```
origami-z, line 5: `mergeA11yProps` was not exported before? so doesn't matter for consumers?

Could put a bit more wording on to the `ref` change, consequence of before / after

libertymayc: updated

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
joshwooding: I think it would be the props we already forward for now, if we need to forward more we can just make a bug fix

packages/lab/src/input-next/InputNext.tsx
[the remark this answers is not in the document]
joshwooding: I would probably forward them to input

.changeset/three-kids-refuse.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/lab": minor
+---
+
+`InputProps` interface changed to extend `InputHTMLAttributes<HTMLDivElement>` ,
```
libertymayc, line 5:
```suggestion
`InputProps` interface changed to extend `Omit<ComponentPropsWithoutRef<"div">,"defaultValue">, Pick<ComponentPropsWithoutRef<"input">, "disabled" | "value" | "defaultValue">` 
```

.changeset/three-kids-refuse.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/lab": minor
+---
+
+`InputProps` interface changed to extend `Omit<ComponentPropsWithoutRef<"div">,"defaultValue">, Pick<ComponentPropsWithoutRef<"input">, "disabled" | "value" | "defaultValue">`
```
libertymayc, line 5:
```suggestion
`InputNext`: `InputProps` interface changed to extend `Omit<ComponentPropsWithoutRef<"div">,"defaultValue">, Pick<ComponentPropsWithoutRef<"input">, "disabled" | "value" | "defaultValue">`
```

PR #1704, opened 2023-05-23
Add component pages sidebar data

JoanaMMoreira:
- Add sidebar metadata to component pages
- Create `SecondarySidebar` component that is used by the `DetailComponent` layout and handles the sidebar links and pills
- Swap examples list view so toggle is initially disabled by default


libertymayc (approved):
just the pills have uppercase as the text transform token isn't defined - it's a problem with Pill, but maybe you could override it here until that's fixed if desired :)

JoanaMMoreira:
> just the pills have uppercase as the text transform token isn't defined - it's a problem with Pill, but maybe you could override it here until that's fixed if desired :)

Good spot! I'll have to do a follow up PR for the mobile view anyway so I'll fix the pills then :)

PR #1706, opened 2023-05-23
[WIP] List cypress test

feliciackh:


packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -104,12 +104,7 @@ export const ListItemNext = forwardRef<HTMLLIElement, ListItemNextProps>(
     );
 
     return isOverflowed ? (
-      <Tooltip
-        disabled={!isOverflowed}
-        open={focused}
-        content={content}
-        hideIcon
-      >
+      <Tooltip disabled={!isOverflowed} content={content} hideIcon>
```
Fercas123, line 107: nice! does this work for focused with keyboard too?

packages/theme/css/characteristics/selectable.css
```diff
@@ -10,59 +10,140 @@
...
-  --salt-selectable-background-selected: var(--salt-palette-interact-background-active);
-  --salt-selectable-background-blurSelected: var(--salt-palette-interact-background-blurSelected);
-  --salt-selectable-background-disabled: var(--salt-palette-interact-background-disabled);
+  --salt-selectable-background-hover: var(
+    --salt-palette-interact-background-hover
+  );
+  --salt-selectable-background-selected: var(
+    --salt-palette-interact-background-active
+  );
+  --salt-selectable-background-selectedDisabled: #a4d5f466;
```
feliciackh, line 47: WIP: to be updated based on latest guidance by @libertymayc

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -106,9 +106,9 @@ export const ListItemNext = forwardRef<HTMLLIElement, ListItemNextProps>(
     return isOverflowed ? (
       <Tooltip
         disabled={!isOverflowed}
-        open={focused}
         content={content}
         hideIcon
+        open={focused || undefined}
```
libertymayc, line 111: not sure if this is the best fix for this one, would expect focused to be well defined regardless of mouse or keyboard but im unsure without looking too deeply

packages/lab/src/list-next/ListItemNext.css
```diff
@@ -40,23 +40,42 @@
...
 /* TODO: do we need to add the role exception to the below classes? */
 .saltListItemNext-focused {
   --list-item-background: var(--salt-selectable-background-hover);
   cursor: var(--salt-selectable-cursor-hover);
 }
 
-.saltListItemNext[aria-selected="true"] {
-  border-block: var(--salt-size-border) var(--salt-container-borderStyle) var(--salt-selectable-borderColor-selected);
+/* active list item when selected */
+.saltListItemNext[aria-selected="true"]:not(.saltListItemNext-disabled) {
```
libertymayc, line 52: we're trying to avoid using 'not', could this be split out into an override if also disabled?

packages/theme/css/characteristics/selectable.css
feliciackh: updated in [PR here ](https://github.com/jpmorganchase/salt-ds/pull/1638)

feliciackh:
> not sure it's in scope for this PR: but tab to focus doesn't show focus on first item (i can't see any focus), and then arrow down goes to 2nd item, so seems focus on first is broken

should be fixed in [PR here ](https://github.com/jpmorganchase/salt-ds/pull/1638)

PR #1708, opened 2023-05-24
List next clean and comment hook

Fercas123:
cleaning hook and leaving comments on things we need to keep tabs on

### DIFF
--- a/packages/lab/src/list-next/useList.ts
+++ b/packages/lab/src/list-next/useList.ts
@@ -71,8 +71,8 @@ export const useList = ({
     }
   };
 
-  const getElementIndex = (option): number => {
-    const optionIndex = getAllOptions().indexOf(option);
+  const getListItemIndex = (item): number => {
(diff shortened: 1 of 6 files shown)


packages/lab/src/list-next/useList.ts
```diff
@@ -28,45 +29,166 @@ export const useList = ({
...
+    // Find last active item
+    const allActiveOptions = getAllActiveOptions();
+    const lastItem = allActiveOptions[allActiveOptions.length - 1];
     if (lastItem) {
-      focusItem(lastItem);
+      focusAndSelect(lastItem);
+    }
+  };
+
+  const getElementIndex = (option): number => {
```
feliciackh, line 74: QQ - would the name `getItemIndex` be a better fit as it seems majority of functions refer to list items here as item, could keep it consistent

feliciackh (approved):

PR #1710, opened 2023-05-24
Banner Changes emphasize to variant

joshwooding:
# Todo 

- [ ] Should we note the two types of alerts: those based on user action and those not
- [ ] Should we have guidance for when to include a title when not


packages/core/src/banner/BannerContent.tsx
```diff
@@ -1,10 +1,13 @@
...
 import "./BannerContent.css";
+import { clsx } from "clsx";
 
 const withBaseName = makePrefixer("saltBannerContent");
 
-export const BannerContent = ({ children }: { children: ReactNode }) => {
-  return <span className={withBaseName()}>{children}</span>;
+export interface BannerContentProps extends ComponentPropsWithoutRef<"span"> {}
+
+export const BannerContent = ({ className, ...rest }: BannerContentProps) => {
```
alycrys, line 11: so what would be the use of the BannerContent if it's the only one left on composition?

joshwooding: It ensures alignment at the moment and it current serves as a place to put the role. Although I would have to check with Stella

joshwooding:
Changed base to main so the builds can run

packages/core/src/banner/BannerContent.tsx
alycrys: the role can be placed directly on Banner. I believe this can be merged with my changes on https://github.com/jpmorganchase/salt-ds/pull/1643 which includes removal of BannerContent

joshwooding: It could be placed but then the button would be announced which might not be desirable

alycrys (approved):

alycrys (dismissed):

packages/core/src/banner/Banner.tsx
```diff
@@ -0,0 +1,40 @@
+import { forwardRef, HTMLAttributes } from "react";
+import { makePrefixer } from "../utils";
+import { StatusIndicator, ValidationStatus } from "../status-indicator";
+import { clsx } from "clsx";
+
+import "./Banner.css";
+
+export interface BannerProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Emphasize the styling by applying a background color: defaults to false
```
alycrys, line 10:
```suggestion
   * Emphasize the Banner by applying variant `secondary`. 
```

PR #1712, opened 2023-05-25
update WindowContext to just be window instead of an object

el-dav:


ImmanuelBaskaran (changes requested):
The type of windowContext is now Window, but the hook useComponentCssInjection still expects WindowLike

We either need to change the type the hook expects to window, or change the context type to WindowLike

ImmanuelBaskaran (approved):

PR #1720, opened 2023-05-30
Accordion

joshwooding:


tooling/theme-editor/src/tokens/foundations/shadow/FlatShadow.tsx
```diff
@@ -29,10 +25,11 @@ export const FlatShadow = (props: FlatShadowProps): ReactElement => {
   const patternParts = getShadowParts(props.shadowPattern);
 
   return (
-    <AccordionSection
+    <Accordion
       key={`${props.themeName}-${props.innerPattern}-accordion`}
```
Fercas123, line 29: why not spreading props above? (same for all the shadows)

Fercas123 (approved):
Nice one! just one tiny one, i'm not sure if we have a convention about it but not a blocker

tooling/theme-editor/src/tokens/foundations/shadow/FlatShadow.tsx
joshwooding: I just copied what was there 😄 

I didn't want to go into a rabbit hole of refactoring the theme editor code 🥲

packages/lab/src/accordion/AccordionHeader.css
```diff
@@ -0,0 +1,42 @@
...
+  --saltIcon-color: var(--salt-text-primary-foreground);
+  text-align: left;
+}
+
+.saltAccordionHeader:focus-visible {
+  background: var(--salt-actionable-primary-background-hover);
+  outline: var(--salt-focused-outline);
+}
+
+@media (hover: hover) and (pointer: fine) {
```
ZarkoPernar, line 24: This is to prevent the hover style on mobile press? We don't do this in other places. Should we start?

ZarkoPernar: But then again, you don't have an :active style. Does that mean the mobile user will see no visual indicator of their press?

ZarkoPernar (approved):

packages/lab/src/accordion/AccordionHeader.css
joshwooding: We should probably do it elsewhere but this is because of a lack of an `:active` style. Accordion wasn't designed with one

PR #1738, opened 2023-05-31
Hide component docs table of contents when only a single example is displayed

JoanaMMoreira:
- Create `AllExamplesViewContext` to allow sidebar component to hide table of contents if only a single example is displayed
- Fix broken link on Getting Started page
- Preview: https://saltdesignsystem-git-fix-toc-in-docs-pages-joshwooding.vercel.app/salt/components/button/examples


origami-z:
What's wrong with single item in ToC?

JoanaMMoreira:
> What's wrong with single item in ToC?

Currently all items are displayed even if only one example is selected. We could have just a single item in the ToC but I'm not sure how useful that would be for users

PR #1741, opened 2023-05-31
Add status pill to site page headers

JoanaMMoreira:


james-nash (dismissed):
LGTM!

james-nash:
Since Mosaic builds in `main` are borked right now, I've rebased this off the Foundation pages branch since:

- Mosaic builds still work there (because it branched off `main` before the style injection stuff was added)
- This feature is primarily needed for those foundation pages, so once merged we can configure and preview them as needed

PR #1746, opened 2023-06-01
[Input v2] Input Adornment button

libertymayc:


packages/core/src/input/adornment-button/AdornmentButton.tsx
```diff
@@ -0,0 +1,34 @@
...
+  useComponentCssInjection({
+    testId: "salt-adornment-button",
+    css: AdornmentButtonCss,
+    window: targetWindow,
+  });
+
+  return (
+    <Button
+      className={withBaseName()}
+      disabled={disabled}
```
alycrys, line 26: Should disabled come from the ff context?

packages/core/src/input/Input.css
```diff
@@ -185,4 +185,17 @@
   display: inline-flex;
   padding-left: var(--salt-size-adornmentGap);
   column-gap: var(--salt-size-adornmentGap);
-}
\ No newline at end of file
+}
+
+/* TODO: Firefox fix */
+.saltInput:has(.saltInput-startAdornmentContainer):has(.saltAdornmentButton) {
+  --saltInput-paddingLeft: calc(0.25 * var(--salt-size-unit));
```
alycrys, line 192: makes sense to use `--salt-spacing-25`?
same for all `--salt-size-unit` calculations here

packages/core/src/input/adornment-button/AdornmentButton.tsx
libertymayc: fixed, leaving comment open as pending also what should happen in read only to a button cc @bhoppers2008

libertymayc: decision to disable if readonly

packages/core/src/input/Input.tsx
```diff
@@ -208,6 +208,11 @@ export const Input = forwardRef<HTMLDivElement, InputProps>(function Input(
           {endAdornment}
         </div>
       )}
+      {endAdornment && (
+        <div className={withBaseName("endAdornmentContainer")}>
+          {endAdornment}
+        </div>
+      )}
```
alycrys, line 215: hmmmm

alycrys:
Would it make sense to add some FF stories with these adornments?

packages/core/src/__tests__/__e2e__/form-field/FormField.cy.tsx
```diff
@@ -222,5 +223,57 @@ describe("GIVEN a FormField", () => {
...
+              startAdornment={<AdornmentButton>Test</AdornmentButton>}
+              data-testid="test-id-3"
+            />
+          </FormField>
+        );
+        cy.findByRole("button").should("be.visible");
+      });
+
+      it("THEN should disable the button when disabled", () => {
+        const clickSpy = cy.stub().as("clickSpy");
```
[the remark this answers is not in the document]
libertymayc: ooh yes, I had it to check on focus you couldn't click before realising you can't focus when disabled anyway :P deleted

Issue #1567, opened 2023-04-24
Checkbox and Radio Button updates


libertymayc:
- Readonly states to be added to form controls Radio Button and Checkbox for the next iteration

- Disabled/readonly to not be allowed with error/warning state (aligning with Input in FF work) - will be deprecated in code

- Add the small tick icon to HD in both RB/CB

- Re-organise the figma component for both (following Andi Wills' feedback on figma- cc @bhoppers2008 @navkaur76 ) Figma comment: https://www.figma.com/file/mmygvNxjCFW9orsBaqOKPE?node-id=2421:471255#462815262

- Add warning state to CB/RB

parsakhan21:
PR for spacing updates
https://github.com/jpmorganchase/salt-ds/pull/1835
 
PRs for removing disabled error styling variant
https://github.com/jpmorganchase/salt-ds/pull/1817 [radio]
https://github.com/jpmorganchase/salt-ds/pull/1815 [checkbox]
 
PRs for readonly (drafted ~ pending tokens/design decision)
https://github.com/jpmorganchase/salt-ds/pull/1826 [radio]
https://github.com/jpmorganchase/salt-ds/pull/1814 [checkbox]
 
& Checkbox HD small tick pending.

parsakhan21:
June 16- in progress; outstanding - read only states - call with ben/darrin/paul to finalize design direction next week

libertymayc:
jun 28 - still pending design feedback

PR #1750, opened 2023-06-01
[Form Field v2] Checkbox and Radiobutton support

libertymayc:
Checkbox and RadioButton integration with FF full support
Part of V2 Form Field features

**PENDING**: Warning and Readonly support (features missing & both needed in CB/RB)

### DIFF
--- a/packages/core/src/radio-button/RadioButton.tsx
+++ b/packages/core/src/radio-button/RadioButton.tsx
@@ -135,7 +135,7 @@ export const RadioButton = forwardRef<HTMLLabelElement, RadioButtonProps>(
           {
             [withBaseName("disabled")]: disabled,
             [withBaseName(validationStatus || "")]: validationStatus,
-            [withBaseName("error")]: error
+            [withBaseName("error")]: error,
           },
           className
         )}
--- a/packages/core/src/radio-button/RadioButton.tsx
+++ b/packages/core/src/radio-button/RadioButton.tsx
@@ -143,8 +143,14 @@ export const RadioButton = forwardRef<HTMLLabelElement, RadioButtonProps>(
         {...rest}
       >
         <input
-          aria-describedby={clsx(radioGroup.a11yProps?.["aria-describedby"], inputDescribedBy)}
-          aria-labelledby={clsx(radioGroup.a11yProps?.["aria-labelledby"], inputLabelledBy)}
+          aria-describedby={clsx(
+            radioGroup.a11yProps?.["aria-describedby"],
+            inputDescribedBy
+          )}
+          aria-labelledby={clsx(
+            radioGroup.a11yProps?.["aria-labelledby"],
+            inputLabelledBy
+          )}
           className={withBaseName("input")}
           {...restInputProps}
           checked={checked}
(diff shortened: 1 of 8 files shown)


packages/core/stories/form-field/form-field.stories.tsx
```diff
@@ -205,6 +205,16 @@ export const WithValidation: ComponentStory<typeof FormField> = (props) => {
...
+      <FormField validationStatus="error" {...props}>
+        <FormLabel>Error Form Field</FormLabel>
+        <Checkbox label="Option 1" />
+        <FormHelperText>Helper text</FormHelperText>
+      </FormField>
+      <FormField validationStatus="error" {...props}>
+        <FormLabel>Error Form Field</FormLabel>
+        <RadioButton label="Option 1" />
+        <FormHelperText>Helper text</FormHelperText>
+      </FormField>
```
alycrys, line 217: I think design can give some more realistic examples to add here, based on their research on forms

libertymayc: agreed, waiting on specs from @bhoppers2008

alycrys:
There is a scenario that is missing, the `readOnly` form field. There was a discussion about read-only states on checkbox and radio button. Not sure what the status is on that. Maybe @navkaur76 or @bhoppers2008 can advise on this?

navkaur76:
> There is a scenario that is missing, the `readOnly` form field. There was a discussion about read-only states on checkbox and radio button. Not sure what the status is on that. Maybe @navkaur76 or @bhoppers2008 can advise on this?

For CB/RB the read-only state has been designed, and as far as I'm aware can be added to the component. It was left as an agenda item for Q2 as there was some ADA questions around it to discuss with Stella.

libertymayc:
> > There is a scenario that is missing, the `readOnly` form field. There was a discussion about read-only states on checkbox and radio button. Not sure what the status is on that. Maybe @navkaur76 or @bhoppers2008 can advise on this?
> 
> For CB/RB the read-only state has been designed, and as far as I'm aware can be added to the component. It was left as an agenda item for Q2 as there was some ADA questions around it to discuss with Stella.

it's tracked by https://github.com/jpmorganchase/salt-ds/issues/1567

packages/core/src/radio-button/RadioButton.tsx
```diff
@@ -90,6 +91,24 @@ export const RadioButton = forwardRef<HTMLLabelElement, RadioButtonProps>(
...
+      a11yProps: {
+        "aria-describedby": formFieldDescribedBy,
+        "aria-labelledby": formFieldLabelledBy,
+      } = {},
+      disabled: formFieldDisabled,
+      validationStatus: formFieldValidationStatus,
+    } = useFormFieldProps();
+
+    const disabled = formFieldDisabled ?? disabledProp;
+    const error = formFieldValidationStatus === "error" ?? errorProp;
```
alycrys, line 110: i think this should be done on the RadioButtonGroup. it's unlikely the radio button will be used directly inside the FF

libertymayc: updated

alycrys (approved):

PR #1757, opened 2023-06-02
fix window issue for SSR

el-dav:


james-nash (approved):
Thanks for doing this! I've verified that it solves the Salt site build issues (I know the site build fails in this branch, but that's due to an unrelated issue - with the fix for that applied, it builds fine!) 🎉

joshwooding (approved):

PR #1778, opened 2023-06-05
[ag-grid-theme] Fix cell truncation

### LINKED ISSUE #1766: Salt Ag Grid Theme Missing Text Truncation
### Package name(s)

AG Grid Theme (@salt-ds/ag-grid-theme)

### Package version(s)

v1.1.4

### Description

Text truncation doesn't work in **Salt** AG grid theme, but works in **UITK** version. This is likely due to `display: flex` applied on the cell.

This mostly impacts on right aligned text (mostly numbers) where most significant digits are getting lost.

Report from MTK team.

### Steps to reproduce

Use demo at https://storybook.saltdesignsystem.com/?path=/docs/documentation-data-grid-ag-grid-theme--page
Shrink any column, no text gets automatically truncated.

### Expected behavior

Text should be truncated when column too narrow, just like UITK theme.

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

ZarkoPernar:
Closes #1766 

### Changes
- change to `display: block` since I couldn't figure out how to get text truncation to work with `display: flex` and direct children being a text node.
- Changing the display messed up with the vertical align of the text on focus since the border causes a layout shift. Had to use some calculations to offset the border thickness change.


origami-z:
I see some icon / text flicks on the "Row Grouping" example before/after selecting a row/cell

ZarkoPernar:
> I see some icon / text flicks on the "Row Grouping" / "Drag Row Order" example before/after selecting a row/cell

Good catch! I just pushed a fix for those two

origami-z:
Worth to check what AG grid out of box theme are doing.?

ZarkoPernar:
> Worth to check what AG grid out of box theme are doing.?

The issue is we are changing the border thickness and ag grid is not. If they have 1px border on "bordered" cell, then they have 1px thick focus ring. So our change messes their internal calculations.

packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
```diff
@@ -21,18 +21,48 @@
   .ag-rtl {
     .ag-has-focus .ag-cell-focus:not(.ag-cell-range-selected),
     .ag-context-menu-open .ag-cell-focus:not(.ag-cell-range-selected),
-    .ag-has-focus .ag-full-width-row.ag-row-focus .ag-cell-wrapper.ag-row-group,
     .ag-cell-range-single-cell,
     .ag-cell-range-single-cell.ag-cell-range-handle {
-      border: 2px dotted var(--agGrid-color-focus) !important;
-      padding-left: ag-param(cell-horizontal-padding) - 2px;
-      padding-right: ag-param(cell-horizontal-padding) - 2px;
+      border: var(--agGrid-cursor-borderWidth) var(--agGrid-cursor-borderStyle)
```
libertymayc, line 26: should these tokens have -cursor in them?

ZarkoPernar: Yep, these are the focus styles, and we are using the border as focus ring

packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
```diff
@@ -21,18 +21,48 @@
...
+    .ag-row-focus .ag-cell-focus .ag-cell-wrapper.ag-row-group,
+    .ag-cell-focus .ag-drag-handle.ag-row-drag,
+    .ag-cell-focus
+      .ag-cell-wrapper
+      > :not(.ag-cell-value):not(.ag-group-value) {
+      height: calc(
+        min(
+            var(--ag-internal-calculated-line-height),
+            var(--ag-internal-padded-row-height)
+          ) - 2px
```
alycrys, line 63: var(--agGrid-cursor-borderWidth)?

alycrys:
does the PR description still make sense? 😄

ZarkoPernar:
> does the PR description still make sense? 😄

@alycrys yep, should be. I had to change to display block, and then adjust the line-height/height on focus

packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
ZarkoPernar: Cursor is the ag-grid/data-grid naming for the focus ring

packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
ZarkoPernar: Yep, will update

packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
libertymayc: since we can call it anything, I wonder if its ok to name it something that makes more sense to devs without having to know that convention of ag grid

ZarkoPernar: Ok, you convinced me :D renamed...

libertymayc (approved):

alycrys (approved):

PR #1785, opened 2023-06-06
Fix the duration foundation tokens having invalid values

joshwooding:
The values were missing the units


ImmanuelBaskaran (approved):
Tested within storybook, all works

libertymayc (dismissed):

libertymayc:
may be worth updating duration.stories.mdx although it does say in ms

PR #1797, opened 2023-06-08
Fix homepage card styling regressions on website

james-nash:
The cards on our site's homepage have turned white. This fixes that.

The root cause is as follows:

* The website's `Card` component (which is _not_ the Salt core `Card` at the moment), was using `useTheme()` to determine light/dark mode and then adding extra CSS classes to change the colors
* For some weird reason, that call to `useTheme()` on the homepage was incorrectly reporting `light`, when it was actually meant to be `dark`, so the wrong card styles were applied. (Weirdly, the elements rendered to the DOM _do_ actually set a `data-mode="dark"` attribute, so as far as CSS is concerned, dark mode is in effect)

My fix removes that JS code and instead leans on our CSS custom properties to apply the correct light/dark styling. Our homepage actually uses some bespoke colors, so I'm doing that via the Hompage's CSS.

While this does fix the card styling, there should be further investigation into _why_ the `<SaltProvider>` doesn't seem to be working as expected.

### DIFF
--- a/site/src/_index/accordion/Accordion.module.css
+++ b/site/src/_index/accordion/Accordion.module.css
@@ -14,7 +14,7 @@
 }
 
 .accordionWrapper {
-  background-color: var(--card-background);
+  background-color: #10212b;
   width: min(55vw, var(--max-container-width));
   padding: calc(var(--salt-size-unit) * 10);
   grid-area: 2 / 1 / 3 / 3;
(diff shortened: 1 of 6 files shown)


origami-z (approved):
This fixes the urgent issue on homepage Card styling. So will merge it in.
Need separate investigation on other issues, e.g. viewport / density responsiveness

PR #1814, opened 2023-06-12
Add readonly variant to Checkbox

libertymayc:
Added readonly variant, that can be toggled by setting the `readOnly` prop.
cc @navkaur76


joshwooding:
Replaced by #1958

PR #1815, opened 2023-06-12
Remove error disabled styling variant in Checkbox

libertymayc:
Removes styles of disabled error in Checkbox [RadioButton will follow #1817 ]
Puts disabled > error in logic

### DIFF
--- a/.changeset/khaki-tigers-battle.md
+++ b/.changeset/khaki-tigers-battle.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": minor
 ---
 
-Give priority to disabled styling over error in Checkbox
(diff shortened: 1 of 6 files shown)


.changeset/khaki-tigers-battle.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/core": minor
+---
+
+**_Significant styling change_**
```
libertymayc, line 5: Is this ok for the changeset? The styling is no longer available, but we don't consider that breaking.. @joshwooding

feliciackh (approved):

alycrys (approved):

PR #1833, opened 2023-06-14
Fix pills being shown on incorrect pages

joshwooding:


james-nash (dismissed):
Yup, seems to be working fine now.

Thanks for figuring this out so quickly!

pseys (dismissed):

james-nash (approved):
LGTM

PR #1840, opened 2023-06-15
Fix icons table on the site

### LINKED ISSUE #1441: Scrollable icon table needs keyboard access
(Btw, this bug exists on both the current site and the Mosaic port)

The i[con sizes table](https://www.saltdesignsystem.com/getting-started/designing#understanding-our-icon-sizes) on the "Designing with Salt" page is wrapped in a `<div>` with `overflow-y: auto`, so that it becomes horizontally scrollable on narrow viewports. aXe has flagged that the `<div>` also needs to be keyboard focusable, so that keyboard-only users can then scroll the table via arrow keys.

They have more info here: https://dequeuniversity.com/rules/axe/4.6/scrollable-region-focusable?application=AxeChrome

Which suggests adding `tabindex="0"` to that wrapper `<div>`. I wonder though, whether we'd also need to add a visual focus ring to the div, so that keyboard users can see that it's currently focused. Perhaps @maj-stella can confirm whether or not that's needed.

joshwooding:
Closes #1441

A minor fix to the table, previously a scroll pane was added to scroll the table on small screens but this created more issues around accessibility. 

This PR changes the table to a fixed layout so it scales with screen size

### DIFF
--- a/site/src/components/getting-started/icons-table/IconsTable.module.css
+++ b/site/src/components/getting-started/icons-table/IconsTable.module.css
@@ -2,15 +2,12 @@
   color: var(--site-tertiary-accent-purple);
 }
 
-.iconsTableContainer {
-  overflow-x: auto;
-}
-
 .iconsTableContainer table thead {
   top: 0;
 }
 
 .iconsTableContainer table {
+  table-layout: fixed;
   caption-side: bottom;
 }
 


origami-z (approved):
Tried on desktop up to 400% zoom, and worked

PR #1839, opened 2023-06-15
Add warning to Checkbox and RadioButton

libertymayc:
Added `validationStatus` prop to RadioButton and Checkbox allowing for error and warning status, added `validationStatus` to group context
Deprecated `error` prop in RadioButton and Checkbox

PR #1844, opened 2023-06-16
Fix for deprecated error prop

libertymayc:
Fix for deprecated error prop in CB and RB; still support styling for now
Accidentally removed in #1839 when adding validationStatus


.changeset/nice-roses-fail.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix for deprecated `error` prop on Checkbox and RadioButton: ensures styling still works for deprecated prop.
```
joshwooding, line 5: I don't think we need a changeset. The deprecation hasn't been released yet

joshwooding (changes requested):

joshwooding:
Could you add the reason why this was needed to the PR please

alycrys (approved):

PR #1846, opened 2023-06-19
Fix Data Grid Dropdown showing when there are no list items

Fercas123:
this change avoids showing a dropdown in a cell editor when there is no source or the source is empty


feliciackh (approved):

joshwooding (changes requested):
Needs a changeset

PR #1878, opened 2023-06-20
[FormField v3] Multi child support

libertymayc:
Multi child support for v3


packages/core/src/form-field/FormFieldControlWrapper.css
```diff
@@ -1,7 +1,18 @@
...
+  border: none;
   display: grid;
   grid-auto-columns: minmax(min-content, 1fr);
   grid-auto-flow: column;
   grid-column-gap: var(--salt-size-unit);
+  margin: 0;
+  padding: 0;
+}
+
+.saltFormFieldControlWrapper-secondary .saltInput {
```
alycrys, line 12: would it be better to maybe pass the variant to the input?

libertymayc: yes, updated

alycrys (approved):

PR #1884, opened 2023-06-21
[Form Field v2] Single Checkbox/RadioButton support

libertymayc:
#1750 didn't account for use case of single checkbox/radio button used in FF where context isn't provided by group


alycrys:
should we have some tests along with this?

packages/core/src/checkbox/Checkbox.tsx
```diff
@@ -77,7 +78,19 @@ export interface CheckboxProps
...
+function getValidationStatus(
+  checkboxGroupStatus: CheckboxValidationStatus,
+  formFieldStatus: CheckboxValidationStatus | "success" | undefined,
+  checkboxStatus: CheckboxValidationStatus
+) {
+  return checkboxGroupStatus ?? formFieldStatus
+    ? formFieldStatus !== "success"
+      ? formFieldStatus
+      : undefined
+    : checkboxStatus;
```
alycrys, line 93: Not sure I understand the logic here. Can you please add a comment?
Why do we allow for 'success' status if we're ignoring it?

alycrys (approved):

Fercas123 (approved):

alycrys (approved):

PR #1890, opened 2023-06-22
Fix breakpoints being ignored if they are out of order

Fercas123:
Prevents grid (and probably others that use responsiveProps)from ignoring some breakpoints if they don't come in the right order.

### DIFF
--- a/packages/core/src/stack-layout/StackLayout.tsx
+++ b/packages/core/src/stack-layout/StackLayout.tsx
@@ -68,7 +68,7 @@ export const StackLayout: StackLayoutComponent = forwardRef(
 
     const flexGap = useResponsiveProp(gap, 3);
     const separatorAlignment = separators === true ? "center" : separators;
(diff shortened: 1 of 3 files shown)


JoanaMMoreira (approved):

JoanaMMoreira (approved):

packages/core/src/stack-layout/StackLayout.tsx
```diff
@@ -68,7 +68,7 @@ export const StackLayout: StackLayoutComponent = forwardRef(
 
     const flexGap = useResponsiveProp(gap, 3);
     const separatorAlignment = separators === true ? "center" : separators;
-    const flexDirection = useResponsiveProp(direction, "column");
+    const flexDirection = useResponsiveProp(direction) || "column";
```
joshwooding, line 71: Are the default changes still needed?

joshwooding (approved):

JoanaMMoreira (approved):

Issue #1729, opened 2023-05-30
Toast v1


parsakhan21:
```[tasklist]
### Checklist
- [x] Density (HD, MD, LD, TD)
- [x] Mode (Light, Dark)
- [x] Basic ADA
- [x] Individual toast, static
- [x] iRAG status (Info/success/warning/error)
- [x] Optional close button (default = visible)
```
```[tasklist]
### Deliverables
- [x] Characteristics spec
- [x] Interactions spec, where necessary
- [x] In Code & Figma (Component + Guide)
- [x] Supporting site documentation
```

```[tasklist]
### Features
- [x] Individual toast, static
- [x] iRAG states
- [x] Optional close button (default = visible)
```

navkaur76:
29/06 - Kick-off meeting scheduled for 30/06

parsakhan21:
@navkaur76 would you mind updating the github issue with kick-off notes - what was discussed, scope of work (features to be built, etc), next steps? Thank you. 
cc @AnniaT @joshwooding

navkaur76:
**KICK-OFF MEETING NOTES (03/07):**

@bhoppers2008 @alycrys @joshwooding @dplsek 

Finalised versions scope:

**V1 (delivery deadline - 17th July EOD)**

- Densities
- Modes
- Basic ADA
- Individual toast, static
- iRAG status (Info/success/warning/error)
- Optional close button (default = visible)

**V2**

- Stacking & positioning (flexible placement on interface)
- Entry & exit animation (duration tokens)

**V3**

- (If there's a strong use case) Filled/secondary variant

**V4**

- Additional status options from status indication pattern (e.g. progress, urgency) - not a strong requirement for now as users can refer to the pattern for guidance

**Additional action items/considerations:** 

**Content** @mikearildbrown @sion-docs : documentation should clearly define the difference in usage between banner + toast ([messaging pattern](https://www.figma.com/file/ecxcWvcxVC5BzMFvhiWjq3/UITK-Patterns-%2F-Messaging?type=design&node-id=75%3A480&mode=design&t=OQ9ErE65HzFFAayV-1) exists as reference for distinction). Important that this is covered in docs for V1 Toast. 

**Placement/positioning logic:** we should ideally provide something to users that makes placement easier, especially with stacking. Useful for @alycrys to provide a demo with supporting doc in storybook, but treated as a late V1/early V2 feature. 

[...]

PR #1892, opened 2023-06-23
Toast

alycrys:
Link to issue: https://github.com/jpmorganchase/salt-ds/issues/1729


packages/lab/stories/toast/toast.stories.tsx
```diff
@@ -0,0 +1,67 @@
...
+export const CustomContent = Template.bind({});
+CustomContent.args = {
+  children: (
+    <div>
+      <div>
+        <strong>Lorem ipsum</strong>
+      </div>
+      <div>Lorem ipsum dolor sit amet consectetur adipiscing elit.</div>
+      <div
+        style={{
```
libertymayc, line 53: could use a layout here

libertymayc:
lgtm, needs changeset

libertymayc (approved):

libertymayc (dismissed):
🍞

packages/lab/src/toast/Toast.css
```diff
@@ -0,0 +1,56 @@
...
+.saltToast {
+  --toast-background: var(--salt-container-primary-background);
+
+  background: var(--saltToast-background, var(--toast-background));
+  border-color: var(--saltToast-borderColor, var(--toast-borderColor));
+  border-width: var(--saltToast-borderWidth, var(--salt-size-border));
+  border-style: var(--saltToast-borderStyle, var(--salt-container-borderStyle));
+
+  box-sizing: border-box;
+  box-shadow: var(--salt-overlayable-shadow-popout);
```
libertymayc, line 11: I feel we could take this work as an opportunity to get the new shadow ramp in (see the sheet) similarly done to https://github.com/jpmorganchase/salt-ds/pull/1939

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return (
+    <div
+      className={clsx(withBaseName(), withBaseName(status))}
+      {...restProps}
+      ref={ref}
+      role="alert"
+    >
+      <StatusIndicator status={status} className={withBaseName("icon")} />
+      <div className={clsx(withBaseName("content"))}>{children}</div>
+      {!hideClose && (
```
joshwooding, line 59: This differs from Banner with the Button being built in, is this intentional?

Making it composable would get rid of two props, and allow for more control over the Button and the Icon

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-toast",
+    css: toastCss,
+    window: targetWindow,
+  });
+
+  return (
+    <div
+      className={clsx(withBaseName(), withBaseName(status))}
```
joshwooding, line 52:
```suggestion
      className={clsx(withBaseName(), withBaseName(status), className)}
```

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+  status?: ValidationStatus;
+  /**
+   * Callback function triggered when the close button is actioned.
+   */
+  onClose?: () => void;
+}
+
+export const Toast = forwardRef(function Toast(
+  {
+    children,
```
joshwooding, line 35:
```suggestion
    children,
    className,
```

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+   */
+  hideClose?: boolean;
+  /**
+   *  A string to determine the current state of the Toast.
+   */
+  status?: ValidationStatus;
+  /**
+   * Callback function triggered when the close button is actioned.
+   */
+  onClose?: () => void;
```
joshwooding, line 30: If this prop is used the type is wrong, it should have an event parameter

packages/lab/stories/toast/toast.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+  children: <div>LEI Updated</div>,
+  hideClose: true,
+};
+
+export const CustomContent = Template.bind({});
+CustomContent.args = {
+  children: (
+    <div>
+      <div>
+        <strong>Lorem ipsum</strong>
```
joshwooding, line 49: Is there any copy we can use here instead of lorem ipsum?

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+}
+
+export const Toast = forwardRef(function Toast(
+  {
+    children,
+    hideClose = false,
+    onClose,
+    status = "info",
+    ...restProps
+  }: ToastProps,
```
joshwooding, line 40: More of a nitpick but these can be passed as generic to forwardRef

joshwooding (changes requested):

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+    <div
+      className={clsx(withBaseName(), withBaseName(status))}
+      {...restProps}
+      ref={ref}
+      role="alert"
+    >
+      <StatusIndicator status={status} className={withBaseName("icon")} />
+      <div className={clsx(withBaseName("content"))}>{children}</div>
+      {!hideClose && (
+        <div className={clsx(withBaseName("close"))}>
```
Fercas123, line 60: is the wrapper needed? you can probably pass the same styles straight into the button

packages/lab/src/toast/Toast.tsx
alycrys: I've started with the assumption that the close button will always be displayed. My bad, will change to composable

alycrys: Optional close button (default = visible).
I'll go back to Nav about this, as there will be no default

packages/lab/stories/toast/toast.stories.tsx
alycrys: will check with Nav

libertymayc (approved):

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+    css: toastCss,
+    window: targetWindow,
+  });
+
+  return (
+    <div
+      className={clsx(withBaseName(), withBaseName(status), className)}
+      {...rest}
+      ref={ref}
+      role="alert"
```
Fercas123, line 36:
```suggestion
      className={clsx(withBaseName(), withBaseName(status), className)}
      role="alert"
      {...rest}
      ref={ref}
```

packages/lab/src/toast/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { makePrefixer, StatusIndicator, ValidationStatus } from "@salt-ds/core";
+import { useWindow } from "@salt-ds/window";
+import { useComponentCssInjection } from "@salt-ds/styles";
+
+import toastCss from "./Toast.css";
+
+const withBaseName = makePrefixer("saltToast");
+
+export interface ToastProps extends ComponentPropsWithoutRef<"div"> {
+  children?: ReactNode;
```
Fercas123, line 12: this should not be needed here

Fercas123 (dismissed):

packages/lab/src/toast/ToastClose.css
```diff
@@ -0,0 +1,8 @@
+/* Styles applied to ToastClose button */
+.saltButton.saltToastClose {
+  margin-right: calc(var(--salt-spacing-50) * -1);
```
joshwooding, line 3: Do we need to use a negative margin here? They tend to cause more issues than they solve

alycrys: Hopefully not. There are still some design discussions taking place and there will be changes. I'll fix in the next PR if you're happy with the API on this one

joshwooding: Will those design conversations also affect whether ToastClose stays?

alycrys: I don't think so. Close button is still optional, don't think there's any intention to change it. There was no mention about it.

packages/lab/src/toast/ToastClose.css
```diff
@@ -0,0 +1,7 @@
+/* Styles applied to ToastClose button */
+.saltButton.saltToastClose {
```
ZarkoPernar, line 2: I see no visual change when I toggle this class in the dev tools. Like you said in the meeting today, I feel like you could completely remove this component.

ZarkoPernar (approved):
The single line Toast looks funky indeed 🤪

libertymayc (approved):

PR #1895, opened 2023-06-23
Forms pattern page

joshwooding:


site/docs/patterns/index.mdx
```diff
@@ -0,0 +1,524 @@
...
+
+<ExampleContainer type="positive">
+- Align the width of a Form Field to the number of columns that best reflect the length of its expected text.
+- Use the Figma "Salt Layout Grid" style library, which aligns to Salt Breakpoints. Simply pick the style that fits to your density and frame width.
+- Use "Compact layout" for very data-dense forms.
+</ExampleContainer>
+
+## Form container
+
+<Diagrams>
```
james-nash, line 35: Do we really need new `<Diagram(s)>` components for this?

Could we not achieve the same thing with existing stuff. For example:

```tsx
<GridLayout ...>
  <GridItem>

    ### Full page

    Embed your form...

    <Image ... />

  <GridItem>
 
  <GridItem>
    ...
  </GridItem>
</GridLayout>
```

That way we have less site-specific stuff to maintain going forwards.

site/src/components/image-switcher/ImageSwitcher.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const [toggle, setToggle] = useState(false);
+
+  const handleToggle = () => {
+    setToggle((old) => !old);
+  };
+
+  const currentImage = toggle ? images[1] : images[0];
+
+  return (
+    <div className={styles.container}>
```
james-nash, line 22: I wonder if we could use the (neutral) `<ExampleContainer>` here since it looks like image switcher uses the exact same border & padding. I.e.

```tsx
<ExampleContainer type="neutral">
  <div className={styles.header}>
  ...
</ExampleContainer>
```

Arguably, the image switcher is a specialised kind of example, so doing it that way also guarantees that `<ExampleContainer>` and `<ImageSwitcher>` will always share the same container styles, in case we ever change them in the future.

site/src/components/image-switcher/ImageSwitcher.module.css
```diff
@@ -0,0 +1,32 @@
...
+
+@media screen and (max-width: 996px) {
+  .container {
+    padding: var(--salt-spacing-100);
+    gap: var(--salt-spacing-200);
+  }
+}
+
+.container .switch {
+  --saltSwitch-height: var(--salt-switch-height);
```
libertymayc, line 18: `--salt-switch-height` doesn't exist

site/docs/patterns/index.mdx
```diff
@@ -0,0 +1,524 @@
...
+
+<ImageSwitcher
+  images={[
+    { src: "/img/patterns/forms/form_layout_1.png", alt: "" },
+    { src: "/img/patterns/forms/form_layout_2.png", alt: "" },
+  ]}
+  label="Show spacing"
+/>
+
+\*You can find more information on the Salt spacing system here.
```
libertymayc, line 25: is this supposed to be a link? :)

site/docs/patterns/index.mdx
```diff
@@ -0,0 +1,524 @@
...
+    { src: "/img/patterns/forms/heading_1.png", alt: "" },
+    { src: "/img/patterns/forms/heading_2.png", alt: "" },
+  ]}
+  label="Show spacing"
+/>
+
+<ExampleContainer type="negative">
+  Space above headings is optional. It shouldn’t be added when:
+  - The heading is at the beginning of the form.
+  - The heading follows a separator. - Using a “Compact form” layout.
```
libertymayc, line 243:
```suggestion
  - The heading follows a separator. 
  - Using a “Compact form” layout.
```

libertymayc:
looking good :)

site/docs/patterns/index.mdx
joshwooding: I still need to go through and add the links everywhere :)

site/docs/patterns/index.mdx
joshwooding: Thanks, there's a weird formatting issue that when prettier runs it messes up all the mdx

site/docs/patterns/index.mdx
```diff
@@ -0,0 +1,524 @@
...
+    - Is 'non-specialist' or infrequently used.
+    - Isn't space constrained/doesn't need to be entirely visible.
+    - Has minimal sections/columns.
+  </Diagram>
+  <Diagram
+    src="/img/patterns/forms/field_background_2.png"
+    alt="Example showing secondary field on secondary background"
+    contentPosition="top"
+  >
+    ### 1b Secondary field on secondary background
```
ZarkoPernar, line 98: Shouldn't this be primary field on secondary background?

site/src/components/image-switcher/ImageSwitcher.module.css
joshwooding: It's not defined here, this is to override a change in Mosaic

site/docs/patterns/index.mdx
joshwooding: Updated thanks

site/docs/patterns/index.mdx
joshwooding: This would need a lot of custom CSS to work which is what I was trying to avoid. I think for now getting the patterns up is more important than trying to be DRY

site/src/components/image-switcher/ImageSwitcher.tsx
joshwooding: I don't really see the need for this but if necessary it can be done at a later date.

bhoppers2008 (approved):

bhoppers2008 (approved):

ZarkoPernar (approved):

PR #1905, opened 2023-06-26
Update Toggle Button and Button to have a fixed icon gap

joshwooding:
A requirement from design where Button and Toggle Button need to have a fixed gap of spacing-50 between the icon and text

### DIFF
--- a/packages/core/src/button/Button.css
+++ b/packages/core/src/button/Button.css
@@ -94,6 +94,7 @@
   color: var(--saltButton-text-color, var(--button-text-color));
   cursor: var(--saltButton-cursor, pointer);
   display: inline-flex;
+  gap: var(--salt-spacing-50);
   justify-content: var(--saltButton-justifyContent, center);
   font-size: var(--saltButton-fontSize, var(--salt-text-fontSize));
   font-family: var(--saltButton-fontFamily, var(--salt-text-fontFamily));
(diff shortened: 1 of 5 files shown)


libertymayc (approved):
loving the long branch name

alycrys (dismissed):

packages/core/src/button/Button.css
```diff
@@ -115,7 +116,6 @@
   -webkit-tap-highlight-color: transparent;
   /* Styles applied to align children*/
   font-weight: var(--saltButton-fontWeight, var(--button-fontWeight));
-  white-space: pre;
```
joshwooding, line 118: Removing this means users don't need to remove the space between the icon and text.

packages/lab/src/toolbar/ToolbarButton.css
```diff
@@ -5,4 +5,5 @@
 .saltToolbarButton:not([data-is-inside-panel]) {
   --saltButton-fontSize: 0;
   --saltButton-letterSpacing: 0;
+  gap: 0;
```
joshwooding, line 8: This is needed for when the button is collapsed and the text is hidden

alycrys (dismissed):

packages/lab/src/toolbar/ToolbarButton.css
alycrys: would this need a changeset?

joshwooding: Not really, it's required due to the Button changes so that ToolbarButton doesn't break in its collapsed state.

alycrys (approved):

PR #1907, opened 2023-06-27
Adds component page MDX template

james-nash:
* Adds template MDX files that can be copied and filled in when adding new component pages to the site
* Aligns the frontmatter with the current Word content template
* Simplifies some of the `$ref` stuff in the per-tab MDX files so that there's less stuff to fill in for new components


site/docs/components/button/usage.mdx
```diff
@@ -40,8 +25,4 @@ import { Button } from "@salt-ds/core";
 
 ### Props
 
-| Name                  | Type      | Description                                                                                          | Default   |
-| --------------------- | --------- | ---------------------------------------------------------------------------------------------------- | --------- |
-| disabled              | `boolean` | If `true``, the button will be disabled.                                                             | -         |
-| focusableWhenDisabled | `boolean` | If `true``, the button will be fosusable when disabled.                                              | -         |
-| variant               | -         | The variant to use. “Primary” is the default value, and the other options are “secondary” and “cta.” | "primary" |
+:propsTable{src="../packages/core/src/button/Button.tsx"}
```
JoanaMMoreira, line 28: I'm not sure it's worth adding this as it may change slightly if we don't use the Mosaic plugin (it will be probably something like `<PropsTable componentName="Button" />`)

templates/component-pages/component-name/usage.mdx
```diff
@@ -0,0 +1,32 @@
...
+sidebar:
+  label: Usage
+data:
+  $ref: ./#/data
+---
+
+### Using the Component Name component
+
+Usage information for this component. Copy the content from the content
+temoplate's "How to use" section here.
```
JoanaMMoreira, line 15:
```suggestion
template's "How to use" section here.
```

JoanaMMoreira (approved):

JoanaMMoreira:
🎉

site/docs/components/button/usage.mdx
james-nash: How about I keep it in as a commented out line for now? That way we have the full path in case we need it. Once we know how we're going to do props tables, we can amend this template accordingly.

JoanaMMoreira: Sounds good!

james-nash: Actually, assuming that your props table PR #1911 to gets merged soon, I may as well update this to be:

```suggestion
<PropsTable componentName="ComponentName" />
```

templates/component-pages/component-name/usage.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+```
+import { ComponentName } from "@salt-ds/core";
+```
+
+### Props
+
+{/* Update componentName below as needed */}
+
+<PropsTable componentName="ComponentName" />
```
JoanaMMoreira, line 31: Should we also add a note about the optional `packageName` prop for non-core components?

james-nash: Yeah good idea. Just did it

templates/component-pages/README.md
```diff
@@ -0,0 +1,8 @@
+# How to use the component page MDX template
+
+To add a new component page to the Salt website:
+
+1. Copy the [`component-name/`](./component-name/) folder into the [`site/docs/components/` directory](../../site/docs/components/) and rename it to your component's name in kebab case. For example `form-field/`.
+   - 💡 **Tip:** If you are using GitHub's website to edit component docs, you can press the <kbd>.</kbd> key to launch a version of VS Code in your browser and copy the directory via that.
+1. Fill in the `index.mdx` file's frontmatter (refer to the comments within that file for guidance)
```
feliciackh, line 7: Are these meant to be numbered 2. and 3. respectively? (Can't seem to find the link to preview on site)

james-nash: This README doesn't appear within our site. It just for when folks are browsing our repo.

When markdown is rendered to HTML, the numbers get replaced with the correct ones automatically: https://github.com/jpmorganchase/salt-ds/tree/add-component-template/templates/component-pages

feliciackh (approved):

PR #1908, opened 2023-06-27
Form Field Group [Pattern] using layouts and shared props

libertymayc:
~Second implementation after discussion on 26/6 regarding Form Field Group~

~This uses Stack Layout to show how a group could easily be made with our layout components pretty cleanly
Uses Form Field Group context as a wrapper to specify label width, placement, and variant~

3rd implementation:

Added stories to show Form Field Group as a pattern, with usage of **Stack and Grid Layout components and shared props** 

Added "right" as an option to `FormFieldLabelPlacement` type used in FormField `labelPlacement` prop


packages/core/stories/form-field/form-field.stories.tsx
```diff
@@ -511,3 +515,281 @@ export const WithInputWithAdornments: ComponentStory<typeof FormField> = (
...
+        labelPlacement: "right"
+      }}
+    >
+    <StackLayout style={{"--formFieldLabel-width": "100px"} as React.CSSProperties}>
+        <FormField {...props}>
+          <FormLabel>Form Field label left</FormLabel>
+          <Input defaultValue="Value" />
+          <FormHelperText>Helper text</FormHelperText>
+        </FormField>
+        <StackLayout direction={"row"}>
```
Fercas123, line 680: Whenever you are adding directions props, you could swap them for `direction={{xs: "column", sm: "row"}}` so they don't squish on resize, and your examples become responsive

libertymayc: fab! thank you

origami-z:
(I talked to Liberty earlier) If this is more of a pattern thing and there is a way we can do without introducing additional context / dependency of code (e.g. `const sharedProp = {position: 'left'}; <FormField1 {...sharedProp} /> <FormField2 {...sharedProp}/>`) and not too much more code consumer going to write, then let's not introducing them.

When we get to look at app wide config / pattern, we *may* be able to do something at the salt provider level by then.

el-dav:
Looks good to me other @origami-z 's comment around whether we need to provide this vs leaving it to consumers

DavieReid:
The legacy FormField (when label was at the top) used to align the label, value and helpertext.  I haven't seen the new design spec but wanted to check that this isnt still a requirement as in the examples added in this PR the value is not in alignment

packages/core/src/form-field/FormField.tsx
```diff
@@ -1,14 +1,13 @@
...
 import { A11yValueProps, FormFieldContext } from "../form-field-context";
 import { makePrefixer, useId, capitalize } from "../utils";
 
-import { useWindow } from "@salt-ds/window";
-import { useComponentCssInjection } from "@salt-ds/styles";
-
 import formFieldCss from "./FormField.css";
 
-export type FormFieldLabelPlacement = "top" | "left";
+export type FormFieldLabelPlacement = "top" | "left" | "right" | undefined;
```
DavieReid, line 10: `undefined` shouldn't be here as it is not a valid label placement.  It should be specified by the *prop* which has the type `FormFieldLabelPlacement`

packages/core/src/input/Input.tsx
```diff
@@ -107,15 +107,15 @@ export const Input = forwardRef<HTMLDivElement, InputProps>(function Input(
...
+  const controlWrapper = useControlWrapper();
 
   const restA11yProps = {
     "aria-activedescendant": ariaActiveDescendant,
     "aria-expanded": ariaExpanded,
     "aria-owns": ariaOwns,
   };
 
-  const variant = controlWrapperVariant ?? variantProp;
+  const variant = controlWrapper?.variant ?? variantProp;
```
DavieReid, line 118: The wrapper takes precedence over the prop someone has added to the input?  That's not the case for `disabled` or `readOnly` props.

packages/core/stories/form-field/form-field.stories.tsx
```diff
@@ -511,3 +516,253 @@ export const WithInputWithAdornments: ComponentStory<typeof FormField> = (
...
+        <FormLabel>Form Field label</FormLabel>
+        <Input defaultValue="Value" />
+        <FormHelperText>Helper text</FormHelperText>
+      </FormField>
+      <FormField {...props}>
+        <FormLabel>Form Field label</FormLabel>
+        <Input defaultValue="Primary Input value" />
+      </FormField>
+    </StackLayout>
+  );
```
DavieReid, line 547: Nice 👍🏻

packages/core/stories/form-field/form-field.stories.tsx
```diff
@@ -511,3 +516,253 @@ export const WithInputWithAdornments: ComponentStory<typeof FormField> = (
...
+};
+
+export const GroupedWithLabelRight: ComponentStory<typeof FormField> = (
+  props
+) => {
+  const groupedProps = { labelPlacement: "right" as FormFieldLabelPlacement };
```
DavieReid, line 584: Dislike the use of `as`.  Is it even needed?

packages/core/src/input/Input.tsx
libertymayc: the wrapper is just for visual purposes (layout, variant). potentially we could add disabled and readonly to it to disable sections at a time, but i feel like that would add more into the context than intended

packages/core/stories/form-field/form-field.stories.tsx
libertymayc: same, it wouldn't let me put it as it said string cant be applied to FormFieldLabelPlacement type. will try to adjust

libertymayc: other than making an enum which doesn't feel right for this or putting the whole obj as `Partial<FormFieldProps>` I can't think of an alternative, always requires casting it from a string.. am i missing something obvious?

DavieReid:
```
import { type FormFieldLabelPlacement } from 'somewhere';

const groupedProps:{labelPlacement: FormFieldLabelPlacement} = { labelPlacement: "right"  };
```

maybe?

packages/core/src/input/Input.tsx
libertymayc: thats interesting, tbh I never considered it that way round but I can see why it might be the case. I assumed wrapper > individual because you wouldn't want a random primary within a secondary wrapper if you're trying to force them all to be a certain variant. I'll raise this in FED meet up tomorrow if there's time and see what people think

+ to be honest, control wrapper could be changed to a stack layout potentially, for the layout, and we have something like a variant context instead..

ImmanuelBaskaran:
Aside from what David has hilighted, everything else looks good

Fercas123 (approved):
LGTM

alycrys (approved):

PR #1911, opened 2023-06-27
Add props table component to site docs

JoanaMMoreira:
Temporarily add the props table manually until we fix/refactor the Mosaic plugin.

Preview at https://saltdesignsystem-git-add-props-table-v3-joshwooding.vercel.app/salt/components/button/usage


site/package.json
```diff
@@ -13,11 +13,12 @@
     "deploy": "yarn mosaic deploy --out snapshots --name latest --config mosaic.config.js",
     "dev": "next dev",
     "start": "next start",
-    "gen:snapshot": "yarn mosaic build --out snapshots --name latest --config mosaic.config.js && yarn copy:sitemap",
+    "gen:snapshot": "yarn mosaic build --out snapshots --name latest --config mosaic.config.js && yarn copy:sitemap && yarn gen:props",
```
james-nash, line 16: Is it correct to run `gen:props` _after_ the Mosaic site build? Shouldn't the props JSON files be generated/updated _before_ we build the site?

site/src/props/core-props.json
james-nash, line 1: Might it be better to exclude these JSON files from git and instead force them to be freshly generated when building or developing the site? Similar to what we already do with the snapshots.

Could avoid confusion around stale props data appearing on pages.

Since `yarn build` already runs `yarn gen:props`, can we not just do the same for `yarn serve` too? I.e.:

```diff
-"serve": "concurrently --kill-others 'yarn dev' 'yarn mosaic serve -c ./mosaic.config.js' -p 8080",
+"serve": "yarn gen:props && concurrently --kill-others 'yarn dev' 'yarn mosaic serve -c ./mosaic.config.js' -p 8080",
```

james-nash (commented):
Mostly looks fine. Just have a couple of queries...

site/src/components/components/PropsTable.module.css
```diff
@@ -0,0 +1,3 @@
+.tableWrapper td > * {
+  margin-bottom: 0;
```
origami-z, line 2: Is this to compensate `p` in a cell? I wonder whether `p` should be used ..?

Or if there is genuine need for multi-paragraph support, should margin be removed?

JoanaMMoreira: That's a good point, the `p` is actually added by `ReactMarkdown` and I can replace it with something else like a `span` but then it wouldn't support multi paragraph very well so I'll just restore the margin

site/propsGen.js
```diff
@@ -0,0 +1,32 @@
...
+const components = componentFiles.map((component) =>
+  docgen.parse(component, options)
+);
+
+const filePaths = componentPackages.map(
+  (package) => `./src/props/${package}-props.json`
+);
+
+components.forEach((component, index) =>
+  writeFile(filePaths[index], JSON.stringify(component, null, 2), (error) => {
```
james-nash, line 25: This currently fails because the `props/` dir does not exist in freshly checked out git repo. (it probably works fine on your machine as you probably added that dir manually at some point while working on this)

I suggest you...

* Either check a dummy file into that `props/` dir into git, so that it's guaranteed to exist. E.g. an empty file like `props/.gitkeep`, or maybe put a `README.md` in there that explains what the purpose of that dir is.
* Or add some logic here into `propGen.js` to check if the dir exists and, if not, create it first.

james-nash (changes requested):

site/propsGen.js
JoanaMMoreira: Good spot, I've added a `props` dir with a `README` file

james-nash (approved):
Nice one! 🙌

alycrys (approved):

origami-z (approved):

PR #1913, opened 2023-06-27
Allow site component examples title to be customised

JoanaMMoreira:
- Add optional `title` prop to `LivePreview` component to allow users to customise the title that is displayed for each example. This will default to the example file name if not provided and will also allow for a cleaner `examples.mdx`
- Add missing `value` to mode toggle button
- Amend mobile drawer transition speed
- Preview at https://saltdesignsystem-git-live-preview-tweaks-joshwooding.vercel.app/salt/components/button/examples

Screenshot 2023-06-27 at 16 43 30
Screenshot 2023-06-27 at 16 43 39

### DIFF
--- a/site/src/components/components/ExamplesListView.tsx
+++ b/site/src/components/components/ExamplesListView.tsx
@@ -6,17 +6,16 @@ import {
   Dropdown,
 } from "@salt-ds/lab";
 import useIsMobileView from "../../utils/useIsMobileView";
+import { Heading3 } from "../mdx/h3";
+import { exampleNameRegex } from "./LivePreview";
 
 import styles from "./ExamplesListView.module.css";
-
-const exampleNameRegex = /[A-Z]?[a-z]+|[0-9]+|[A-Z]+(?![a-z])/g;
 
 type ExamplesListViewProps = { examples: ReactElement[] };
 
 const ExamplesListView: FC<ExamplesListViewProps> = ({ examples }) => {
-  const examplesList: string[] = Children.map(
-    examples,
-    ({ props }) => props.exampleName
+  const examplesList: string[] = Children.map(examples, ({ props }) =>
+    props.title ? props.title : props.exampleName
   );
 
   const [selectedItem, setSelectedItem] = useState<string | null>(
@@ -36,7 +35,7 @@ const ExamplesListView: FC<ExamplesListViewProps> = ({ examples }) => {
     examplesArray[0];
 
   const {
-    props: { children: exampleCopy, ...restProps },
+    props: { children: exampleCopy, title, exampleName, ...restProps },
     ...rest
   } = selectedExample;
 
@@ -67,11 +66,17 @@ const ExamplesListView: FC<ExamplesListViewProps> = ({ examples }) => {
     </div>
   );
 
-  const componentExample = { props: { list, ...restProps }, ...rest };
+  const componentExample = {
+    props: { list, exampleName, ...restProps },
+    ...rest,
+  };
 
   return (
     <>
       {componentExample}
+      <Heading3>
+        {title ? title : exampleName.match(exampleNameRegex)?.join(" ")}
+      </Heading3>
       {exampleCopy}
     </>
   );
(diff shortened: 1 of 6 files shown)


site/src/layouts/DetailComponent/MobileDrawer.module.css
```diff
@@ -7,7 +7,7 @@
   margin-top: calc(-1 * calc(var(--salt-size-unit) * 2));
   box-shadow: var(--salt-overlayable-shadow-modal);
   padding-left: calc(var(--salt-size-unit) * 2);
-  transition: max-height 1s ease-in-out;
+  transition: max-height 0.6s ease-in-out;
```
origami-z, line 10: `--salt-duration-*` variables could help?

site/src/components/components/ExamplesListView.tsx
```diff
@@ -67,11 +66,17 @@ const ExamplesListView: FC<ExamplesListViewProps> = ({ examples }) => {
...
+  const componentExample = {
+    props: { list, exampleName, ...restProps },
+    ...rest,
+  };
 
   return (
     <>
       {componentExample}
+      <Heading3>
+        {title ? title : exampleName.match(exampleNameRegex)?.join(" ")}
```
origami-z, line 78: This may end up as empty `h3`?

james-nash (approved):
Great!

site/src/layouts/DetailComponent/MobileDrawer.module.css
james-nash: I was wondering the same thing, but [the documentation](https://storybook.saltdesignsystem.com/?path=/docs/theme-foundations-duration--page) states that they are "duration between the moment a user performs an action and the moment feedback is given by the system".

Since this is the duration of an animation, not the delay before something happens on-screen, I'm not sure they're appropriate.

(Aside: This does make me wonder whether there should be tokens whose intent is how long an animation runs for. Thoughts, @pseys & @dplsek?)

james-nash: (Also: Is it OK for a component to reference _foundation_ tokens? I've always assumed those are kind of private to the theme. Is that not the case?)

libertymayc: cc @navkaur76

libertymayc: > (Also: Is it OK for a component to reference foundation tokens? I've always assumed those are kind of private to the theme. Is that not the case?)

Yes, foundations are part of the theme and can be used directly, with the exception of color [and the intermediate palette] which are given meaning within characteristics
see tooling/stylelint/correct-theme-token-usage/index.js

james-nash: Ah cool. Thanks for clarifying that @libertymayc

site/src/components/components/ExamplesListView.tsx
JoanaMMoreira: Good spot! Should be fixed

libertymayc (approved):

site/src/layouts/DetailComponent/MobileDrawer.module.css
JoanaMMoreira: I've replaced the animation with `--salt-duration-perceptible`, let me know how that looks :)

ZarkoPernar (approved):

libertymayc (approved):

origami-z (approved):

james-nash (approved):

site/src/layouts/DetailComponent/MobileDrawer.module.css
navkaur76: > I was wondering the same thing, but [the documentation](https://storybook.saltdesignsystem.com/?path=/docs/theme-foundations-duration--page) states that they are "duration between the moment a user performs an action and the moment feedback is given by the system".
> 
> Since this is the duration of an animation, not the delay before something happens on-screen, I'm not sure they're appropriate.
> 
> (Aside: This does make me wonder whether there should be tokens whose intent is how long an animation runs for. Thoughts, @pseys & @dplsek?)

Paul may be able to better assist on this, but generally I think the plan is to treat animation separately from duration - it would instead fall under 'Motion' which is a foundation yet to be created. However, if a duration token is needed as a temporary solution for the drawer, then using duration-perceptible as Jo has applied makes sense. Out of the existing tokens, it feels the most acceptable timing for the drawer to expand/collapse upon interaction.

PR #1918, opened 2023-06-29
Change replacement token in docs to hex values

libertymayc:
Uses the previous hex value rather than N/A


origami-z (approved):
Is there anything we need to modify / update within past changelog / changeset?

alycrys (approved):

origami-z (approved):

PR #1921, opened 2023-06-30
Add TabstripNext and TabNext

joshwooding:


packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,93 @@
...
+    onClick?.(event);
+  };
+
+  const handleFocus = (event: FocusEvent<HTMLButtonElement>) => {
+    focus(value);
+    onFocus?.(event);
+  };
+
+  useEffect(() => {
+    registerTab({ value, label: children });
```
ZarkoPernar, line 68: If the user passes jsx as children this will cause infinite loop, no? Why not store the references in a ref? Besides, why is this register/unregister even necessary? Can't we map/filter the children?

joshwooding: Mapping/filtering the children is brittle. Have you tested the infinite loop theory? I tested the implementation and I didn't see any issues

packages/lab/src/tabs-next/TabNext.css
```diff
@@ -0,0 +1,72 @@
...
+.saltTabNext {
+  align-items: center;
+  justify-content: center;
+  appearance: none;
+  display: inline-flex;
+  background: var(--salt-navigable-primary-background);
+  border: 0 solid transparent;
+  border-radius: 0;
+  white-space: pre;
+  min-width: 4em;
```
alycrys, line 11: what's the rule on magic numbers?

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -0,0 +1,93 @@
...
+    isSelected,
+    select,
+    isFocused,
+    focus,
+    disabled: tabstripDisabled,
+    unregisterTab,
+    registerTab,
+  } = useTabs();
+  const selected = isSelected(value);
+  const focusable = isFocused(value);
```
alycrys, line 54: `focusable` or `focused`?

packages/lab/src/tabs-next/TabNext.css
joshwooding: I think this one is fine. It's a shared value between Input and Tabs

packages/lab/src/tabs-next/TabNext.tsx
joshwooding: It's pretty minor but I'll update the wording to be consistent

joshwooding: I've updated it

alycrys (approved):

navkaur76 (approved):

PR #1922, opened 2023-06-30
Remove browser versions from README.md

origami-z:
Firefox version is not an accurate description on what's supported by our code, e.g. css gap, focus visible.

It's better we put it back once true version support is defined by product

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,13 +73,6 @@ browsers, outlined in the table below.
 |                  | Chrome      | JAWS          |
 | macOS            | Safari      | VoiceOver     |
 
-## Browser compatibility
-
-| Web Browser  | Version         |
-| ------------ | --------------- |
-| Firefox      | 38+             |
-| Chrome, Edge | last 2 versions |
-
 ## Thanks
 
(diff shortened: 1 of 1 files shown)


joshwooding (approved):

ZarkoPernar:
@origami-z compatibility is whatever is defined by the code, not by product? I don't think anyone suggested we are going to replace all usages of flexbox gap. I think we should be clear what we support **now**, rather than what we want to support.

libertymayc (approved):

origami-z:
> @origami-z compatibility is whatever is defined by the code, not by product? I don't think anyone suggested we are going to replace all usages of flexbox gap. I think we should be clear what we support **now**, rather than what we want to support.

There will be some unique business requirement (in our industry) which we need to meet, e.g. ever green browser support won't be possible. Those will ultimately come from product.

I agree we need to find out what we support now (which would help PO to decide), but unless we put in guard (e.g. linting) to prevent using 'newer' features in our code, "current" state will be a moving target.

Issue #1664, opened 2023-05-15
Stepped Tracker v1


joshwooding:
```[tasklist]
### Checklist
- [ ] Density (HD, MD, LD, TD)
- [ ] Mode (Light, Dark)
- [ ] Fill (Primary, Secondary, Tertiary)
- [ ] State (Default, Hover, Selected/Active, Disabled, Read-only, Other)
- [ ] Controlled API
- [ ] UITK accessibility parity (screen reader + keyboard interactions)
```
```[tasklist]
### Deliverables
- [x] Characteristics spec
- [x] Interactions spec, where necessary
- [ ] In Code & Figma (Component + Guide)
- [ ] Supporting site documentation
```

```[tasklist]
### Features
- [x] Horizontal oreintation
- [x] Basic stepped tracker (Default, In progress, Complete)
- [x] Label truncation + tooltip support
```
*Dependency*:
- Color update requires basic validation
- Progress Indicators

parsakhan21:
June 30 - kick off 6/30

el-dav:
@joshwooding @bhoppers2008 @ivan-calderon 

Could I get clarification on something. Is it possible to move onto a step without the previous one being completed. E.g. something like: (complete, incomplete, active, incomplete)?

✓-----ο-----●-----ο

I could see this being useful if some steps are skippable, or if you're able to fill out the steps out of order, but I couldn't see this behaviour in the UITK examples

bhoppers2008:
> @joshwooding @bhoppers2008 @ivan-calderon
> 
> Could I get clarification on something. Is it possible to move onto a step without the previous one being completed. E.g. something like: (complete, incomplete, active, incomplete)?
> 
> ✓-----ο-----●-----ο
> 
> I could see this being useful if some steps are skippable, or if you're able to fill out the steps out of order, but I couldn't see this behaviour in the UITK examples

@el-dav would suggest pushing to a later version... my rationale is that navigating forward and backward through non-consecutive steps is probably something worth supporting but will require extra work (both dev and design). I'd opt to keep as a strict linear flow for the time being and enhance it with non-consecutive navigation at a later date.

el-dav:
For me I'd ideally want it to be considered earlier as currently the UITK implementation can be a bit confusing.

For example:

```jsx
<SteppedTracker activeStep={2}>
  <TrackerStep completed={false} />
  <TrackerStep  />
  <TrackerStep  />
</SteppedTracker
```

[...]

PR #1959, opened 2023-07-11
Stepped tracker

el-dav:
PR for [v1 of the SteppedTracker component](https://github.com/jpmorganchase/salt-ds/issues/1664) which includes the following features:

- Horizontal orientation
- Basic stepped tracker (Default, In progress, Complete)
- Label truncation + tooltip support

Should be pretty straightforward. The more interesting part is detecting text truncation in the steps and showing Tooltips in this case (see [useDetectTruncatedText](https://github.com/jpmorganchase/salt-ds/pull/1959/files#diff-4ab7ab333221599eba2c9543533e91b8d8a6bee46e22d9726cf233c3af069d88)).

A couple of points for discussion:
- Later versions of SteppedTracker will also need to support a `message` which is an additional level of text below the label. I believe this will require us to wrap each text element to to enable truncation detection.   
  - This will probably require an API change I'm thinking either:

```jsx
 <SteppedTracker activeStep={0}>
  <TrackerStep>
    <StepLabel>Step 1</StepLabel>
    <StepMessage>Additional Message</StepMessage>
  </TrackerStep>
 <TrackerStep>
    <StepLabel>Step 2</StepLabel>
    <StepMessage>Additional Message</StepMessage>
  </TrackerStep>
</SteppedTracker>
```

or

```jsx
<SteppedTracker activeStep={0}>
  <TrackerStep message="additional message">Step 1</TrackerStep>
  <TrackerStep message="additional message">Step 2</TrackerStep>
</SteppedTracker>
```
  - Thoughts welcome, since this is going into labs initially, hopefully it shouldn't block the PR


packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const [hasTooltips, setHasTooltips] = useState(false);
+
+    // A factory function used to get a callback ref for checking truncation.
+    const getOverflowRef = useDetectTruncatedText(setHasTooltips);
+
+    const totalSteps = Children.count(children);
+
+    return (
+      <ul {...restProps} className={clsx(withBaseName(), className)} ref={ref}>
+        {Children.map(children, (child, i) => {
```
el-dav, line 77: I've used `Children.map` and `cloneElement` to inject the index for each child. As discussed recently we could do this in other ways. @joshwooding would be good to get your input

el-dav: Another way I considered was using Children.map and a context e.g.
```jsx
{Children.map(children, (child, i) => {
  <TrackerStepContext
    activeStep: activeStep,
    overflowRef: getOverflowRef(i),
    hasTooltip: hasTooltips,
    totalSteps: totalSteps,
    stepNumber: i,
  />
    {child}
  <TrackerStepContext>
}
```

packages/lab/src/stepped-tracker/TrackerStep/TrackerStep.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+    const isActive = activeStep === stepNumber;
+    const Icon = getStateIcon({ isActive, state });
+    const resolvedState = getState({ isActive, state });
+    const connectorState = activeStep > stepNumber ? "active" : "default";
+    const hasConnector = stepNumber < totalSteps - 1;
+
+    const Inner = (
+      <li
+        {...restProps}
```
Fercas123, line 145: should these move a bit below so they can be overriden?

packages/lab/src/stepped-tracker/SteppedTracker.css
```diff
@@ -0,0 +1,11 @@
+.saltSteppedTracker {
+  margin: 0;
+  padding: 0;
+  text-indent: 0;
+  list-style-type: 0;
```
Fercas123, line 5: this one gets overriden in the next line

Fercas123 (commented):
Looks good, just tiny comments

packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -0,0 +1,92 @@
...
+          if (!isValidElement(child)) {
+            return child;
+          }
+
+          return cloneElement(child, {
+            _activeStep: activeStep,
+            _overflowRef: getOverflowRef(i),
+            _hasTooltip: hasTooltips,
+            _totalSteps: totalSteps,
+            _stepNumber: i,
```
ZarkoPernar, line 86: Not sure I love this api. Is the intention behind underscore prefix just trying to "hide" certain props from the user of TrackerStep? In that case you can either wrap TrackerStep with context or render a totally different component, like so: 

```tsx
createElement(InternalTrackerStep, child.props)
```

ZarkoPernar (commented):
I find the overflow ref implementation quite convoluted. There is a lot of indirection going on there. Just trying to understand the requirements you were working with... Do you want to enable tooltip for every step if any of them have overflowing text?

packages/lab/src/stepped-tracker/SteppedTracker.tsx
ZarkoPernar: I think I prefer context in this case mainly because of the underscore props.

el-dav:
> I find the overflow ref implementation quite convoluted. There is a lot of indirection going on there. Just trying to understand the requirements you were working with... Do you want to enable tooltip for every step if any of them have overflowing text?

Yep, that's the behaviour in the existing component

packages/lab/src/stepped-tracker/SteppedTracker.tsx
el-dav: Thanks. Think I'll change to use context

packages/lab/src/stepped-tracker/SteppedTracker.tsx
el-dav: I've updated to use context, let me know what you think

ZarkoPernar (dismissed):
The only issue with the current design is that it is hard to tell if the completed step is also active. But if the design is ok with that then we can merge.

packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -0,0 +1,130 @@
...
+  defaultSteppedTrackerContext as unknown as SteppedTrackerContextType
+);
+
+export const useSteppedTrackerContext = () => useContext(SteppedTrackerContext);
+
+type TrackerStepNumberContextType = number;
+
+const TrackerStepContext = createContext<TrackerStepNumberContextType>(0);
+
+export const useTrackerStepContext = () => useContext(TrackerStepContext);
```
alycrys, line 51: can you please separate the context into it's own file

packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -0,0 +1,130 @@
...
+  activeStep: number;
+  /**
+   * Should be one or more TrackerStep components
+   */
+  children: ReactNode;
+}
+
+const useCheckInvalidChildren = (children: ReactNode) => {
+  useEffect(() => {
+    if (process.env.NODE_ENV !== "production") {
```
alycrys, line 68: what's the reason behind this check?

packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -0,0 +1,130 @@
...
+      testId: "salt-stepped-tracker",
+      css: steppedTrackerCss,
+      window: targetWindow,
+    });
+
+    useCheckInvalidChildren(children);
+    const [hasTooltips, setHasTooltips] = useState(false);
+    // A factory function used to get a callback ref for checking truncation.
+    const getOverflowRef = useDetectTruncatedText(setHasTooltips);
+    const totalSteps = Children.count(children);
```
alycrys, line 101: do we want to consider only valid children maybe?

packages/lab/src/stepped-tracker/TrackerConnector/TrackerConnector.css
```diff
@@ -0,0 +1,22 @@
+.saltTrackerConnector {
+  --trackerConnector-style-default: var(--saltTrackerConnector-style-default, var(--salt-track-borderStyle-incomplete));
+  --trackerConnector-style-active: var(--saltTrackerConnector-style-active, var(--salt-track-borderStyle-active));
+  --trackerConnector-color: var(--saltTrackerConnector-color, var(--salt-track-borderColor));
+  --trackerConnector-thickness: var(--saltTrackerConnector-thickness, var(--salt-track-borderWidth));
+  --trackerConnector-margin: var(--saltTrackerConnector-margin, var(--salt-spacing-50));
+}
```
alycrys, line 7: don't think you need to use these vars, they're not changing their values so you can just use the value directly

packages/lab/src/stepped-tracker/TrackerStep/TrackerStep.css
```diff
@@ -0,0 +1,49 @@
+.saltTrackerStep {
+  --trackerStep-icon-active-color: var(--saltTrackerStep-icon-active-color, var(--salt-status-info-foreground));
+  --trackerStep-icon-completed-color: var(--saltTrackerStep-icon-completed-color, var(--salt-status-success-foreground));
```
alycrys, line 3:
```suggestion
  --trackerStep-icon-color-active: var(--saltTrackerStep-icon-color-active, var(--salt-status-info-foreground));
  --trackerStep-icon-color-completed: var(--saltTrackerStep-icon-color-completed, var(--salt-status-success-foreground));
```

packages/lab/src/__tests__/__e2e__/stepped-tracker/SteppedTracker.cy.tsx
alycrys, line 1: I feel there are few tests missing

packages/lab/src/stepped-tracker/SteppedTracker.tsx
el-dav: Any particular reason?

packages/lab/src/stepped-tracker/SteppedTracker.tsx
el-dav: Could do but I think at that point we're already in trouble as the component won't work correctly

packages/lab/src/stepped-tracker/SteppedTracker.tsx
el-dav: Do you mean the hook or the `process.env.NODE_ENV !== "production"`?

The hook is to alert the consumer if they're doing something they shouldn't. E.g. adding random components within the `SteppedTracker` which would break it. Not sure how useful it is since I didn't want to disallow all other React components in case the consumer created a 'WrappedTrackerStep' with the same API which could be valid.

the condition here is so the end-user doesn't get those warnings in a production app.

packages/lab/src/stepped-tracker/TrackerConnector/TrackerConnector.css
el-dav: I added these as they give consumers a way to override certain aspects of the design without needing a complex selector. Personally I think we should do this more, but we should have a general consensus on it.

packages/lab/src/__tests__/__e2e__/stepped-tracker/SteppedTracker.cy.tsx
el-dav: Anything specific?

packages/lab/src/stepped-tracker/TrackerStep/TrackerStep.css
el-dav: Is this something we're consistent on? Think I was following the naming of the theme tokens e.g. `status-success-foreground` vs `status-foreground-success`

PR #1974, opened 2023-07-13
ListNext controlled list

Fercas123:
add controlled list, that needed a prop to control highlighted items, and it's example


packages/lab/src/list-next/ListNext.tsx
```diff
@@ -22,6 +22,8 @@ export interface ListNextProps extends HTMLAttributes<HTMLUListElement> {
    */
   disabled?: boolean;
   /* Value for the uncontrolled version. */
+  highlightedIndex?: number;
```
el-dav, line 25: Should the comment be 'Value for the controlled version'?

Fercas123: oh true, it is the other way around :P will change it

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -55,7 +58,7 @@ export const ListNext = forwardRef<HTMLUListElement, ListNextProps>(
     const listId = useId(id);
     const listRef = useRef<HTMLUListElement>(null);
     const handleRef = useForkRef(listRef, ref);
-
+    const controlled = Boolean(selected || highlightedIndex);
```
el-dav, line 61: Could there be an issue if the first item was highlighted `highlightedIndex === 0` would zero being falsey accidentally make `controlled` `false`

el-dav:
Up and down arrows don't seem to be working anymore for list navigation

packages/lab/src/list-next/ListNext.tsx
Fercas123: good point, i fixed this in the hook but forgot it here, I think I'll check for undefined here and send back controlled to the hook instead

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -58,7 +67,9 @@ export const ListItemNext = forwardRef<HTMLLIElement, ListItemNextProps>(
     const focused = isFocused(itemId);
 
     const handleClick = (event: MouseEvent<HTMLLIElement>) => {
-      select(event);
+      if (!disabled) {
```
alycrys, line 70: i'd change it to only have a handler if !disabled

packages/lab/src/list-next/ListItemNext.tsx
```diff
@@ -78,7 +87,7 @@ export const ListItemNext = forwardRef<HTMLLIElement, ListItemNextProps>(
         aria-selected={selected || undefined}
         id={itemId}
         data-value={value}
-        onClick={handleClick}
+        onClick={disabled ? undefined : handleClick}
```
libertymayc, line 90: what's our go to for this? the above, or putting an if !disabled in the handleClick?

.changeset/long-hats-count.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/lab": patch
+---
+
+Added controlled version of ListNext
+
+- Added `highlightedIndex` prop to `ListNext`
```
libertymayc, line 7: add description of what this is used for

packages/lab/src/list-next/ListItemNext.tsx
Fercas123: not sure we have one, but @alycrys suggested this implementation, seems like we are already doing it in accordion too

packages/lab/src/list-next/useList.ts
```diff
@@ -13,43 +14,58 @@ interface UseListProps {
...
    */
   disabled?: boolean;
-  /* Value for the uncontrolled version. */
+  /* Value for the controlled version. */
+  highlightedIndex?: number;
+  /* Value for the controlled version. */
   selected?: string;
-  /* Initial value for the uncontrolled version. */
+  /* Initial value for the controlled version. */
   defaultSelected?: string;
```
libertymayc, line 22: I think the defaultSelected could still be used for uncontrolled version, is that true?

packages/lab/src/list-next/useList.ts
```diff
@@ -172,16 +197,22 @@ export const useList = ({
...
   const blurHandler = () => {
     setShowFocusRing(false);
+    setFromMouse(false);
   };
 
   const mouseDownHandler = () => {
     // When list gets focused, we can't guarantee that focus happens after click event.
-    // If first focus (where !activeDescendant) happens from a click, list shouldn't render focus ring in the first element.
-    setShowFocusRing(false);
+    // If focus happens from a click, list shouldn't render focus ring.
```
libertymayc, line 205: on controlled story I get a focus ring on mouse click. is that intended on controlled vs un?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -98,7 +112,7 @@ export const ListNext = forwardRef<HTMLUListElement, ListNextProps>(
           id={listId}
           className={clsx(withBaseName(), className)}
           role="listbox"
-          tabIndex={disabled ? -1 : 0}
+          tabIndex={disabled || controlled ? -1 : 0}
```
origami-z, line 115: I'm wondering why `controlled` would change behaviour of a component?

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -55,7 +64,10 @@ export const ListNext = forwardRef<HTMLUListElement, ListNextProps>(
     const listId = useId(id);
     const listRef = useRef<HTMLUListElement>(null);
     const handleRef = useForkRef(listRef, ref);
-
+    const controlled = useMemo(
```
el-dav, line 67: Nit: I don't think the useMemo is necessary

packages/lab/stories/list-next/list.stories.tsx
```diff
@@ -1,32 +1,22 @@
 import { ComponentMeta, Story } from "@storybook/react";
-import { ListNext, ListItemNext, ListNextProps } from "../../src";
+import { ListItemNext, ListNext, ListNextProps } from "../../src";
 import { Button, FlexLayout } from "@salt-ds/core";
 import { useState } from "react";
+import { usStateExampleData } from "../assets/exampleData";
+import { ArrowDownIcon, ArrowUpIcon } from "@salt-ds/icons";
 
 export default {
```
el-dav, line 8: I don't think this is new in the PR but I'm noticing an issue in each of the stories. When you first mousedown on a ListItem (try it on an item which isn't the first) the first ListItem is highlighted, even if that isn't the one you are doing mousedown on.

Seems to only occur the first time an item is selected. Can fix in a separate PR if needed

el-dav: Ignore this. Hadn't fetched the latest

packages/lab/src/list-next/ListNext.tsx
```diff
@@ -16,14 +18,19 @@ import { ListNextContext } from "./ListNextContext";
...
-export interface ListNextProps extends HTMLAttributes<HTMLUListElement> {
+export interface ListNextProps
+  extends Omit<ComponentPropsWithoutRef<"ul">, "onChange"> {
   /**
    * If true, all items in list will be disabled.
    */
   disabled?: boolean;
-  /* Value for the uncontrolled version. */
+  /* Value for the controlled version. */
+  highlightedIndex?: number;
```
el-dav, line 28: Is there a logic as to why highlighted is controlled by the index, but selected is controlled by the value of the relevant item? Maybe these should be aligned?

One potential issue is you might want to control a list using the `onChange` prop of `List` (in the current example you use the `onClick` of `ListItem`). But the `onChange` event of `List` doesn't pass the index of the clicked item as part of the callback data. If we do keep it as an index then at least we should also pass the clicked index in that callback to allow you to do something like:

```
const handleChange = (_, { value, index }) => {
    setSelectedItem(value);
    setHighlightedIndex(index);
  };
```
but I do think it would be nicer if you could simply do:
```
const handleChange = (_, { value }) => {
    setSelectedItem(value);
    setHighlightedItem(value);
  };
```

el-dav: Having looked a bit more it seems like it wouldn't be possible to use the onChange callback for this since it only triggers when a different item is selected.

packages/lab/src/list-next/ListNext.tsx
libertymayc: if it is, add a description ;)

packages/lab/src/list-next/useList.ts
Fercas123: i need to double check this with Design

packages/lab/src/list-next/ListNext.tsx
Fercas123: changed for disableFocus

packages/lab/stories/list-next/list.stories.tsx
```diff
@@ -49,8 +39,81 @@ export const Default: Story<ListNextProps> = ({ children, ...rest }) => {
   );
 };
 
-Default.args = {
-  style: { width: "200px" },
+Default.args = {};
+
+export const Controlled: Story<ListNextProps> = ({ children, ...rest }) => {
```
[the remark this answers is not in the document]
Fercas123: will follow up with a PR for this :)

packages/lab/stories/list-next/list.stories.tsx
```diff
@@ -49,8 +39,81 @@ export const Default: Story<ListNextProps> = ({ children, ...rest }) => {
...
+  const [highlightedIndex, setHighlightedIndex] = useState<number | undefined>(
+    undefined
```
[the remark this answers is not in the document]
Fercas123: can you explain a bit more?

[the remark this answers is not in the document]
alycrys: when you have 
```
const [highlightedIndex, setHighlightedIndex] = useState<number | undefined>(0)
```
I'd expect that the first item is highlighted. Is that not the case?

packages/lab/src/list-next/useList.ts
```diff
@@ -84,6 +103,13 @@ export const useList = ({
     [setSelectedItem, updateActiveDescendant]
   );
 
+  // Effect to move the cursor when items change controlled.
+  // this could be following active descendant if there is no better way of doing it when controlled
+  useEffect(() => {
+    const activeOptions = getOptions();
+    highlightedIndex && updateScroll(activeOptions[highlightedIndex]);
+  }, [highlightedIndex, getOptions, updateScroll]);
```
origami-z, line 111: is there a way to directly drive `activeDescendant` via `useControlled`'s `controlled` prop?

packages/lab/src/list-next/ListNextContext.tsx
```diff
@@ -5,8 +5,10 @@ export interface ListNextContextValue {
   disabled?: boolean;
   id?: string;
   select: (event: SyntheticEvent<HTMLLIElement>) => void;
-  isSelected: (id: string) => boolean;
-  isFocused: (id: string) => boolean;
+  isSelected: (value: string) => boolean;
+  isFocused: (value: string) => boolean;
+  highlight: (event: SyntheticEvent<HTMLLIElement>) => void;
+  isHighlighted: (id: string) => boolean;
```
origami-z, line 11: Is some of these work on `id` others work on `value`?

packages/lab/src/list-next/useList.ts
```diff
@@ -1,54 +1,77 @@
...
-  /* Value for the uncontrolled version. */
+  /* Value for the controlled version. */
+  highlightedIndex?: number;
+  /* Value for the controlled version. */
   selected?: string;
-  /* Initial value for the uncontrolled version. */
+  /* Initial value for the controlled version. */
   defaultSelected?: string;
-  /* Callback for the controlled version. */
+  /* Callback for the uncontrolled version. */
```
origami-z, line 24: This is needed for both controlled and uncontrolled

packages/lab/src/list-next/useList.ts
```diff
@@ -1,54 +1,77 @@
...
-import { useControlled } from "@salt-ds/core";
+import { useControlled, useIsFocusVisible } from "@salt-ds/core";
 
 interface UseListProps {
   /**
    * If true, all items in list will be disabled.
    */
   disabled?: boolean;
-  /* Value for the uncontrolled version. */
+  /* Value for the controlled version. */
```
origami-z, line 18: Could be better to mention what it controls. The word of "Controlled" can be multiple use cases.

packages/lab/src/list-next/useList.ts
```diff
@@ -1,54 +1,77 @@
...
 }: UseListProps) => {
-  const [activeDescendant, setActiveDescendant] = useState<string | undefined>(
-    undefined
-  );
+  const getOptions: () => HTMLElement[] = useCallback(() => {
+    return Array.from(
+      ref.current?.querySelectorAll('[role="option"]:not([aria-disabled])') ??
+        []
+    );
+  }, [ref]);
```
origami-z, line 46: Don't think `ref` is meant to be used in dep array?

packages/lab/src/list-next/useList.ts
```diff
@@ -1,54 +1,77 @@
...
+    name: "ListNextHighlighted",
+    state: "activeDescendant",
+  });
 
-  const [showFocusRing, setShowFocusRing] = useState<boolean>(false);
   const [selectedItem, setSelectedItem] = useControlled({
     controlled: selectedProp,
     default: defaultSelected,
-    name: "ListNext",
+    name: "ListNextSelected",
```
origami-z, line 65: `useControlled` will put `name` + `state` together

packages/lab/src/list-next/useList.ts
Fercas123: there isn't, but active descendant always follows highlightedIndex (changed to to highlighteditem to move them all with value)

packages/lab/src/list-next/ListNextContext.tsx
Fercas123: updated this

packages/lab/src/list-next/ListNext.css
```diff
@@ -5,7 +5,6 @@
   border-color: var(--salt-container-primary-borderColor);
   border-style: var(--salt-container-borderStyle);
   border-width: var(--salt-size-border);
-  height: 100%;
```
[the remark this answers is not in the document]
Fercas123: this was something to do with how it drives the height of the container, but I'll add it in another change to make sure it is actually needed.

libertymayc (approved):

libertymayc:
when i tab into the controlled list, what is supposed to focus? currently nothing is focusing for me

.changeset/long-hats-count.md
```diff
@@ -0,0 +1,11 @@
+---
+"@salt-ds/lab": patch
+---
+
+Added controlled version of ListNext
+
+- Added `highlightedItem`, `selected` and `onChange` props to `ListNext` in order to support a controlled list.
```
feliciackh, line 7:
```suggestion
- Added `highlightedItem`, `selectedItem` and `onChange` props to `ListNext` in order to support a controlled list.
```

feliciackh (approved):

Fercas123:
> when i tab into the controlled list, what is supposed to focus? currently nothing is focusing for me I approved because otherwise it's good - maybe can fix in a new pr, but can probably chuck it in here

that example has focusDisable in, how CB and DD are going to use it, so the focus should always be on what controls the list

PR #1984, opened 2023-07-14
List fix for double borders

Fercas123:
This PR has the fix to list's double border replicating dataGrid styles, 
one key difference is that i am adding margin to elements, which means there will be a px between the top and bottom borders of the container and the first/last items, to account for their borders.

I see this happens in Grid aswell, but just want to double check if it is what we intend in list too. 

image

Another difference is the shared 1px gap between elements.

image

cc. @ivan-calderon @bhoppers2008


libertymayc (approved):

libertymayc:
in 2nd screenshot, is Arkansas hovered and california focused? should there still be a gap between arkansas and arizona like there is between ark/cal?

libertymayc:
makes sense :)

feliciackh (approved):

PR #1996, opened 2023-07-17
Stepped Tracker V1

el-dav:
PR for [v1 of the SteppedTracker component](https://github.com/jpmorganchase/salt-ds/issues/1664) which includes the following features:

- Horizontal orientation
- Basic stepped tracker (Default, In progress, Complete)
- Label truncation + tooltip support


Fercas123 (approved):

packages/lab/src/stepped-tracker/StepLabel/StepLabel.tsx
```diff
@@ -0,0 +1,48 @@
...
+      window: targetWindow,
+    });
+
+    const { getOverflowRef } = useSteppedTrackerContext();
+    const overflowRef = useMemo(() => getOverflowRef(id), [id, getOverflowRef]);
+
+    const combinedRef = useForkRef(overflowRef, ref);
+
+    return (
+      <label
```
libertymayc, line 39: can we use Text label here?

packages/lab/src/stepped-tracker/TrackerStep/TrackerStep.css
```diff
@@ -0,0 +1,50 @@
...
+  margin: 0;
+  padding: 0;
+  text-indent: 0;
+  list-style-type: none;
+  flex: 1;
+  position: relative;
+  display: flex;
+  align-items: center;
+  flex-direction: column;
+  width: var(--trackerStep-width);
```
libertymayc, line 22: should we put a fallback for this one in case undefined? I also think that when we inject it from the component styles we usually use the API token --saltTrackerStep-width

packages/lab/src/stepped-tracker/TrackerStep/TrackerStep.tsx
```diff
@@ -0,0 +1,135 @@
...
+    const resolvedState = getState({ isActive, state });
+    const connectorState = activeStep > stepNumber ? "active" : "default";
+    const hasConnector = stepNumber < totalSteps - 1;
+
+    const Inner = (
+      <li
+        className={clsx(withBaseName(), withBaseName(resolvedState), className)}
+        style={
+          {
+            ...props.style,
```
libertymayc, line 109: you could take the style out from restProps, I think otherwise they're being added again below?
I would take it out and make `const innerStyle = {...style, "--saltTrackerStep-width": `${100 / totalSteps}%`}` like we do for the other components

el-dav: Good spot! I had moved the spreading of ..rest after the other props in response to feedback where this wouldn't be an issue, but hadn't spotted to update this.

packages/lab/src/stepped-tracker/StepLabel/StepLabel.tsx
el-dav: I had tried this previously and seemed to have an issue with getting the overflow behaviour, but will try again.

libertymayc (approved):

Fercas123 (approved):

PR #2013, opened 2023-07-19
Badge V1

lukeac123:
- rename badgeContent to value 
- add inline styling when no children are declared
- enable alpha numeric characters 
- truncation on numbers according to max value 
- truncation on strings greater than 4 characters

### DIFF
--- a/packages/lab/src/badge/Badge.css
+++ b/packages/lab/src/badge/Badge.css
@@ -1,21 +1,3 @@
-/* Default variables applied to the root element */
-.saltBadge {
-  /* Colors */
-  --badge-content-text-color: var(--saltBadge-content-text-color, var(--salt-accent-foreground));
-  --badge-content-icon-color: var(--saltBadge-content-icon-color, var(--salt-accent-background));
-  /* Font, Typography */
-  --badge-content-borderRadius: 16px;
-  --badge-content-lineHeight: 15px;
-  --badge-content-padding: 0px 4px;
-  --badge-fontSize: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
-
-  /* Derived */
-  /* Same as content padding in x */
-  --badge-offset-inline: -4px;
-
-  --badge-sequence-gap: var(--salt-size-unit);
-}
-
 /* Style applied to the root element */
 .saltBadge {
   -webkit-font-smoothing: antialiased;
@@ -24,44 +6,35 @@
   position: relative;
   flex-shrink: 0;
   vertical-align: middle;
+
+  /* Should this vary according to touch size */
+  padding: 0px var(--salt-spacing-50);
   line-height: var(--salt-text-lineHeight);
-  font-size: var(--salt-text-fontSize);
+  min-width: 10px;
+  line-height: 15px;
+  border-radius: 16px;
+
+  font-size: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
+  font-weight: var(--salt-accent-fontSize);
+
+  background: var(--saltBadge-content-icon-color, var(--salt-accent-background));
+  color: var(--saltBadge-content-text-color, var(--salt-accent-foreground));
 }
 
-/* Give automatic margin when they sit next to each other */
-.saltBadge + .saltBadge {
-  margin-left: var(--badge-sequence-gap);
-}
-
-/* Style applied to the badge in the corner of the node */
-.saltBadge-badge {
-  color: var(--badge-content-text-color);
-  display: inline-block;
-  font-weight: 700; /* TODO: Use characteristic once aligned */
-  font-size: var(--badge-fontSize);
-  /* default small type */
-  width: auto;
-  min-width: 15px;
-  line-height: var(--badge-content-lineHeight);
-  border-radius: var(--badge-content-borderRadius);
-  text-align: center;
-
-  transform: translate(var(--badge-offset-inline), -50%);
-  transform-origin: 100% 0%;
-  /*
-    We want offset the next inline element to be strictly next to the badge.
-    This negative margin makes sure this node "appears" to be narrower.
-   */
-  margin-right: var(--badge-offset-inline);
-
-  padding: var(--badge-content-padding);
-  background: var(--badge-content-icon-color);
-
+.saltBadge-inline {
   position: relative;
-
-  flex-wrap: wrap;
   align-items: center;
-  align-content: center;
-  flex-direction: row;
   justify-content: center;
+  display: inline-flex;
+  margin-bottom: 4px;
+  margin-left: var(--salt-spacing-50);
+}
+
+.saltBadge-topRight {
+  top: -8px;
+  right: 5px;
+}
+
+.saltBadge-wrap {
+  position: relative;
 }
--- a/packages/lab/src/badge/Badge.css
+++ b/packages/lab/src/badge/Badge.css
@@ -3,38 +3,38 @@
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   display: inline-flex;
-  position: relative;
   flex-shrink: 0;
-  vertical-align: middle;
+  justify-content: center;
+}
 
+.saltBadge-badge {
   /* Should this vary according to touch size */
   padding: 0px var(--salt-spacing-50);
   line-height: var(--salt-text-lineHeight);
   min-width: 10px;
-  line-height: 15px;
-  border-radius: 16px;
+  max-width: 50px;
+  border-radius: 999px;
 
-  font-size: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
-  font-weight: var(--salt-accent-fontSize);
-
-  background: var(--saltBadge-content-icon-color, var(--salt-accent-background));
-  color: var(--saltBadge-content-text-color, var(--salt-accent-foreground));
-}
-
-.saltBadge-inline {
-  position: relative;
   align-items: center;
   justify-content: center;
-  display: inline-flex;
+  display: inline-block;
   margin-bottom: 4px;
-  margin-left: var(--salt-spacing-50);
+
+  font-size: var(--salt-text-label-fontSize);
+  font-weight: var(--salt-accent-fontSize);
+  background: var(--salt-accent-background);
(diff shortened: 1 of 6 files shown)


packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
   position: relative;
   flex-shrink: 0;
   vertical-align: middle;
```
ZarkoPernar, line 8: I would replace this with 
```suggestion
  justify-content: center;
  align-items: center;
```
So we can center the content of the badge

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
   position: relative;
```
ZarkoPernar, line 6: I would replace this with 
```suggestion
  position: absolute;
```

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
-  min-width: 15px;
-  line-height: var(--badge-content-lineHeight);
-  border-radius: var(--badge-content-borderRadius);
-  text-align: center;
+  /* Should this vary according to touch size */
+  padding: 0px var(--salt-spacing-50);
+  line-height: var(--salt-text-lineHeight);
+  min-width: 10px;
+  line-height: 15px;
+  border-radius: 16px;
```
ZarkoPernar, line 15: If we want a full border radius no matter the height of the badge, this should be 
```suggestion
  border-radius: 9999px;
```

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
-  /* default small type */
-  width: auto;
-  min-width: 15px;
-  line-height: var(--badge-content-lineHeight);
-  border-radius: var(--badge-content-borderRadius);
-  text-align: center;
+  /* Should this vary according to touch size */
+  padding: 0px var(--salt-spacing-50);
+  line-height: var(--salt-text-lineHeight);
+  min-width: 10px;
```
ZarkoPernar, line 13: This should come from the characteristics.

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
+  line-height: 15px;
```
ZarkoPernar, line 14: Remove this line, you already defined it above.

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
-    We want offset the next inline element to be strictly next to the badge.
-    This negative margin makes sure this node "appears" to be narrower.
-   */
-  margin-right: var(--badge-offset-inline);
+  font-size: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
+  font-weight: var(--salt-accent-fontSize);
 
-  padding: var(--badge-content-padding);
-  background: var(--badge-content-icon-color);
+  background: var(--saltBadge-content-icon-color, var(--salt-accent-background));
```
ZarkoPernar, line 20: This variable `--saltBadge-content-icon-color` doesn't make sense. This should probably be something like `--saltBadge-background`. And color should be `--saltBadge-foreground`.

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
-  flex-direction: row;
   justify-content: center;
+  display: inline-flex;
+  margin-bottom: 4px;
+  margin-left: var(--salt-spacing-50);
+}
+
+.saltBadge-topRight {
+  top: -8px;
+  right: 5px;
```
ZarkoPernar, line 35: I think these two will change once you swap to position absolute.

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
   position: relative;
-
-  flex-wrap: wrap;
   align-items: center;
-  align-content: center;
-  flex-direction: row;
   justify-content: center;
+  display: inline-flex;
+  margin-bottom: 4px;
+  margin-left: var(--salt-spacing-50);
```
ZarkoPernar, line 30: I would not have margin on badge itself. When it is displayed "inline", the parent should be responsible for its positioning. So not a concern of the badge component.

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
 
+.saltBadge-inline {
   position: relative;
-
-  flex-wrap: wrap;
   align-items: center;
-  align-content: center;
-  flex-direction: row;
   justify-content: center;
+  display: inline-flex;
```
ZarkoPernar, line 28: This is not necessary

packages/lab/src/badge/Badge.tsx
```diff
@@ -60,32 +34,39 @@ export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
...
-  );
+  let valueText = value;
 
-  let badgeContentValue = badgeContent;
-  if (badgeContentValue > max) {
-    badgeContentValue = `${max}+`;
+  if (typeof value === "number" && max) {
+    valueText = value > max ? `${max}+` : value;
+  } else if (typeof value === "string") {
+    valueText = value.length > 4 ? `${value.slice(0, 4)}...` : value;
```
ZarkoPernar, line 42: I would remove this. Just display the string contents. Truncation can be done with css.

packages/lab/src/badge/Badge.tsx
```diff
@@ -60,32 +34,39 @@ export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
...
-      {...rest}
-    >
-      {children && isValidElement<HTMLAttributes<HTMLElement>>(children)
-        ? cloneElement(children, { id: childId })
-        : children}
-      <span id={badgeId} className={clsx(withBaseName("badge"))}>
-        {badgeContentValue}
+  if (!children) {
+    return (
+      <div>
```
ZarkoPernar, line 47: You don't need this div. Make sure to create a storybook example where you are rendering the badge inline with some text and no `display: flex` on the parent. You will  immediately notice the issue. Divs have `display: block` by default meaning they will take up the whole line.

```tsx
<span>some text <Badge value="hello" /></span>
```

ZarkoPernar: When rendering inline, you can simply return the <span/> below directly. No need for a wraper.

packages/lab/src/badge/Badge.tsx
```diff
@@ -60,32 +34,39 @@ export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
...
+  } else
+    return (
+      <span
+        className={clsx(withBaseName("wrap"), className)}
+        ref={ref}
+        {...rest}
+      >
+        {children}
+        <span
+          className={clsx(withBaseName(), withBaseName(`topRight`), className)}
```
ZarkoPernar, line 66: You should not apply className prop to both elements. Always to the top most one.

ZarkoPernar (commented):
I think you slightly missed the structure I was suggesting. When children gets passed we want to wrap the children with an element that has position relative. This means anything within the wrapper gets absolute positioned "relative" to that wrapper. ( I know the naming is a bit whack 😅 )

So, this is the structure we are looking for...

```tsx
<span className="wrapper">
  {children}
  <span className="badge">{value}</span>
</span>
```

```css
.wrapper {
  position: relative;
}

.badge {
  position: absolute;
  top: 0;
  right: 0;
}
```

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
 
-  transform: translate(var(--badge-offset-inline), -50%);
-  transform-origin: 100% 0%;
-  /*
-    We want offset the next inline element to be strictly next to the badge.
-    This negative margin makes sure this node "appears" to be narrower.
-   */
-  margin-right: var(--badge-offset-inline);
+  font-size: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
+  font-weight: var(--salt-accent-fontSize);
```
libertymayc, line 18:
```suggestion
  font-weight: var(--salt-accent-fontWeight);
```

packages/lab/src/badge/Badge.css
```diff
@@ -24,44 +6,35 @@
...
+  border-radius: 16px;
 
-  transform: translate(var(--badge-offset-inline), -50%);
-  transform-origin: 100% 0%;
-  /*
-    We want offset the next inline element to be strictly next to the badge.
-    This negative margin makes sure this node "appears" to be narrower.
-   */
-  margin-right: var(--badge-offset-inline);
+  font-size: var(--saltBadge-fontSize, var(--salt-text-label-fontSize));
```
libertymayc, line 17:
```suggestion
  font-size: var(--saltBadge-fontSize, var(--salt-accent-fontSize));
```

packages/lab/src/badge/Badge.tsx
```diff
@@ -60,32 +34,39 @@ export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
...
-      role="img"
-      aria-labelledby={clsx(childId, badgeId)}
-      {...rest}
-    >
-      {children && isValidElement<HTMLAttributes<HTMLElement>>(children)
-        ? cloneElement(children, { id: childId })
-        : children}
-      <span id={badgeId} className={clsx(withBaseName("badge"))}>
-        {badgeContentValue}
+  if (!children) {
```
libertymayc, line 45: I'm not sure about this conditional return. it can be done inline on the changes. something like
```return (
<span
          className={clsx(withBaseName(), {[withBaseName("inline")]: children, [withBaseName("wrap")]: !children}, className)}
          ref={ref}
          {...rest}
        >
          {children}
          <span
          className={clsx({[withBaseName(`topRight`)]: children})}> 
          {children}
          </span>
        </span>
);
```

packages/lab/src/badge/Badge.css
lukeac123: does adding "position: absolute" not position the element relative to the DOM. So in this case would appear top right of the screen ?

ZarkoPernar: No, that is `position: fixed`

PR #2016, opened 2023-07-19
MultilineInput character count

libertymayc:
Added a story to show how character count can be implemented as an adornment.


alycrys (approved):

alycrys (approved):

packages/lab/stories/multiline-input/multiline-input.doc.stories.mdx
```diff
@@ -112,6 +112,16 @@ Adornments can be added at the beginning or end of the text area using the `star
   <Story id="lab-multiline-input--with-adornments" />
 </Canvas>
 
+### Character Count
+
+Use a controlled Multiline Input to enforce a character count on the user input. Add the character count as a span adornment to present the remaining characters out of the given limit.
+
+<Canvas>
+  <Story id="lab-multiline-input--controlled" />
```
origami-z, line 120: wrong id?

packages/lab/stories/multiline-input/multiline-input.stories.tsx
```diff
@@ -100,6 +100,55 @@ export const Disabled: ComponentStory<typeof MultilineInput> = (args) => {
...
+
+  const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
+    const newVal = event.target.value;
+
+    if (newVal.length < MAX_CHARS) {
+      const value = newVal;
+      setValue(value);
+    }
+
+    if (MAX_CHARS - newVal.length === 50) {
```
joshwooding, line 119: Has this announcer logic been tested? Additions like that can actually harm accessibility rather than improve it.

Also as it is currently written this logic is broken. You can only write 9 characters and the only announcements you can achieve are:

10 characters remaining and once at the maximum value you can type (9 characters) you can press another key to then have it has 0 characters remaining but nothing gets updated/

packages/lab/stories/multiline-input/multiline-input.doc.stories.mdx
libertymayc: thanks

packages/lab/stories/multiline-input/multiline-input.stories.tsx
libertymayc: doh, I did test it, but then changed it to 10 characters as thought it would be a clearer demo then didnt update the announcer to match 🤣 
I've deleted it in the hope to get this out asap, we can add something back if recommended by stella in the future

joshwooding (dismissed):

packages/lab/stories/multiline-input/multiline-input.stories.tsx
```diff
@@ -100,6 +100,38 @@ export const Disabled: ComponentStory<typeof MultilineInput> = (args) => {
...
+    }
+  };
+
+  return (
+    <MultilineInput
+      {...args}
+      value={value}
+      onChange={handleChange}
+      endAdornment={
+        <span
```
joshwooding, line 123: More of a minor thing but would using Text have been better here? They wouldn't need to know so much about the CSS variables

libertymayc: changed :) good shout

joshwooding (approved):

Fercas123 (approved):

PR #2026, opened 2023-07-20
Replace `Scrim` with `FloatingFocusManager` in Drawer

JoanaMMoreira:
- Replace `Scrim` with `FloatingFocusManager`
- Add `Scrim` styles to `FloatingOverlay`
- Remove `disableScrim`, `disableAnimations` and `scrimProps` props
- Rename prop `isOpen` to `open` and add `onOpenChange` prop to allow component to be dismissed when clicking outside or pressing `Esc`
- Add storybook docs and update examples
- Update cypress tests


packages/lab/stories/drawer/drawer.doc.stories.mdx
```diff
@@ -0,0 +1,94 @@
...
+<Canvas>
+  <Story id="lab-drawer--bottom" />
+</Canvas>
+
+## Configuring Drawer
+
+### API
+
+```
+import { Drawer } from "@salt-ds/core";
```
feliciackh, line 75:
```suggestion
import { Drawer } from "@salt-ds/lab";
```

libertymayc (dismissed):

.changeset/wild-keys-obey.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/lab": patch
+---
+
+Drawer
+
+- Replace `Scrim` with `FloatingFocusManager`
+- Remove `disableScrim`, `disableAnimations` and `scrimProps` props
+- Rename prop `isOpen` to `open` and add `onOpenChange` prop to allow component to be dismissed when clicking outside or pressing `Esc`
```
libertymayc, line 9: clicking outside or Esc dont work for me in the stories :(

packages/lab/src/drawer/Drawer.tsx
```diff
@@ -71,58 +64,59 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(function Drawer(
 
   const [isAnimating, setIsAnimating] = useState(false);
 
+  const { context, floating } = useFloatingUI({
+    open,
+    onOpenChange,
+  });
+
+  const floatingRef = useForkRef(floating, ref) as Ref<HTMLDivElement>;
```
joshwooding, line 72:
```suggestion
  const floatingRef = useForkRef<HTMLDivElement>(floating, ref);
```
would this work?

.changeset/wild-keys-obey.md
JoanaMMoreira: oh no:( let me take a look

JoanaMMoreira: Should be working now. I've moved most of the floating-ui logic into a separate hook called `useDrawer`

feliciackh (dismissed):

libertymayc (approved):

packages/lab/src/__tests__/__e2e__/drawer/Drawer.cy.tsx
```diff
@@ -7,12 +7,12 @@ const { Default, Top, Right, Bottom } = composedStories;
...
 describe("GIVEN a Drawer", () => {
   describe("WHEN no props are provided", () => {
-    it("THEN it should display a scrim by default", () => {
+    it("THEN it should display an overlay by default", () => {
       cy.mount(<Default />);
 
       cy.findByRole("button", { name: /Open Drawer/i }).click();
 
-      cy.get(".saltScrim").should("be.visible");
+      cy.get(".saltDrawer-overlay").should("be.visible");
```
joshwooding, line 15: Is there a way to check for a role instead?

packages/lab/src/__tests__/__e2e__/drawer/Drawer.cy.tsx
```diff
@@ -94,31 +74,24 @@ describe("GIVEN a Drawer", () => {
...
-  describe("WHEN animations are enabled", () => {
-    it("THEN it should display animations", () => {
-      cy.mount(<Default disableAnimations={false} />);
+  describe("WHEN a drawer is open", () => {
+    it("THEN it should be able to close", () => {
+      cy.mount(<Default />);
 
       cy.findByRole("button", { name: /Open Drawer/i }).click();
 
       cy.get(".saltDrawer").should("have.class", "saltDrawer-enterAnimation");
```
joshwooding, line 83: I would check for the presence and removal of a role here rather than classes

packages/lab/src/__tests__/__e2e__/drawer/Drawer.cy.tsx
JoanaMMoreira: there's no role on the overlay, we only add `role="dialog"` to the actual Drawer. Is there an additional role we should be using here?

joshwooding: Sorry I meant just in general, checking this for overlay is fine :)

packages/lab/src/drawer/Drawer.tsx
```diff
@@ -71,58 +62,53 @@ export const Drawer = forwardRef<HTMLDivElement, DrawerProps>(function Drawer(
 
   const [isAnimating, setIsAnimating] = useState(false);
```
joshwooding, line 63: More of a general question, setIsAnimating is only set to true in the code and will never go back to false. Is this an issue?

JoanaMMoreira: I was refactoring this and just realised it's not even needed anymore (I think it was a leftover from the `disableAnimations` prop we removed) so I've removed it

joshwooding (approved):

origami-z (approved):

PR #2033, opened 2023-07-21
ToastGroup

alycrys:


ZarkoPernar:
Do you need the overflow auto on the container? It cuts off the shadow. If you do maybe add some padding to it so that the entirety of the shadow is visible.

ZarkoPernar:
Overall it looks good. I know it is very basic functionality but at least it opens it up for a lot of further enhancements down the line.

alycrys:
> Do you need the overflow auto on the container? It cuts off the shadow. If you do maybe add some padding to it so that the entirety of the shadow is visible.

Should look better now. Thanks Zarko.

libertymayc (approved):

feliciackh (approved):

PR #2039, opened 2023-07-24
Radiobutton design fix

libertymayc:
Minor fix noticed by @pseys 
Not a huge deal - but the actual control in the RadioButton doesn't match the size of the icon presented


joshwooding (changes requested):
I believe this breaks the Checkbox alignment.

joshwooding:
> > I believe this breaks the Checkbox alignment.
> 
> how so?

Applying align-items to Checkbox means that when there is a long label the checkbox aligns to the centre rather than the baseline.

joshwooding (changes requested):
What issue is this solving, can you raise a bug

.changeset/soft-ligers-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed control size in RadioButton to match the shown icon size
```
joshwooding, line 5:
```suggestion
Fixed control size in Radio Button to match the shown icon size.
```

joshwooding (approved):

joshwooding (approved):

Fercas123 (approved):

PR #2040, opened 2023-07-24
uses conditional styling for inline and logic to truncate long strings

lukeac123:
- rename "BadgeContent" to "value"
- truncate strings on > 4 characters 
- inline styling when no child it passed to the component


lukeac123:
need to change author

PR #2054, opened 2023-07-25
Add Characteristics pages

maleeha13:
- Creates a charcteristics section .
- Has a search table that allows users to copy/ lookup any css names or values.
- Has a characteristics type page that displays accordian view of the different types of characteristics, with description and visual representation. 
- Should display banner if no data is loaded.
- Responsive on mobile devices.


site/docs/characteristics/accordianView.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Characteristic types
+layout: DetailTechnical
+sidebar:
+  label: Characteristic types
+---
+
+<AccordianView />
```
lilyvc, line 8: should be Accordion?

site/src/components/css-display/Accordian_view.tsx
```diff
@@ -0,0 +1,193 @@
...
+const text: string[] = new Array("fontStyle");
+const outline: string[] = new Array("outline");
+const shadow: string[] = new Array("shadow");
+const align: string[] = new Array("textAlign");
+const textDecoration: string[] = new Array("textDecoration");
+const transform: string[] = new Array("textTransform");
+const fontFamily: string[] = new Array("fontFamily");
+const lineHeight: string[] = new Array("lineHeight");
+
+const MyComponent: React.FC<{ name: string }> = ({ name }) => {
```
lilyvc, line 56: what is 'MyComponent'?

site/src/components/css-display/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Css_table";
+export * from "./Accordian_view";
```
lilyvc, line 2: most of the other components in the site are using PascalCasing for the filenames. For example AccordionView.tsx

site/src/components/css-display/Accordian_view.tsx
maleeha13: changed the naming to BlockView,, basically decides which type of block to use when displaying the css vars

site/docs/characteristics/accordianView.mdx
maleeha13: its the custom css accordian view

packages/theme/css/cssVariables.json
```diff
@@ -0,0 +1,11 @@
+{
+  "position": "relative",
+  "right": "0",
+  "top": "50%",
+  "transform": "translateY(-50%)",
+  "width": "unset",
+  "display": "none",
+  "margin-bottom": "5%",
+  "font-size": "15px",
+  "max-width": "768px) {\n  .viewColumn {\n    display: none"
```
libertymayc, line 10: typo

packages/theme/css/cssVariables.json
```diff
@@ -0,0 +1,11 @@
+{
+  "position": "relative",
+  "right": "0",
+  "top": "50%",
+  "transform": "translateY(-50%)",
+  "width": "unset",
+  "display": "none",
+  "margin-bottom": "5%",
```
libertymayc, line 8: could we use spacing tokens?

site/src/components/css-display/descriptions.js
```diff
@@ -0,0 +1,78 @@
...
+    value:
+      "Components that may form a hierarchy and stack upon or beneath other elements of the UI.    ",
+  },
+  {
+    key: "selectable",
+    value:
+      "Group of components that allow single or multiple selection, Selection can trigger actions.    ",
+  },
+  {
+    key: "seperable",
```
libertymayc, line 47:
```suggestion
    key: "separable",
```

packages/theme/css/cssVariables.json
maleeha13: this file was a test generation so i deleted it

site/docs/characteristics/accordionView.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Characteristic types
+layout: DetailTechnical
+sidebar:
+  label: Characteristic types
+---
+
+<CssAccordionView />
```
JoanaMMoreira, line 8: We already have a site `Accordion` component, could we use that instead?

JoanaMMoreira: Also it would be good to the page's content in mdx as much as possible so anyone can edit it

maleeha13: Most of the content in rendered dynamically vs the static view on storybook, thus why everything is done inside components rather than mdx. So any changes in css vars over time will automatically get updated rather than manually editting it.

site/docs/index.mdx
```diff
@@ -17,6 +17,8 @@ frameOverrides:
         link: /salt/components/index
       - title: Patterns
         link: /salt/patterns/forms
+      - title: Characteristics
```
JoanaMMoreira, line 20: Do we want this to be a stand alone section or should it be in `Foundations`?

site/cssGen.js
```diff
@@ -0,0 +1,51 @@
...
+const jsonData = JSON.stringify(allCssVariables, null, 2);
+const cssFolderPath = path.resolve(
+  __dirname,
+  "../site/src/components/css-display"
+);
+
+const outputPath = path.join(cssFolderPath, "cssVariables.json");
+
+try {
+  fs.writeFileSync(outputPath, jsonData, "utf8");
```
JoanaMMoreira, line 48: Should the output file be added to `.gitignore`?

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,193 @@
...
+const outline: string[] = new Array("outline");
+const shadow: string[] = new Array("shadow");
+const align: string[] = new Array("textAlign");
+const textDecoration: string[] = new Array("textDecoration");
+const transform: string[] = new Array("textTransform");
+const fontFamily: string[] = new Array("fontFamily");
+const lineHeight: string[] = new Array("lineHeight");
+
+const BlockView: React.FC<{ name: string }> = ({ name }) => {
+  if (color.some((c) => name.includes(c))) {
```
JoanaMMoreira, line 57: Would a switch statement work better here?

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,193 @@
...
+      } catch (err) {
+        console.error("Error fetching JSON data:", err);
+      }
+    };
+
+    fetchJsonData();
+  }, []);
+
+  if (!cssVariablesData) {
+    return <Banner>Failed to laod characteristic data</Banner>;
```
JoanaMMoreira, line 109:
```suggestion
    return <Banner>Failed to load characteristic data</Banner>;
```

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,193 @@
...
+    });
+
+    return groupedData;
+  };
+
+  const groupedData = groupByType(cssVariablesData);
+
+  return (
+    <div>
+      <div className="story-root">
```
JoanaMMoreira, line 135: What is this class for?

site/src/components/index.ts
```diff
@@ -1,3 +1,5 @@
+import { ExampleContainer } from "./example-container";
```
JoanaMMoreira, line 1: Is this being used?

site/cssGen.js
maleeha13: i dont think it should matter, because it will overwrite the file if it exists or generate one if it doesn't

site/src/components/css-display/AccordionView.tsx
maleeha13: removed it

site/cssGen.js
origami-z: The file is not *source code*, and the file committed may not reflect the latest information needed. Being source controlled may be misleading

site/src/components/css-display/AccordionView.tsx
libertymayc: i think a switch will be more readable due to the amount here

maleeha13: okie, done :)

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,193 @@
...
+      } catch (err) {
+        console.error("Error fetching JSON data:", err);
+      }
+    };
+
+    fetchJsonData();
+  }, []);
+
+  if (!cssVariablesData) {
+    return <Banner>Failed to load characteristic data</Banner>;
```
JoanaMMoreira, line 111: It's good that we're displayed an error message when there's no data but I can also see this message for a couple seconds when the page is loading so maybe we also need a loading state?

site/src/components/css-display/AccordionView.tsx
```diff
@@ -100,15 +100,22 @@ export const CssAccordionView: React.FC<{ value: string }> = ({ value }) => {
         const data = require("./cssVariables.json");
         setCssVariablesData(data);
       } catch (err) {
-        console.error("Error fetching JSON data:", err);
+        <Banner>Failed to load characteristic data</Banner>;
```
libertymayc, line 103:
```suggestion
        <Banner><BannerContent>Failed to load characteristic data.</BannerContent></Banner>;
```

libertymayc: would need to import BannerContent

site/src/components/css-display/CssTable.tsx
```diff
@@ -19,15 +19,22 @@ export const CssVariablesTable: React.FC = () => {
         const data = require("./cssVariables.json");
         setCssVariablesData(data);
       } catch (err) {
-        console.error("Error fetching JSON data:", err);
+        <Banner>Failed to laod characteristic data</Banner>;
```
libertymayc, line 22: same here with BannerContent

maleeha13: done :))

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+### Token name convention
+
+When defining characteristic design tokens, they should generally be in the format `--salt-<characteristic name>-<variant>-<scale>-<density>`
+
+1. Always begin with `--salt-`
+2. Follow this by the foundation name, e.g. `---salt-actionable-`, `---salt-text-`, `--salt-focused-`
+3. Append the variant if applicable, e.g. `--salt-actionable-cta-`, `--salt-text-h2-`
+4. Append the attribute, e.g. `--salt-actionable-cta-background`, `--salt-text-h2-fontSize`, `--salt-focused-outlineColor`
+5. Append the state if applicable, e.g. `--salt-actionable-cta-background-hover`, `--salt-focused-outlineColor-warning`
```
libertymayc, line 43:
```suggestion
5. Append the state if applicable, e.g. `--salt-actionable-cta-background-hover`
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+```css
+.saltButton {
+  background: var(--salt-actionable-primary-background);
+  font-weight: var(--salt-actionable-fontWeight);
+}
+```
+
+Characteristics are both light/dark and density scoped, meaning that the value the characteristic points to will change according to the values given in the nearest `SaltProvider`. Considering the above, we would get that in light
+theme, --salt-actionable-primary-background resolves to `var(--salt-action-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
```
libertymayc, line 59:
```suggestion
theme, `--salt-actionable-primary-background` resolves to `var(--salt-action-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+```css
+.saltButton {
+  background: var(--salt-actionable-primary-background);
+  font-weight: var(--salt-actionable-fontWeight);
+}
+```
+
+Characteristics are both light/dark and density scoped, meaning that the value the characteristic points to will change according to the values given in the nearest `SaltProvider`. Considering the above, we would get that in light
+theme, `--salt-actionable-primary-background` resolves to `var(--salt-action-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
```
libertymayc, line 59:
```suggestion
theme, `--salt-actionable-primary-background` resolves to `var(--salt-palette-interact-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+  font-weight: var(--salt-actionable-fontWeight);
+}
+```
+
+Characteristics are both light/dark and density scoped, meaning that the value the characteristic points to will change according to the values given in the nearest `SaltProvider`. Considering the above, we would get that in light
+theme, `--salt-actionable-primary-background` resolves to `var(--salt-palette-interact-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
+
+### Simple customization using characteristics
+
+In the following example, we will show how to customize the Card component through characteristics. Some components, like Button and Panel, have different variants such as 'cta' and 'secondary'. Currently, Card only has one variant, and uses tokens such as `--salt-container-primary-background` by default for all of its styling purposes. Here's how the Salt Card looks contained within a secondary Panel:
```
libertymayc, line 63: "Here's how the Salt Card looks contained within a secondary Panel:"

needs image following

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+```css
+.my-app .saltCard {
+  --saltCard-background: var(--salt-accent-background);
+  --saltCard-borderColor: var(--salt-accent-borderColor);
+  --saltCard-color: var(--salt-accent-foreground);
+}
+```
+
+Note how no class name is needed to be passed in. We have just redefined the Card's styling at a higher specificity, to produce the following styling:
```
libertymayc, line 88: needs image following

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+
+It is important to note that just because you have assigned a characteristic to a component, does not mean it consumes _all_ of those characteristics. We can freely pick which characteristics we want. A good example of this may be to notice that the `focused` characteristic offers both a border and outline style. From this you would probably want to use an outline if you're setting the outline attribute, and of course border for the border attribute. Yet, there is nothing stopping you from writing `border-style: var(--salt-focused-outlineStyle)` instead.
+
+Think carefully when deciding which characteristics to use. It's up to you to choose which characteristic is most appropriate - for example, should the background color of your component be coming from container, because that's its general purpose, or should it be coming from selectable, since it can be selected? Many characteristics offer the same attribute (like background color), and it's up to you to decide what part of your component you are styling as to which characteristic it should come from. Taking time to consider this will ensure you have the most consistent styling possible across your app.
```
libertymayc, line 101:
```suggestion
It is important to note that just because you have assigned a characteristic to a component, does not mean it consumes _all_ of those characteristics. We can freely pick which characteristics we want. A good example of this may be to notice that the `focused` characteristic offers both a border and outline style. From this you would want to use the outline token if you're setting the outline attribute, and the border token if setting the border attribute.

Think carefully when deciding which characteristics are most appropriate. For example, should the background color of your component be coming from container, because that's its general purpose, or should it be coming from selectable, since it can be selected? Many characteristics host for the same attribute (e.g. background color), and it's important to make the correct decision as to which characteristic your component should adopt the token from. Taking time to consider this will ensure you have the most consistent styling possible across your app, especially when different themes are applied.
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,110 @@
...
+Think carefully when deciding which characteristics are most appropriate. For example, should the background color of your component be coming from container, because that's its general purpose, or should it be coming from selectable, since it can be selected? Many characteristics host for the same attribute (e.g. background color), and it's important to make the correct decision as to which characteristic your component should adopt the token from. Taking time to consider this will ensure you have the most consistent styling possible across your app, especially when different themes are applied.
+
+## How do I know which characteristics the Salt components are using?
+
+There are two easy ways to check this, depending on what you're interested in:
+
+- Each component has a breakdown of the characteristics used on its component story.
+- Each characteristic has a list of components using one or more of that characteristics token, which can be found at the bottom of the characteristics docs page.
+
+You can of course also dive into the components CSS files to see exactly how they are styled.
```
libertymayc, line 110:
```suggestion
Each component has a breakdown of the characteristics used on its documentation page within Storybook. You can also dive into the components CSS files to see exactly how they are styled.
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,105 @@
...
+Characteristics offer the following benefits:
+
+- Ability to theme the Salt inline with third-parties or CIB sub brand guidelines.
+- Increase consistency across all components, with the knock-on effect of improving the overall user experience e.g. reduced cognitive load.
+- Simplified design of new components and patterns e.g. drag-and-drop behavior.
+- Streamlined evolution – a single change to a characteristic will automatically cascade down through all related components within the library.
+
+### Token name convention
+
+When defining characteristic design tokens, they should generally be in the format `--salt-<characteristic name>-<variant>-<scale>-<density>`
```
libertymayc, line 37:
```suggestion
When defining characteristic design tokens, they should generally be in the format `--salt-<characteristic name>-<variant>-<attribute>-<state>`
```

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,105 @@
...
+For example, a Button is a component on which a user can perform an action, so you could use the following characteristics to style the component:
+
+```css
+.saltButton {
+  background: var(--salt-actionable-primary-background);
+  font-weight: var(--salt-actionable-fontWeight);
+}
+```
+
+Characteristics are both light/dark and density scoped, meaning that the value the characteristic points to will change according to the values given in the nearest `SaltProvider`. Considering the above, we would get that in light
```
libertymayc, line 58: SaltProvider is in code format `SaltProvider` but there are other components mentioned below e.g. Panel, Card in plain text. how do we want to present them?

site/docs/foundations/characteristics/index.mdx
```diff
@@ -0,0 +1,105 @@
...
+theme, `--salt-actionable-primary-background` resolves to `var(--salt-palette-interact-primary-background)`, which in light theme is `var(--salt-color-gray-60)` and in dark theme is `var(--salt-color-gray-300)`, of which foundation tokens then in turn resolve to their respective RGBA values.
+
+### Simple customization using characteristics
```
libertymayc, line 61: I think we should actually delete this whole section until we agree as a team on how this is going to be dealt with (hopefully ASAP, and this will encourage the conversation)

e.g., --saltCard-borderRadius token used doesn't even exist anymore, since this doc was written a while ago, needs to be updated. we need designers to also give input on which attributes we want to expose, if we do go down the CSS API route
cc @bhoppers2008 @pseys

maleeha13: yea for now i had just moved over the docs from storybook,, but lmk what all needs to be removed or changed

libertymayc: just the bit from this heading to the next should be deleted, the rest is okay :) I made minor tweaks where it needed refreshing

maleeha13: the stuff under "Simple customization using characteristics" ?

site/docs/foundations/characteristics/index.mdx
libertymayc: i'm not sure, would be good for content to take a look - cc @jpmorganchase/salt-content

site/docs/foundations/characteristics/index.mdx
```diff
@@ -53,62 +53,6 @@ For example, a Button is a component on which a user can perform an action, so y
...
-<Panel className="my-app" variant="secondary">
-  <Card>
-    <H1 style={{ margin: "0", lineHeight: "1.3em" }}>
-      Card with accent styles
-    </H1>
-    <Text>Here is some content</Text>
-  </Card>
-</Panel>
-
-### How do I know which characteristics to apply?
```
libertymayc: still need this heading

james-nash:
This is looking pretty cool!

Is there a reason we can't have the filter on the characteristic types page and then remove the search page entirely? They each contain the same info, just presented in different ways. Why not combine the best of both into one?

Also, for the tables showing tokens:
* Perhaps the right column should be labelled "_Default_ value", since those are the values our default Salt theme happens to assign to the characteristics. A different theme could have different values
* I'd suggest removing the copy button from the right column. My understanding is that we should be encouraging folks to use characteristic tokens in their designs & code and _not_ hard-coding values. That way their stuff becomes theme/mode/density agnostic. So, if we only make it easy to copy the characteristic, we encourage good behaviour :-)

maleeha13:
Design changes and search will be added later !

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,223 @@
...
+const handleCopyToClipboard = (text: string) => {
+  const textField = document.createElement("textarea");
+  textField.innerText = text;
+  document.body.appendChild(textField);
+  textField.select();
+  document.execCommand("copy");
+  textField.remove();
+};
+
+export const CssAccordionView: React.FC<{ value: string }> = ({ value }) => {
```
origami-z, line 105: Name the file with the same component name if there is only 1 component being exported.

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+    return groupedData;
+  };
+
+  const groupedData = groupByType(cssVariablesData);
+
+  return (
+    <div>
+      {Object.entries(groupedData).map(([groupName, groupData]) => (
+        <Accordion key={groupName} value={groupName}>
```
origami-z, line 159: Misalignment between chevron and heading text

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,223 @@
...
+                            <BlockView name={name} />
+                          </>
+                        }
+                        onOpenChange={function _l() {}}
+                      >
+                        <Link className={styles.tooltip}> {name}</Link>
+                      </Tooltip>
+                      <div className={styles.hideView}>
+                        {name}{" "}
+                        <Button
```
origami-z, line 202: Misalignment between name and button

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,223 @@
...
+                          className={styles.alignButton}
+                          variant="secondary"
+                          onClick={() => handleCopyToClipboard(name)}
+                        >
+                          <CopyIcon />
```
origami-z, line 207: When icon alone is used, prefer to have `aria-label` on button. See button doc.

site/src/components/css-display/AccordionView.tsx
```diff
@@ -0,0 +1,223 @@
...
+    case fontFamily.some((fw) => name.includes(fw)):
+      return <TextBlock fontFamily={name} />;
+    case lineHeight.some((fw) => name.includes(fw)):
+      return <TextBlock lineHeight={name} />;
+    default:
+      return <p>{name}</p>;
+  }
+};
+
+const handleCopyToClipboard = (text: string) => {
```
origami-z, line 96: Not a blocker, would be good to know whether `navigator.clipboard` API would work, and the impact of doing `execCommand` approach (e.g. focus management)

site/src/components/css-display/CssTable.tsx
```diff
@@ -0,0 +1,117 @@
+import React, { useEffect, useState } from "react";
+import { Table } from "../mdx/table";
+import { CopyIcon, SearchIcon } from "@salt-ds/icons";
+import styles from "./Characteristics.module.css";
```
origami-z, line 4: Keep CSS file name the same with tsx as well.

joshwooding (approved):

Fercas123 (approved):

PR #2063, opened 2023-07-26
Add docs around versioning and component milestones

joshwooding:


site/docs/about/component-milestones.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+To help us track the state of a component, we use a set of milestones:
+
+- Legacy
+- Experimental
+- In Progress
+- RC (Release Candidate)
+- Stable
+
+Legacy, Experimental, In Progress and RC are all in `@salt-ds/lab`. Stable components are considered production ready and are in `@salt-ds/core`.
```
libertymayc, line 14: we have RC in core atm?

joshwooding: Hi, you don't need to review this yet. This is around a new approach and it will be shared soon.

libertymayc: np, should it be drafted? :)

site/docs/about/component-milestones.mdx
james-nash, line 1: I appreciate that this is still work-in-progress, but just wanted to offer a little suggestion (since it might also influence the process): It would be great if we can explain to our consumers how they can identify which `lab` components are legacy / experimental / in progress / RC.

I don't know if that's intended to be conveyed via component name, docs, or something else - but whatever it is, having a consistent way of doing it and documenting that here would be really beneficial to consumers IMHO.

PR #2075, opened 2023-07-27
Update Accordion styling

joshwooding:


.changeset/wise-pigs-push.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Update Accordion styling
```
origami-z, line 5: Worth to talk about border bottom -> top, and new `status` prop?

packages/core/src/accordion/AccordionHeader.css
```diff
@@ -6,37 +6,47 @@
...
-  min-height: var(--salt-size-base);
+  min-height: calc(var(--salt-size-base) + var(--salt-spacing-100));
   background: var(--salt-actionable-secondary-background);
   font-size: var(--salt-text-fontSize);
   line-height: var(--salt-text-lineHeight);
   font-family: var(--salt-text-fontFamily);
   color: var(--salt-text-primary-foreground);
-  --saltIcon-color: var(--salt-text-primary-foreground);
+  font-weight: var(--salt-text-fontWeight-strong);
+  --saltIcon-color: currentcolor;
```
origami-z, line 16: Seeing this more often these days makes me think whether this should be the default

packages/core/src/accordion/AccordionPanel.css
```diff
@@ -1,18 +1,20 @@
...
 }
 
 .saltAccordionPanel-inner {
   overflow: hidden;
 }
 
 .saltAccordionPanel-content {
-  padding: var(--salt-spacing-200);
   padding-left: calc(var(--salt-spacing-200) + var(--salt-size-icon));
+  padding-block: var(--salt-spacing-300);
```
origami-z, line 19: Just realize `padding-block` comes in from Chrome 87, may have trouble for min browser conv

origami-z (commented):
Need new story for status?

Also group story may need an update, args are different so storybook controls doesn't work.

packages/core/src/accordion/AccordionHeader.css
joshwooding: I have already brought it up, but haven't been able to chat with a designer

.changeset/wise-pigs-push.md
joshwooding: I did the further updates quite quickly, so this PR will receive some more polish e.g. adding a QA story to have visual regression, updating the docs and adding a story for status.

packages/core/stories/accordion/accordion.stories.tsx
```diff
@@ -14,88 +19,69 @@ export default {
...
     <Accordion {...props}>
-      <AccordionHeader>Mountains and hills</AccordionHeader>
+      <AccordionHeader>Accordion label</AccordionHeader>
       <AccordionPanel>
-        <MountainsAndHills />
+        <FlowLayout>
+          This is content inside of an Accordion.
+          <FormField labelPlacement="left">
+            <FormLabel>Disclosure ID</FormLabel>
+            <Input />
```
origami-z, line 35: All inputs right focus ring is cut off.

origami-z (approved):

packages/core/src/accordion/AccordionHeader.css
```diff
@@ -34,9 +30,37 @@
...
+.saltAccordionHeader-warning:focus-visible {
+  outline-color: var(--salt-status-warning-foreground);
+}
+
+.saltAccordionHeader-success {
+  background: var(--salt-status-success-background);
+}
+
+.saltAccordionHeader-success:focus-visible {
+  outline-color: var(--salt-status-error-foreground);
```
libertymayc, line 61:
```suggestion
  outline-color: var(--salt-status-error-borderColor);
```
for these outlines?

joshwooding: Probably, I was copying what has been done in other components. We'll probably have to update those since this will introduce inconsistency

libertymayc (approved):

PR #2076, opened 2023-07-27
Fix `Drawer` docs 

JoanaMMoreira:
Render Drawer stories in Storybook docs on an iframe instead of inline


ZarkoPernar (approved):

ZarkoPernar:
Is there an issue with having two drawers in one window? Or two floating ui elements?

JoanaMMoreira:
> Is there an issue with having two drawers in one window? Or two floating ui elements?

I've tested both scenarios in the stories and they work fine (maybe I can add this to the tests as well) but the issue seems to be related to the Storybook components (`Canvas` and `Story`)

origami-z (approved):

PR #2099, opened 2023-07-28
remove truncation

ZarkoPernar:


packages/lab/src/badge/Badge.tsx
```diff
@@ -40,33 +40,18 @@ export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
     valueText = value > 999 ? "999+" : value;
   } else if (typeof value === "number" && max) {
     valueText = value > max ? `${max}+` : value;
```
alycrys, line 42: i think it would make sense to default max to 999 and simplify this?

alycrys:
Truncated story would need to be removed also

packages/lab/src/badge/Badge.css
```diff
@@ -1,33 +1,38 @@
...
+  display: inline-flex;
   margin: auto;
 
   font-size: var(--salt-accent-fontSize);
   font-weight: var(--salt-accent-fontWeight);
   background: var(--salt-accent-background);
   color: var(--salt-accent-foreground);
+
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
```
lukeac123, line 30: are we writing to support Firefox as as well as chrome ?

joshwooding: This was already there, I just moved it. But yes, we also support Firefox

packages/lab/src/badge/Badge.css
```diff
@@ -1,33 +1,38 @@
...
 }
 
 .saltBadge-topRight {
-  position: relative;
-  top: -7px;
-  right: 4px;
+  position: absolute;
+  top: 0;
+  right: 0;
+  transform: translate(50%, -50%);
```
lukeac123, line 37: This positioning translates the badge by 50% of the badge size, so as the badge length varies with strings and numerical values the positioning isn't consistent. The way i found to achieve consistency was with relative positioning and displacing by a specified number of pixels.

packages/lab/src/badge/Badge.tsx
```diff
@@ -24,7 +24,7 @@ export interface BadgeProps extends HTMLAttributes<HTMLSpanElement> {
 const withBaseName = makePrefixer("saltBadge");
 
 export const Badge = forwardRef<HTMLSpanElement, BadgeProps>(function Badge(
-  { value, max, className, children, ...rest },
+  { value, max = 999, className, children, ...rest },
```
lukeac123, line 27: Yep specifying 999 as the max value makes it a lot cleaner

packages/lab/src/badge/Badge.css
joshwooding: This positioning scales with density so arguably is better. I'm unsure whether consistency in positioning is an issue here. This needs design consideration anyway

lukeac123: if we offset with one of the spacing vars from the theme, will that no also scale with density and keep the positioning more consistent independent of the content

joshwooding: We've had some guidance from design. I'll ping you

.changeset/loud-trains-remember.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/lab": patch
+---
+
+Remove string truncation from Badge.
+Change default max to `999` for Badge.
```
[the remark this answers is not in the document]
origami-z: Yes, they are used to fill in the props table, especially the `default` column

packages/lab/stories/badge/badge.stories.tsx
```diff
@@ -9,76 +14,63 @@ export default {
...
-  value: 1,
-  children: "Lorem Ipsum",
+export const ListStory: ComponentStory<typeof Badge> = () => {
+  return (
+    <List aria-label="Declarative List example">
+      <ListItem>Level 1</ListItem>
+      <ListItem>Level 2</ListItem>
+      <ListItem>Level 3</ListItem>
+      <ListItem>
+        Level 4<Badge value={"NEW"} />{" "}
```
origami-z, line 72: Is `{" "}` meant to be between `4` and `Badge`?

origami-z (approved):

PR #2101, opened 2023-07-30
Make ToggleButton `value` prop optional instead of `undefined`

origami-z:
In later versions of TypeScript (e.g. 5.1.6 & 4.9.5), `<ToggleButton />` will give below TS error. Update `value` prop to be optional to achieve the same effect.

```
Property 'value' is missing in type '{ children: (string | Element)[]; }' but required in type 'Omit<ToggleButtonProps, "ref">'.ts(2741)
```


joshwooding:
I think `value` should probably just be required, and we delete `undefined` from the union.

origami-z:
> I think `value` should probably just be required, and we delete `undefined` from the union.

Standalone storybook example of toggle button doesn't contain them?

PR #2102, opened 2023-07-31
Nav item

JoanaMMoreira:
- Add nav item props for `blurSelected` state (when a nav item has active children), `IconComponent` and `BadgeComponent`
- Update component styling to match designs
- Remove `ChevronDownIcon` from horizontal expansion button as per designs
- Update storybook examples to match designs


DavieReid:
Can you have a badge and dropdown on the same nav item?

When a nav item dropdown is open, tabbing skips through all nav items before hitting the list.  I think it would be nicer to go into the dropdown...maybe?

JoanaMMoreira:
> Can you have a badge and dropdown on the same nav item?
> 
> When a nav item dropdown is open, tabbing skips through all nav items before hitting the list. I think it would be nicer to go into the dropdown...maybe?

Yeah we can have a badge and a dropdown e.g. 

Screenshot 2023-08-01 at 16 23 04

Good point about the dropdown, I'll make some tweaks

JoanaMMoreira:
> > Can you have a badge and dropdown on the same nav item?
> > When a nav item dropdown is open, tabbing skips through all nav items before hitting the list. I think it would be nicer to go into the dropdown...maybe?
> 
> Yeah we can have a badge and a dropdown e.g.
> 
> Screenshot 2023-08-01 at 16 23 04
> Good point about the dropdown, I'll make some tweaks

@DavieReid should be fixed

Fercas123 (approved):
LGTM

PR #2105, opened 2023-07-31
Adds Badge component docs to site

james-nash:
Adds the Badge component docs from the completed Word template to the site

### Notes for reviewers:

* You can preview this page here: https://saltdesignsystem-git-add-badge-docs-joshwooding.vercel.app/salt/components/badge/usage

### DIFF
--- a/site/docs/components/badge/examples.mdx
+++ b/site/docs/components/badge/examples.mdx
@@ -1,0 +1,38 @@
+---
+# Leave the frontmatter as is
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Examples
+data:
+  $ref: ./#/data
+---
+
+<LivePreviewControls>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Icon badge">
+
+Badge is positioned in the top right corner when placed on a child component, for example with Icon.
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Words Badge">
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Badge with maximum number">
+
+Specify the maximum number to be displayed with `max`. If `value` exceeds `max`, suffix `+` will be appended.
+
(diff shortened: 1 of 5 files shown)


site/docs/components/badge/examples.mdx
```diff
@@ -0,0 +1,44 @@
...
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Badge with string">
+
+`value` can accept both number and string.
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Badge with truncated string">
+
+When a string of greater than 4 characters is passed to `value` it is truncated and appended with `...`.
```
lukeac123, line 38: this truncation has also been removed - same as above will change and push put a new commit

lukeac123 (commented):
Badge characteristics have changed since this was agreed by content. Will push up a new commit with updated characteristics

lukeac123 (approved):

site/docs/components/badge/usage.mdx
```diff
@@ -0,0 +1,38 @@
...
+layout: DetailComponent
+sidebar:
+  label: Usage
+data:
+  $ref: ./#/data
+---
+
+### Using the Badge component
+
+#### When not to use Component Name
```
feliciackh, line 14:
```suggestion
#### When not to use Badge
```

site/docs/components/badge/usage.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+import { Badge } from "@salt-ds/core";
+```
+
+### Props
+
+{/* TODO: Update this to grab from core package when component is released */}
+
+<PropsTable packageName="labs" componentName="Badge" />
```
feliciackh, line 33:
```suggestion
<PropsTable packageName="core" componentName="Badge" />
```
might be a good idea to have it as packageName="core" (in alignment with what's mentioned above in Import section, and all other component docs)

origami-z:
Will be covered by a new PR by #2199

PR #2106, opened 2023-07-31
Adds Stepped Tracker component docs to site

james-nash:
Adds the Stepped Tracker component docs from the completed Word template to the site

### Notes for reviewers:

* You can preview this page here: https://saltdesignsystem-git-add-stepped-tracker-docs-joshwooding.vercel.app/salt/components/stepped-tracker/usage

### DIFF
--- a/site/docs/components/stepped-tracker/examples.mdx
+++ b/site/docs/components/stepped-tracker/examples.mdx
@@ -1,0 +1,34 @@
+---
+# Leave the frontmatter as is
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Examples
+data:
+  $ref: ./#/data
+---
+
+<LivePreviewControls>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Basic Features">
+
+The SteppedTracker contains multiple TrackerStep child components. Each TrackerStep indicates its status through its icon. The connectors indicate current active step progress through the process. You should add a label to the TrackerStep using the StepLabel component as follows:
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Step Progression">
+
+In normal circumstances, once a step is completed you should both advance the active step, and change the status of the current active step in unison.
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Auto Progression">
+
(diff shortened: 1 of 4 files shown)


site/docs/components/stepped-tracker/examples.mdx
```diff
@@ -0,0 +1,34 @@
...
+  label: Examples
+data:
+  $ref: ./#/data
+---
+
+<LivePreviewControls>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Basic Features">
+
+The SteppedTracker contains multiple TrackerStep child components. Each TrackerStep indicates its status through its icon. The connectors indicate current active step progress through the process. You should add a label to the TrackerStep using the StepLabel component as follows:
```
james-nash, line 16: I think these example blurbs (and perhaps the examples themselves?) need re-writing. Unlike the StoryBook docs, where this all gets rendered as a continuous block of content, on the Salt site, individual examples can be viewed stand-alone. Therefore each example + blurb should be self-contained.

Also, examples are rendered _above_ the blurb (so "...as follows:" doesn't make sense - perhaps "...as shown in this example's code" would be better?)

site/docs/components/stepped-tracker/index.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Stepped Tracker
+data:
+  description: The Stepped Tracker visually communicates a user’s progress through a linear process. It gives the user context about where they are in the process, as well as an indication of remaining steps and the state of all steps.
+
+  # Fill in the info from the content template's "Metadata" table below.
+  # To omit optional items, comment them out with #
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/core/src/stepped-tracker"
+  package:
+    name: "@salt-ds/core"
```
el-dav, line 10: FYI for the moment the component is still in labs

site/docs/components/stepped-tracker/usage.mdx
```diff
@@ -0,0 +1,53 @@
...
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Usage
+data:
+  $ref: ./#/data
+---
+
+### Using the Component Name component
```
feliciackh, line 12:
```suggestion
### Using the Stepped Tracker component
```

feliciackh (approved):

james-nash:
Replaced by #2291

PR #2108, opened 2023-07-31
Adds Pill component docs to site

james-nash:
Adds the Pill component docs from the completed Word template to the site

### Notes for reviewers:

* You can preview the page here: https://saltdesignsystem-git-add-pill-docs-joshwooding.vercel.app/salt/components/pill/examples


site/docs/components/pill/usage.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+#### When not to use Pill
+
+- When the content is a system message and cannot be interacted with. Use Badge instead.
+
+### Import
+
+{/* Update the text and code snippet below as needed: */}
+
+To import Component Name from the core Salt package, use:
```
origami-z, line 28: "Component Name"?

site/docs/components/pill/index.mdx
```diff
@@ -0,0 +1,26 @@
...
+data:
+  description: Pill is a small visual element that contains versatile content. You can use Pill to label, tag or categorize content. With Pill, users can trigger actions, make selections or filter results. Actions that may change depending on the context, and are not fixed, are better represented by Pill than Button.
+
+  # Fill in the info from the content template's "Metadata" table below.
+  # To omit optional items, comment them out with #
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/blob/pill-next/packages/core/src/pill/"
+  package:
+    name: "@salt-ds/core"
+  alsoKnownAs: ["Tag", "Chip", "Badge"]
+  relatedComponents: [
```
origami-z, line 12: Should "Contains" display nothing or the text should not be there as well?

site/docs/components/pill/usage.mdx
james-nash: Oops! That's a leftover from the template I forgot to replace. Fixing that now...

site/docs/components/pill/index.mdx
james-nash: The heading should disappear. It's a bug in the component page layout. I'm working on a separate PR to fix that and a few other minor issues I've found on the component page layout.

feliciackh (approved):

james-nash:
Replaced by #2300

PR #2109, opened 2023-07-31
Adds Progress component docs to site

james-nash:
Adds the Component component docs from the completed Word template to the site

### Notes for reviewers:

* You can preview this page here: https://saltdesignsystem-git-add-progress-docs-joshwooding.vercel.app/salt/components/progress/examples


site/docs/components/progress/usage.mdx
```diff
@@ -0,0 +1,43 @@
...
+### Import
+
+{/* Update the text and code snippet below as needed: */}
+
+To import Component Name from the core Salt package, use:
+
+```js
+import { CircularProgress } from "@salt-ds/lab";
+// or
+import { LinearProgress } from "@salt-ds/lab";
```
libertymayc, line 29:
```suggestion
import { CircularProgress } from "@salt-ds/core";
// or
import { LinearProgress } from "@salt-ds/core";
```
assuming this isnt going in until Progress is in core, re the PR description and above comment

libertymayc (commented):
At which point are we transferring teams docs to the site? seems like there's a lot that could come over now.. so curious why Progress has been chosen when there's a few in core ready to go

site/docs/components/progress/usage.mdx
james-nash: Good spot, thanks!

site/docs/components/progress/examples.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Circular progress">
+
+Use linear or circular depending on the context, layout and functionality of an application. Use the same indicator for the same type of process to ensure consistency.
+
+</LivePreview>
+
+<LivePreview componentName="component-name" exampleName="ComponentExample" title="Linear progress">
+
+Use linear or circular depending on the context, layout and functionality of an application. Use the same indicator for the same type of process to ensure consistency.
```
lilyvc, line 22: does this line need to be repeated?

site/docs/components/progress/index.mdx
```diff
@@ -0,0 +1,19 @@
+---
+title: Progress
+data:
+  description: "A Progress indicator gives the user an understanding of how long a system operation will take. It should be used when the operation will take more than a second to complete. Two variants are available to accommodate different layout constraints: linear and circular."
```
origami-z, line 4: The casing of "A Progress indicator" is a little weird, should it be "A progress indicator" or "Progress is an indicator"?

site/docs/components/progress/accessibility.mdx
```diff
@@ -0,0 +1,16 @@
...
+sidebar:
+  label: Accessibility
+data:
+  $ref: ./#/data
+---
+
+Improve accessibility by customizing the `ariaLabel` prop to provide additional context about what is loading. For example:
+
+```jsx
+<LinearProgress ariaLabel="loading settings panel 0%" />
```
origami-z, line 15: `ariaLabel` -> `aria-label`

Given "0%" is used, should `value` be used?

site/docs/components/progress/accessibility.mdx
```diff
@@ -0,0 +1,16 @@
...
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Accessibility
+data:
+  $ref: ./#/data
+---
+
+Improve accessibility by customizing the `ariaLabel` prop to provide additional context about what is loading. For example:
```
origami-z, line 12: `ariaLabel` -> `aria-label`

site/docs/components/progress/examples.mdx
james-nash: I need to check that with the folks who authored this component (was it you by any chance?). The examples can be viewed individually (this is the default view on the component pages, unless the user flips the "All examples" switch). Since this blurb seemed relevant to both examples, I just copied it for now (I did wonder, whether moving it to the "Usage" tab might be better though), but my intent was to flag this and work with the component authors and content team to decide what to do.

site/docs/components/progress/index.mdx
james-nash: True. I just copy-pasted from the Word doc the content team had reviewed. I'm guessing only "Progress" is capitalized since that's the name of the component. But, I think you're "Progress is an indicator..." suggestion might read better. Thoughts @jpmorganchase/salt-content ?

lilyvc: Yeah tbh I copied this from toolkit. Either  "A progress indicator" or "Progress indicators give...." is better

site/docs/components/progress/examples.mdx
lilyvc: okay that makes sense, I wasn't really aware of how the layout would be when writing the word doc.

site/docs/components/progress/index.mdx
lilyvc, line 4:
```suggestion
  description: "A progress indicator gives the user an understanding of how long a system operation will take. It should be used when the operation will take more than a second to complete. Two variants are available to accommodate different layout constraints: linear and circular."
```

site/docs/components/progress/usage.mdx
```diff
@@ -0,0 +1,43 @@
...
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Usage
+data:
+  $ref: ./#/data
+---
+
+### Using the Component Name component
```
lilyvc, line 12:
```suggestion
### Using the Progress component
```

site/docs/components/progress/usage.mdx
```diff
@@ -0,0 +1,43 @@
...
+
+It should be used when the operation will take more than a second to complete, to reassure users that the operation is in progress.
+
+#### When not to use Component Name
```
lilyvc, line 16:
```suggestion
#### When not to use Progress
```

lilyvc (approved):

site/docs/components/progress/accessibility.mdx
lilyvc: 0 is default

feliciackh (approved):

PR #2111, opened 2023-07-31
Adds Drawer component docs

james-nash:
Adds the Badge component docs from the completed Word template to the site

### Notes for reviewers:

* You can preview this page here: https://saltdesignsystem-git-add-drawer-docs-joshwooding.vercel.app/salt/components/drawer/

### DIFF
--- a/site/docs/components/drawer/usage.mdx
+++ b/site/docs/components/drawer/usage.mdx
@@ -1,0 +1,49 @@
+---
+# Leave the frontmatter as is
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  label: Usage
+data:
+  $ref: ./#/data
+---
+
+### Using the Component Name component
+
+Use this component by wrapping it around your content and setting the desired position. For example:
+
+```jsx
+<Drawer isOpen={open} position="right">
+  <DrawerContentExample />
+</Drawer>
+```
+
+Provide a component like a Button so that users can open and close the Drawer.
+
+The Drawer displays a visual overlay, also known as a scrim, over the existing content. It also plays relevant animations when opening and closing. You can disable all of these visuals via props.
+
+#### When to use Drawer
+
+Use Drawer when you need additional content, such as a form or sidebar navigation, to be displayed in a specific area of the screen.
+
+#### When not to use Component Name
+
+If you want your content split between pages or tabs, consider Deck Layout instead.
+
+### Import
+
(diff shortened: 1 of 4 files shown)


site/docs/components/drawer/usage.mdx
```diff
@@ -0,0 +1,49 @@
...
+data:
+  $ref: ./#/data
+---
+
+### Using the Component Name component
+
+Use this component by wrapping it around your content and setting the desired position. For example:
+
+```jsx
+<Drawer isOpen={open} position="right">
```
origami-z, line 17: `isOpen={open}` may be better just be `isOpen` here

site/docs/components/drawer/usage.mdx
```diff
@@ -0,0 +1,49 @@
...
+  <DrawerContentExample />
+</Drawer>
+```
+
+Provide a component like a Button so that users can open and close the Drawer.
+
+The Drawer displays a visual overlay, also known as a scrim, over the existing content. It also plays relevant animations when opening and closing. You can disable all of these visuals via props.
```
origami-z, line 24: Is there a need to point out scrim is not only a "visual" thing but also focus management for accessibility?

site/docs/components/drawer/usage.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+#### When not to use Component Name
+
+If you want your content split between pages or tabs, consider Deck Layout instead.
+
+### Import
+
+{/* Update the text and code snippet below as needed: */}
+
+To import Component Name from the core Salt package, use:
```
origami-z, line 38: "Component Name"?

james-nash: D'oh! Fixing it now (and for all the other MDX conversion PRs too)

site/docs/components/drawer/usage.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+Provide a component like a Button so that users can open and close the Drawer.
+
+The Drawer displays a visual overlay, also known as a scrim, over the existing content. It also plays relevant animations when opening and closing. You can disable all of these visuals via props.
+
+#### When to use Drawer
+
+Use Drawer when you need additional content, such as a form or sidebar navigation, to be displayed in a specific area of the screen.
+
+#### When not to use Component Name
```
libertymayc, line 30:
```suggestion
#### When not to use Drawer
```

PR #2112, opened 2023-07-31
ListNext fix keyboard nav and expose hook setters

Fercas123:
In this PR: 

- The previous change in useList meant that any key not purposefully declared in the hook would be prevented to do it's default, which meant `Tab` had to be explicitly declared.
- exposing `setSelectedItem` and `setHighlightedItem` for controlled usage.
- since `keydownHandler` is not guaranteed to be in a list when controlled, changed type
- The hook is setting highlighted when selected, so the example doesn't need to do it anymore.
- setting selectedItem on input change in the example so deleting from the input deletes the selection.

### DIFF
--- a/packages/lab/src/list-next/useList.ts
+++ b/packages/lab/src/list-next/useList.ts
@@ -238,7 +238,7 @@ export const useList = ({
   };
 
   // takes care of keydown when using keyboard navigation
-  const keyDownHandler = (event: KeyboardEvent<HTMLUListElement>) => {
+  const keyDownHandler = (event: KeyboardEvent<HTMLElement>) => {
     const { key } = event;
     const currentItem = getActiveItem();
     let nextItem = currentItem;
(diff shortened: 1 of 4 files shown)


packages/lab/src/list-next/useList.ts
```diff
@@ -278,6 +278,8 @@ export const useList = ({
           onChange?.(event, { value: nextItem.dataset.value || "" });
         }
         break;
+      case "Tab":
+        break;
       case "PageDown":
       case "PageUp":
       default:
```
origami-z, line 285: I'll probably only do `event.preventDefault();` on the PageUp/Down instead of `default` instead of adding `Tab` logic

origami-z (approved):

joshwooding (approved):

ZarkoPernar (approved):

PR #2116, opened 2023-07-31
Add supported platform docs

mark-tate:
Added a page which describes the `Supported Platforms`, this page can be used to describe browser and platform versions.
Initially just a placeholder to align us all on the minimum browser versions.

Whilst creating this page, noticed that `active` mode was not supported by the site, so you had to change the snapshots to edit a page. With the `.env.local` file, you can not edit the page sources, locally and refresh the browser.

Also, noticed the IntelliJ (.idea) files were not ignored correctly

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,10 +31,6 @@ storybook-static
 
 # misc
 .DS_Store
-.env.local
-.env.development.local
-.env.test.local
-.env.production.local
 .eslintcache
 
 npm-debug.log*
--- a/.gitignore
+++ b/.gitignore
@@ -38,8 +38,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # IDEs
-.idea/*
-!.idea/runConfigurations
+.idea
 .vscode/launch.json
 /.VSCodeCounter
 *.iml
--- a/site/.env.local
+++ b/site/.env.local
@@ -1,0 +1,2 @@
+MOSAIC_ACTIVE_MODE_URL=http://localhost:8080
+MOSAIC_MODE=active
--- a/site/docs/about/supported-platforms.mdx
+++ b/site/docs/about/supported-platforms.mdx
@@ -15,7 +15,7 @@ Dependent on the minimum supported browsers of your application, you may need to
 
 ## React
 
-Salt supports the most recent, stable versions of React, as a minimum we support 16.0.0 but recommend the newest version that you application will support.
+Salt supports the most recent, stable versions of React, as a minimum we support 16.14.0 but recommend the newest version that your application will support.
 
 ## Server Side Rendering (SSR)
 
--- a/site/docs/getting-started/supported-platforms.mdx
+++ b/site/docs/getting-started/supported-platforms.mdx
@@ -1,0 +1,22 @@
+---
+title: Supported Platforms
+layout: DetailTechnical
+---
+
+Salt supports the latest, stable releases of all major browsers and platforms.  
+
+## Browser
+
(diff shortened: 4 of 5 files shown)


site/docs/getting-started/supported-platforms.mdx
joshwooding, line 1: Wondering if this should be an "About" page instead

site/.env.local
```diff
@@ -0,0 +1,2 @@
+MOSAIC_ACTIVE_MODE_URL=http://localhost:8080
```
mark-tate, line 1: when run locally using `yarn serve` Mosaic will pickup changes to the original source docs

This means you do not need to edit snapshot files, which are designed to be immutable.

site/docs/getting-started/supported-platforms.mdx
joshwooding: Sorry, I meant I wonder if it should live under the About section rather than getting started

mark-tate: that works... moved

joshwooding (approved):

.gitignore
```diff
@@ -31,19 +31,14 @@ storybook-static
...
-.env.production.local
 .eslintcache
 
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
 # IDEs
-.idea/*
-!.idea/runConfigurations
```
joshwooding, line 46: This was to support having runConfigurations checked in. Was there any specific issues?

site/docs/getting-started/supported-platforms.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+Dependent on the minimum supported browsers of your application, you may need to provide your own polyfills.
+
+| Chrome | Firefox | Edge  | Safari (macOS) | Safari (iOS) | IE                |
+| ------ | ------- | ----- | -------------- | ------------ | ----------------- |
+| >= 86  | >= 86   | >= 85 | >= 15.4        | >= 15.4      | Un-supported / 11 |
+
+## React
+
+Salt supports the most recent, stable versions of React, as a minimum we support 16.0.0 but recommend the newest version that you application will support.
```
origami-z, line 18: 16.14.0 is minimum, see #2002

site/docs/about/supported-platforms.mdx
```diff
@@ -0,0 +1,22 @@
...
+
+Dependent on the minimum supported browsers of your application, you may need to provide your own polyfills.
+
+| Chrome | Firefox | Edge  | Safari (macOS) | Safari (iOS) | IE                |
+| ------ | ------- | ----- | -------------- | ------------ | ----------------- |
+| >= 86  | >= 86   | >= 85 | >= 15.4        | >= 15.4      | Un-supported / 11 |
+
+## React
+
+Salt supports the most recent, stable versions of React, as a minimum we support 16.4.0 but recommend the newest version that you application will support.
```
joshwooding, line 18:
```suggestion
Salt supports the most recent, stable versions of React, as a minimum we support 16.14.0 but recommend the newest version that your application will support.
```

origami-z (approved):

JoanaMMoreira (approved):

.gitignore
mark-tate: removed for now as undesirable to maintain

site/docs/about/supported-platforms.mdx
mark-tate: done

PR #2120, opened 2023-08-01
Fix Linear Progress's bar not animating to the beginning on reset

lilyvc:
Fix to stop the progress bar from disappearing on reset (instead of animating back to the start)

### DIFF
--- a/.changeset/dull-goats-vanish.md
+++ b/.changeset/dull-goats-vanish.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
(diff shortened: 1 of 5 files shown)


joshwooding (changes requested):
The bar is visible when the progress is 0. Seems like the styling need to be updated.

lilyvc:
> The bar is visible when the progress is 0. Seems like the styling need to be updated.

made a change but can't tell if there's a difference

joshwooding (approved):

PR #2135, opened 2023-08-02
Navigation item rename and component rendering

JoanaMMoreira:
- Rename `NavItem` to `NavigationItem`
- Change component so it either renders a `Button` or a `Link` depending on whether it has children
- Rename `ExpansionButton` to `ExpansionIcon` as this is no longer a button
- Update storybook examples and remove the following:
`HorizontalGroupWithDropdown`,  `NestedCardGroup` and `HorizontalExpandableGroup`
- Small styling tweaks
- Add storybook docs


packages/lab/src/navigation-item/ExpansionIcon.tsx
```diff
@@ -14,17 +13,13 @@ const iconExpansionMap = {
...
 
-export function ExpansionButton({
+export function ExpansionIcon({
   expanded = false,
   orientation = "horizontal",
-  ...rest
-}: Pick<NavItemProps, "expanded" | "orientation"> &
+  className,
+}: Pick<NavigationItemProps, "expanded" | "orientation"> &
   ComponentPropsWithoutRef<"button">) {
```
feliciackh, line 21: think we might not need this line now that button is no longer present

packages/lab/stories/navigation-item/navigation-item.doc.stories.mdx
```diff
@@ -0,0 +1,126 @@
+import { ArgsTable, Canvas, Meta, Story } from "@storybook/addon-docs";
+import { NavigationItem } from "@salt-ds/lab";
+import { withFlexGap } from "docs/decorators/withFlexGap";
```
Fercas123, line 3: nit: unused import

Fercas123 (approved):

PR #2137, opened 2023-08-02
Fix Accordion padding

joshwooding:


joshwooding:
This PR doesn't need a Chang set because Accordion hasn't gone out yet

libertymayc (approved):

PR #2138, opened 2023-08-02
Fix Switch disabled thumb

joshwooding:


libertymayc (commented):
looks good, needs changeset?

joshwooding:
> looks good, needs changeset?

This is fix something that hasn't gone out yet so doesn't need a change set

libertymayc (approved):

libertymayc:
oh, I think myself (and others) have been always adding even when not released

lukeac123 (approved):

PR #2150, opened 2023-08-07
Data Grid scrolling fixes

### LINKED ISSUE #2064: Data grid not scrolling with mouse wheel
### Package name(s)

Data Grid (@salt-ds/data-grid)

### Package version(s)

Current

### Description

Can't scroll data grid using the mouse wheel, only by finding the scroll bar on the edge

### Steps to reproduce

Attempt to scroll data grid using mouse wheel

### Expected behavior

Scroll using the mouse wheel

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### LINKED ISSUE #2156: DataGrid with scrollbar blocks page scrolling by mousewheel
### Package name(s)

Data Grid (@salt-ds/data-grid)

### Package version(s)

Current

### Description

While hovering over a DataGrid which itself has a scrollbar, all page scrolling is disabled.

Note, a separate issue exists to cover the issue of a DataGrid with _no_ scrollbar from preventing page scrolling: #2064

### Steps to reproduce

1. Hover your mouse/touchpad cursor above a DataGrid that has a scrollbar
2. Use the mousewheel (or equivalent - e.g. 2 finger swipe on touchpad) to scroll up or down

The DataGrid rows will scroll (as expected). However, once you've reached the start or end of the rows, and try scroll past them, nothing happens.

### Expected behavior

Scrolling past the first or last row of the grid should probably make the page scroll up or down.

(Need to confirm with design team, but that is the usual behaviour for any kind of nested, scrollable region of a page)

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [X] Firefox
- [ ] Edge

libertymayc:
closes #2064 
closes #2156

tests attempted:
- Using [scrollTo](https://docs.cypress.io/api/commands/scrollTo) by targeting grid-scrollable, grid-middle-part, and entire container, using %s and px's, to scroll past the grid. Attempted with multiple grid sizes, making grid full height of viewport, making grid full width+height of viewport, etc
- Attempted hovering/clicking over grid and then scrolling
- Using next div 'should.be.visible' doesn't work, because of how visibility is tested in cypress, if it is possible that next div can be visible (ie user can scroll to it) then it will pass, so tried to add custom commands isInViewport/isNotinViewport but after trying to scroll past its not visible... due to next point:
- Key note from docs: "It is [unsafe](https://docs.cypress.io/guides/core-concepts/retry-ability#Only-queries-are-retried) to chain further commands that rely on the subject after .scrollTo()"
- Putting onKeyDown into container div and testing if that was called when attempting to scroll past grid
- Using scrollIntoView of next div that is out of viewport, but that is obviously just going to scroll it into view regardless 
- Catching if event listener default/propagation was called or not (struggled with being able to do this in cypress with our impl)


origami-z (commented):
Can we add a test?

libertymayc:
> Can we add a test?

have been trying to write a test that fails on the old version and passes on this but haven't been successful (on making it fail for old) so far - will keep trying

libertymayc:
@jpmorganchase/salt-developer after 2+ days I'm struggling to write a test that works (current one in files is one attempt of many..) does anyone have any ideas on how this can definitely work in cypress?

packages/data-grid/src/Grid.tsx
```diff
@@ -421,9 +421,19 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
       if (s) {
         s.scrollLeft += deltaX;
         s.scrollTop += deltaY;
+
+        if (
+          !(
+            (
+              s.clientHeight - s.scrollHeight - s.scrollTop <= 0 || // reached the bottom
```
joshwooding, line 428: scrollTop was a decimal for me and caused the old logic to break. This would avoid that

origami-z (dismissed):

packages/data-grid/src/Grid.tsx
libertymayc: logic not working for me and @origami-z , perhaps it would work if we rounded it?

packages/data-grid/src/Grid.tsx
```diff
@@ -421,9 +421,18 @@ export const Grid = function Grid<T>(props: GridProps<T>) {
       if (s) {
         s.scrollLeft += deltaX;
         s.scrollTop += deltaY;
+        if (
+          !(
+            (
```
feliciackh, line 426: would it be beneficial to add a comment here to detail what the next few lines check for - for future dev references

Fercas123 (approved):

origami-z (approved):

feliciackh (approved):

Issue #2145, opened 2023-08-05
Checkbox and Radio has gap within selected square on Safari


origami-z:
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

-

### Description

Checkbox has a vertical gap within square. 

<img width="182" alt="image" src="https://github.com/jpmorganchase/salt-ds/assets/5257855/57ce5f89-72c8-4743-976a-e44878cbb755">
<img width="156" alt="image" src="https://github.com/jpmorganchase/salt-ds/assets/5257855/869eb160-f85a-4009-81dc-9c33d57c21e2">

When zoomed in (Cmd+=, 3-4 times)

<img width="177" alt="image" src="https://github.com/jpmorganchase/salt-ds/assets/5257855/6c5d5690-7054-4d98-b754-59f12e38c65f">

Also on Radio Button

<img width="144" alt="image" src="https://github.com/jpmorganchase/salt-ds/assets/5257855/4c2ac3ef-92a1-4358-837f-b8a1473480d7">

### Steps to reproduce

Go to checkbox doc storybook: https://storybook.saltdesignsystem.com/?path=/docs/documentation-core-checkbox--page
or radio button doc storybook: https://storybook.saltdesignsystem.com/?path=/story/documentation-core-radio-button--page

MacOS 13.4.1 (c) (22F770820d)
Safari Version 16.5.2 (18615.2.9.11.10)

### Expected behavior

No visual gap

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [X] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

james-nash:
Bug ranking score: **40** (using rubric v3)

Priority: **medium**

```
========== BEGIN BUG RUBRIC SCORE DATA ===========
eyJ2YWx1ZXMiOnsiYWZmZWN0ZWRQZW9wbGVPcHRpb25zIjoyNS
wiYWZmZWN0ZWRDb21wb25lbnRzT3B0aW9ucyI6MTIsImZyZXF1
ZW5jeU9wdGlvbnMiOjE5LCJpbXBhY3RPcHRpb25zIjoxLCJhZm
[...]

PR #2157, opened 2023-08-08
Fix svg positioning bug when used as icon

Fercas123:
Fixes [issue 2145](https://github.com/jpmorganchase/salt-ds/issues/2145)


joshwooding (dismissed):
Needs a changeset :)

JoanaMMoreira (approved):

Issue #2065, opened 2023-07-26
Data grid error on last page of paginated view


shinytoyrobots:
### Package name(s)

Core (@salt-ds/core), Data Grid (@salt-ds/data-grid)

### Package version(s)

Current

### Description

Error thrown on last page of paginated data table

### Steps to reproduce

Using a paginated grid:

- Scroll to the bottom of the grid
- Navigate to the last page of the grid
- The Grid will crash

### Expected behavior

Should correctly display the data grid on the final page

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

shinytoyrobots:
@joshwooding - from current conversation with James Sol.

joshwooding:
We'll need to contact James Sol for more information on how to reproduce this

solcratius:
Thanks for looking into this.

https://adl-sandbox.apps.test.na-1z.gap.jpmchase.net/#/data-grid-page

In this page Datagrid is scrollable with in. Scroll all the way down in data grid to see the last row of data and then click on page 4 (last page of the grid). Error will appear. But when you don't scroll down and go to the last page it will show the last few rows of data correctly.

https://bitbucketdc.jpmchase.net/projects/DXD/repos/access-design-library/browse/packages/sandbox/src/data-grid-page/index.tsx?at=refs%2Fheads%2Fdevelop

james-nash:
James Sol has shared more details with me. We can't post them here, as they reference internal code. I'll share on Teams internally, but can also email a copy to whoever investigates this one.

joshwooding:
@feliciackh I've added reproduction steps here

james-nash:
Has this been fixed yet?

libertymayc:
@james-nash  #2171 has been ready for a while but still pending getting it merged in

PR #2159, opened 2023-08-08
Data Grid Pagination Bug Fix - #2065

lukeac123:
Error: when scrolling to the bottom of a page and then clicking on a page where scroll is not enabled, (e.g a page with a single row), causing error in NumberRange class where start > end.
Fix: Reset scrollTop = 0 when rowData changes, e.g when pagination occurs

Reproduce Error: 
pagination story > scroll to bottom of grid > click page 8 > error


feliciackh (commented):
could you share the way to reproduce the error for reviewers? would like to check as no idea why the +1 was there myself 
edit: [it's here!](https://github.com/jpmorganchase/salt-ds/issues/2065)

PR #2160, opened 2023-08-08
Add zindex foundation

pseys:


site/docs/foundations/zindex.mdx
```diff
@@ -0,0 +1,77 @@
...
+`--salt-zIndex-notification` allows an object to sit over the entire user interface in order to provide a status update or message, such as [Toast](https://storybook.saltdesignsystem.com/?path=/docs/core-toast--default).  
+
+## Drag object
+`--salt-zIndex-dragObject` is intended to be used by any object that can be lifted off of the app layout and dragged over the interface.
+
+## Context menu
+`--salt-zIndex-contextMenu` a menu which appears over the UI on right-click.
+
+## Flyover
+`--salt-zIndex-contextMenu` an object that appears over the UI on hover.
```
joshwooding, line 59:
```suggestion
`--salt-zIndex-flyover` an object that appears over the UI on hover. 
```

site/docs/foundations/zindex.mdx
```diff
@@ -0,0 +1,77 @@
...
+| `--salt-zIndex-popout`          | 1,000        |
+| `--salt-zIndex-default`         | 1            |
+
+## Default
+`--salt-zIndex-default` represents the lowest layer in the stack. Its intended for any object that sits within the application layout, not over it. 
+
+## Popout
+`--salt-zIndex-popout`is used for any object that appears over the app layout in response to a user interaction and remains rooted to the element that triggered it, such as a [Tooltip](https://storybook.saltdesignsystem.com/?path=/docs/documentation-core-tooltip--page). 
+
+## Docked
```
joshwooding, line 37: I'm not really sure what this would be used for

pseys: Yes, I have no idea what Docked is used for! Do we have anyway of finding where its used in code? If we don't have a use for it we can remove it from the foundation guidance and as a token.

pseys:
@bhoppers2008 @joshwooding I've put Tooltip as an example of Popover, but I'm not clear on the difference between Popout and Flyover, and if I've used the correct example. Do you know when they're both used and the reasoning for both?

site/docs/foundations/zindex.mdx
```diff
@@ -0,0 +1,73 @@
...
+
+A component with a greater z-index, or stack order, will always appear in front of an element with a lower stack order.
+
+Z-index is often used in combination with [Shadow](shadow) to help distinguish each layer in the stack.
+
+## Z-Index tokens
+
+| Token                           | Value        |
+| ------------------------------- | ------------ |
+| `--salt-zIndex-flyover`         | 1,500        |
```
joshwooding, line 20: Not sure where this comes from?

site/docs/foundations/zindex.mdx
```diff
@@ -0,0 +1,73 @@
...
+| `--salt-zIndex-contextMenu`     | 1,450        |
```
joshwooding, line 21: I'm not sure we need this. We probably could use `popout`

site/docs/foundations/zindex.mdx
```diff
@@ -0,0 +1,73 @@
...
+  status: In progress
+---
+
+Z-index determines the stacking order of elements. By providing consistent z-indexes, the design system ensures a predictable hierarchy of visual elements when stacked on top of one another. 
+
+A component with a greater z-index, or stack order, will always appear in front of an element with a lower stack order.
+
+Z-index is often used in combination with [Shadow](shadow) to help distinguish each layer in the stack.
+
+## Z-Index tokens
```
joshwooding, line 16: I'm a bit worried that we have a lot of z-index tokens here that are quite specific to UI elements

pseys: I agree. Lets discuss this in tomorrows call.

Issue #1575, opened 2023-04-24
Simplify/prettify the URLs (post Mosaic rearchitecture)


shinytoyrobots:
The current url structure - https://www.saltdesignsystem.com/salt/index - should be corrected so that we don't have those additional namespace suffixes.

james-nash:
Indeed. This is a [known limitation of Mosaic](https://github.com/jpmorganchase/salt-ds/pull/1481#issuecomment-1497673782). I'll get something added to their backlog to make cleaner URLs possible.

DavieReid:
This has been addressed in Mosaic a while ago --> https://github.com/jpmorganchase/mosaic/pull/311

It was released in version beta 30.

james-nash:
Just tried this locally and it's working great.

Any suggestions on how to remove the trailing `index` parts though? I tried mucking about with `rewrites` in `next.config.js` to map URL path like `/foo/bar` to something like `/foo/bar/index`, but wasn't able to get it to work properly (disclaimer, I'm a novice at Next.js, so I may well have been doing it wrong).

james-nash:
Thanks. I'll write up an issue shortly.

james-nash:
@DavieReid done: https://github.com/jpmorganchase/mosaic/issues/442

PR #2161, opened 2023-08-08
Make roadmap link on components page more robust

james-nash:
Currently, both of these URLs will display our components page:

* `https://www.saltdesignsystem.com/salt/components/index`
* `https://www.saltdesignsystem.com/salt/components`

Since the "roadmap" link under "Need a component that's not listed?" uses a relative URL, it works for the first URL but leads to a 404 on the 2nd URL.

(Really, the proper fix would be to ensure that we always redirect to single, canonical URL for each page, but for now we can avoid this scenario by just using an absolute path in the roadmap link)


Fercas123 (approved):

feliciackh (approved):

JoanaMMoreira (approved):

DavieReid:
Would you consider this a mosaic issue?

james-nash:
> Would you consider this a mosaic issue?

Not quite sure.

The fact that `/salt/components` isn't an HTTP redirect to `/salt/components/index` and instead is a 200 response with the actual page, might be an issue with our redirect configuration which would be more of a Next.js thing (Mosaic is built on top of Next.js).

Note that the `/salt/` prefix and `index` suffix that gets wrapped around to most of the Salt site's URL paths are definitely due to how Mosaic works. They're mostly harmless, but I'm a stickler for clean URLs, so I'd love to be able to eliminate them at some point (and we do have an [open issue covering that](https://github.com/jpmorganchase/salt-ds/issues/1575)). A side-benefit of that we wouldn't need to fiddle about with redirects and brittle links like the one fixed by this PR would be avoidable.

We work really closely with the Mosaic team, so I'm confident that's a matter of when not if. It's just not a huge priority right now. Shipping more Salt components is a lot more valuable to our consumers than prettifying URLs :-)

DavieReid:
> > Would you consider this a mosaic issue?
> 
> Not quite sure.
> 
> The fact that `/salt/components` isn't an HTTP redirect to `/salt/components/index` and instead is a 200 response with the actual page, might be an issue with our redirect configuration which would be more of a Next.js thing (Mosaic is built on top of Next.js).
> 
> Note that the `/salt/` prefix and `index` suffix that gets wrapped around to most of the Salt site's URL paths are definitely due to how Mosaic works. They're mostly harmless, but I'm a stickler for clean URLs, so I'd love to be able to eliminate them at some point (and we do have an [open issue covering that](https://github.com/jpmorganchase/salt-ds/issues/1575)). A side-benefit of that we wouldn't need to fiddle about with redirects and brittle links like the one fixed by this PR would be avoidable.
> 
> We work really closely with the Mosaic team, so I'm confident that's a matter of when not if. It's just not a huge priority right now. Shipping more Salt components is a lot more valuable to our consumers than prettifying URLs :-)

prefixDir has been optional in Mosaic since Beta30 so you could remove "salt" from the URL.

You could also make use of the Mosaic `BrokenLinksPlugin` to catch broken urls like this.  This isn't currently documented but the source is here https://github.com/jpmorganchase/mosaic/blob/main/packages/plugins/src/BrokenLinksPlugin.ts.  I can help you set it up if you need assistance.

james-nash:
> prefixDir has been optional in Mosaic since Beta30 so you could remove "salt" from the URL.

Ooooh! Nice!

And, yes, I'd love to take you up on that offer of setting up the BrokenLinksPlugin. Thanks @DavieReid

PR #2167, opened 2023-08-09
Fix images having a filter in dark and removed the url prefix

joshwooding:


joshwooding:
> Site Preview keeps spinning

Yep, I have made a fix in Mosaic for this

libertymayc:
what's happening with this PR? :) @joshwooding

james-nash:
The image issue was fixed in a separate PR, so those changes should be removed from this PR. Removing the `/salt/` prefix on the URLs would still be worthwhile, but - once this PR has been updated with the latest stuff in `main` - we might need to check if any other links need updating too (perhaps some component pages link to other pages in the site?).

james-nash:
Another thing is that it would be nice to remove those `index` suffixes on the overview pages. That requires [a change to Mosaic](https://github.com/jpmorganchase/mosaic/issues/442) though. If that's likely to be implemented soon, it may be worth waiting and then fully cleaning up our URLs (i.e. remove `salt` prefix _and_ `index` suffix at the same time) in one go. But if that's likely to take a long time, then maybe we get this fix in now and deal with the `index` stuff at a later date.

Either way, when these changes go live, it would be nice to update any internal Go links or links on Confluence pages accordingly (yes, I know we have redirects in place, but it would cleaner to use the canonical URLs wherever we can)

PR #2171, opened 2023-08-09
Fix Data Grid crashing when a user scrolls to the bottom and navigates to the last page

### LINKED ISSUE #2065: Data grid error on last page of paginated view

lukeac123:
Error: when scrolling to the bottom of a page and then clicking on a page where scroll is not enabled, (e.g a page with a single row), causing error in NumberRange class where start > end.
Fix: Add guard when start > end to return empty Number range positioning the VisibleRowRange at the top

Reproduce Error:
pagination story > scroll to bottom of grid > click page 8 > error

fixes #2065

further investigation created #2321


packages/data-grid/src/internal/gridHooks.tsx
```diff
@@ -185,18 +185,26 @@ export function useVisibleRowRange(
...
     if (start === 0) {
       endPos -= 1;
     }
     const end = Math.min(
       rowCount,
       Math.max(start, Math.ceil(endPos / rowHeight))
     );
+
+    // Scroll Top not returning to 0 after pagination. Guard to ensure issues with ScrollTop throw error
+    if (start > end) {
```
libertymayc, line 204: wondering if it should be >=? also will need cs :)

lukeac123: NumberRange doesn't throw an error when when start = end so i think either is fine
cs = cypress test ?

joshwooding: a changeset

lukeac123: added a changeset - doesn't seem to have been picked up by the github action. Am i doing something wrong ?

libertymayc: I think it automatically updates when you add one, not sure why it hasn't, do you know @joshwooding ? could just be taking time

joshwooding: Not sure, it should just work automatically

.changeset/kind-humans-explode.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/data-grid": patch
+---
+
+bug - error due to scrollTop not resetting on each pagination causing start > end when passed into NumberValue
+fix - guard to return an empty NumberValue when start > end preventing an error when scrollTop doesn't reset
```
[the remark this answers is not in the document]
libertymayc: something like this is best for consumers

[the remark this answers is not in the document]
lukeac123: does a changeset not also help to provide visibility to developers without them needing to dig into the code ?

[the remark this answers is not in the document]
libertymayc: hmm, I think the general user wouldnt care how it was fixed just that it is fixed, and if they do then would probably dig into it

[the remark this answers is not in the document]
joshwooding: The changeset is meant to explain to a consumer what has changed and how it would impact them. For this change I would probably stick with:

```suggestion
Fixed the Data Grid crashing when a user scrolls to the bottom and navigates to the last page
```

.changeset/kind-humans-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/data-grid": patch
+---
+
+Fixed the Data Grid crashing when a user scrolls to the bottom and navigates to the last page
```
origami-z, line 5: should pagination be mentioned somewhere here?

origami-z (commented):
Add a new bug for underlying issue and mention it in PR description

lilyvc (dismissed):

.changeset/kind-humans-explode.md
joshwooding: It's sort of inferred when it says "navigates to the last page". You could add "using the pagination" at the end I guess to make it clearer.

joshwooding, line 5:
```suggestion
Fixed the Data Grid crashing when a user scrolls to the bottom and uses the pagination to navigate to the last page.
```

origami-z (approved):

libertymayc (approved):

PR #2174, opened 2023-08-10
Fix inverted images on Salt site

james-nash:
Adds a temporary CSS workaround for Mosaic CSS which is inverting images on our site. Once they release a fix, we can remove this again.

### DIFF
--- a/site/src/css/utilities/patches.css
+++ b/site/src/css/utilities/patches.css
@@ -1,6 +1,6 @@
 /* Temporary (hopefully!) patches and work-arounds to override/undo issues from 3rd party CSS */
 
 /* Prevent Mosaic theme from inverting all images */
-.salt-theme[data-mode=dark] main img:not(.no-filter) {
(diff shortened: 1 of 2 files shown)


site/src/css/utilities/patches.css
```diff
@@ -0,0 +1,6 @@
+/* Temporary (hopefully!) patches and work-arounds to override/undo issues from 3rd party CSS */
+
+/* Prevent Mosaic theme from inverting all images */
+[data-mode=dark] main img:not(.no-filter) {
```
joshwooding, line 4: Is this specific enough? In my change, I use this:

```suggestion
.salt-theme[data-mode=dark] main img:not(.no-filter) {
```

so no matter the CSS order, it will be more specific

joshwooding (changes requested):

joshwooding (approved):

feliciackh (approved):

Issue #2179, opened 2023-08-11
Update Input read-only state


bhoppers2008:
### Description
The current Salt Input has the following attributes in a read-only state:

- No padding-left/-right
- Is not focusable.

This was a decision based on design exploration and ADA SME feedback from the ETF Instruct team. However, the decision causes alignment issues and is difficult to justify to the teams that query the decision.

As a result, this issue will:

- Add spacing-100 padding to left and right of Input in read-only state.
- Allow read-only state to be focusable by default.

These changes will apply to Input and any dependant/ or related components, including:

- Input
- Combo Box (dependant)
- Multiline (related)
- Dropdown (dependant)

### Tasks
- [ ] Update Input code 
- [ ] Update Multiline code @libertymayc 
- [ ] Align Combo Box code @Fercas123 
- [ ] Align Dropdown code @feliciackh 
- [ ] Update all Figma specs @bhoppers2008 
- [ ] Update all Figma components @bhoppers2008

libertymayc:
"Allow read-only state to be focusable by default."
Do we want to add a prop to make it optionally not focusable?

libertymayc:
added also for CB/RB @joshwooding

joshwooding:
> added also for CB/RB but its not in the description - just checking that was intended? @joshwooding @bhoppers2008

The behaviour should be consistent across controls

PR #2184, opened 2023-08-11
Make readonly focusable in Input and MultilineInput

libertymayc:
related #2179 
Q for dev reviewers:  Do we want to add a prop to make it optionally not focusable?


packages/core/src/input/Input.tsx
```diff
@@ -196,11 +196,11 @@ export const Input = forwardRef<HTMLDivElement, InputProps>(function Input(
         readOnly={isReadOnly}
         ref={inputRef}
         role={role}
-        tabIndex={isReadOnly || isDisabled ? -1 : 0}
+        tabIndex={isDisabled ? -1 : 0}
         onBlur={handleBlur}
         onChange={handleChange}
         // From ADA: read-only field doesn't need to be focusable since it's not a field but text
```
feliciackh, line 202: might not need this comment anymore

feliciackh (approved):

Fercas123 (approved):

origami-z:
> Q for dev reviewers: Do we want to add a prop to make it optionally not focusable?

I don't think we need. Focusable readonly is aligned with HTML `input`. If the code works as intended, the user should be able to pass in  `tabIndex` themselves to override when see fit

PR #2227, opened 2023-08-14
Change Navigation Item alignment and update docs/examples

JoanaMMoreira:
- Change `NavigationItem` flex alignment so it works well with longer labels
- Rename `blurSelected` prop to `blurActive`
- Add new `--salt-size-indicator` to `--navigationItem-bar-size` css var
- Update description on `blurSelected` prop
- Move `ConditionalWrapper` outside `NavigationItem`
- Add longer label example to storybook
- Add `FlexLayout` to storybook examples
- Update docs following content review


origami-z (commented):
When using keyboard 'space' to toggle group expansion, should focus remain on the parent item given no navigation would happen?

JoanaMMoreira:
> When using keyboard 'space' to toggle group expansion, should focus remain on the parent item given no navigation would happen?

Good point, I'll amend it

packages/lab/src/navigation-item/NavigationItem.tsx
```diff
@@ -89,13 +91,20 @@ export const NavigationItem = forwardRef<HTMLDivElement, NavigationItemProps>(
       onExpand?.(event);
     };
 
+    const buttonRef = useRef<HTMLButtonElement | null>(null);
+
+    useEffect(() => {
+      expanded && buttonRef.current?.focus();
```
origami-z, line 97: I don't think this is the solution to focus being lost.

It's very likely due to `ConditionalWrapper` where a new Button is created. Can try refactor the code so `Button` stays the same

JoanaMMoreira: Should be fixed now 😄

Fercas123 (dismissed):

packages/lab/src/navigation-item/NavigationItem.tsx
```diff
@@ -50,9 +50,48 @@ export interface NavigationItemProps extends ComponentPropsWithoutRef<"div"> {
...
+interface ConditionalWrapperProps
+  extends Pick<
+    NavigationItemProps,
+    "parent" | "expanded" | "onExpand" | "active" | "href"
+  > {
   children: ReactNode;
   className: string;
+}
+
+const ConditionalWrapper = ({
```
alycrys, line 62: would it make sense to have it in a separate file?

.changeset/empty-tips-thank.md
```diff
@@ -0,0 +1,8 @@
+---
+"@salt-ds/lab": patch
+---
+
+- Change `NavigationItem` flex alignment so it works well with longer labels
+- Add new `--salt-size-indicator` to `--navigationItem-bar-size` css var
```
alycrys, line 6: don't think we need to mention the css changes. Or do we?

alycrys (dismissed):

.changeset/empty-tips-thank.md
JoanaMMoreira: probably not, I'll remove this

alycrys (approved):

PR #2181, opened 2023-08-11
Controlled combobox

Fercas123:
In This PR: 
- changed `itemRenderer` for `ListItem`
- added examples for controlled and input props
- fixed and ordered controlled props 
- fixed skipped test

PR #2233, opened 2023-08-14
Add onSelect to List Next

Fercas123:
Splitting https://github.com/jpmorganchase/salt-ds/pull/2181#issue-1846550676
into more contained changes.

in this PR: 
- handle onSelect for list. 

Currently the difference is not much but it will be once we have multiSelect, as onChange would return the full selectedItem and onSelect will return only the selection.


packages/lab/src/list-next/useList.ts
```diff
@@ -21,8 +21,10 @@ export interface UseListProps {
   selected?: string;
   /* Initial selected value for when the list is controlled. */
   defaultSelected?: string;
-  /* Callback for when the list is controlled. */
-  onChange?: (e: SyntheticEvent, data: { value: string }) => void;
+  /* Callback for when the list is controlled. Returns the event and full selection. */
```
origami-z, line 24: Not controlled only

packages/lab/src/list-next/useList.ts
```diff
@@ -173,10 +176,11 @@ export const useList = ({
         activeOptions.findIndex((i) => i.id === event.currentTarget.id) !== -1;
       if (newValue && selectedItem !== newValue && isActiveOption) {
         selectItem(event.currentTarget);
-        onChange?.(event, { value: newValue });
+        onChange?.(event, { value: selectedItem });
+        onSelect?.(event, { value: newValue });
```
origami-z, line 180: I think onSelect should fire no matter new item or not

origami-z (commented):
Can add a test to cover new prop?

packages/lab/src/combo-box-next/ComboBoxNext.tsx
```diff
@@ -97,7 +97,10 @@ export interface ComboBoxNextProps<T>
    */
   itemFilter?: (source: T[], filterValue?: string) => T[];
   /* Callback for change event in input. */
-  onChange?: (event: SyntheticEvent, data: { value: string }) => void;
+  onChange?: (
```
libertymayc, line 100: when I was playing around, I wonder why we split it into data: {value}, rather than just value, seems a bit unnecesssary to have to do that data.value all the time
but then also if its just going to be the event.target.value you can get from the event anyway, can we keep it simpler and just have onChange: (event) => void?

Fercas123: Given the keyboard and click events are different, the value is in a different place in the event, so extracting the value saves you from having to check outside if it was from the keyboard or mouse to check for target or currentTarget.
about the data object, i would imagine it is to standardise outputs of handlers for cases in which we might send more than one parameter... idk if we do it anywhere else, i guess we could just send `value` but might be a good one for today's discussion

origami-z (approved):

packages/lab/stories/combo-box-next/combo-box-next.stories.tsx
```diff
@@ -50,7 +50,10 @@ const customItemFilter = (source: LargeCity[], filterValue?: string) =>
   );
 
 const ComboBoxTemplate: Story<ComboBoxNextProps<any>> = (args) => {
-  const handleChange = (event: SyntheticEvent, data: { value: string }) => {
+  const handleChange = (
+    event: SyntheticEvent,
+    data: { value: string | undefined }
+  ) => {
     console.log("input value changed", data);
```
origami-z, line 57: if we do `args.onChange`, storybook's Action panel would show action as well

packages/lab/src/list-next/useList.ts
```diff
@@ -21,8 +21,10 @@ export interface UseListProps {
   selected?: string;
   /* Initial selected value for when the list is controlled. */
   defaultSelected?: string;
-  /* Callback for when the list is controlled. */
-  onChange?: (e: SyntheticEvent, data: { value: string }) => void;
+  /* Callback for change event. Returns the event and full selection. */
+  onChange?: (e: SyntheticEvent, data: { value: string | undefined }) => void;
+  /* Callback for select event. Returns the event and new selection. */
```
joshwooding, line 26:
```suggestion
  /* Callback for select event. This is called when any selection occurs, even if a previously selected value is selected again. */
```

packages/lab/src/list-next/useList.ts
```diff
@@ -52,7 +55,7 @@ export const useList = ({
 
   const [highlightedItem, setHighlightedItem] = useControlled({
     controlled: highLightedItemProp,
```
joshwooding, line 57:
```suggestion
    controlled: highlightedItemProp,
```

joshwooding (approved):

origami-z (approved):

PR #2234, opened 2023-08-15
Dialog story changes

libertymayc:
feedback from @ivan-calderon 
Make Dialog story content configurable
Fix width of container


packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -1,5 +1,5 @@
 import { PropsWithChildren, useState } from "react";
-import { Button, StackLayout } from "@salt-ds/core";
+import { Button, capitalize, StackLayout } from "@salt-ds/core";
```
origami-z, line 2: Are you using `capitalize`?

packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -166,14 +174,8 @@ const AlertDialogTemplate: ComponentStory<typeof Dialog> = ({
...
       >
-        <DialogTitle>
-          {status![0].toUpperCase()}
-          {status!.slice(1)}
-        </DialogTitle>
-        <DialogContent>
-          A Dialog is a window that opens over the application content, focusing
-          the user's attention on a particular task or piece of information.
-        </DialogContent>
+        <DialogTitle>{title}</DialogTitle>
```
origami-z, line 177: If `title` defaults to `undefined`, is there a case to not render `DialogTitle`?

packages/lab/stories/dialog/dialog.stories.tsx
libertymayc: i was but removed, will take out

packages/lab/stories/dialog/dialog.stories.tsx
libertymayc: I didn't mean to default it to undefined (was trying to see if i can override the default props) - the default should be the args given for each story and there is always one
i think conditional rendering title and content would be a different story altogether showcasing how it looks when empty (like empty formfield with no control) which wouldnt be encouraged, so i wont add it (at least for now, but will note to Ivan, if there's use cases for no title/content)

origami-z (approved):

ZarkoPernar (approved):

alycrys (approved):

origami-z (approved):

PR #2235, opened 2023-08-15
Checkbox and RadioButton readonly states made focusable by default

libertymayc:
related #2179


origami-z (commented):
Can we add a test. Behaviour code change should fail test to help validate future decision change

alycrys (approved):

PR #2251, opened 2023-08-16
Combo Box controlled

Fercas123:
Last split of https://github.com/jpmorganchase/salt-ds/pull/2181

Changes to Combo Box:

- Fix controlled implementation.
- Removed `InputProps`.
- Added `onListChange` and `onSelect` props.

### DIFF
--- a/packages/lab/src/combo-box-next/useComboBox.tsx
+++ b/packages/lab/src/combo-box-next/useComboBox.tsx
@@ -1,8 +1,11 @@
-import { FocusEvent, KeyboardEvent, SyntheticEvent, useState } from "react";
+import { FocusEvent, KeyboardEvent, SyntheticEvent, useEffect } from "react";
 import { useList, UseListProps } from "../list-next/useList";
 import { useComboboxPortal, UseComboBoxPortalProps } from "./useComboboxPortal";
+import { useControlled } from "@salt-ds/core";
 
 interface UseComboBoxProps {
+  inputValue?: string;
+  defaultInputValue?: string;
   onBlur?: (event: FocusEvent<HTMLInputElement>) => void;
   onFocus?: (event: FocusEvent<HTMLInputElement>) => void;
   onMouseOver?: (event: SyntheticEvent) => void;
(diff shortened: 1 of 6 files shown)


packages/lab/src/combo-box-next/useComboBox.tsx
```diff
@@ -1,8 +1,11 @@
-import { FocusEvent, KeyboardEvent, SyntheticEvent, useState } from "react";
+import { FocusEvent, KeyboardEvent, SyntheticEvent, useEffect } from "react";
 import { useList, UseListProps } from "../list-next/useList";
 import { useComboboxPortal, UseComboBoxPortalProps } from "./useComboboxPortal";
+import { useControlled } from "@salt-ds/core";
 
 interface UseComboBoxProps {
+  inputValue?: string;
+  defaultInputValue?: string;
```
libertymayc, line 8: will the hook be used by users? should we put these props in CS?

JoanaMMoreira (approved):

libertymayc (approved):

packages/lab/stories/combo-box-next/combo-box-next.doc.stories.mdx
```diff
@@ -58,6 +58,23 @@ Tip: For recommendations on variant choice, see the <Link href="https://www.salt
...
+You can use a placeholder to prompt user input if no default value is provided.
+
+<Canvas>
+  <Story id="lab-combo-box-next--placeholder" />
+</Canvas>
+
+### Controlled
+
+You can control the `value` and `open` and `selected` props of Combo Box.
+Learn more about <Link href="https://react.dev/learn/sharing-state-between-components#controlled-and-uncontrolled-components">controlled vs. uncontrolled components</Link> in the official React docs.
```
libertymayc, line 72: in uitk we have in the doc: 
"When used as a controlled ... how Combo Box interacts with user input. In this example, the selected text is transformed to upper case."

I think we could put here what/how it's strictly controlled by the buttons, to be more clear on the controlled implementation
I wonder if we should be similar with our controlled stories, e.g. this/dropdown using buttons whereas Input controlled story isnt really different from the default, could probably be improved

Fercas123: will follow up with this update

PR #2283, opened 2023-08-23
Make site fonts work in portals

james-nash:
The Salt site uses Next's font optimisations and therefore needs to set the `--salt-typography-fontFamily*` vars to the font family names generated by Next.

For this to work, we had to add a couple of classnames, which are generated by Next, to our outermost `<SaltProvider>`, so that its CSS vars (`--site-font-family*`) were visible to the Salt theme.

Unfortunately, things like dropdown lists which render in portals, are outside of that element in the DOM and so those `--site-font-family*` vars are invisible to them, leading to text not using the correct font.

This PR fixes that, by hoisting those `--site-font-family*` declarations into the `:root` selector, so that they're available anywhere on the page and thus work inside portals too.


joshwooding (approved):

joshwooding (changes requested):
Prettier needs to be run

joshwooding (approved):

feliciackh (approved):

shey-v (approved):

PR #2301, opened 2023-08-25
Add Link documentation to the site

### LINKED ISSUE #2228: Link Documentation

joshwooding:
closes #2228


site/docs/components/link/usage.mdx
```diff
@@ -0,0 +1,54 @@
...
+#### When to use Link
+
+- To provide navigation to a page on the same or different site.
+- To link to documents, email addresses or phone numbers.
+- To send users to a specific section on the same page.
+- To provide users with more detailed information or assistance e.g., terms and conditions, help sections or contact information.
+
+#### When not to use Link
+
+To trigger an action, such as submitting a form or opening a Dialog. Instead, use [Button](../../button).
```
feliciackh, line 26:
```suggestion
To trigger an action, such as submitting a form or opening a dialog. Instead, use [Button](../../button).
```
unless its meant to refer to the Dialog component? in which case we'll also need to link the component

joshwooding: Ultimately I think it will refer to the component but for now it's safer to make it lowercase

shey-v (dismissed):
With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.

james-nash:
> With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.

The trouble is, the live previews contain real links and once someone clicks them, their browser will display the visited state and there's no way for us to undo that. (Technically we could fake it via some extra CSS, but that would then become part of the examples code and be distracting and confusing to folks)

It just so happens that, in the visited state, the primary and secondary links use the exact same color (not sure if that's intentional or a bug/oversight in our component code).

Currently, the example links point to the Salt site homepage, so there's a high chance they'll appear in the visited state since someone looking at the component pages is likely to have been on the homepage in the past. I therefore suggest changing them to fragment links (you could just do `href="#"`). It won't change the fact that once someone clicks on them, they'll appear in the visited state, but it does at least reduce the chances of them appearing visited when the user first lands on this component page. (It also has the benefit that clicking them is a bit less disruptive since it just jumps you to the top of the page instead of taking you away to the homepage)

site/src/examples/link/Variant.tsx
```diff
@@ -0,0 +1,13 @@
+import { ReactElement } from "react";
+import { Link, StackLayout } from "@salt-ds/core";
+
+export const Variant = (): ReactElement => (
+  <StackLayout>
+    <Link href="/" variant="primary">
```
james-nash, line 6:
```suggestion
    <Link href="#" variant="primary">
```
 Reduce likelihood of the link appearing in visited state when someone visits this component page.

site/src/examples/link/Variant.tsx
```diff
@@ -0,0 +1,13 @@
...
+      Primary Variant
+    </Link>
+    <Link href="/" variant="secondary">
```
james-nash, line 9:
```suggestion
    <Link href="#" variant="secondary">
```

site/src/examples/link/Default.tsx
```diff
@@ -0,0 +1,4 @@
+import { ReactElement } from "react";
+import { Link } from "@salt-ds/core";
+
+export const Default = (): ReactElement => <Link href="/">Link</Link>;
```
james-nash, line 4:
```suggestion
export const Default = (): ReactElement => <Link href="#">Link</Link>;
```

site/src/examples/link/OpenInANewTab.tsx
```diff
@@ -0,0 +1,26 @@
...
+      href="https://github.com/jpmorganchase/salt-ds"
+      target="_blank"
+      rel="noopener"
+      IconComponent={GithubIcon}
+    >
+      Salt GitHub repository
+    </Link>
+    <Link href="#" target="_blank" rel="noopener" IconComponent={null}>
+      This link has no icon
+    </Link>
```
james-nash, line 24: The links are a bit inconsistent here, one describes how the link will open, the next where it goes and the last what it looks like. What if put some descriptive text outside the link and made the actual link text describe where it goes (as per our guidance). For example:

```suggestion
 <p>
    Using default icon:
    {" "}
    <Link
      href="https://www.saltdesignsystem.com"
      target="_blank"
      rel="noopener"
    >
      Open the Salt website in a new tab
    </Link>
  </p>
  <p>
    Using a custom icon:
    {" "}
    <Link
      href="https://github.com/jpmorganchase/salt-ds"
      target="_blank"
      rel="noopener"
      IconComponent={GithubIcon}
    >
      Open the Salt GitHub repository in a new tab
    </Link>
  </p>
  <p>
    With the icon removed:
    {" "}
    <Link href="" target="_blank" rel="noopener" IconComponent={null}>
      Open this page in a new tab
    </Link>
```

james-nash (commented):
Made a few suggestions. Do with them as you wish :-)

feliciackh:
> @feliciackh do the changes suggested by @james-nash look okay to you? can we merge?

Changes look good to me, it's best to get @joshwooding's input as I haven't been closely following any updates/discussion had for Link component

site/src/examples/link/OpenInANewTab.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <Link
+      href="https://github.com/jpmorganchase/salt-ds"
+      target="_blank"
+      rel="noopener"
+      IconComponent={GithubIcon}
+    >
+      Open the Salt GitHub repository in a new tab
+    </Link>
+  </p>
+  <p>
```
joshwooding, line 30: This is missing a close tag

site/src/examples/link/OpenInANewTab.tsx
```diff
@@ -0,0 +1,37 @@
+import { ReactElement } from "react";
+import { StackLayout, Link } from "@salt-ds/core";
+import { GithubIcon } from "@salt-ds/icons";
+
+export const OpenInANewTab = (): ReactElement => (
+  <StackLayout>
+ <p>
+    Using default icon:
```
joshwooding, line 8: We should decide if we think having them in sentences like this is what we want versus in sections with theses as subtitles

lilyvc:
> With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.

I applied James's changes - is this better for you now?

shey-v:
> > With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.
> 
> I applied James's changes - is this better for you now?

I can see the difference now between the two states, however, as James mentions above, it doesn't stop everything from appearing in the 'visited link' state once I've clicked on them. Even after I refresh my browser.

shey-v:
When all examples toggle is switched on, I can see the right hand nav 'On this page' appearing. Should this be removed? There is a huge inconsistency across all of the docs pages when this toggle is on, some have the right hand nav and some do not... @origami-z FYI.

origami-z:
> When all examples toggle is switched on, I can see the right hand nav 'On this page' appearing. Should this be removed? There is a huge inconsistency across all of the docs pages when this toggle is on, some have the right hand nav and some do not... @origami-z FYI.

A site bug, could be related to #2310. Should be looked at separately.

lilyvc:
> > > With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.
> > 
> > 
> > I applied James's changes - is this better for you now?
> 
> I can see the difference now between the two states, however, as James mentions above, it doesn't stop everything from appearing in the 'visited link' state once I've clicked on them. Even after I refresh my browser.

is this an issue with the component itself? or do we want it to behave differently in the docs compared to when people are using it?

shey-v:
> > > > With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.
> > > 
> > > 
> > > I applied James's changes - is this better for you now?
> > 
> > 
> > I can see the difference now between the two states, however, as James mentions above, it doesn't stop everything from appearing in the 'visited link' state once I've clicked on them. Even after I refresh my browser.
> 
> is this an issue with the component itself? or do we want it to behave differently in the docs compared to when people are using it?

I feel as though for the docs it should behave differently, otherwise it is a bit misleading when users return to the other examples and they are all in a 'visited link' style. @bhoppers2008 what is your call on this?

lilyvc:
> > > > > With the 'Variant' example, I do not see a visual difference between Primary and Secondary. They are both currently appearing in the visited link style, could we amend this please.
> > > > 
> > > > 
> > > > I applied James's changes - is this better for you now?
> > > 
> > > 
> > > I can see the difference now between the two states, however, as James mentions above, it doesn't stop everything from appearing in the 'visited link' state once I've clicked on them. Even after I refresh my browser.
> > 
> > 
> > is this an issue with the component itself? or do we want it to behave differently in the docs compared to when people are using it?
> 
> I feel as though for the docs it should behave differently, otherwise it is a bit misleading when users return to the other examples and they are all in a 'visited link' style. @bhoppers2008 what is your call on this?

I can override it so they never turn purple, I agree it makes sense for the users to see the not visited style. But, then they are never in a visited state - so do you think we need an example to demonstrate this? Or some words explaining it

site/docs/components/link/accessibility.mdx
```diff
@@ -0,0 +1,21 @@
...
+---
+
+- Screen reader users will use the Tab key to jump from link to link or pull up a list of links on a page. Avoid using the same link text to link to different content and ensure that links make sense out of context.
+- When a link contains an icon and text, pass `aria-hidden=”true"` to the icon so that its label is not announced by screen readers
+- When a link has an icon and no text, pass an `aria-label` to Link to describe the content the link leads to.
+
+<KeyboardControls>
+<KeyboardControl keyOrCombos="Enter">
+
+If the link has a href, the Enter key press navigates the user.
```
JoanaMMoreira, line 18:
```suggestion
If the link has an href, the Enter key press navigates the user.
```

feliciackh (approved):

origami-z (approved):

feliciackh (approved):

PR #2327, opened 2023-09-01
Fix site build

origami-z:
Broke since f8ac3bff7c1bf95c64f06c5864436637ff331ad6


site/propsGen.js
```diff
@@ -4,6 +4,8 @@ const { writeFile } = require("fs");
 
 const componentPackages = ["lab", "core", "data-grid", "icons"]; // Adding more packages here will generate additional prop files
 
+console.log("Generating props in packages:", componentPackages);
```
origami-z, line 7: This script is slow, it often appears to be not running. Add some log.

feliciackh (approved):

JoanaMMoreira (approved):

Issue #2265, opened 2023-08-21
Component documentation feedback


bhoppers2008:
A place to capture feedback on the docs template or the site page itself. Where are the challenges? Is there anything missing? Is it difficult to find a place for content you want to add? Etc...

bhoppers2008:
- Update token in preview area to --container-primary-background token.

el-dav:
Note the displayName issue seems to be due to the components using arrow functions rather than named functions, so we should create a ticket to go though and update all components to use named functions. Updating FlowLayout to the following fixed that component for me:

```tsx
export const FlowLayout: FlowLayoutComponent = forwardRef(function FlowLayout<
  T extends ElementType = "div"
>({ children, ...rest }: FlowLayoutProps<T>, ref?: PolymorphicRef<T>) {
  return (
    <FlexLayout direction="row" ref={ref} wrap {...rest}>
      {children}
    </FlexLayout>
  );
});
```
from
```tsx
export const FlowLayout: FlowLayoutComponent = forwardRef(
  <T extends ElementType = "div">(
    { children, ...rest }: FlowLayoutProps<T>,
    ref?: PolymorphicRef<T>
  ) => {
    return (
      <FlexLayout direction="row" ref={ref} wrap {...rest}>
        {children}
      </FlexLayout>
    );
  }
);

```

james-nash:
> 9. Clicking on the chevron in the side nav should only expand/collapse the item, not open the page at the same time.

I believe this'll be fixed when we replace the current vertical nav component (which comes from Mosaic) with our own, based on the `NavigationItem` component.

Not sure there's much point trying to "fix" this in the Mosaic component now, given that we're due to replace it anyway.

james-nash:
FYI, I've opened an issue to note some additional functionality for the `LivePreview` component (the thing that displays our component examples) that has been discussed in the past but never actually written down: #2348

PR #2330, opened 2023-09-01
Change all true values to boolean in docs

libertymayc:
fixes no.20 in #2265


site/docs/components/avatar/accessibility.mdx
```diff
@@ -9,6 +9,6 @@ data:
 
 ### Best practices
 
-Decorative Avatars (Avatars that repeat the information already expressed by the accompanied text) should be hidden from the screen reader to avoid repetition. If you are adding an Avatar for decorative purposes, pass `aria-hidden=”true”` to hide its label from the screen reader.
+Decorative Avatars (Avatars that repeat the information already expressed by the accompanied text) should be hidden from the screen reader to avoid repetition. If you are adding an Avatar for decorative purposes, pass `aria-hidden={true}` to hide its label from the screen reader.
```
libertymayc, line 12: aria-hidden is Booleanish, but doesn't hurt to change it to true

james-nash (dismissed):
👏

site/docs/components/avatar/accessibility.mdx
joshwooding: I'm tempted to say this whole sentence should be something like:

```suggestion
Decorative Avatars (Avatars that repeat the information already expressed by the accompanying text) should be hidden from the screen reader to avoid repetition. If you are adding an Avatar for decorative purposes, use `aria-hidden` to hide its label from the screen reader.
```

joshwooding: cc @jpmorganchase/salt-content

site/docs/components/button/accessibility.mdx
```diff
@@ -11,7 +11,7 @@ data:
 
 ### Best practices
 
-- When a Button has both an icon and text, pass `aria-hidden=”true”` to the icon so that its text is not announced by screen readers.
+- When a Button has both an icon and text, pass `aria-hidden={true}` to the icon so that its text is not announced by screen readers.
```
joshwooding, line 14:
```suggestion
- When a Button has both an icon and text, use`aria-hidden` on the icon so that screen readers do not announce its text.
```

site/docs/components/avatar/accessibility.mdx
joshwooding: It's actually not so simple, I'll chat with you

libertymayc: agreed, will change to Joshs suggestion if its ok? @mikearildbrown

lilyvc (approved):

alycrys (approved):

PR #2351, opened 2023-09-07
Showcase all tokens on the site

libertymayc:
**This PR ...** https://saltdesignsystem-git-deprecated-tokens-docs-fed-team.vercel.app/salt/foundations/tokens/index

Showcases foundations, characteristics, palette and deprecated tokens.
Improves and extends general code in site/docs/foundations

We need to document all tokens on the site to be able to educate the theme better, rather than just listing the characteristics out solely, when there are a lot of other tokens that build up the theme. Documenting the following helps;

- Foundations: Users may want a list of all the tokens to easily find them, rather than having to go to the individual page. I suggest this info could move to each foundation page eventually, but for now we will just do what was done for characteristics
- Deprecated tokens: Allows for users to become aware of deprecated tokens without having to read changelogs or this information getting lost. Also helps us have these tokens written in a clear list for our own sake, to be able to keep track, and will help designers know what has been deprecated in code and what hasn't, especially in this limbo stage of theme freezing in code where we are no longer in complete sync [cc @pseys]
- Palette tokens: The palette may well be where the majority of customisation is done in terms of what colours are used in themes, rather than at the characteristic level. Beginning to surface these tokens will help this process.

**Help needed/future work**

- Token values are currently only light/dark theme (I think dark). Need to be able to show both.
- The tokens are all currently listed under the Foundations tab, where I renamed the old 'Characteristics' title in the sidebar to 'Tokens' and split out the 4 different token groups (foundations, deprecated, palette, characteristics). I think this is going to cause a lot of confusion if we keep it here for too long. _Characteristics, the palette, and (some) deprecated tokens are not foundations._  What is the plan for where we document characteristics and the palette? Will they be in their own section completely, or should we tweak the currently IA and re-title this tab 'Theme' for now to avoid confusion? cc @james-nash @jpmorganchase/salt-content  @pseys 
- I moved the general characteristics overview to the Characteristics page (with tokens then listed at the bottom), since keeping this info on the 'Tokens' page made no sense. Foundations, palette, and deprecated tokens don't have this level of detail though and currently only have the accordion of tokens. What could/should we add here? And on the overview page, how should this be (for now, considering we are likely not going to keep this IA?) @jpmorganchase/salt-content


libertymayc:
> Hi @libertymayc, thanks for sending through.
> 
> I just spoke to @pseys about this. We have a meet on the 19th to discuss content strategy for the foundations/theming/token-related pages, that should hopefully answer question 2. He can jump in if I'm incorrect here, but as I understand there's upcoming dev work that will improve the existing tokens and pave the way for theming capabilities. The content strategy won't be implemented until we've done the dev work to avoid spending time on soon-to-be-outdated information.
> 
> This PR addresses an issue that the meeting on the 19th will hopefully resolve—how will we structure our IA and content to meet users' needs, particularly for foundations/tokens/theming, and how do we ensure that the content encourages teams to use the DS in the intended way?
> 
> Personally I'd prefer to hold this PR until after the meet/after a strategy is decided so we can discuss first what information is required and how best we can represent it.
> 
> Ccing @AnniaT and @shinytoyrobots for thoughts on this.

thanks! sounds good, I'd be interested to be in that meeting as it's exciting how we will document these tokens! 

After a discussion with the devs yesterday, we thought it would be best to document all tokens, it doesn't make sense to just do the characteristics (and at least especially show the deprecated ones). After all, the characteristic tokens are already there displayed in a way that's not going to be the ultimate layout, this just lists all of them in the same way. I think it would be best to get it all out since it will benefit early adopters - if that meeting is only ~10 days away, seems like this IA, although not ideal, won't be on the site for too long.

ZarkoPernar:
Don't let perfect be the enemy of good. This is good enough for now @mikearildbrown. We need to stop hiding important information from our users, like component documentation pages for example, because it is not in some ideal state.

lilyvc (dismissed):

feliciackh (approved):

origami-z:
Pending #2571 updates to go in first

site/src/components/css-variables/TokenTable.tsx
```diff
@@ -0,0 +1,53 @@
...
+        {Object.entries(data).map(([name, value]) => (
+          <tr key={name}>
+            <td className={styles.viewColumn}>
+              <BlockView name={name} />
+            </td>
+            <td>
+              <Tooltip
+                className={styles.tooltip}
+                content={<BlockView name={name} />}
+                onOpenChange={function _l() {}}
```
origami-z, line 31: ?

site/src/components/css-variables/json/foundationsVariables.json
```diff
@@ -0,0 +1,241 @@
+{
```
origami-z, line 1: These files are gitignored?

origami-z (commented):
Anything page description we can add above accordion to the new pages?

site/docs/foundations/tokens/deprecated.mdx
```diff
@@ -0,0 +1,9 @@
+---
+title: Deprecated tokens
+layout: DetailTechnical
+sidebar:
+  label: Deprecated tokens
+  priority: 1
```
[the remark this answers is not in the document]
feliciackh: @mikearildbrown is there a preference as to how the pages are ordered? following order of layers and/or having deprecated ones last in the list

joshwooding:
> @pseys now the big theme update is out can we merge this PR? is there anything we need to change?

We spoke about this either last week or the week before. I think Paul wanted to propose something different.

joshwooding:
We're currently looking into how we document the theme, characteristics, foundations and other design language-related work. We'll refer to this as a potential direction, but it's unlikely we'll take this PR directly. I'll close this for now. Thanks @libertymayc

PR #2365, opened 2023-09-08
Display component docs on the site

JoanaMMoreira:
- Change components page layout so it displays sidebar nav
- Remove storybook logo from table header and display it only on storybook links
- Add table links to site docs
- Preview at https://saltdesignsystem-git-display-component-docs-on-site-fed-team.vercel.app/salt/components/index


origami-z (dismissed):

Fercas123:
Not related to this PR but i noticed the Progress examples are not rendering correctly 
https://saltdesignsystem-git-display-component-docs-on-site-fed-team.vercel.app/salt/components/progress/examples

might be worth fixing them before we release the docs showing

Fercas123 (dismissed):

site/src/components/components/ComponentsList.tsx
```diff
@@ -56,11 +56,16 @@ const componentsListSorting = (
...
 const ComponentNameData = ({ component }: { component: ComponentDetails }) => {
-  const { devStatus, name, storybookUrl } = component;
+  const { name, docsUrl } = component;
 
-  return devStatus === ComponentStatus.READY && storybookUrl ? (
-    <Link href={storybookUrl}>
+  return docsUrl ? (
+    <Link
+      href={docsUrl}
+      target={isExternalLink(docsUrl) ? "_blank" : undefined}
```
james-nash, line 67: We don't need `_target="blank"` here - linking to another website is no reason to force opening a new tab. Our own Link component documentation also advises against doing this.

james-nash (changes requested):
Mostly great. But forcing Storybook links to open in a new tab contradicts our own Link guidance.

alycrys (dismissed):

JoanaMMoreira:
> Not related to this PR but i noticed the Progress examples are not rendering correctly https://saltdesignsystem-git-display-component-docs-on-site-fed-team.vercel.app/salt/components/progress/examples
> 
> might be worth fixing them before we release the docs showing

Since these are still work in progress it's worth hiding them for now

site/src/components/components/ComponentsList.tsx
JoanaMMoreira: We wanted to differentiate the storybook links from the internal ones by using the tear out icon (which only makes sense if we open on a new page) so maybe we need a different approach? I can add the storybook icon instead cc @origami-z

origami-z: Can be any visual difference, so user wouldn't be surprised to not finding them on the side bar.

james-nash: Yeah, a Storybook icon could work. Alternatively, just append " (on Storybook)" or something similar to the link text. E.g. "AG Grid theme (on Storybook)".

It's unambiguous and potentially more meaningful for screenreader users, who might be listening to the link texts out of context.

JoanaMMoreira: I've added the icon which I think looks a bit better on the table but let me know your thoughts 😄

james-nash (approved):
LGTM. Ship it! 🛳️🎉👍

origami-z (approved):

JoanaMMoreira:
> Hey guys, thanks for getting this out! Is the long-term plan to keep both lists of components on the index page? Side nav and "available and upcoming components" lists. Visually it looks like we're repeating the lists.

I don't think this is the long term plan but we will need to spend time designing and building a new page so this was the quickest solution to get the docs live for now

james-nash:
> Hey guys, thanks for getting this out! Is the long-term plan to keep both lists of components on the index page? Side nav and "available and upcoming components" lists. Visually it looks like we're repeating the lists.

As an interim, might it help if we changes the heading above the table to "Component status"? The status info _is_ something additional it provides, besides the (duplicate) links to the component pages.

There are precedents for that in other DS sites too:

* [GitHub Primer](https://primer.style/design/guides/status)
* [Dell Design System](https://www.delldesignsystem.com/updates/component-status/)
* [Decathlon Design System](https://www.decathlon.design/726f8c765/p/93ca3c-component-status)


If you agree, I could sneak in that change as part of PR #2377, which updates the status info in that table.

PR #2462, opened 2023-09-13
Add a color prop to Icon

joshwooding:
This PR adds a color prop to Icon that supports: primary, secondary and inherit.

There should be another PR to align Text. 

But the reason for this change is to make working with Icon a lot easier, usually Icon will always following the text colour used and currently we have to be very explicit when trying to achieve this. Changing Icon to inherit the text colour by default would reduce a lot of extra code and improve compatibility with the UITK button

### DIFF
--- a/.changeset/bright-tips-tell.md
+++ b/.changeset/bright-tips-tell.md
@@ -4,10 +4,11 @@
 
 Add a `color` prop to Icon.
 
-This can be used to specify the color used by Icon and supports: primary, secondary and inherit. The default value is secondary.
+This can be used to specify the color used by Icon and supports: primary, secondary and inherit. The default value is "inherit".
 
 ```tsx
 <HomeIcon color="primary" />
 <UserIcon color="secondary" />
 <CloseIcon color="inherit" />
 ```
+*Note:* This changes the default icon color from secondary to primary due to it inheriting the default text style. The previous default of secondary was an error.
--- a/.changeset/bright-tips-tell.md
+++ b/.changeset/bright-tips-tell.md
@@ -1,0 +1,13 @@
+---
+"@salt-ds/icons": minor
+---
+
+Add a `color` prop to Icon.
+
+This can be used to specify the color used by Icon and supports: primary, secondary and inherit. The default value is secondary.
+
+```tsx
+<HomeIcon color="primary" />
+<UserIcon color="secondary" />
+<CloseIcon color="inherit" />
+```
--- a/.changeset/bright-tips-tell.md
+++ b/.changeset/bright-tips-tell.md
@@ -11,4 +11,5 @@ This can be used to specify the color used by Icon and supports: primary, second
 <UserIcon color="secondary" />
 <CloseIcon color="inherit" />
 ```
-*Note:* This changes the default icon color from secondary to primary due to it inheriting the default text style. The previous default of secondary was an error.
+
+_Note:_ This changes the default icon color from secondary to primary due to it inheriting the default text style. The previous default of secondary was an error.
--- a/packages/icons/src/icon/Icon.css
+++ b/packages/icons/src/icon/Icon.css
@@ -1,6 +1,6 @@
 /* Style applied to the root element */
 .saltIcon {
-  --icon-color: var(--saltIcon-color, var(--salt-text-secondary-foreground));
+  --icon-color: currentColor;
   --icon-size-multiplier: var(--saltIcon-size-multiplier, 1);
   --icon-base-size: var(--salt-icon-size-base, 12px);
   /**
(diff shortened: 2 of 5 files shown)


ZarkoPernar:
What is the difference between `currentColor` and `inherit` in practice in this particular case? Wouldn't the final result be the same?

alycrys:
would this need tests?

james-nash (dismissed):

james-nash:
> What is the difference between `currentColor` and `inherit` in practice in this particular case? Wouldn't the final result be the same?

They'd be different in this case. `inherit` would just cause the `--icon-color` to inherit its value, which would probably be nothing. `currentColor` on the other hand resolves to whatever the element's CSS `color` has been set to - i.e. it matches the text color.

joshwooding:
> would this need tests?

It has visual regression tests, e.g. Chromatic

packages/icons/src/icon/Icon.css
```diff
@@ -1,6 +1,6 @@
 /* Style applied to the root element */
 .saltIcon {
-  --icon-color: var(--saltIcon-color, var(--salt-text-secondary-foreground));
+  --icon-color: currentColor;
```
libertymayc, line 3: --saltIcon-color is used by a lot of components e.g. Button, we can't just remove it here

joshwooding: It's not removed

.changeset/bright-tips-tell.md
```diff
@@ -0,0 +1,14 @@
...
+Add a `color` prop to Icon.
+
+This can be used to specify the color used by Icon and supports: primary, secondary and inherit. The default value is "inherit".
+
+```tsx
+<HomeIcon color="primary" />
+<UserIcon color="secondary" />
+<CloseIcon color="inherit" />
+```
+*Note:* This changes the default icon color from secondary to primary due to it inheriting the default text style. The previous default of secondary was an error.
```
libertymayc, line 14: I believe when we were doing the theme @pseys said icons should be secondary, has that changed now?

packages/icons/src/icon/Icon.css
joshwooding: We would also see things showing up in the visual diff if it was removed

libertymayc: gotcha I missed that, didnt check chromatic :D

packages/icons/src/icon/Icon.tsx
```diff
@@ -14,6 +14,10 @@ export const makePrefixer =
 const withBaseName = makePrefixer("saltIcon");
 
 export interface IconProps extends SVGAttributes<SVGSVGElement> {
+  /*
+   * The color of the icon. Defaults to "secondary".
```
JoanaMMoreira, line 18:
```suggestion
   * The color of the icon. Defaults to "inherit".
```

.changeset/bright-tips-tell.md
joshwooding: I've had this agreed by design yeah

libertymayc (dismissed):

libertymayc:
> > > What is the difference between `currentColor` and `inherit` in practice in this particular case? Wouldn't the final result be the same?
> > 
> > 
> > They'd be different in this case. Since it's being applied to the SVG's `fill` property, `inherit` would cause it to inherit whatever the `fill` value would be - probably nothing or `initial` in this case.`currentColor` on the other hand resolves to whatever the element's CSS `color` has been set to - i.e. it matches the text color.
> 
> As James says, currentColor matches text colour, inherit would inherit a fill which won't be what we want

will "inherit" make sense as the prop if it behaves slightly differently to currentColor? would it be better to be "current"?

site/src/examples/icon/IconColor.tsx
```diff
@@ -1,13 +1,11 @@
...
-        "--saltIcon-color": "var(--salt-status-info-foreground)",
-      } as CSSProperties
-    }
-    size={2}
-  />
+  <StackLayout>
+    <AddDocumentIcon color="inherit" size={2} />
+    <AddDocumentIcon color="primary" size={2} />
+    <AddDocumentIcon color="secondary" size={2} />
+  </StackLayout>
```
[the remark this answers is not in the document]
libertymayc: the change works in storybook, does the site get the changes automatically within a PR?

packages/icons/src/icon/Icon.tsx
joshwooding: Thanks!

site/src/examples/icon/IconColor.tsx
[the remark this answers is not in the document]
joshwooding: Mosaic CSS strikes again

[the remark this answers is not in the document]
JoanaMMoreira: Is it worth removing/hiding this example until we fix the override in Mosaic and upgrade the site?

alycrys (dismissed):

JoanaMMoreira (approved):

libertymayc (approved):

libertymayc:
approved, but my comment above still worth considering :)

feliciackh (approved):

PR #2465, opened 2023-09-14
fix bug in GridLayout where columnGap and rowGap couldn't be zero

el-dav:
Fixing bug in GridLayout where columnGap/rowGap couldn't be set to zero, as logic was determining zero as falsey causing the default (non-zero) value to be used

### DIFF
--- a/.changeset/sour-fireants-crash.md
+++ b/.changeset/sour-fireants-crash.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


JoanaMMoreira (approved):
👌

james-nash (approved):

joshwooding:
For bonus points this also fixes some lint issues :)

origami-z (approved):
good to have a test?

PR #2492, opened 2023-09-21
Upgrade TypeScript packages

joshwooding:


origami-z:
Does this impact users with lower TS version using our types? If so, should we produce types for multiple version (`typesVersions`)?

joshwooding:
There's an issue with polymorphic components that are built using other polymorphic components. They can't inherit their polymorphism.

PR #2500, opened 2023-09-25
Fix FormField adds extra "padding" below the input

### LINKED ISSUE #2497: FormField adds extra "padding" below the input
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

1.8.1

### Description

FormField declares three grid template areas, even when the elements for those areas are not rendered. For example, this is the grid-template-area by default:

```css
.saltFormField-labelTop {
grid-template-areas:
        "label"
        "controls"
        "helperText";
}
```

The issue is gap property  is always applied to all three areas, even when we are rendering no helper text which results in additional "padding" below form field.

Screenshot 2023-09-22 at 17 28 00

### Steps to reproduce

Check out an example without helper text here https://www.saltdesignsystem.com/salt/components/form-field/examples

### Expected behavior

The gap between the input and helper text should not be applied when there is no helper text. Only one gap should be applied, between the label and the input.

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

libertymayc:
closes #2497

ideally would use :has selector instead

Note that the same would apply above if no label is provided. I've only added the fix for helper text. I don't see why a label wouldn't be provided, the control would probably just be used itself in that case?

### DIFF
--- a/.changeset/young-timers-explain.md
+++ b/.changeset/young-timers-explain.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixes extra padding below a control in Form Field when no helper text provided
--- a/packages/core/src/form-field/FormField.css
+++ b/packages/core/src/form-field/FormField.css
@@ -19,18 +19,16 @@
 .saltFormField-labelLeft {
   align-self: center;
   grid-template-columns: var(--formField-label-width, 40%) 1fr;
-  grid-template-areas:
-    "label controls";
+  grid-template-areas: "label controls";
 }
 
 .saltFormField-labelRight {
   align-self: center;
   grid-template-columns: var(--formField-label-width, 40%) 1fr;
-  grid-template-areas:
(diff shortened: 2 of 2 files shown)


ZarkoPernar (dismissed):

packages/core/stories/form-field/styles.css
joshwooding, line 1: Is this empty file intended?

joshwooding (changes requested):

lilyvc (dismissed):

packages/core/stories/form-field/styles.css
libertymayc: oops! good spot i was testing something else

lilyvc (approved):

ZarkoPernar (approved):

PR #2503, opened 2023-09-26
Improve lab components visibility on the site

JoanaMMoreira:
- Update copy on components overview page
- Add "In progress" status to lab components page heading
- Add 🚧 to sidebar label for lab components


lilyvc (dismissed):

joshwooding (commented):
I thought the table on the page was also being removed?

joshwooding (commented):
@mikearildbrown The copy for labs is pretty hidden on the page and is easily missed

JoanaMMoreira:
> I thought the table on the page was also being removed?

I thought we were keeping it until we have the docs for ag grid theme but I can remove it if that's not the case

JoanaMMoreira:
> it looks like the sort order is the same, is this something we can change?

@joshwooding did we agree we would make the change in Mosaic rather than changing the order for all core component pages individually?

libertymayc (approved):

site/docs/components/combo-box/index.mdx
```diff
@@ -26,6 +28,7 @@ data:
       { name: "List", relationship: "contains" },
       { name: "Input", relationship: "contains" },
     ]
+  status: In progress
```
origami-z, line 31: Probably something more clear to lab, e.g., "Lab component"

origami-z: in progress doesn't tell direct association to lab component, especially "do not use in production". It's likely different than the "In process" in foundation actual mean

JoanaMMoreira: @mikearildbrown @origami-z is there a consensus on this? I personally like "in progress" and I think the sidebar icon provides enough context but let me know either way so we can get this merged :)

feliciackh (approved):

PR #2515, opened 2023-09-28
[SaltProvider] Add defaultViewport prop

ZarkoPernar:
## Problem
Currently, the layout components offer responsive props as a useful feature which in turn use useViewport to determine which responsive prop to apply. The viewport is set dynamically after first render in the browser by measuring the window size, and the default value is 0. This default value causes a flash of different layout when server-side rendering on anything larger than the smallest mobile screen.

To support ssr I added a defaultViewport value which can be set on the server using either client hints or by parsing the user agent and naively setting the value based on the device. There is a way to do it as a user but it is not ideal.

## Alternative
There is a way to do this as a user by importing the ViewportContext and setting the value yourself however, you must remember to set it back to null on first render otherwise the viewport will always be the set to your initial value because ViewportProvider inherits from itself. And it has to be null specifically, not undefined.

Here is how I did it in my project.

```
import { useState, useEffect } from 'react';
import { ViewportContext } from '@salt-ds/core';

export function MyViewportProvider({
  children,
  defaultViewport
}: {
  children: React.ReactNode;
  defaultViewport?: number;
}) {
  const [value, setValue] = useState(defaultViewport);
  useEffect(() => {
    setFirstRender(null);
  }, []);

  return (
    <ViewportContext.Provider value={value}>
      {children}
    </ViewportContext.Provider>
  );
}
```

### DIFF
--- a/packages/core/src/salt-provider/SaltProvider.tsx
+++ b/packages/core/src/salt-provider/SaltProvider.tsx
@@ -10,7 +10,7 @@ import React, {
 import { AriaAnnouncerProvider } from "../aria-announcer";
 import { Breakpoints, DEFAULT_BREAKPOINTS } from "../breakpoints";
 import { Density, Mode, ThemeName } from "../theme";
-import { ViewportProvider } from "../viewport";
+import { ViewportProvider, ViewportProviderProps } from "../viewport";
 import { useIsomorphicLayoutEffect } from "../utils";
 
 import saltProviderCss from "./SaltProvider.css";
@@ -94,7 +94,7 @@ type SaltProviderBaseProps = {
   theme?: ThemeName;
   mode?: Mode;
   breakpoints?: Breakpoints;
-};
+} & Pick<ViewportProviderProps, "defaultViewport">;
 
 interface SaltProviderThatAppliesClassesToChild extends SaltProviderBaseProps {
   children: ReactElement;
@@ -122,6 +122,7 @@ export function SaltProvider({
   theme: themeProp,
   mode: modeProp,
   breakpoints: breakpointsProp,
+  defaultViewport,
 }: SaltProviderProps) {
   const inheritedDensity = useContext(DensityContext);
   const { theme: inheritedThemes, mode: inheritedMode } = useTheme();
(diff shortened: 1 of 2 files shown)


DavieReid:
We had a similar issue with Image widths in Mosaic.  We have a [hook](https://github.com/jpmorganchase/mosaic/blob/e55f3e68d2a8a9e8eaa2823bd46cb3fdc359a74b/packages/components/src/ThemeProvider.tsx#L8-L16) to check when the page has hydrated.  This feels similar to what you are doing in the Alternative approach.


How does this solution stop the flash of a different layout?  Would it not entirely depend on what device you were using?  So if I set a really high value and access it on a desktop it will look better than it does now, but a really high value on a mobile device would show the desktop view until the hook can run client-side...right?


P.S. 
Really weird that the `ViewportProvider` is just giving you back the viewport width.  I don't think `defaultValue` is what you want to call the prop.  It's unclear what it should be.

ZarkoPernar:
Using the client hint I don't get the "flash of incorrect layout" because the viewport width is consistent between ssr and client no matter the device.

ZarkoPernar:
Also, the guess function will possibly return a slightly different number between the server and an iphone browser, but if it is in the correct "responsive range" it should not cause any layout shift.

DavieReid:
What I was getting at in my comment is:

If someone arbitrarily sets a large `defaultValue` because they are developing on a desktop then without the client hint  or some other indication of device size then they *would* see layout shift on smaller devices.  You have identified this and used the client hint so I think we are in agreement.

Perhaps a comment on the prop to explain how the `defaultValue` should be used would be good.  Or rename the prop to indicate that its only to be used for SSR?  `ssrWidth` or something like that?

ZarkoPernar:
Ah sorry, yes @DavieReid you are right. This prop can definitely be misused in such a way. I imagined only the devs who cared enough about this sort of UX would even dig up this prop 😂 

And yeah, I was thinking of calling it ssrWidth initially but decided against it. I could rename it but like you say, I think documenting this prop properly is the key to it being useful.

joshwooding:
Considering that we probably will remove ViewportProvider and the responsive props from the layout components, is it worth considering a more future-proof option?

ZarkoPernar:
What is the proposed alternative @joshwooding?

PR #2507, opened 2023-09-26
AG grid theme changes

libertymayc:
Replaced borders (which was a bit hacked - had to give transparent border to avoid content shifting) to outlines

PR #2569, opened 2023-10-09
remove uitk theme files

alycrys:
Closes https://github.com/jpmorganchase/salt-ds/issues/2568


origami-z (commented):
Need to think about how to manage breaking change for this one. when to bump version, do we need another branch?

alycrys:
> Need to think about how to manage breaking change for this one. when to bump version, do we need another branch?

Not sure what would be the right way to deal with this breaking change. 
Is it necessary? Could we just leave the UITK theming as is?

origami-z:
> > Need to think about how to manage breaking change for this one. when to bump version, do we need another branch?
> 
> Not sure what would be the right way to deal with this breaking change. Is it necessary? Could we just leave the UITK theming as is?

With the new SASS approach, existing UITK theme code won't compile? If so, we have to remove them (at least until we find out adding back them with the new approach will be relatively easy?)

With the breaking change, we're not sure how long v29+ work will be done. To support potential changes like #2507 needed before v2 release, do we need something like new grid theme v2 branch?

Issue #2487, opened 2023-09-21
Combobox Next / Dropdown Next is mis-using blur event to close dropdown list


origami-z:
### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

[@salt-ds/lab@1.0.0-alpha.17](https://github.com/jpmorganchase/salt-ds/releases/tag/%40salt-ds%2Flab%401.0.0-alpha.17)

### Description

[Combobox Next](https://github.com/jpmorganchase/salt-ds/blob/8a57226922f9fcecdff5d27a55e920c3c517757a/packages/lab/src/combo-box-next/useComboBox.tsx#L72) and [Dropdown Next](https://github.com/jpmorganchase/salt-ds/blob/8a57226922f9fcecdff5d27a55e920c3c517757a/packages/lab/src/dropdown-next/useDropdownNext.tsx#L129) are both using trigger elements' blur event to close dropdown list, rather than detecting click outside of list (`useInteractions` `useDismiss`) and selection on list. This causes several issues, e.g., in desktop where the list is dismissed before selection handler can detect and fire, or clicking/dragging scrollbar of the list will dismiss the list.

Thanks to @jchristou for raising this and suggesting alternative.

### Steps to reproduce

See storybook
- http://localhost:6006/?path=/story/lab-dropdown-next--default
- http://localhost:6006/?path=/story/lab-combo-box-next--default

### Expected behavior

/

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

lukeac123:
pr merged - https://github.com/jpmorganchase/salt-ds/pull/2551

joshwooding:
PR, up needs reviews

PR #2600, opened 2023-10-12
use floating ui useDismiss hook in favour of on blur event

lukeac123:
Dropdown and Combobox - Add FloatingUIs useDismiss hook to Dropdown + remove onBlur handler
Combobox not longer opens on focus - maintains consistency with dropdown
Add event handlers into FloatingUI's getReferenceProps, as advised in the docs - https://floating-ui.com/docs/useInteractions
https://github.com/jpmorganchase/salt-ds/issues/2487

### DIFF
--- a/packages/lab/src/combo-box-next/ComboBoxNext.tsx
+++ b/packages/lab/src/combo-box-next/ComboBoxNext.tsx
@@ -164,6 +164,7 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
     clickHandler,
     focusHandler,
     selectHandler,
+    blurHandler,
     setSelectedItem,
     setHighlightedItem,
     mouseOverHandler,
@@ -182,7 +183,7 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
     setOpen,
     floating,
     reference,
-    getTriggerProps,
+    getReferenceProps,
     getPortalProps,
     getPosition,
   } = portalProps;
@@ -241,12 +242,16 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
           tabIndex: disabled ? -1 : 0,
           onFocus: focusHandler,
           onKeyDown: keyDownHandler,
-          onClick: clickHandler,
         }}
         role="combobox"
         variant={variant}
         value={inputValue}
-        {...getTriggerProps()}
+        ref={reference}
+        {...getReferenceProps}
+        {...getReferenceProps({
+          onBlur: () => blurHandler(),
+          onClick: () => clickHandler(),
+        })}
         {...restInputProps}
       />
 
--- a/packages/lab/src/combo-box-next/useComboBox.tsx
+++ b/packages/lab/src/combo-box-next/useComboBox.tsx
@@ -36,7 +36,7 @@ export const useComboBox = ({
     floating,
     reference,
     getPortalProps,
-    getTriggerProps,
+    getReferenceProps,
     getPosition,
   } = useComboboxPortal(PortalProps);
 
@@ -77,6 +77,10 @@ export const useComboBox = ({
   };
 
   const selectHandler = () => {
+    setOpen(false);
+  };
+
+  const blurHandler = () => {
     setOpen(false);
   };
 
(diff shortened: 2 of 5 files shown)


packages/lab/src/combo-box-next/useComboBox.tsx
```diff
@@ -64,25 +63,23 @@ export const useComboBox = ({
   }, [selectedItem]);
 
   const focusHandler = (event: FocusEvent<HTMLInputElement>) => {
-    setOpen(true);
     listFocusHandler(event);
     onFocus?.(event);
   };
 
-  const blurHandler = (event: FocusEvent<HTMLInputElement>) => {
```
Fercas123, line 72: check keyboard nav, you might still need blur

lukeac123: In what respect, enter and esc are handled by the keyDownHandler or are you referring to a user being able to tab onto the next component when the combobox is open ?

Fercas123: yeah, tab in keyboardhandler might do

lukeac123: Added blur event for keyboard nav

packages/lab/src/combo-box-next/ComboBoxNext.tsx
```diff
@@ -247,7 +246,12 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
         role="combobox"
         variant={variant}
         value={inputValue}
-        {...getTriggerProps()}
+        ref={reference}
+        {...getReferenceProps}
```
origami-z, line 250: why 2 `getReferenceProps`

packages/lab/src/combo-box-next/ComboBoxNext.tsx
```diff
@@ -247,7 +246,12 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
         role="combobox"
         variant={variant}
         value={inputValue}
-        {...getTriggerProps()}
+        ref={reference}
```
origami-z, line 249: If this is needed for floating ui to work, should it be at least `triggerRef`?
Or `inputRef` should already make it work?

packages/lab/src/combo-box-next/ComboBoxNext.tsx
```diff
@@ -247,7 +246,12 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
...
+        {...getReferenceProps}
+        {...getReferenceProps({
+          onBlur: () => blurHandler(),
```
origami-z, line 252: Are these the same effect compared to before? i.e. blur of input drives open to close?

Test with click on scrollbar of list

packages/lab/src/combo-box-next/ComboBoxNext.tsx
```diff
@@ -162,15 +161,16 @@ export const ComboBoxNext = forwardRef(function ComboBoxNext<T>(
     activeDescendant,
     focusVisibleRef,
     keyDownHandler,
+    clickHandler,
```
origami-z, line 164: These names are not descriptive enough, click on input? click on list?

origami-z (changes requested):
I don't think this changes the behaviour. We don't want portal to be closed because of blur event of the trigger element

feliciackh (commented):
noticed these: for DropdownNext, on chevron mouse click to open/close list, focus ring flickers/doesn't show. Also the list gets dismissed on clicking/dragging list scrollbar

packages/lab/src/combo-box-next/ComboBoxNext.tsx
lukeac123: yep is needed for floating-ui to work, before this was stated in trigger ref but to incorporate the event handlers into floating-ui i've stated it here instead

origami-z:
FYI, clicking scrollbar is a simulation for Desktop use cases

packages/lab/src/combo-box-next/ComboBoxNext.tsx
lukeac123: so the ref needs to be attached to the input - so will change to trigger ref

PR #2609, opened 2023-10-13
Spinner Doc Updates

### LINKED ISSUE #2591: Spinner v2: Supporting site documentation

lilyvc:
closes #2591 

Added documentation for small size.

Also updated code to align with design, having `small`, `medium` and `large` sizes instead of `small`, `default` and `large`. `default` still remains and maps to `medium`, to avoid a breaking change.


lilyvc:
@joshwooding is this change to the size prop okay?

cc @dplsek

.changeset/sixty-cameras-visit.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added 'medium' size to Spinner component with 'default' size now being an alias for 'medium'
```
joshwooding, line 5: We should document an intention to deprecate and remove "default"

lilyvc, line 5:
```suggestion
Added 'medium' size to Spinner component with 'default' size now being an alias for 'medium'. 'Default' is to be deprecated.
```

packages/core/src/spinner/Spinner.tsx
```diff
@@ -13,11 +13,23 @@ import { useDensity } from "../salt-provider";
  * Spinner component, provides an indeterminate loading indicator
  *
  * @example
- * <Spinner size="default | "large" | "small" />
+ * <Spinner size="small" | "medium" | "large" />
```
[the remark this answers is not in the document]
lilyvc: I think we want people to use medium so there's no need to add it to the comment

feliciackh (approved):

lukeac123 (approved):

PR #2613, opened 2023-10-13
remove repeated tokens & changeset from theme

feliciackh:
PR for Spinner and Progress introduced same foundations/size `--salt-size-bar-small` token, remove repeated reference of it

### DIFF
--- a/.changeset/afraid-tigers-rest.md
+++ b/.changeset/afraid-tigers-rest.md
@@ -1,9 +1,0 @@
----
-"@salt-ds/theme": minor
----
-
-New token added in `size.css` foundations:
-
-```diff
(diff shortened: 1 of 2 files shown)


origami-z (changes requested):
These are used in progress?

origami-z (approved):

feliciackh:
> These are used in progress?

Yeah used in both Spinner & Progress components, which was merged into main one after another so might have been overlooked

origami-z:
Updated release note (https://github.com/jpmorganchase/salt-ds/releases/tag/%40salt-ds%2Ftheme%401.9.0), remove the code in next relevant PR

PR #2615, opened 2023-10-16
add github link type to pattern resources

Fercas123:
We need to add a link to a github resource in navigation pattern, but the current implementation adds the figma link by default.

### DIFF
--- a/site/src/layouts/DetailPattern/Resources.tsx
+++ b/site/src/layouts/DetailPattern/Resources.tsx
@@ -4,26 +4,14 @@ import styles from "./Resources.module.css";
 import { Image } from "@jpmorganchase/mosaic-site-components";
 import { Link, Text } from "@salt-ds/core";
 
-type ResourceType = "figma" | "github";
-
-const LinkWithLogo = ({
-  href,
-  label,
-  resourceType,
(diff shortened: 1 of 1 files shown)


joshwooding (commented):
This is more a broader question, even though it could be considered "too smart". Is there a reason to not use the URL to determine what image to show? It seems a lot simpler and would avoid mistakes where a resourceType doesn't match the URL.

Fercas123:
> This is more a broader question, even though it could be considered "too smart". Is there a reason to not use the URL to determine what image to show? It seems a lot simpler and would avoid mistakes where a resourceType doesn't match the URL.

sounds good, so look for github or figma, else don't add anything sounds good?

joshwooding:
> > This is more a broader question, even though it could be considered "too smart". Is there a reason to not use the URL to determine what image to show? It seems a lot simpler and would avoid mistakes where a resourceType doesn't match the URL.
> 
> sounds good, so look for github or figma, else don't add anything sounds good?

Yeah, exactly. We can hide the icon if it's not something we would expect too

feliciackh (approved):

alycrys (approved):

Issue #2286, opened 2023-08-24
The `variant` prop for Text and Link should be deprecated and renamed to `color`


joshwooding:
### Area

UI Components

### The problem

The attribute they are changing is referred to as `color` in most of our documentation and the usage of `variant` here doesn't map to any of our other components

### The solution

Renamed `variant` to `color` in a backwards-compatible way and remove `variant` in v2.

### Alternatives and examples

N/A

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

libertymayc:
> the usage of variant here doesn't map to any of our other components

Variant elsewhere is changing between the primary and secondary tokens, granted these only change color, I don't see how it's different?

origami-z:
`variant` is the correct term derived from characteristics. `color` in Icon was more recent and not aligned?

joshwooding:
@origami-z This decision has already been discussed with design. If you have any concerns please start a chat on Teams

origami-z:
Done in #2619

PR #2619, opened 2023-10-17
adding new color property in Text and Link component

tejasnikam:
Fix issues [#2286 ](https://github.com/jpmorganchase/salt-ds/issues/2286)

### DIFF
--- a/packages/core/src/text/Text.tsx
+++ b/packages/core/src/text/Text.tsx
@@ -40,6 +40,7 @@ export type TextProps<T extends ElementType> = PolymorphicComponentPropWithRef<
      * Change text color palette
      */
     variant?: "primary" | "secondary";
+    color?: "primary" | "secondary";
   }
 >;
 
@@ -60,6 +61,7 @@ export const Text: TextComponent = forwardRef(
       style,
       styleAs,
       variant = "primary",
+      color = "primary",
       ...restProps
     }: TextProps<T>,
     ref?: PolymorphicRef<T>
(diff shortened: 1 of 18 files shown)


packages/core/src/text/Text.tsx
```diff
@@ -37,6 +37,7 @@ export type TextProps<T extends ElementType> = PolymorphicComponentPropWithRef<
      * Change text color palette
      */
     variant?: "primary" | "secondary";
+    color?: "primary" | "secondary";
```
joshwooding, line 40: We should copy Icon here

https://github.com/jpmorganchase/salt-ds/blob/5e48416c048e474b3750f405d677b608853b10ef/packages/icons/src/icon/Icon.tsx#L20

joshwooding (changes requested):
Thanks for this, we need to do a bit more work here:

- Add two Storybook stories one for Link and Text showcasing the prop
- Update our QA stories to prevent visual regressions

There's some other steps too but we can tackle those after

tejasnikam:
> Thanks for this, we need to do a bit more work here:
> 
> * Add two Storybook stories one for Link and Text showcasing the prop
> * Update our QA stories to prevent visual regressions
> 
> There's some other steps too but we can tackle those after

Thanks for the quick review

-  Have added two Storybook stories one for Link and Text showcasing the prop
-  Updated QA stories to prevent visual regressions
-  Also updated documentation

packages/core/src/text/Text.tsx
tejasnikam: Needful has been done

tejasnikam:
Hello Team,
Any updates on this.

joshwooding (approved):

Fercas123 (approved):

PR #2624, opened 2023-10-18
Remove table from component doc index page

### LINKED ISSUE #2516: Remove components table from index page
The components index page has a table that lists all available and upcoming components. The vertical nav and roadmap mostly duplicate the table's content, with two key exceptions:

- Table shows which components are still in lab
- Table links to a few docs that are still only available in Storybook

Once the below tasks are complete, the table will be redundant and we can remove it.

```[tasklist]
### Tasks
- [ ] #2502 
- [ ] #2506 
- [ ] #2212 
- [ ] #2229 
- [ ] #2228 
- [ ] Remove table from components/index page
```

origami-z:
Saves us manually maintain the status. Data Grid and Theme are the 2 items not in side bar yet, which i think is fine for now.

Although secondary nav doesn't work in mobile view, there's still the "next page" at the bottom in the footer, which would be a valid workaround.

Closes #2516


site/src/components/components/components-list.tsx
james-nash, line 1: Are we happy to completely discard all the meta-data here? (e.g. which version a component was first released in and the Figma status)

If that's been agreed then fine. Otherwise, it might be worth adding those bits of info into the frontmatter of the corresponding component docs pages. Although we don't (yet) display it on the site, we'll at least have it available in the future if we ever want to.

origami-z: first release is already in index's frontmatter. no need for Figma status, they should be aligned with code going forward

james-nash: Cool. Happy for this to get nuked then :-P

james-nash (approved):

feliciackh (approved):

Fercas123 (approved):

lilyvc (approved):

PR #2629, opened 2023-10-20
switch to core

lukeac123:
Mosaic Components has a dep on Switch in Labs, so this has been patched for now. Can be removed when mosaic-components has been updated.


Fercas123 (approved):

Fercas123 (dismissed):

site/src/components/components/components-list.tsx
```diff
@@ -670,9 +670,9 @@ export const componentDetails: ComponentDetails[] = [
     name: "Switch",
     description:
       "Allows the user to turn a specific value on or off from two choices—with the action having an immediate impact.",
-    devStatus: ComponentStatus.IN_LAB,
+    devStatus: ComponentStatus.READY,
     designStatus: ComponentStatus.READY,
-    availableInCodeSince: "1.0.0-alpha.15",
+    availableInCodeSince: "1.2.0",
```
james-nash, line 675: Given that `@salt-ds/core` is already on version 1.**10**.0, I'd expect this PR to go into the next minor release, so this should probably be 1.**11**.0.

It's a minor thing though. You could leave as is for now and correct this once a new version of Core that contains Switch has been released. Then whatever the actual Core version is will be known.

lukeac123: If you look at Spinner and Split Layout they're also listed as "1.2.0" so was following that - do all these need to be changed or is this the convention ?

Fercas123 (approved):

site/src/components/components/components-list.tsx
james-nash: The value should be whatever version the component first appeared in core. In the case of Spinner & Split Layout that was 1.2.0 (https://github.com/jpmorganchase/salt-ds/releases/tag/%40salt-ds%2Fcore%401.2.0), so it's correct.

However, for Switch, it will be whichever core release ends up containing this PR once it gets merged.

Like I say, it's something we can update later on, so I'll go ahead and approve the PR (everything else looks good). Also, I just saw there's another PR to remove this table altogether, so this info might become redundant anyway.

james-nash (approved):

Issue #2630, opened 2023-10-20
[Ag Grid Theme] Column headers are wrapping instead of ellipsis


alycrys:
### Package name(s)

AG Grid Theme (@salt-ds/ag-grid-theme)

### Package version(s)

salt-ds v1.3.3

### Description

i noticed that we have a minor bug with ag-grid-theme v1.3.3. when the column header text is longer than the column width, we used have ellipsis effect. on the new version however, it will cause the text to wrap and go to the next line

### Steps to reproduce

Navigate to https://storybook.saltdesignsystem.com/?path=/story/data-grid-ag-grid-theme--wrapped-header.
Observe column headers wrapping.

### Expected behavior

Column headers should show ellipsis when content is wider than the header cell.

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

PR #2189, opened 2023-08-11
Bug fixes for ag-grid-theme

libertymayc:
based on MTK feedback
closes #2185 

Fixed bugs:

- Filter icon incorrect color in HD compact theme
- Vertical alignment of cells
- Column group cell
- Alignment of header cell separators with column group cell separators
- Status bar style enhancements
- Clipping of sort arrow in header row
- Floating filter new design [cc @bhoppers2008 ]
- Alignment of cell borders
- React container styles for custom components

Issue #2185, opened 2023-08-11
AG-Grid Theme feedback fixes


libertymayc:
1. Floating filters 
(Has been redesigned, but MTK unsure about redesign, need @bhoppers2008 to perhaps chat with MTK designer to discuss)
2. Column group cells
- one continuous vertical line instead of short open line, needs fix
3. Header Cell - need design input @bhoppers2008  @dplsek 
- the short vertical line is getting shorter (HD compact) 
- misaligned with the column group cell vertical lines by 2px, needs fix
4. Status bar - need design input @bhoppers2008  @dplsek
- right side text has a smaller font size, needs fix
- the left side looks shifted up and to the left, needs margin
5. Wrong padding in HD/HD compact
- Ticket 8349 (Figma file for DataGrid says each cell should have 4px padding either side, but the AgGrid theme applies 7px)

joshwooding:
@bhoppers2008 to talk to the team about feedback, @libertymayc looking at the blotter demo.

@jpmorganchase/salt-developer to review the PR

PR #2631, opened 2023-10-20
[HOLD][Ag Grid Theme] Column headers are wrapping instead of ellipsis

alycrys:
Closes https://github.com/jpmorganchase/salt-ds/issues/2630

Introduced in https://github.com/jpmorganchase/salt-ds/pull/2189 while fixing bugs https://github.com/jpmorganchase/salt-ds/issues/2185

### DIFF
--- a/packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
+++ b/packages/ag-grid-theme/css/_salt-ag-theme-mixin.scss
@@ -296,11 +296,6 @@
     margin-left: var(--salt-spacing-100);
   }
 
-  .ag-header-cell-auto-height .ag-header-cell-text {
-    overflow: hidden;
-    white-space: pre-wrap;
-  }
-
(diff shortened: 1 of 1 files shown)


origami-z (commented):
Wrapped header is needed with specific configuration, the story demos the capability. 
https://blog.ag-grid.com/wrapping-column-header-text/

alycrys:
> Wrapped header is needed with specific configuration, the story demos the capability. https://blog.ag-grid.com/wrapping-column-header-text/

If this was a new feature, then it should have been signaled as such. This seems to be a bug to both MTK and @bhoppers2008 .

origami-z:
What does MTK expects when auto height is turned on for a header? I'd assume auto height is only needed when multi-line header is needed

libertymayc:
this was originally requested by MTK, surprised to hear its now being seen as a bug

PR #2648, opened 2023-10-26
Turn off prefer-nullish-coalescing

origami-z:
A lot of false negative between `??` and `||`

Only 13 errors reduced

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -44,8 +44,12 @@ module.exports = {
         ],
         "@typescript-eslint/no-misused-promises": "off",
         "@typescript-eslint/no-unsafe-assignment ": "off",
-        // nullish coalescing introduces a lot false negative
-        "@typescript-eslint/prefer-nullish-coalescing": "off",
+        "@typescript-eslint/prefer-nullish-coalescing": [
+          "error",
+          {
+            ignoreConditionalTests: true,
+          },
+        ],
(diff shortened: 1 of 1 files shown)


Fercas123 (dismissed):

james-nash (dismissed):

.eslintrc.js
```diff
@@ -44,6 +44,8 @@ module.exports = {
         ],
         "@typescript-eslint/no-misused-promises": "off",
         "@typescript-eslint/no-unsafe-assignment ": "off",
+        // nullish coalescing introduces a lot false negative
+        "@typescript-eslint/prefer-nullish-coalescing": "off",
```
joshwooding, line 48:
```suggestion
    		"@typescript-eslint/prefer-nullish-coalescing": [
          		"error",
     		     {
    		        ignoreConditionalTests: true,
  		        },
```

Have you considered only turning of instances that have a high rate of false positives?

joshwooding (changes requested):

.eslintrc.js
joshwooding: Actually a better option would be:

```suggestion
        "@typescript-eslint/prefer-nullish-coalescing": [
        "error",
        {
        	ignorePrimitives: { boolean: true }
        }],
```

origami-z:
Back to draft until i have time to look at this again

PR #2655, opened 2023-10-27
CSS Extraction & Disable Style Injection

DavieReid:
The appDir version of Mosaic sites suffers from layout shift and flash of un-styled content (FOUC) due to the CSS injection feature of Salt components.  Essentially SSR'd apps are rendered on the server and have no salt styling until they are rendered on the client. 

You can also see a [workaround](https://gist.github.com/ZarkoPernar/719e95492654df2042da1eb2ee8d247b) that @ZarkoPernar has implemented here that imports component css files directly from the dist folder.

This PR is a POC that aims to start the discussion on salt component CSS files and disabling the CSS injection feature.  It includes:

- A way to enable/disable component CSS injection using a `SaltProvider`

A user would be able to disable CSS injection for their app by:

```
<SaltProvider enableStyleInjection={false} /** other props*/ >// children</SaltProvider>
```

### Questions/Observations

1. Should we bundle the component CSS into 1 file?
EDIT: we are bundling into 1 css file per package e.g. core css and labs css

At present, a user would need to go through their app and understand all the components they are using and manually add an CSS import for the component into their app.  This could get tricky with composite components e.g. `Dropdown` which would need it's own styles plus those of `Button` and `List`

3. `SaltProvider` apply classes to functionality

EDIT: These classes need to be added manually

This is still dynamically adding a theme and density class to a DOM element...again this might mean the FOUC will still happen so it might mean that these classes need to be manually added by the user.  Will try run it and see what happens

4. Change output Directory?

The CSS files are still output to a `dist` folder.  We might want to output to a more user friendly location...

5. Code organisation and naming

The `useComponentCssInjection` function is inside a folder called `use-style-injection` and the folder name is the same as the new hook I have added.  Might be a good idea to rename a few things to avoid confusion


ZarkoPernar:
As far as build tooling for this is concerned, I believe the css extraction as you intended would be done by tsup as part of [this work Josh started](https://github.com/jpmorganchase/salt-ds/pull/2508). But that might be a bigger piece, and your work could be an interim solution. Not sure what the rest of the team thinks.

DavieReid:
> As far as build tooling for this is concerned, I believe the css extraction as you intended would be done by tsup as part of [this work Josh started](https://github.com/jpmorganchase/salt-ds/pull/2508). But that might be a bigger piece, and your work could be an interim solution. Not sure what the rest of the team thinks.

CSS extraction is eximental in `tsup`

DavieReid:
I have tried this out on the new Mosaic app router branch.  The changes I made were:

## Update site-preset-styles package

The core and lab css files are included among the other CSS files Mosaic considers the preset styling.

```
import '@salt-ds/core/css/salt-core.css';
import '@salt-ds/lab/css/salt-lab.css';
```

## Update Mosaic ThemeProvider

set `enableStyleInjection` to be `false` and manually apply `salt-theme` and `salt-density-medium` classes

joshwooding:
> As far as build tooling for this is concerned, I believe the css extraction as you intended would be done by tsup as part of [this work Josh started](https://github.com/jpmorganchase/salt-ds/pull/2508). But that might be a bigger piece, and your work could be an interim solution. Not sure what the rest of the team thinks.

I've been debating this myself. I think there's a risk to this approach where technically you could end up with different CSS if you used the bundled output vs the injected CSS. I don't think it's a huge issue though but it will be hard to catch

DavieReid:
> > As far as build tooling for this is concerned, I believe the css extraction as you intended would be done by tsup as part of [this work Josh started](https://github.com/jpmorganchase/salt-ds/pull/2508). But that might be a bigger piece, and your work could be an interim solution. Not sure what the rest of the team thinks.
> 
> I've been debating this myself. I think there's a risk to this approach where technically you could end up with different CSS if you used the bundled output vs the injected CSS. I don't think it's a huge issue though but it will be hard to catch

Can you give an example of how the CSS could be different?  Is it the ordering?

ZarkoPernar:
With the runtime injection, the order is based on order of rendering of components, and this is different on each page/layout. At least with bundling we control the ordering (to whatever extent the tool lets us). I think bundling would make it easier for us to catch bugs in our css, like for example if we tried overwriting .saltInput in Dropdown and didn't do it properly (using css vars or higher specificity) and missed the bug because the ordering of injected css happened to be a certain way in our storybook/tests.

docs/components/QAContainerNoStyleInjection.tsx
```diff
@@ -0,0 +1,35 @@
+import { useEffect } from "react";
+import { QAContainer, QAContainerProps } from "./QAContainer";
+
+// eslint-disable-next-line import/no-unresolved
+import coreCss from "../../dist/salt-ds-core/css/salt-core.css?inline";
```
DavieReid, line 5: This means you will have to have run a build before running storybook 😢

.storybook/preview.tsx
```diff
@@ -106,6 +106,15 @@ export const globalTypes: GlobalTypes = {
       title: "Scaffold",
     },
   },
+  styleInjection: {
```
origami-z, line 109: Is this toggle here for us to see components without css applied? Given in theory, QA stories can handle different situation without global setting.

packages/core/scripts/build.mjs
```diff
@@ -0,0 +1,57 @@
...
+const absWorkingDir = path.resolve(__dirname, "..");
+const entryFile = "src/salt-core.css";
+
+/** MAIN */
+deleteSync([buildFolder], { force: true });
+createCssEntryFile(createEntryFile);
+runBuild();
+
+function createCssEntryFile(shouldCreate) {
+  if (shouldCreate) {
```
origami-z, line 22: Should this always run as part of build process? Otherwise it can be easily forgotten when new components are added

origami-z (commented):
Component fixes and this new feature would need changeset?

packages/core/src/salt-core.css
```diff
@@ -0,0 +1,42 @@
+@import "accordion/Accordion.css";
```
ZarkoPernar, line 1: This is not automated, right? Seems like we could easily forget to add files here if the runtime injection is the default

packages/core/src/switch/Switch.css
```diff
@@ -102,7 +102,8 @@
 }
 
 @media (prefers-reduced-motion) {
-  .saltSwitch-thumb {
+  .saltSwitch-input:focus-visible + .saltSwitch-track .saltSwitch-thumb,
+  .saltSwitch:hover .saltSwitch-thumb {
```
ZarkoPernar, line 106: Why the component changes?

packages/core/stories/banner/banner.qa.stories.tsx
```diff
@@ -37,31 +42,36 @@ const ErrorBanner = () => <BasicBannerExample status={"error"} />;
...
+    <WarningBanner />
+    <SuccessBanner />
+  </QAContainer>
 );
 
 ExamplesGrid.parameters = {
   chromatic: { disableSnapshot: false },
 };
+
+export const NoStyleInjectionGrid: StoryFn<QAContainerNoStyleInjectionProps> = (
```
ZarkoPernar, line 58: Could we not automate this, and build it in to the QAContainer?

packages/lab/scripts/build.mjs
```diff
@@ -0,0 +1,57 @@
+import esbuild from "esbuild";
```
ZarkoPernar, line 1: any reason we couldn't reuse these scripts? Seem to be identical between core and lab, with a few arguments

packages/core/stories/banner/banner.qa.stories.tsx
ZarkoPernar: So you don't have to change all the QA stories

packages/core/scripts/build.mjs
DavieReid: If we can add a new build step in do it first then sure...the issue I had was that 3 build tasks get kicked off in parrallel all deleting and recreating the same file so the build failed a lot

packages/core/src/salt-core.css
DavieReid: You can forget to do it yes....but there is a command in package.json that will recreate the entry file for you. Will see if we can change the order of the build commands

packages/core/src/switch/Switch.css
DavieReid: with the bundled css and a switch that is checked by default, you see the transition of the switch thumb from left to right.  The change ensures that the transition only happens if the switch input is focused

packages/core/stories/banner/banner.qa.stories.tsx
DavieReid: I wanted to split it out so that I could be sure that the components are 100% being styled by the CSS or vice versa

packages/lab/scripts/build.mjs
DavieReid: yeah probably...where would you keep them?

.storybook/preview.tsx
DavieReid: Yes...found it useful when testing things so left it in

packages/lab/scripts/build.mjs
ZarkoPernar: Something like `/tooling/css-bundle/build.mjs`?

origami-z (dismissed):

ZarkoPernar (approved):

packages/core/scripts/build.mjs
DavieReid: Fixed

origami-z (approved):

ZarkoPernar (approved):

joshwooding:
The gen:css scripts don't work on Windows :(

PR #2656, opened 2023-10-27
[Ag Grid Theme] Remove added styles when autoHeaderHeight is applied

alycrys:


feliciackh (dismissed):

lilyvc (approved):

feliciackh (approved):

packages/ag-grid-theme/stories/examples/WrappedHeader.tsx
```diff
@@ -67,6 +67,7 @@ const WrappedHeader = (props: AgGridReactProps) => {
           defaultColDef={{
             ...propsColDefs,
             autoHeaderHeight: true,
+            wrapHeaderText: true,
```
origami-z, line 70: This is only v28 feature? https://blog.ag-grid.com/whats-new-in-ag-grid-28/

The theme supports v25-28?

alycrys: Dang! You're right. One sec

alycrys: Ok. So as this feature was provided by ag grid only in version 28, it will not be available in the earlier versions. I wanted to check to see if there would be a way to work it somehow, but there isn't. So if Ag Grid doesn't supply the feature, we won't have it either. 
This PR ensures that the feature works correctly by removing our override and configuring our example accordingly.
So yes, version 28 only.

PR #2659, opened 2023-10-30
[WIP] - make muti-line input scrollable and resizeable

lukeac123:
Multi-line Input is not scrollable when content overflows.
1. Make Input Scrollable when content overflows in the y direction 
2. Make Input Resizeable


lukeac123:
Confirmation required from design

PR #2663, opened 2023-10-31
Expand Sidebar to Current Page

lilyvc:
This PR adds logic to expand sidebar to whatever page is selected. This behaviour doesn't work by default as Mosaic uses breadcrumbs data to determine which page to expand, but we have disabled the Breadcrumbs plugin.

It is just a draft to try and diagnose a problem with nested pages collapsing in the Sidebar.

### DIFF
--- a/site/src/components/PageNavigation.tsx
+++ b/site/src/components/PageNavigation.tsx
@@ -1,0 +1,42 @@
+import React from "react";
+import { SidebarItem, useRoute, useStore } from "@jpmorganchase/mosaic-store";
+import { VerticalNavigation } from "@jpmorganchase/mosaic-site-components";
+
+export const PageNavigation = () => {
+  const menu = useStore((state) => state.sidebarData) || [];
+  let { route } = useRoute();
+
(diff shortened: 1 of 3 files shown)


lilyvc:
> @lilyvc Could you add more content where this change has come from? Does this fix the active page state on the navigation menu?

Oh is there a problem with the active page state? This could help, but not sure of the issue. This PR is just because @feliciackh was adding a nested page in the sidebar and noticed it kept collapsing. We figured it is because of two things - it's reloading for some reason and when it reloads it doesn't know what should be expanded. This fixes the latter so I wanted to share so we could see the preview and if the former issue persists.

feliciackh:
To provide more context, [Dashboard pattern published recently](https://www.saltdesignsystem.com/salt/patterns/dashboards/index) uses accordion in sidebar. 
 
Steps to reproduce the issue @lilyvc noted above:
1. click on Dashboards in the sidebar, accordion expands
2. click on Analytical Dashboard, accordion collapses when it should remain expanded on the active page

PR #2665, opened 2023-11-01
Viewport provider fix

lilyvc:
This PR ensures Viewport is read and set immediately. Previously, Viewport was sometimes 0 initially causing a flash of layouts.

### DIFF
--- a/packages/core/src/viewport/ViewportProvider.tsx
+++ b/packages/core/src/viewport/ViewportProvider.tsx
@@ -26,6 +26,7 @@ const ViewportProvider = ({ children }: ViewportProviderProps) => {
       );
 
       observer.observe(document.body);
+      setViewport(document.body.getBoundingClientRect().width);
     }
 
     return () => {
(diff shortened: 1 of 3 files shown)


ImmanuelBaskaran (approved):

packages/core/src/viewport/ViewportProvider.tsx
```diff
@@ -26,6 +26,7 @@ const ViewportProvider = ({ children }: ViewportProviderProps) => {
       );
 
       observer.observe(document.body);
+      setViewport(document.body.getBoundingClientRect().width);
```
origami-z, line 29: Will this break in SSR?

lilyvc: I think it being inside `useIsomorphicLayoutEffect` means it shouldn't break

origami-z:
Did you log the output? This will still get `0`

Issue #2671, opened 2023-11-02
Utilizing Salt and UITK In combination prevents styling from applying correctly


adamrideout:
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

    "@salt-ds/core": "^1.10.0",

### Description

We're in the process of migrating our application from using UITK to using Salt. Since not all components are available yet, we have a mixture of the two. The problem is, the component styling from each is interfering with each other. For example, I'm trying to add a group of salt RadioButtons, but they end up completely misaligned. When I remove the `import "@jpmuitk/style/css/jpmuitk.css";` from our app.tsx, the styling returns to normal. However, without it the styling for all of our UITK components becomes completely messed up. What is the recommended process for adding salt styling and UITK styling without them interfering with each other? Thanks

### Steps to reproduce

Created a react application
Install UITK
Install SALT
Add 
```
import "@jpmuitk/style/css/jpmuitk.css";

import "@salt-ds/theme/index.css";
```
to your application.
Add salt radio buttons.

### Expected behavior

I expected the styling of the radio buttons to be correct, with the buttons aligned with the label

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

___

Affected Components

- [x] Checkbox
- [x] Radio Button
- [ ] Banner

james-nash:
Bug ranking score: **52** (using rubric v3)

Priority: **high**

```
========== BEGIN BUG RUBRIC SCORE DATA ===========
eyJ2YWx1ZXMiOnsiYWZmZWN0ZWRQZW9wbGVPcHRpb25zIjoxOS
wiYWZmZWN0ZWRDb21wb25lbnRzT3B0aW9ucyI6MTIsImZyZXF1
ZW5jeU9wdGlvbnMiOjE5LCJpbXBhY3RPcHRpb25zIjoyMCwiYW
ZmZWN0ZWREc09wdGlvbnMiOjE3LCJ1cmdlbmN5T3B0aW9ucyI6
MTksInhGYWN0b3JPcHRpb25zIjoxOH0sInJ1YnJpY1ZlcnNpb2
4iOjN9
=========== END BUG RUBRIC SCORE DATA ============
```

[...]

PR #2679, opened 2023-11-03
Fix Salt / UITK text property compatibility

origami-z:
Added text styles appearing in UITK's `body` to storybook (not website), so all subsequent components we're working on will have that checked out. 

Try to fix all core component's defect as well in this PR, mostly rely on Chromatic snapshot comparison.

Given Chromatic snapshot are rendered using `QAContainer`, which has `SaltProvider` wrapped inside, additional `body` styles will not be taken effect (unlike the order of`:root.salt-theme body`). So I also added additional text styles to `QAContainer` as well (`.background-item-wrapper`)

#2671

## Alternative Approaches A

Add default text styles to a fixed global className (similar to `.saltFocusVisible`), add apply that to components needed.

I didn't choose this with 2 reasons

1. Not all components needs the same default, e.g. for Button, which becomes an override of bold text on top of default value. It will either pollute relying more on CSS order, or increased CSS specificity 
1. A few components need to expose component CSS API for override, e.g. `--saltComponent-fontSize`, which this approach doesn't help

## Alternative Approaches B

Wrap every component with Text component which provides this.

1. I think most component we write should be atomic, i.e. less dependency between each other, just like we likely not use layout components ourselves. This helps isolated update and maintenance. 
1. User override will be harder, given additional nested DOM structure introduced

### DIFF
--- a/.changeset/orange-bulldogs-glow.md
+++ b/.changeset/orange-bulldogs-glow.md
@@ -1,0 +1,16 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly impacted by external global styles, which should follow text characteristic from Salt theme.
+
+- Avatar
+- Badge
+- Banner
+- Card
+- Form Field
+- Panel
+- Switch
+- Text
+- Toast
+- Tooltip
--- a/.changeset/orange-bulldogs-glow.md
+++ b/.changeset/orange-bulldogs-glow.md
@@ -2,7 +2,7 @@
 "@salt-ds/core": patch
 ---
 
-Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly impacted by external global styles, which should follow text characteristic from Salt theme.
+Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly inheriting external global styles, which should follow the text characteristic from the Salt theme.
 
 - Avatar
 - Badge
--- a/.changeset/young-pets-exist.md
+++ b/.changeset/young-pets-exist.md
@@ -1,0 +1,14 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly impacted by external global styles, which should follow text characteristic from Salt theme.
+
+- Content Status
+- Dialog
+- Form Field Legacy
+- List
+- List Next
+- Tab
+- Tab Next
+- Toolbar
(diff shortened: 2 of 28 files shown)


packages/core/src/card/Card.css
```diff
@@ -6,6 +6,11 @@
   border-style: var(--salt-container-borderStyle);
   border-color: var(--salt-container-primary-borderColor);
   padding: var(--saltCard-padding, var(--salt-spacing-300));
+
+  font-family: var(--salt-text-fontFamily);
```
el-dav, line 10: If utik components were placed inside the Card would they then inherit the font-family etc from the Card instead of body and therefore still have the compatibility issues?

origami-z: Good question, I checked a couple components from UITK. They all have text properties defined in their own stylesheet (not inheriting from `body`), so won't be impacted. i.e. UITK components are styled to UITK theme/styles. `body` styles are just an impact for consumer's own components.

This does mean that the intent of this change is less about (specifically) UITK, but more to make Salt components adhere to Salt themes.

.changeset/orange-bulldogs-glow.md
```diff
@@ -0,0 +1,15 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed components text properties impacted by global styles
```
libertymayc, line 5: did we want to say 'external global styles'? we also have our own global text styles in global.css (applied to the .salt-theme class) which should technically be correct for the below components so it might seem like thats not the case

origami-z: Updated with more details

.changeset/orange-bulldogs-glow.md
```diff
@@ -0,0 +1,16 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly impacted by external global styles, which should follow text characteristic from Salt theme.
```
joshwooding, line 5:
```suggestion
Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly inheriting external global styles, which should follow the text characteristic from the Salt theme.
```

joshwooding (dismissed):

.changeset/young-pets-exist.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly impacted by external global styles, which should follow text characteristic from Salt theme.
```
origami-z, line 5:
```suggestion
Fixed component text properties (`font-weight`, `font-family`, `font-size`, `line-height`) incorrectly inheriting external global styles, which should follow the text characteristic from the Salt theme.
```

joshwooding (approved):

ImmanuelBaskaran (approved):

PR #2685, opened 2023-11-07
Add QA stories to all core components

origami-z:
Add missing QA stories to core components, so can do comparison in #2679

I specifically named all group name as `<Component> QA` so it's much clearer on Chromatic which component contains change


feliciackh (dismissed):
would it benefit to update all other core components QA story name to <component QA> as well 🤔

origami-z:
> would it benefit to update all other core components QA story name to as well 🤔

updated core ones

feliciackh (approved):

joshwooding (approved):

PR #2687, opened 2023-11-07
Fix Tooltip status icon alignment

origami-z:
LD / TD top margin looks off

LD / TD tooltip icon alignment off

### DIFF
--- a/packages/core/src/tooltip/Tooltip.css
+++ b/packages/core/src/tooltip/Tooltip.css
@@ -3,7 +3,7 @@
   --tooltip-background: var(--saltTooltip-background, var(--salt-container-primary-background));
   --tooltip-zIndex: var(--saltTooltip-zIndex, var(--salt-zIndex-flyover));
   /* apply icon margin based on the text line height so it's aligned on all densities */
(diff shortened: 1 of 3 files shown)


feliciackh (approved):

packages/core/src/tooltip/Tooltip.css
```diff
@@ -3,7 +3,7 @@
   --tooltip-background: var(--saltTooltip-background, var(--salt-container-primary-background));
   --tooltip-zIndex: var(--saltTooltip-zIndex, var(--salt-zIndex-flyover));
   /* apply icon margin based on the text line height so it's aligned on all densities */
-  --tooltip-icon-marginTop: calc((var(--salt-text-lineHeight) - 12px) / 2);
+  --tooltip-icon-marginTop: calc((var(--salt-text-lineHeight) - max(var(--salt-icon-size-base), 12px)) / 2);
```
origami-z, line 6: We have to compute `max` here, `--icon-size` variable of Icon component is not accessible at Tooltip level

libertymayc (approved):

PR #2697, opened 2023-11-08
Fix Combo Box Next `onChange` using stale values

el-dav:
Fix issue where combobox onChange was lagging behind the text typed into the combobox by one render. 

This was caused by sending inputValue to the onInputChange callback which had been set via setInputValue, but won't be updated till the next render.

Instead this change passes the value from event/target.value which will be the updated value.

### DIFF
--- a/.changeset/ninety-queens-retire.md
+++ b/.changeset/ninety-queens-retire.md
@@ -2,4 +2,4 @@
 "@salt-ds/lab": patch
 ---
 
-Fix combobox onchange using stale value
+Fix Combo Box Next `onChange` using stale values
--- a/.changeset/ninety-queens-retire.md
+++ b/.changeset/ninety-queens-retire.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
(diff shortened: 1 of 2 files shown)


joshwooding (changes requested):
Could you add a changeset please

.changeset/ninety-queens-retire.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fix combobox onchange using stale value
```
origami-z, line 5: Probably need "next"

origami-z (approved):

.changeset/ninety-queens-retire.md
joshwooding, line 5:
```suggestion
Fix Combo Box Next `onChange` using stale values
```

el-dav: Ah yes, thanks

origami-z (approved):

joshwooding (approved):

PR #2703, opened 2023-11-10
Card Secondary Variant

### LINKED ISSUE #2658: Card secondary variant

lilyvc:
closes #2658

### DIFF
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -12,7 +12,7 @@
 }
 
 .saltCard-secondary {
-  background: var(--salt-container-secondary-background);
+  background: var(--saltCard-background, var(--salt-container-secondary-background));
   border-color: var(--salt-container-secondary-borderColor);
 }
 
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -8,6 +8,11 @@
   padding: var(--saltCard-padding, var(--salt-spacing-300));
 }
 
+.saltCard-secondary {
+  background: var(--salt-container-secondary-background);
+  border-color: var(--salt-container-secondary-borderColor);
+}
+
 /*
  * **Deprecated:** The following styles are deprecated
  *  Use Interactable Card component instead
--- a/packages/core/src/card/Card.css
+++ b/packages/core/src/card/Card.css
@@ -1,11 +1,14 @@
 /* Styles applied to the root element */
 .saltCard {
-  background: var(--saltCard-background, var(--salt-container-primary-background));
   box-shadow: var(--saltCard-boxShadow, var(--salt-overlayable-shadow));
   border-width: var(--salt-size-border);
   border-style: var(--salt-container-borderStyle);
-  border-color: var(--salt-container-primary-borderColor);
   padding: var(--saltCard-padding, var(--salt-spacing-300));
+}
+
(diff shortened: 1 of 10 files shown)


Fercas123 (changes requested):
looks good! just 1 repeated class

packages/core/src/card/InteractableCard.css
```diff
@@ -11,6 +11,9 @@
   padding: var(--saltCard-padding, var(--salt-spacing-300));
 }
 
+.saltInteractableCard-secondary {
+  background: var(--salt-container-secondary-background);
+}
```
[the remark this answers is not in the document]
lilyvc: thanks for noticing!

Fercas123 (dismissed):

ImmanuelBaskaran (approved):

feliciackh (dismissed):

packages/core/src/card/Card.css
```diff
@@ -8,6 +8,11 @@
   padding: var(--saltCard-padding, var(--salt-spacing-300));
 }
 
+.saltCard-secondary {
```
origami-z, line 11: Should `primary` variant sharing same API?

lilyvc: yeah good point, I will update this

alycrys:
Would it make sense to add an example to storybook also? Maybe to QA story for chromatic testing? Not sure what will happen to the storybook docs, so maybe we don't need an example there.

packages/core/src/card/Card.css
origami-z: `--saltCard-background` not the same after update?

packages/core/src/card/InteractableCard.css
```diff
@@ -1,16 +1,25 @@
 /* Styles applied to InteractableCard */
 .saltInteractableCard {
-  background: var(--saltCard-background, var(--salt-container-primary-background));
```
origami-z, line 3: `--saltCard-background` will be gone?

packages/core/src/card/InteractableCard.css
```diff
@@ -64,14 +73,30 @@ a:link .saltInteractableCard * {
...
+  border-color: var(--saltCard-borderColor-disabled, var(--salt-container-primary-borderColor-disabled));
+  background: var(--saltCard-background-disabled, var(--salt-container-primary-background-disabled));
+}
+
+/* Styles applied to InteractableCard secondary variant if `disabled={true}` */
+.saltInteractableCard-secondary.saltInteractableCard-disabled,
+.saltInteractableCard-secondary.saltInteractableCard-disabled:focus,
+.saltInteractableCard-secondary.saltInteractableCard-disabled:hover,
+.saltInteractableCard-secondary.saltInteractableCard-disabled:active {
+  border-color: var(--salt-container-secondary-borderColor-disabled);
```
origami-z, line 96: Different API here as well

packages/core/src/card/Card.css
lilyvc: I've updated - is this what you mean? or should it be a different token like:

```
background: var(--saltCard-secondary-background, var(--salt-container-secondary-background));
```

Fercas123 (approved):

feliciackh (approved):

PR #2571, opened 2023-10-09
Theme 1.0 updates

feliciackh:
- Please review commit by commit, i.e. (1-10) refers to excel rows 1 to 10; (47-54) refers to excel rows 47 to 54
- There's been design reverts that differ slightly from excel sheet, commit messages reflect that

Closes #2530, #1951, #2489, #2518, #2014, #2190, #1417

PR #2705, opened 2023-11-10
add variants Notation and Action to Text component

feliciackh:
Blocked by #2571 
part of Theme 1.0 updates


feliciackh (commented):
cs

packages/core/src/text/Text.tsx
```diff
@@ -32,17 +28,17 @@ export type TextProps<T extends ElementType> = PolymorphicComponentPropWithRef<
...
      * Change text color palette
      */
     variant?: "primary" | "secondary";
   }
 >;
 
-type TextComponent = <T extends ElementType = "div">(
-  props: TextProps<T>
-) => ReactElement | null;
+type TextComponent = <T extends ElementType = "div">(props: TextProps<T>) => ReactElement | null;
```
joshwooding, line 41: What caused this formatting to be different?

joshwooding:
This should probably go to main rather than theme-updates. We can merge this after the other PR

.changeset/little-pillows-fry.md
```diff
@@ -0,0 +1,13 @@
+---
+"@salt-ds/core": minor
+---
+
+`Text` component has two new variants: `Action` and `Notation`. See [Storybook](https://storybook.saltdesignsystem.com/?path=/story/core-text-text-qa--all-variants-grid) for more information.
+
+```tsx
```
joshwooding, line 7: Are we avoiding mentioning Action here?

packages/core/src/text/Action.tsx
```diff
@@ -0,0 +1,22 @@
...
+const withBaseName = makePrefixer("saltText");
+
+export const Action = forwardRef<
+  HTMLSpanElement,
+  Omit<TextProps<"span">, "as">
+>(function Action({ children, className, ...rest }, ref) {
+  return (
+    <Text
+      as="span"
+      className={clsx(className, withBaseName(`action`))}
```
joshwooding, line 15:
```suggestion
      className={clsx(className, withBaseName("action"))}
```

packages/core/src/text/Text.tsx
```diff
@@ -32,7 +32,9 @@ export type TextProps<T extends ElementType> = PolymorphicComponentPropWithRef<
       | "label"
       | "display1"
       | "display2"
-      | "display3";
+      | "display3"
+      | "notation"
```
joshwooding, line 36: This won't work without the CSS being updated

.changeset/little-pillows-fry.md
feliciackh: originally kept it succinct in theme PR but now that it's separated i've added Action too

packages/core/src/text/Display.tsx
```diff
@@ -12,7 +12,7 @@ export const Display1 = forwardRef<
   return (
     <Text
       as="span"
-      className={clsx(className, withBaseName(`display1`))}
+      className={clsx(className, withBaseName("display1"))}
```
joshwooding, line 15: I know this was here already but isn't this equivalent to styleAs="display1" ?

.changeset/little-pillows-fry.md
joshwooding: changesets should always cover the complete set of new additions

packages/core/src/text/Display.tsx
feliciackh: yes they are, good spot i'll update these to use styleAs to encourage APi use

PR #2720, opened 2023-11-14
Fix height of closable pill next

Fercas123:
adding max-width to the close button inside pill to avoid specificity issues when styles are applied from an external wrapper

### DIFF
--- a/.changeset/moody-weeks-hunt.md
+++ b/.changeset/moody-weeks-hunt.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fix height of close button in `Pill`
--- a/packages/lab/src/pill-next/PillNext.css
+++ b/packages/lab/src/pill-next/PillNext.css
@@ -47,6 +47,7 @@
 }
 
 .saltPillNext-close-button {
+  max-height: calc(var(--salt-size-base) - var(--salt-spacing-100));
   --saltButton-height: calc(var(--salt-size-base) - var(--salt-spacing-100));
   --saltButton-padding: var(--salt-spacing-50);
 }


origami-z (commented):
The change feels a little random. Is there a need to add a test or something else to show why this is needed?

Fercas123:
> The change feels a little random. Is there a need to add a test or something else to show why this is needed?

Screenshot 2023-11-14 at 13 40 43

The issue happens when styles are applied to `.saltButton` externally (from multilineinput in this case) and override the class passed within the pill.
Multiline input applies a general style rule because it is used in all buttons inside end and start adornments

origami-z:
> The issue happens when styles are applied to `.saltButton` externally (from multilineinput in this case) and override the class passed within the pill.

If this comes from `Button`, should there be a height CSS API exposed, and overridden in `PillNext` and adornment accordingly? Applying `max-height` feels like a last hope solution

lilyvc (approved):

Fercas123:
> > The issue happens when styles are applied to `.saltButton` externally (from multilineinput in this case) and override the class passed within the pill.
> 
> If this comes from `Button`, should there be a height CSS API exposed, and overridden in `PillNext` and adornment accordingly? Applying `max-height` feels like a last hope solution

both of them use `--saltButton-height` already

origami-z:
> both of them use `--saltButton-height` already, the thing is, even though pill uses it within a class, multiline input uses it inside `.saltMultilineInput .saltButton`, which wins the race, i could do `.saltPillNext .saltPillNext-close-button.saltButton`

Is Multiline input targeting too widely? Should it only targeting buttons within adornment?

joshwooding:
> > both of them use `--saltButton-height` already, the thing is, even though pill uses it within a class, multiline input uses it inside `.saltMultilineInput .saltButton`, which wins the race, i could do `.saltPillNext .saltPillNext-close-button.saltButton`
> 
> Is Multiline input targeting too widely? Should it only targeting buttons within adornment?

I raised this to design last week, the overrides applied with button input adornments are broken and too restrictive. They are currently very broken and should be removed.

Fercas123:
> > > both of them use `--saltButton-height` already, the thing is, even though pill uses it within a class, multiline input uses it inside `.saltMultilineInput .saltButton`, which wins the race, i could do `.saltPillNext .saltPillNext-close-button.saltButton`
> > 
> > 
> > Is Multiline input targeting too widely? Should it only targeting buttons within adornment?
> 
> I raised this to design last week, the overrides applied with button input adornments are broken and too restrictive. They are currently very broken and should be removed.
> 
> I don't think this change should go ahead and I think that root problem should be solved

makes sense, should we have an issue to reflect that ongoing conversation?

Fercas123:
Closing this PR as the issue has already been raised in relation to input styles.

PR #2738, opened 2023-11-16
Fix CSS bundle scripts on Windows

joshwooding:


packages/core/scripts/build.mjs
```diff
@@ -61,8 +61,8 @@ function runBuild() {
         path.join(buildFolder, outfileName),
         path.join(cssFolder, outfileName)
       );
-
-      //delete generated entry file
-      deleteSync([entryFile], { force: true });
-    });
+    }).finally(() => {
+    //delete generated entry file
+    deleteSync([entryFile], { force: true });
```
joshwooding, line 66: Moved deleting from finally, so if the script errors it still cleans up. Before if it failed and you didn't delete it would leave an invalid file that would then cause the next run to fail.

origami-z (approved):

DavieReid (approved):
Thanking you

PR #2741, opened 2023-11-17
Remove minWidth of PillNext

Fercas123:
Removing the minWidth: 4em style so the pill adapts to it's content


origami-z:
Changeset to release. BTW, `em` is not character width, but relative unit to font size

feliciackh (approved):

Fercas123:
intended qa change for "extra extra long pill label example" as the min-width was setting the flex basis, truncation should be intentional like in `story/lab-pill-next--truncated`

ImmanuelBaskaran (approved):

PR #2744, opened 2023-11-17
[Theme 1.0] Text component updates

feliciackh:
#2571
part of Theme 1.0 updates : add variants Notation and Action to Text component


origami-z (commented):
The code looks generally fine. I'm not sure exporting a component called `Action` is doing us any good, v.s. `ActionText` or something like that

feliciackh:
> The code looks generally fine. I'm not sure exporting a component called `Action` is doing us any good, v.s. `ActionText` or something like that

fair point - i'm thinking `TextAction` and `TextNotation` to emphasise that they're `Text` components first before Action/Notation. Label will benefit from the same update i reckon `TextLabel` to keep names aligned, what do you think? @origami-z

feliciackh:
> > The code looks generally fine. I'm not sure exporting a component called `Action` is doing us any good, v.s. `ActionText` or something like that
> 
> fair point - i'm thinking `TextAction` and `TextNotation` to emphasise that they're `Text` components first before Action/Notation. Label will benefit from the same update i reckon `TextLabel` to keep names aligned, what do you think? @origami-z

made changes only to Action (new `TextAction`) and Notation (new `TextNotation`) as it does make sense to use `Label` on its own

.changeset/little-pillows-fry.md
```diff
@@ -0,0 +1,15 @@
+---
+"@salt-ds/core": minor
+---
+
+`Text` component has two new variants: `TextAction` and `TextNotation`. See [Storybook](https://storybook.saltdesignsystem.com/?path=/story/core-text-text-qa--all-variants-grid) for more information.
```
libertymayc, line 5: would it be better to link to the site? also wondering if this links like this change in the future how that would impact changelogs

packages/core/src/text/Display.tsx
```diff
@@ -1,21 +1,12 @@
...
-import { clsx } from "clsx";
 import { forwardRef } from "react";
 import { Text, TextProps } from "./Text";
 
-const withBaseName = makePrefixer("saltText");
-
 export const Display1 = forwardRef<
   HTMLSpanElement,
   Omit<TextProps<"span">, "as">
 >(function Display1({ children, className, ...rest }, ref) {
```
libertymayc, line 7:
```suggestion
>(function Display1({ children, className, styleAs = "display1", ...rest }, ref) {
```

packages/core/src/text/Display.tsx
```diff
@@ -1,21 +1,12 @@
...
   return (
-    <Text
-      as="span"
-      className={clsx(className, withBaseName(`display1`))}
-      ref={ref}
-      {...rest}
-    >
+    <Text as="span" styleAs="display1" ref={ref} {...rest}>
```
libertymayc, line 9:
```suggestion
    <Text as="span" styleAs={styleAs} ref={ref} {...rest}>
```

along with above change, i think this is slightly better as could be misleading otherwise

.changeset/little-pillows-fry.md
feliciackh: good shout - will update link + add them to site's Text examples

alycrys (approved):

site/src/examples/text/Styles.tsx
```diff
@@ -23,5 +25,7 @@ export const Styles = (): ReactElement => (
     <H4>H4</H4>
     <Text>Text</Text>
     <Label>Label</Label>
+    <TextNotation>Notation</TextNotation>
```
origami-z, line 28: Example description (`site/docs/components/text/examples.mdx`) needs update as well?

.changeset/little-pillows-fry.md
```diff
@@ -0,0 +1,15 @@
...
+
+```tsx
+<Text styleAs="notation" as="span">Lorem ipsum</Text>
+<Text styleAs="action" as="span">Lorem ipsum</Text>
+```
+
+```tsx
+<TextNotation>Lorem ipsum</TextNotation>
```
joshwooding, line 13: Do we want TextNotation, TextAction

feliciackh: @joshwooding do you mean we might not want these components altogether  `TextNotation / TextAction`, and have them available only through `styleAs` prop, and shown in examples on site?

joshwooding: Exactly yeah, Notation is one I'm unsure about but I don't really see people using TextAction often?

feliciackh: Hmm I understand where you're coming from - I'm just conscious if we derail from the "format" of offering all available `styleAs` variants in individual components form, could it be seen as a mistake/error on our end that maybe our documentations just aren't up to date etc. If we do remove the components TextNotation/TextAction we'll have document that well so users won't assume at first glance components they can import are the only ones available for use, and to use action or notation the only way is through `styleAs`

joshwooding: I don't know if we've ever had the stance that we'll always have equivalence. It's a fair enough assumption, though but it creates redundancy. We can keep TextNotation/TextAction and always reassess this decision later.

alycrys (approved):

joshwooding (approved):

PR #2767, opened 2023-11-21
Fix Pill Next's className prop not being applied to the root element

Fercas123:
classname should be applied to wrapper component


.changeset/clean-jeans-grab.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+`className` is now forwarded to `PillNext`'s wrapper instead of the nested button.
```
joshwooding, line 5:
```suggestion
Pill Next's `className` prop is now forwarded to its wrapper instead of the nested button.
```

joshwooding (approved):

ImmanuelBaskaran (approved):

PR #2778, opened 2023-11-23
Fix nested Salt Providers not inheriting a parent provider's `enableStyleInjection` value

DavieReid:
Root `StyleInjectionProvider` overrides nested `StyleInjectionProvider`

### DIFF
--- a/packages/core/src/salt-provider/SaltProvider.tsx
+++ b/packages/core/src/salt-provider/SaltProvider.tsx
@@ -209,7 +209,7 @@ function InternalSaltProvider({
 }
 
 export function SaltProvider({
-  enableStyleInjection = true,
+  enableStyleInjection,
   ...restProps
 }: SaltProviderProps) {
   return (
--- a/packages/core/stories/style-injection/style-injection.disabled.stories.tsx
+++ b/packages/core/stories/style-injection/style-injection.disabled.stories.tsx
@@ -1,0 +1,23 @@
+import { Button } from "@salt-ds/core";
+import { StoryFn } from "@storybook/react";
+import { SaltProvider } from "@salt-ds/core";
+
+export default {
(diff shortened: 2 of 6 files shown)


packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -209,7 +209,7 @@ function InternalSaltProvider({
 }
 
 export function SaltProvider({
-  enableStyleInjection = true,
+  enableStyleInjection,
```
DavieReid, line 212: default comes from context now

packages/core/stories/style-injection/style-injection.disabled.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+export default {
+  title: "Core/Style Injection",
+  component: SaltProvider,
+};
+
+export const RootDisabled: StoryFn = () => (
+  <SaltProvider enableStyleInjection={false}>
+    <Button variant="primary">Root - style injection disabled</Button>
+    <SaltProvider>
+      <Button variant="primary">Nested - style injection enabled</Button>
```
ZarkoPernar, line 14: disabled?
```suggestion
      <Button variant="primary">Nested - style injection disabled</Button>
```

DavieReid: No, it's nested inside a provider that has style injection enabled (line 13).

This test should show 2 buttons with no styles as the root style provider trumps the nested style provider

DavieReid: Updated the text to make clarify the expectations

joshwooding (changes requested):
Can you add a changeset please :) Other than that looks good

joshwooding (approved):

PR #2779, opened 2023-11-23
Updated KeyboardControl site component to allow alternate keycombos to be separated by "or"

james-nash:
Currently, alternate key combinations provided to the `<KeyboardControl>` component (which is used on the site's component pages in the accessibility tab) need to be separated by a slash (`/`). Based on a discussion with @mikearildbrown, we decided it would be more readable if we used the word "or" instead.

This PR allows you to use `/` or `or`, so that existing docs which all use `/` continue to work. However, they can now gradually be updated to use `or` instead.

### DIFF
--- a/site/src/components/keyboard-controls/KeyboardControl.tsx
+++ b/site/src/components/keyboard-controls/KeyboardControl.tsx
@@ -9,8 +9,8 @@ export interface KeyboardControlProps {
    * Use "+" to indcate combos - i.e. when more than one key
    * need to pressed at the same time (e.g. "CTRL + A").
    *
-   * Use "/" to separate alternative keys or combos.
-   * E.g. "SPACE / ENTER"
+   * Use "/" or "or" to separate alternative keys or combos.
+   * E.g. "SPACE / ENTER" or "SPACE or ENTER"
    *
    * You can combine alternate combos too.
    * E.g.: "CTRL + A / CMD + A"
@@ -20,7 +20,7 @@ export interface KeyboardControlProps {
   className?: string;
 }
 
-const keyOrComboSeparator = /\s*([\+\/])\s*/;
+const keyOrComboSeparator = /\s*([\+\/]|or)\s*/i;
 
 function splitCombo(keyOrCombo: string): string[] {
   return keyOrCombo.split(keyOrComboSeparator);


site/src/components/keyboard-controls/KeyboardControl.tsx
```diff
@@ -9,8 +9,8 @@ export interface KeyboardControlProps {
    * Use "+" to indcate combos - i.e. when more than one key
    * need to pressed at the same time (e.g. "CTRL + A").
    *
-   * Use "/" to separate alternative keys or combos.
-   * E.g. "SPACE / ENTER"
+   * Use "/" or "or" to separate alternative keys or combos.
```
lilyvc, line 12: I just raised a related PR, didn't see this one (although mine is more focused on styling).

While I was doing it, I noticed this example using a comma to separate: https://github.com/jpmorganchase/salt-ds/blob/main/site/docs/components/toast/accessibility.mdx. Should this be updated to use a slash or 'or', or should we add a comma as a separator? (A lot of ors in one sentence there)

james-nash: Right now (unless this PR is merged), only `/` and `+` are considered separators. They're used to break up the string provided to the `keyOrCombos` prop, so that the actual names of keys can be wrapped in `<kbd>` tags in the output.

For example, `SPACE / ENTER` would end up being rendered as `<kbd>SPACE</kbd> / <kbd>ENTER</kbd>.

So, if you're doing `keyOrCombos="Tab, Tab + Shift"` you'll end up with `<kbd>Tab, Tab</kbd> + <kbd>Shift</kbd>`, which is not what you want.

Fro your PR I would therefore suggest just swapping the comma for a slash for now. If/when this PR is merged, the slash can eventually be replaced by the word "or".

james-nash: Ah sorry, just took a closer look at your PR and I see that you added comma to the list of separators.

Based on a chat @mikearildbrown and I had the other day, we decided that using the work "or" would be clearer though, so maybe you can swap that change for the one I was doing in this PR (i.e. allow "or" to be a separator instead of a comma) and then fixup any MDX files that might be using commas. I can then close this PR since it'll be redundant.

james-nash:
> Hey @james-nash what happens if someone uses / in a key combo? it's read as a separator right? would it be better to cut / as an option entirely?
> 
> same issue maybe doesn't exist with + as you need to press shift to make a + anyway, so = would refer to the key.

Yes, the current setup wouldn't allow for a key combo that involved pressing the <kbd>/</kbd> key. However, to my knowlegde, none exist right now. As we had discussed, I thought it would be best to continue allowing `/` _for now_, so that we don't have to update all the MDX docs at the same time. Once they're all using "or", we could always come back and remove support for the `/` separator.

site/src/components/keyboard-controls/KeyboardControl.tsx
lilyvc: Okay cool - makes sense. The lack of separation became more noticeable with the styling. Happy to just get this PR in first and update mine.

lilyvc (approved):

alycrys (approved):

PR #2810, opened 2023-11-30
Add component token usage per characteristic to site docs

libertymayc:
Adds to [the token page](https://www.saltdesignsystem.com/salt/foundations/characteristics/accordionview) a table of all usage of a certain characteristic - e.g., '**actionable**' characteristic is used by  _"Accordion", "Button", "ToggleButton", "Dropdown"._.. etc

It also generates which components use the foundations directly, (e.g. z-index is used by "Badge", "Checkbox", "RadioButton", "Switch"... etc) which we may want to utilise in the future

As well as educating consumers how components relate to the theme, this PR will help designers to easily identify which components will be affected by the changes they make in the theme at a characteristic level
It will be helpful to identify the same for the palette layer, which is the true layer the theme would be customised at by designers, cc @dplsek @Jiempreecha


joshwooding:
Thanks for the pr @libertymayc, I think this needs to feed into how we think theming should work and the best way to show users. Therefore more discussion is needed before we merge this.

libertymayc:
> Thanks for the pr @libertymayc, I think this needs to feed into how we think theming should work and the best way to show users. Therefore more discussion is needed before we merge this.

agreed :) the meeting today should start the discussion on this but will ofc need a few conversations

joshwooding:
We're currently looking into how we document the theme, characteristics, foundations and other design language-related work. We'll refer to this as a potential direction, but it's unlikely we'll take this PR directly. I'll close this for now. Thanks @libertymayc

PR #2816, opened 2023-12-01
POC Theme name as array

origami-z:
One of partner teams wanted to conditionally passing in more than 1 theme names, based on conditions

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "type": "root"
   },
   "scripts": {
-    "start": "yarn storybook",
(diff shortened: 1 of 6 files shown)


DavieReid:
Can we not jut use `clsx` for this and relax the type of the `theme` prop on the `SaltProvider`?

origami-z:
> Can we not jut use `clsx` for this and relax the type of the `theme` prop on the `SaltProvider`?

Potentially yes, some code path was not happy about seeing space in theme. I didn't spend much time thinking implementation, but only trying out to see how much work might be needed.

PR #2832, opened 2023-12-08
FileDropZone

Fercas123:


packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,75 @@
...
+  return (
+    <>
+      <Button
+        data-testid="file-input-button"
+        onClick={handleClick}
+        disabled={disabled}
+        ref={useForkRef(ref, buttonRef)}
+        {...rest}
+      >
+        Browse files
```
joshwooding, line 60: We should allow this to be customisable. Can this be children instead?

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+  const handleClick = (event: SyntheticEvent<HTMLButtonElement>) => {
+    event.stopPropagation();
+    fileInputRef.current && fileInputRef.current.click();
+  };
+
+  return (
+    <>
+      <Button
+        data-testid="file-input-button"
```
joshwooding, line 54: Do we need this data-testid ?

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,75 @@
...
+        onClick={handleClick}
+        disabled={disabled}
+        ref={useForkRef(ref, buttonRef)}
```
joshwooding, line 57: Can we pull this into a variable please

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,75 @@
...
+  FileDropZoneTriggerProps
+>(function FileDropZoneTrigger({ disabled, accept, onChange, ...rest }, ref) {
+  const buttonRef = useRef<HTMLButtonElement>(null);
+  const fileInputRef = useRef<HTMLInputElement>(null);
+
+  // As an ADA requirement when dialog is closed and the focus is returned to the input, we need to
+  // move focus back on the button element so that all labels can be announced correctly
+  const handleFocus = (event: FocusEvent<HTMLInputElement>) => {
+    event.stopPropagation();
+    buttonRef.current && buttonRef.current.focus();
```
joshwooding, line 43: I'm not a huge fan of this pattern, 

```suggestion
    buttonRef.current?.focus();
```

packages/lab/src/file-drop-zone/FileDropZoneIcon.tsx
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export const FileDropZoneIcon = forwardRef<
+  SVGSVGElement,
+  FileDropZoneIconProps
+>(function FileDropZoneIcon({ status, size = 2, ...rest }, ref) {
+  return (
+    <div>
+      {status ? (
+        <StatusIndicator ref={ref} status={status} size={size} {...rest} />
```
joshwooding, line 19: Couple questions here, does this mean the UploadIcon size is fine by default?
props are only applied when there's a status and the ref is only valid when there's a status

packages/lab/src/file-drop-zone/FileDropZone.tsx
```diff
@@ -1,115 +1,36 @@
...
+      status,
       className,
       children,
-      description,
       disabled,
-      id: idProp,
-      validate,
-      onFilesAccepted,
-      onFilesRejected,
+      onDragOver: onDragOverProp,
```
joshwooding, line 31: Do we need to rename here? This doesn't follow our current pattern since it's not already present.

packages/lab/src/file-drop-zone/FileDropZone.tsx
```diff
@@ -120,193 +41,56 @@ export const FileDropZone = forwardRef<HTMLDivElement, FileDropZoneProps>(
...
-            [withBaseName("error")]: isRejected,
+            [withBaseName("error")]: status === "error",
+            [withBaseName("success")]: status === "success",
             [withBaseName("active")]: isActive,
             [withBaseName("disabled")]: disabled,
           },
           className
         )}
-        onDragLeave={handleDragLeave}
+        onDragLeave={!disabled ? handleDragLeave : undefined}
```
joshwooding, line 87: Can we move this logic to the handleDragLeave function please. It being here makes it harder to follow the logic

packages/lab/src/file-drop-zone/FileDropZone.tsx
```diff
@@ -120,193 +41,56 @@ export const FileDropZone = forwardRef<HTMLDivElement, FileDropZoneProps>(
...
         onDragOver={handleDragOver}
         onDrop={handleDrop}
```
joshwooding, line 89: Should these event handlers also be affected by the disabled state

packages/lab/src/file-drop-zone/FileDropZone.tsx
```diff
@@ -120,193 +41,56 @@ export const FileDropZone = forwardRef<HTMLDivElement, FileDropZoneProps>(
...
 
       if (disabled) {
         event.dataTransfer && (event.dataTransfer.dropEffect = "none");
         return;
-      } else {
-        // Not using 'move', otherwise apps like Outlook will delete the item
-        event.dataTransfer && (event.dataTransfer.dropEffect = "copy");
       }
-
+      event.dataTransfer && (event.dataTransfer.dropEffect = "copy");
```
joshwooding, line 57: Not a huge fan of this shortcut, can we change this to an if statement please

packages/lab/src/file-drop-zone/FileDropZone.css
```diff
@@ -1,106 +1,47 @@
...
+  pointer-events: none;
 }
-
-/* Styles applied if `isRejected={true}`  */
+/*!* Styles applied if `status="error" *!*/
 .saltFileDropZone-error {
-  --fileDropZone-background: var(--salt-status-error-background);
-  --fileDropZone-borderColor: var(--salt-status-error-borderColor);
-  --fileDropZone-icon-color: var(--salt-status-error-foreground);
+  border: var(--saltFileDropZone-error-border, var(--salt-size-border-strong) var(--salt-target-borderStyle) var(--salt-status-error-borderColor));
```
joshwooding, line 29: I don't think we should be creating "variants" of variables. There should only be one, --saltFileDropZone-border

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,172 @@
...
+  FileDropZoneTrigger,
+  FileDropZoneTriggerProps,
   FilesValidator,
 } from "@salt-ds/lab";
 import { AllRenderer } from "docs/components";
+import {
+  containsFiles,
+  extractFiles,
+  validateFiles,
+} from "../../src/file-drop-zone/internal/utils";
```
joshwooding, line 33: Examples shouldn't contain imports from /src/

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,172 @@
...
-  onFilesAccepted,
-  onFilesRejected,
-  ...args
-}) => {
-  const width = {
-    high: 150,
-    medium: 180,
-    low: 210,
-    touch: 250,
+  useEffect(() => {
```
joshwooding, line 64: I think it would be better to have a button that resets status.

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,172 @@
...
-    <>
-      <FileDropZone
-        data-testid="file-drop-zone-example"
-        onFilesAccepted={handleFilesAccepted}
-        {...args}
-      >
-        Some custom text
+      <FileDropZone {...args}>
+        <FileDropZoneIcon />
+        <strong>Drop files here or</strong>
```
lukeac123, line 46: Minor thing but could this be text with the styleas prop instead of <strong> ?

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,172 @@
...
-  return (
-    <>
-      <FileDropZone
-        data-testid="file-drop-zone-example"
-        onFilesAccepted={handleFilesAccepted}
-        {...args}
-      >
-        Some custom text
+      <FileDropZone {...args}>
+        <FileDropZoneIcon />
```
lukeac123, line 45: What's the logic behind passing this down to the FileDropZone and Trigger as children instead of heaving this within the File Dropzone Component and passing down props ?

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,78 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Callback for input change event
+   */
+  onChange?: (event: SyntheticEvent<HTMLInputElement>) => void;
+  /**
+   * Status indicator to be displayed.
+   */
+  status?: ValidationStatus;
```
lukeac123, line 29: can't see this being used in the component, can it be removed ?

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
joshwooding: strong text can't be implemented using styleas

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
joshwooding: Salt's ethos is composable APIs

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,78 @@
...
+        {...rest}
+      >
+        {children ?? "Browse files"}
+      </Button>
+      <input
+        accept={accept}
+        className="input-hidden"
+        data-testid="file-input"
+        disabled={disabled}
+        multiple
```
joshwooding, line 70: Do we want multiple to be the default? It feels like this sound be configurable

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+const composedStories = composeStories(fileDropZoneStories);
+const { Default } = composedStories;
+
+describe("Given a file drop zone", () => {
+  checkAccessibility(composedStories);
+
+  it("should render an input that accepts multiple files", () => {
+    cy.mount(<Default />);
+    cy.findByTestId("file-input").should("have.attr", "type", "file");
```
joshwooding, line 13: Rather than using a test id, it would be better if we're able to use a role

packages/lab/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -0,0 +1,78 @@
...
+   */
+  accept?: string;
+  /**
+   * Disable all trigger elements.
+   */
+  disabled?: boolean;
+  /**
+   * Callback for input change event
+   */
+  onChange?: (event: SyntheticEvent<HTMLInputElement>) => void;
```
joshwooding, line 25: Do you think it would be better for DX if we pass files as a second parameter?

packages/lab/src/file-drop-zone/FileDropZone.css
joshwooding: Looks like this is still unresolved?

packages/lab/src/file-drop-zone/FileDropZone.css
```diff
@@ -1,106 +1,46 @@
...
-
-/* Styles applied to description if `isRejected={true}`  */
-.saltFileDropZone-error .saltFileDropZone-description {
-  margin-bottom: calc(var(--salt-size-unit) * 2);
+.saltFileDropZone-active * {
+  pointer-events: none;
 }
-
-/* Styles applied if `isRejected={true}`  */
+/*!* Styles applied if `status="error" *!*/
```
joshwooding, line 27: Is there any specific reason we're using *!* here?

packages/lab/src/file-drop-zone/FileDropZone.css
```diff
@@ -1,106 +1,46 @@
...
+  font-size: var(--saltFileDropZone-fontSize, var(--salt-text-fontSize));
+  font-family: var(--saltFileDropZone-fontFamily, var(--salt-text-fontFamily));
+  line-height: var(--saltFileDropZone-lineHeight, var(--salt-text-lineHeight));
 }
 
 /* Styles applied if `isActive={true}` */
-.saltFileDropZone-active:not(.saltFileDropZone-disabled) {
-  --fileDropZone-background: var(--salt-target-background-hover);
+.saltFileDropZone-active,
+.saltFileDropZone-active {
```
joshwooding, line 20: ?

packages/lab/src/file-drop-zone/FileDropZone.css
```diff
@@ -1,106 +1,46 @@
...
+  background: var(--salt-target-background-hover);
+  border: var(--saltFileDropZone-border-hover, var(--salt-size-border-strong) var(--salt-target-borderStyle-hover) var(--salt-target-borderColor-hover));
 }
-
-/* Styles applied to description if `isRejected={true}`  */
-.saltFileDropZone-error .saltFileDropZone-description {
-  margin-bottom: calc(var(--salt-size-unit) * 2);
+.saltFileDropZone-active * {
+  pointer-events: none;
```
joshwooding, line 25: pointer-events none is really destructive, are we 100% this makes sense?

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
```diff
@@ -0,0 +1,61 @@
...
+      action: "drag-drop",
+    });
+    cy.get("@dropFilesSpy").should("have.been.called");
+  });
+  it("should be disabled if disabled prop is passed", () => {
+    cy.mount(<Default disabled />);
+    cy.findByTestId("file-input").should("be.disabled");
+    cy.findByTestId("file-input-trigger").should("be.disabled");
+  });
+  it("should focus on the button on initial focus", () => {
```
joshwooding, line 53: Is this the right behaviour? I would probably not expect the input or receive focus

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
```diff
@@ -0,0 +1,61 @@
...
+        action: "drag-drop",
+      }
+    );
+    cy.findByTestId("file-drop-zone-example").should(
+      "have.class",
+      "saltFileDropZone-success"
+    );
+  });
+  it("should trigger onDrop when files are dropped", () => {
+    const dropFilesSpy = cy.stub().as("dropFilesSpy");
```
joshwooding, line 34:
```suggestion
    const dropSpy = cy.stub().as("dropSpy");
```
I prefer these to match the callback name

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
Fercas123: the input is hidden, so if you tab in the focus will go into the button (trigger) what would you expect instead?

packages/lab/src/file-drop-zone/FileDropZone.css
Fercas123: changed for a ref check on the handler :)

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
joshwooding: Oh sorry, that's what I expect. I got confused with the rest button included :P

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
```diff
@@ -0,0 +1,80 @@
+import { composeStories } from "@storybook/react";
+import * as fileDropZoneStories from "@stories/file-drop-zone/file-drop-zone.stories";
+import { checkAccessibility } from "../../../../../../cypress/tests/checkAccessibility";
```
lilyvc, line 3: not really related but I am generally wondering why checkAccessibility is in a completely different location when it's only imported in files in this location

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -348,61 +181,57 @@ interface ResultCardProps {
   result: ResultCardType | undefined;
 }
 
-export const ResultCard = ({ result }: ResultCardProps) => {
+export const Results = ({ result }: ResultCardProps) => {
```
joshwooding, line 184: Is Results, intended to be a story?

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,169 @@
...
 export default {
   title: "Lab/File Drop Zone",
   component: FileDropZone,
 } as Meta<typeof FileDropZone>;
 
-export const All: StoryFn<typeof FileDropZone> = ({
-  onFilesAccepted,
-  ...args
-}) => {
+export const All: StoryFn<typeof FileDropZone> = ({ onDrop, ...args }) => {
```
joshwooding, line 35: I think we should remove the "All" story and update the QA story to match it

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
Fercas123: it was there before so i left it, but happy to remove it

lilyvc:
On the example, it uploads when I drag and drop but not when I click browse and open: https://3fd3b3bf.saltdesignsystem-storybook.pages.dev/?path=/story/lab-file-drop-zone--default

packages/lab/src/__tests__/__e2e__/file-drop-zone/FileDropZone.cy.tsx
Fercas123: probably because the cypress folder would be a good place to add general tests, also they get used in core and lab, but @joshwooding might have a much better sounding explanation :)

Fercas123:
> On the example, it uploads when I drag and drop but not when I click browse and open: https://3fd3b3bf.saltdesignsystem-storybook.pages.dev/?path=/story/lab-file-drop-zone--default

fixed, thanks for spotting it!!

lilyvc:
> > On the example, it uploads when I drag and drop but not when I click browse and open: https://3fd3b3bf.saltdesignsystem-storybook.pages.dev/?path=/story/lab-file-drop-zone--default
> 
> fixed, thanks for spotting it!!

I think it's only working the other way round now? 😅

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,151 @@
...
+    setResult({ errors });
+    setStatus("error");
+  };
 
-  const validateFileName: FilesValidator = (files) =>
-    files.map((file) => {
-      if (file.name.length > MAX_CHARS) {
-        return `File name ${file.name} is longer than ${MAX_CHARS} chars.`;
+  const handleFilesDrop = (event: SyntheticEvent, files: File[]) => {
+    if (!containsFiles(event as DragEvent)) {
```
joshwooding, line 60: This could check files instead?

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,151 @@
...
-      });
+      return handleFilesAccepted(files, event);
+    }
+  };
 
-      onFilesRejected?.(errors, event);
-    },
-    [onFilesRejected]
-  );
+  const statusTitles = {
```
joshwooding, line 73: If these are static they should be moved out of the component

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -1,336 +1,151 @@
...
+WithMultipleValidations.args = {
+  accept: "image/*",
+  children: (
+    <p>
+      Images only.
+      <br />
+      500KB total file size limit.
+      <br />
+      36 chars File name limit.
+    </p>
```
joshwooding, line 146:
```suggestion
    <Text>
      Images only.
      <br />
      500KB total file size limit.
      <br />
      36 chars File name limit.
    </Text>
```

packages/lab/stories/file-drop-zone/file-drop-zone.stories.tsx
```diff
@@ -348,61 +163,57 @@ interface ResultCardProps {
...
-          ))}
-        </ul>
-      </div>
-    ),
+    (errors: readonly string[]) =>
+      errors.map((error) => {
+        const longLabel = error.split("").length > 36;
+
+        return (
+          <Banner status="error" variant={"secondary"} key={error}>
```
joshwooding, line 195:
```suggestion
          <Banner status="error" variant="secondary" key={error}>
```

.changeset/tough-pugs-hope.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/lab": minor
+---
+
+Refactored FileDropZone.
+
+- FileDropZone has been split into three composible pieces: FileDropZone, DileDropZoneIcon and FileDropZoneTrigger.
```
feliciackh, line 7: `FileDropZoneIcon`

joshwooding:
```suggestion
- FileDropZone has been split into three composible pieces: FileDropZone, FileDropZoneIcon and FileDropZoneTrigger.
```

joshwooding (approved):

PR #2855, opened 2023-12-18
Remove redundant icon overrides

joshwooding:


alycrys (approved):

lukeac123:
How's this change come about, is it from new changes to the theme ?
Could be worth a mention in the changeset why these have now become redundant.

james-nash (approved):
Always nice to see a PR that removes more code than it adds :-)

👏

Issue #2190, opened 2023-08-11
Change --salt-palette-neutral-backdrop value


pseys:
In light mode change `--salt-palette-neutral-backdrop` value from `--color-black` to `color-white`

```[tasklist]
### Tasks
- [x] Change value in Figma docs
- [x] Change value in Light mode figma library
- [x] Change in code
```

mark-tate:
How do we communicate theme changes like this so we can correlate code and figma specs together ?

pseys:
This has been resolved as part of #2571

PR #2866, opened 2024-01-02
Update backdrop CSS tokens

lilyvc:
This PR adds a definition for `--salt-color-white-fade-backdrop` which is currently not defined.

Noticed this issue when looking at [Scrim](https://storybook.saltdesignsystem.com/?path=/story/lab-scrim--default). As you can see, it has no background colour when opened.

### DIFF
--- a/packages/theme/css/foundations/fade.css
+++ b/packages/theme/css/foundations/fade.css
@@ -50,7 +50,7 @@
   --salt-color-gray-800-fade-background-readonly: rgba(36, 37, 38, var(--salt-palette-opacity-background-readonly));
 
   --salt-color-white-fade-backdrop: rgba(255, 255, 255, var(--salt-palette-opacity-backdrop));
-  --salt-color-black-fade-backdrop: rgba(36, 37, 38, var(--salt-palette-opacity-backdrop));
+  --salt-color-black-fade-backdrop: rgba(0, 0, 0, var(--salt-palette-opacity-backdrop));
 
   --salt-color-blue-100-fade-fill: rgba(100, 177, 228, var(--salt-palette-opacity-disabled));
   --salt-color-blue-600-fade-fill: rgba(21, 92, 147, var(--salt-palette-opacity-disabled));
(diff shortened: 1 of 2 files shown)


packages/theme/css/foundations/fade.css
```diff
@@ -49,6 +49,7 @@
   --salt-color-gray-600-fade-background-readonly: rgba(47, 49, 54, var(--salt-palette-opacity-background-readonly));
   --salt-color-gray-800-fade-background-readonly: rgba(36, 37, 38, var(--salt-palette-opacity-background-readonly));
 
+  --salt-color-white-fade-backdrop: rgba(255, 255, 255, var(--salt-palette-opacity-backdrop));
   --salt-color-black-fade-backdrop: rgba(36, 37, 38, var(--salt-palette-opacity-backdrop));
```
lilyvc, line 53: Should this be black? I can't find info about "fade" in the theme characteristics Figma, but it is used here`--salt-palette-neutral-backdrop: var(--salt-color-black-fade-backdrop);` and in the Figma it looks like `--salt-palette-neutral-backdrop` should be `color-black, opacity 70%`.

cc @feliciackh

packages/theme/css/foundations/fade.css
```diff
@@ -49,6 +49,7 @@
   --salt-color-gray-600-fade-background-readonly: rgba(47, 49, 54, var(--salt-palette-opacity-background-readonly));
   --salt-color-gray-800-fade-background-readonly: rgba(36, 37, 38, var(--salt-palette-opacity-background-readonly));
 
+  --salt-color-white-fade-backdrop: rgba(255, 255, 255, var(--salt-palette-opacity-backdrop));
```
lilyvc, line 52: Not sure this should be white, doesn't seem dark enough to be the overlay

packages/theme/css/foundations/fade.css
joshwooding: You should follow Figma

feliciackh: @lilyvc good catch - i've message Paul to enquire further into specific values for said token. In the Theme updates excel sheet, line 63, as well as this [gh issue](https://github.com/jpmorganchase/salt-ds/issues/2190), noted the change from using `color-black` to `color-white` so that was intended. Looks like we never had `var(--salt-color-white-fade-backdrop)` defined to begin with, let's see what Paul says

libertymayc (approved):

feliciackh (approved):

packages/theme/css/foundations/fade.css
pseys: Just to confirm I've replied to Felicia directly. `white-fade` should be #FFFFF and `black-fade` should be #000000, currently its incorrectly set to the equivalent of grey-800.

PR #2901, opened 2024-01-10
fix tooltip disabled status evaluator

Fercas123:
Fixing issue with tooltip:
If formfield disabled is undefined, the disabled prop being explicitly passed into tooltip gets ignored.


.changeset/plenty-trees-laugh.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix logic for tooltip disabled to evaluate disabledProp is passed
```
origami-z, line 5:
```suggestion
Fixed Tooltip not honoring `disabled` prop when used in Form Field
```

origami-z (dismissed):

packages/core/src/tooltip/Tooltip.tsx
```diff
@@ -89,7 +89,8 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
       validationStatus: formFieldValidationStatus,
     } = useFormFieldProps();
 
-    const disabled = formFieldDisabled ?? disabledProp;
+    const disabled = Boolean(formFieldDisabled) || disabledProp;
```
alycrys, line 92: Can you add a story or a test to show the bug? I tried 
`
  <FormField disabled>
    <Tooltip {...props}>
      <Button>Hover</Button>
    </Tooltip>
  </FormField>
`
and it doesn't show the tooltip.

origami-z: Is this expected? disabled tooltip won't show?

alycrys: do you need to display a tooltip even if no truncation?

packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
```diff
@@ -57,6 +57,16 @@ describe("GIVEN a Tooltip", () => {
     });
   });
 
+  describe.only("WHEN disabled", () => {
```
feliciackh, line 60:
```suggestion
  describe("WHEN disabled", () => {
```

packages/core/src/tooltip/Tooltip.tsx
Fercas123: exactly

packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
Fercas123: nice spot!!!

packages/core/src/tooltip/Tooltip.tsx
joshwooding: Ultimately the logic here is wrong, with two booleans they should be "OR"d but in the case it's using the undefined check

packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
```diff
@@ -57,6 +57,16 @@ describe("GIVEN a Tooltip", () => {
...
+    it("should not render when form field context disabled is undefined", () => {
+      cy.mount(
+        <FormField>
+          <Default disabled />
```
joshwooding, line 64: I believe this is a false positive, the Tooltip hasn't been opened yet so won't be visible.

Fercas123: ah true, will use an open one

alycrys (approved):

joshwooding (approved):

PR #2902, opened 2024-01-11
Navigation Item

alycrys:


packages/lab/src/navigation-item/NavigationItem.css
```diff
@@ -1,168 +1,114 @@
...
+  --saltButton-letterSpacing: none;
+  --saltButton-fontWeight: var(--salt-text-fontWeight);
+  --saltButton-width: 100%;
 
-/* Vars applied to NavigationItem component when  blur active */
-.saltNavigationItem-blurActive,
-.saltNavigationItem-active.saltNavigationItem-nested {
-  --navigationItem-fontWeight: var(--salt-text-fontWeight);
-}
+  --saltIcon-color: var(--salt-content-primary-foreground);
```
joshwooding, line 10: Do we really need to change the icon color?

feliciackh (dismissed):

Fercas123 (approved):

packages/lab/src/navigation-item/NavigationItem.tsx
```diff
@@ -75,23 +75,17 @@ export const NavigationItem = forwardRef<HTMLDivElement, NavigationItemProps>(
...
-        ref={ref}
-        style={style}
-        className={clsx(withBaseName(), className)}
-        {...rest}
-      >
+      <div ref={ref} style={style} {...rest}>
         <ConditionalWrapper
           className={clsx(
-            withBaseName("wrapper"),
+            withBaseName(),
```
joshwooding, line 81: I find it confusing that the root classname was moved. The style prop and the rest of the props appear on the root element?

joshwooding (changes requested):
There's a bug when the Navigation Item is in active state the parent icon disappears

packages/lab/src/navigation-item/ConditionalWrapper.tsx
```diff
@@ -25,7 +25,7 @@ export const ConditionalWrapper = ({
     onExpand?.(event);
   };
 
-  return parent ? (
+  return parent || href === undefined ? (
     <Button
```
joshwooding, line 29: I think we are better off refactoring this to not use Salt's Button directly. Dependencies like this cause issues with CSS specificity and make overrides confusing to consumers.

alycrys:
> There's a bug when the Navigation Item is in active state the parent icon disappears

Which icon? I'm not seeing it

joshwooding:
> > There's a bug when the Navigation Item is in active state the parent icon disappears
> 
> Which icon? I'm not seeing it

My suggestion to stop using `<Button>`\`<Link>` fixed this. The issue was probably around Button's active styles being applied to nav item.

Fercas123 (approved):

PR #2910, opened 2024-01-15
move pill from labs to core

Fercas123:
Moving pillnext to core


Fercas123:
cleaning in another PR

Issue #2913, opened 2024-01-15
Country Symbols use a "custom" implementation to generate ids.


joshwooding:
### Package name(s)

Core (@salt-ds/core)

### Description

Country Symbols use a "custom" implementation to generate ids.

https://github.com/jpmorganchase/salt-ds/blob/bbc4ccb1afe1b773b4bfe4c20b2b8c3ec1bdf7a9/packages/countries/src/components/GB.tsx#L9

This should be updated to use the useId hook. This change will need to happen on the template and for the symbols to be regenerated.

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [ ] I am an employee of JPMorgan Chase & Co.

PR #2916, opened 2024-01-16
Update country symbol id generation to use useid

Fercas123:
closes https://github.com/jpmorganchase/salt-ds/issues/2913


packages/countries/scripts/templateCountrySymbol.mustache
```diff
@@ -7,7 +8,7 @@ export type {{componentName}}Props = CountrySymbolProps;
 const {{componentName}} = forwardRef<SVGSVGElement, {{componentName}}Props>(
   function {{componentName}}(props: {{componentName}}Props, ref) {
 
-    const [uid] = useState(() => props.id || Math.random().toString());
+    const [uid] = useState(useId(props.id));
```
joshwooding, line 11: useId is a standalone hook and doesn't need to be used like this.

```suggestion
    const uid = useId(props.id);
```

feliciackh (approved):

alycrys (approved):

lilyvc (approved):

PR #2925, opened 2024-01-18
Scrim to core

lilyvc:
Move Scrim into core.


alycrys:
needs changesets

site/docs/components/scrim/examples.mdx
```diff
@@ -11,39 +11,31 @@ data:
...
 
-</LivePreview>
-
-<LivePreview componentName="scrim" exampleName="CloseOnClick">
-
-### Close on click
+#### Best practices
 
-Pass an `onClick` handler to a scrim to close it on click. This is useful for modals or any floating elements where clicking on the scrim area (outside the main content) should dismiss the view.
+Ensure you have explicitly set a position attribute for the parent container.
```
feliciackh, line 30:
```suggestion
Ensure you have explicitly set a `position` attribute for the parent container.
```

site/docs/components/scrim/examples.mdx
```diff
@@ -11,39 +11,31 @@ data:
...
 </LivePreview>
 
-<LivePreview componentName="scrim" exampleName="Fixed">
+<LivePreview componentName="scrim" exampleName="WithChild">
 
-### Fixed
+### With Child
 
-A fixed scrim fills the entire viewport, regardless of the parent container.
+You can pass children to a scrim, such as a spinner to indicate loading state. You can incorporate other elements as children, to display them above the scrim and draw attention to them.
```
feliciackh, line 38: it might be beneficial to link to spinner component

packages/core/src/scrim/Scrim.tsx
```diff
@@ -1,6 +1,6 @@
 import { clsx } from "clsx";
 import { ComponentPropsWithoutRef, forwardRef } from "react";
-import { makePrefixer } from "@salt-ds/core";
+import { makePrefixer } from "../../src/utils/makePrefixer";
```
alycrys, line 3:
```suggestion
import { makePrefixer } from "../utils";
```

packages/core/stories/scrim/scrim.stories.tsx
alycrys, line 1: Is it worth having a QA story?

site/docs/components/scrim/examples.mdx
lilyvc:
```suggestion
You can pass children to a scrim, such as a [`Spinner`](../spinner) to indicate loading state. You can incorporate other elements as children, to display them above the scrim and draw attention to them.
```

site/src/examples/scrim/FillContainer.tsx
```diff
@@ -0,0 +1,120 @@
...
+        </Button>
+      )}
+      {open && (
+        <Button
+          style={{ width: "fit-content", alignSelf: "center" }}
+          onClick={handleClose}
+          variant="cta"
+        >
+          Hide scrim
+        </Button>
```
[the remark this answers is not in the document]
lilyvc: thanks, updated

packages/core/stories/scrim/scrim.stories.tsx
lilyvc: added

packages/core/stories/scrim/scrim.qa.stories.tsx
alycrys, line 1: I think we need something different for Scrim, as it doesn't change based on density.

site/docs/components/scrim/examples.mdx
```diff
@@ -11,39 +11,31 @@ data:
 
 <LivePreviewControls>
 
-<LivePreview componentName="scrim" exampleName="Default">
+<LivePreview componentName="scrim" exampleName="FillViewport">
 
-### Default
+### Fill Viewport
```
lukeac123, line 16: What's the use case for Scrim to fill the viewport, would we not recommend to use the Dialog component instead ?

packages/core/stories/scrim/scrim.qa.stories.tsx
lilyvc: Yeah good point

site/docs/components/scrim/examples.mdx
joshwooding: Scrim is different to Dialog. Dialog is a semantic element that contains a focus trap and other behaviour. Scrim's only purpose is to obscure what's behind it. Dialog comes with a Scrim built in

lukeac123: Do we have recommendations  on when to use one vs the other and should this be mentioned in the Usage.mdx ?

joshwooding: I don't think it's high priority to add that guidance. We can add it when we work on Dialog this sprint if it's needed.

packages/core/stories/scrim/scrim.qa.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Scrim open />
+    </Card>
+  </AllModes>
+);
+
+InContainer.parameters = {
+  chromatic: { disableSnapshot: false },
+};
+
+export const WithChildren: StoryFn<typeof Scrim> = () => (
```
alycrys, line 33: I think this one is enough. It has all elements that are visually important. Maybe replace the spinner with a static element, so we don't get unnecessary diffs.

site/src/examples/scrim/WithChild.tsx
```diff
@@ -0,0 +1,110 @@
...
+} from "@salt-ds/core";
+import { ArrowDownIcon, ArrowUpIcon } from "@salt-ds/icons";
+
+export const WithChild = (): ReactElement => {
+  const [open, setOpen] = useState(false);
+  const [buttonText, setButtonText] = useState("Show scrim");
+
+  const handleClick = () => {
+    setOpen(!open);
+    setButtonText(!open ? "Hide scrim" : "Show scrim");
```
joshwooding, line 20: Nit: we shouldn't use state for stuff like this. https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state

It can be derived from state, so where buttonText is used should just be {!open ? "Hide scrim" : "Show scrim"}

site/src/examples/scrim/WithChild.tsx
```diff
@@ -0,0 +1,110 @@
...
+  Text,
+} from "@salt-ds/core";
+import { ArrowDownIcon, ArrowUpIcon } from "@salt-ds/icons";
+
+export const WithChild = (): ReactElement => {
+  const [open, setOpen] = useState(false);
+  const [buttonText, setButtonText] = useState("Show scrim");
+
+  const handleClick = () => {
+    setOpen(!open);
```
joshwooding, line 19: This should use a function https://react.dev/reference/react/useState#setstate

```suggestion
    setOpen(old => !old);
```

joshwooding (approved):

feliciackh (approved):

alycrys (approved):

PR #2949, opened 2024-01-24
Improve the error detection logic in Salt Provider

### LINKED ISSUE #2841: SaltProvider does not allow you to apply classes to HTML element when used in a new window
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

1.9.0

### Description

There are cases where a developer wants to apply the theme classes to the html element of a newly created window. However this is impossible at the moment because of the checks the salt provider peforms to see if it is the top level provider.

### Steps to reproduce

I used a window provider to pass the new window to the salt provider via context and passed in "root" to the applyClassesTo prop in the salt provider. However I get the following error message "SaltProvider can only apply CSS classes to the root if it is the root level SaltProvider."

### Expected behavior

I expected the classes to be applied to the html element of the newly created window.

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

joshwooding:
Closes #2841 

Updates the error detection logic so you can have a root provider in a new window.

### DIFF
--- a/packages/core/src/salt-provider/SaltProvider.tsx
+++ b/packages/core/src/salt-provider/SaltProvider.tsx
@@ -180,7 +180,7 @@ function InternalSaltProvider({
         targetWindow.document.documentElement.dataset.mode = mode;
       } else {
         console.warn(
-          "\nSaltProvider can only apply CSS classes to the root if it is the root level SaltProvider."
+          "SaltProvider can only apply CSS classes to the root if it is the root level SaltProvider."
         );
       }
     }
--- a/packages/core/src/salt-provider/SaltProvider.tsx
+++ b/packages/core/src/salt-provider/SaltProvider.tsx
@@ -14,7 +14,7 @@ import { ViewportProvider } from "../viewport";
 import { useIsomorphicLayoutEffect } from "../utils";
 
 import saltProviderCss from "./SaltProvider.css";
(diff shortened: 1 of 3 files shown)


packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -153,20 +164,13 @@ function InternalSaltProvider({
...
-  });
-
   useIsomorphicLayoutEffect(() => {
     const themeNames =
       themeName === DEFAULT_THEME_NAME
         ? [DEFAULT_THEME_NAME]
         : [DEFAULT_THEME_NAME, themeName];
     if (applyClassesTo === "root" && targetWindow) {
-      if (isRoot) {
+      if (isRoot && inheritedWindow != targetWindow) {
```
ImmanuelBaskaran, line 173: if isRoot is false, then this whole condition will go the the else branch?

ImmanuelBaskaran:
Just confirmed the fix works.

ImmanuelBaskaran (approved):

origami-z (approved):

Fercas123 (approved):

ImmanuelBaskaran (approved):

PR #2954, opened 2024-01-25
Add Sizes and Accent to Card

### LINKED ISSUE #2944: Card reconciliation
**Description**
Currently Card and Interactable Card have distinctly different styling solutions. This creates inconsistency when used together or in proximity of one another.

This issue will merge styling solutions for the two cards into one. Both Card and Interactable Card will have border and shadow treatments, with an optional T,R,B,L accent.

Interactable Card will require additional styles for interactive states (hover, active, disabled states).

**To discuss**
The guidance when to use Interactable Card vs Card. We should review whether we use a polymorphic component (do we need to wrap it with a link? what about if we want it to behave like a button, e.g., not navigation).

Do we merge the two variants into one single component instead? If it's used as a link/ button, the interactive states would be displayed (in code).

**Use cases to consider**
MCP have card use cases in their Dashboard (it should also work in our own Salt Dashboard pattern). Speak to Paula from MCP.

The updates will enable Card to be used in Salt's Selectable Card pattern. Consider the requirements of that pattern when making these updates.

-----
**Issue update:** 

2 Feb - In a discussion today between design and dev, we’ve restructured our approach to the Card component, which will need to roll into the next sprint.
 
We’ve decided that we will provide users with 3 x cards:
**Basic Card** - this will give the user the option to apply an accent bar in any placement (top, right, bottom, left), the card will also be hoverable whereby the accent changes state. States include: Default and Hover.
**Clickable Card** - which will use our selectable tokens, similarly to our selectable cards, this card allows the user to make a selection: single-select or multi-select. States include: Default, Hover, Selected, Focused, Disabled and Selected-disabled.
**Link card** - where the whole card navigates a user somewhere. States: Default, Hover, Active, and Focused.
 
We will include the Clickable Card in the ‘selectable card pattern’ and also include an example of a Clickable Card Group.
 
Token updates will be needed: container-primary-borderColor & container-secondary-borderColor to gray40 in light mode and equivalent in dark mode.
 
This sprint we can aim to reconcile the border and the accent in both cards (Basic card and Link card). Next sprint we will aim to deliver the token updates, the basic card, clickable card and link card examples.

The selectable card pattern will add complexity, but it's what teams are asking for (flexibility). If feedback is too complex, then we can review the ‘selectable card’ component requirement at later date.

lilyvc:
closes #2944

Reconciling interactable card and card.

- Interactable Card now has border to align with Card
- sizes prop added to both Card and Interactable Card to adjust padding
- Storybook examples extended to show all variations of Card and Interactable Card


joshwooding (changes requested):
This approach won't work. Deletion of a component is a breaking change. I think we also need to wait for design to decide what direction they want.

packages/core/src/card/Card.tsx
```diff
@@ -53,6 +58,7 @@ export const Card = forwardRef<HTMLDivElement, CardProps>(function Card(
       className={clsx(
         withBaseName(),
         withBaseName(variant),
+        accentPlacement && withBaseName(`accent${capitalize(accentPlacement)}`),
```
joshwooding, line 61: We've not been using this syntax to achieve conditional classes. Could we change this to the object notation please.

packages/core/src/card/InteractableCard.css
```diff
@@ -1,50 +1,78 @@
...
 /* Styles applied to InteractableCard when variant="secondary" */
 .saltInteractableCard-secondary {
   background: var(--saltCard-background, var(--salt-container-secondary-background));
+  border-color: var(--salt-container-secondary-borderColor);
 }
 
 /* Styles applied to InteractableCard if `accentPlacement="bottom"` (default) */
-.saltInteractableCard-accentBottom {
-  --card-borderWidth: 0 0 var(--salt-size-accent) 0;
+.saltInteractableCard-accentBottom::after {
```
joshwooding, line 28: Was this change done because of the diagonal border artifact?

packages/core/src/card/InteractableCard.css
```diff
@@ -1,50 +1,78 @@
...
+  content: "";
```
joshwooding, line 29: There's a lot of duplicated styling here. You should consider structuring this differently:

- Have a ::after selector that has the common styles with the only styles in placement applying positioning or visibility.

packages/core/src/card/InteractableCard.tsx
```diff
@@ -66,7 +66,7 @@ export const InteractableCard = forwardRef<
       className={clsx(
         withBaseName(),
         withBaseName(variant),
-        withBaseName(`accent${capitalize(accentPlacement)}`),
+        accentPlacement && withBaseName(`accent${capitalize(accentPlacement)}`),
```
joshwooding, line 69: Same comment about the syntax.

packages/core/src/card/InteractableCard.tsx
```diff
@@ -31,7 +31,7 @@ export const InteractableCard = forwardRef<
   InteractableCardProps
 >(function InteractableCard(props, ref) {
   const {
-    accentPlacement = "bottom",
+    accentPlacement,
```
joshwooding, line 34: We should decide if this is a breaking change. Arguably it is. But will be up to design to decide.

packages/core/stories/card/card.stories.tsx
```diff
@@ -341,3 +342,55 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
+  return (
+    <StackLayout direction={"row"}>
+      <StackLayout>
+        <H1>Interactable Card</H1>
+        <GridLayout rows={2} columns={2}>
+          {exampleData.map((example, index) => {
+            return (
+              <InteractableCard
+                accentPlacement={
+                  placements[index] as "left" | "right" | "top" | "bottom"
```
joshwooding, line 357: You should be able to avoid this by making the placements array type as const.

packages/core/stories/card/card.stories.tsx
```diff
@@ -341,3 +342,55 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
+                }
+                key={index}
+              >
+                <H1 styleAs="h3">{example.title}</H1>
```
joshwooding, line 361: Why wouldn't we use H3 here?

packages/core/stories/card/card.stories.tsx
```diff
@@ -341,3 +342,55 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
+                <Text>{example.content}</Text>
+              </InteractableCard>
+            );
+          })}
+          <InteractableCard>
+            <H1 styleAs="h3">{exampleData[0].title}</H1>
```
joshwooding, line 367: Same here

packages/core/src/card/InteractableCard.css
lilyvc: yes so it is not diagonal. is there an alternative way?

packages/core/stories/card/card.stories.tsx
```diff
@@ -166,30 +170,33 @@ export const DefaultWithLinkAndImage: StoryFn<typeof Card> = () => (
       className="card-demo-image"
       style={{ width: "-webkit-fill-available" }}
     />
-    <div
+    <StackLayout
       style={{
         padding:
           "0px var(--salt-size-container-spacing) var(--salt-size-container-spacing) var(--salt-size-container-spacing)",
```
joshwooding: --salt-size-container-spacing was deprecated and replaced with --salt-spacing-300

packages/core/stories/card/card.stories.tsx
```diff
@@ -334,10 +334,60 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
 };
+
+export const CardAndInteractableCardSizes: StoryFn<typeof Card> = () => {
+  const sizes = ["small", "medium", "large"];
+  return (
+    <StackLayout style={{ width: 600 }}>
+      {sizes.map((size, index) => {
+        return (
+          <StackLayout direction="row">
+            <Card size={size as "small" | "medium" | "large"} key={index}>
```
joshwooding, line 365: Pretty sure we can get rid of this cast by fixing the type of the size array above. Making is "as const" should do it.

packages/core/stories/card/card.stories.tsx
```diff
@@ -334,10 +334,60 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
   );
 };
+
+export const CardAndInteractableCardSizes: StoryFn<typeof Card> = () => {
+  const sizes = ["small", "medium", "large"];
+  return (
+    <StackLayout style={{ width: 600 }}>
+      {sizes.map((size, index) => {
+        return (
+          <StackLayout direction="row">
```
joshwooding, line 364: The key should be here instead of on InteractableCard and we should avoid using indexes as keys and use the size instead.

packages/core/stories/card/card.stories.tsx
```diff
@@ -334,10 +334,60 @@ export const InteractableAsBlockLinkWithImageBackground: StoryFn<
...
+  );
+};
+
+export const CardAndInteractableCardVariants: StoryFn<typeof Card> = () => {
+  const variants = ["primary", "secondary"];
+  return (
+    <StackLayout style={{ width: 600 }}>
+      {variants.map((variant, index) => {
+        return (
+          <StackLayout direction="row">
```
joshwooding, line 393: same comment as above about keys

packages/core/src/card/InteractableCard.css
```diff
@@ -1,50 +1,85 @@
...
 }
 
 /* Styles applied to InteractableCard when variant="secondary" */
 .saltInteractableCard-secondary {
   background: var(--saltCard-background, var(--salt-container-secondary-background));
+  border-color: var(--salt-container-secondary-borderColor);
+}
+
+.saltInteractableCard-small {
+  padding: var(--saltCard-padding, var(--salt-spacing-100));
```
joshwooding, line 28: To be consistent it's probably better to do the same thing as card-accent-color and have a CSS variable get assigned here and keep the padding application in the main class.

packages/core/stories/card/card.stories.tsx
```diff
@@ -166,30 +177,32 @@ export const DefaultWithLinkAndImage: StoryFn<typeof Card> = () => (
...
-    <div style={{ paddingBottom: "var(--salt-size-unit)" }}>
-      <H1 styleAs="h3">{exampleData[0].title}</H1>
-      <Text>{exampleData[0].content}</Text>
-    </div>
-    <Button onClick={() => window.open("#")}>View funds</Button>
+    <StackLayout gap={1}>
+      <H3>{exampleData[0].title}</H3>
+      <StackLayout gap={2}>
+        <Text>{exampleData[0].content}</Text>
+        <Button onClick={() => window.open("#")}>View funds</Button>
```
joshwooding, line 203: We should remove the onClick from the buttons.

packages/core/stories/card/card.stories.tsx
```diff
@@ -102,18 +103,22 @@ const exampleData = [
```
joshwooding, line 103: We should get rid of All Default

Issue #2114, opened 2023-07-31
Dialog


joshwooding:
```[tasklist]
### Checklist
- [x] Density (HD, MD, LD, TD)
- [x] Mode (Light, Dark)
- [ ] State (Default, Hover, Selected/Active, Disabled, Read-only, Other)
- [ ] Controlled API
- [ ] UITK accessibility parity (screen reader + keyboard interactions)
```
```[tasklist]
### Deliverables
- [x] Characteristics spec
- [x] Interactions spec, where necessary
- [x] React component
- [ ] Supporting site documentation
- [x] Figma component + guide
```

```[tasklist]
### Features
- [x] Standardise width
- [x] Scrolling
- [x] Update useId call to not use hardcoded fallback and allow custom id
- [x] Review API
- [ ] Ensure aria-labelledby and aria-describedby are always valid even when components are missing
- [ ] Check if we need to update aria-describedby behaviour
- [ ] Check Dialog tests
- [ ] Transition needs standardising to ease in out with perceptible token (plus updating in characteristic spec)
- [x] Update to use Scrim
- [x] Update Dialog to allow description to be optional depending on the content
```

bhoppers2008:
Current labs dialog supports states for:
- Default/ Generic
- Information, Error, Warning, Success

It has feature support for:
- Scrolling area
- Title
- Accent bar
- Optional close button (as DialogAction)

Figma specs can be found [here](https://www.figma.com/file/mwOfVisFMPyiQxtkanggsb/Dialog?type=design&node-id=41%3A15441&mode=design&t=rYJhuxbhujy0OKNp-1).
Figma component is already released.

What is required (in addition to features in description)?
- [ ] Desktop support to be reviewed (with partners)
- [ ] Standardized dialog width to be spec'd (design) and implemented (possibly) as S, M, L across breakpoints
- [ ] Visual affordance for scrolling needs updating (in line with prefs dialog)
- [ ] Border missing from default dialog (spec to be updated–design), plus should be inner border.
- [ ] Contextual examples (design) and site descriptions
- [ ] Review Figma component (after build updates) and update as required.

origami-z:
Reported bug by MTK #2970

PR #2958, opened 2024-01-26
Dialog to RC

lukeac123:
```[tasklist]
### Changes
- [x] Standardise width
- [x] Only dialog content scrollable + styles
- [x] removed useId 
- [x] Review API
- [x] aria-describedby behaviour removed, user can pass via {...rest} props
- [x] Check Dialog tests
- [x] Update to use Scrim
- [x] Update to use FloatingComponent for desktop compatibility
- [x] update docs with realistic examples 

```

closes:
https://github.com/jpmorganchase/salt-ds/issues/2114


packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +104,46 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
-      </SaltProvider>
-    </FloatingPortal>
+    <Scrim>
+      <DialogContext.Provider value={contextValue}>
+        <FloatingComponent
+          id={`${dialogId}-FLoatigComponent`}
+          open={open}
+          aria-modal="true"
+          aria-labelledby={`${dialogId}-header`}
+          aria-describedby={`${dialogId}-description`}
```
lukeac123, line 119: Is aria-labelledby / describedby only needed on the parent component, of should title and content also have these aria labels ?

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
joshwooding, line 1: Can we change the filename to be PascalCase please

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
...
+        cy.mount(<Default />);
+
+        cy.findByRole("button", { name: /Open Dialog/i }).click();
+
+        cy.get(".saltDialog").should("have.class", "saltDialog-medium-xl");
+      }
+    );
+  });
+
+  describe("WHEN a size is provided", () => {
```
joshwooding, line 45: This only tests the class is applied and not the actual visuals. Is it possible to use a chromatic test here?

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
+import { composeStories } from "@storybook/react";
+import * as dialogStories from "@stories/dialog/dialog.stories";
+
+const composedStories = composeStories(dialogStories);
+
+const { Default, Accent, InfoStatus } = composedStories;
+
+describe("GIVEN a Dialog", () => {
+  describe("WHEN no props are provided", () => {
+    it("THEN it should display an overlay by default", () => {
```
joshwooding, line 10: The terminology overlay is confusing here

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  describe("WHEN a Dialog is open", () => {
+    it("THEN it should close when the close button is clicked", () => {
+      cy.mount(<Default />);
+
+      cy.findByRole("button", { name: /Open Dialog/i }).click();
+
+      cy.get(".saltDialog").should("be.visible");
+
+      cy.findByLabelText("Close dialog").click();
```
joshwooding, line 85: Is it possible to use the role selector here?

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
...
+    );
+  });
+
+  describe("WHEN a Dialog is open", () => {
+    it("THEN it should close when the close button is clicked", () => {
+      cy.mount(<Default />);
+
+      cy.findByRole("button", { name: /Open Dialog/i }).click();
+
+      cy.get(".saltDialog").should("be.visible");
```
joshwooding, line 83: Is it possible to use a role selector here?

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+      cy.realPress("Escape");
+
+      cy.get(".saltDialog").should("not.exist");
+    });
+
+    it("THEN it should close when clicking outside the dialog", () => {
+      cy.mount(<InfoStatus />);
+      cy.findByRole("button", { name: /Open Dialog/i }).click();
```
joshwooding, line 104: Do we have to use a regex selector here? It has performance implications

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
```diff
@@ -0,0 +1,154 @@
...
+        .eq(3)
+        .should("be.focused")
+        .should("have.attr", "aria-label", "Close dialog");
+      cy.realPress("Tab");
+      // back to the first button
+      cy.findAllByRole("button").eq(0).should("be.focused");
+    });
+  });
+
+  describe("WHEN a status is passed", () => {
```
joshwooding, line 130: These should all probably be visual tests.

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +107,47 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
+            aria-labelledby={`${dialogId}-header`}
+            aria-describedby={`${dialogId}-description`}
+            ref={floatingRef}
+            focusManagerProps={{
+              context: context,
+            }}
+            {...getFloatingProps()}
+            className={clsx(
+              withBaseName(),
+              withBaseName(size as string, currentbreakpoint),
```
ZarkoPernar, line 134: Why can't you just use css media queries? Unless you are making size a responsive prop which it doesn't look like you are.

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
lukeac123: yup - will look into the chromatic tests

packages/lab/src/__tests__/__e2e__/dialog/dialog.cy.tsx
lukeac123: advantages of using the role selector over className ?

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,27 +62,38 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
+    open = false,
     onOpenChange,
     status,
-    initialFocus,
+    id,
+    size = "medium",
     ...rest
   } = props;
-  const dialogId = useId() || "dialog";
+  const dialogId = useId(id) ?? "saltDialog";
```
joshwooding, line 75: Why do we need a default here?

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,27 +62,38 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
+
   const [showComponent, setShowComponent] = useState(false);
 
-  const { floating, context, getFloatingProps } = useDialog({
+  const { context, floating } = useFloatingUI({
     open,
     onOpenChange,
   });
 
+  const click = useClick(context);
```
joshwooding, line 92: We should copy how Tooltip does this and call them directly in the useInteractions array

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +107,47 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
     [dialogId, status]
   );
 
+  if (!open) {
```
joshwooding, line 113: This seems weird, I think FloatingComponent is normally our "source" of whether something is open or not. This would mean FloatingComponent is only every mounted when open. Not sure of the side effects on desktop support.

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +107,47 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
-    </FloatingPortal>
+    open && (
+      <Scrim id={`${dialogId}-scrim`}>
+        <DialogContext.Provider value={contextValue}>
+          <FloatingComponent
+            id={`${dialogId}-FLoatigComponent`}
+            open={open}
+            aria-modal="true"
+            aria-labelledby={`${dialogId}-header`}
+            aria-describedby={`${dialogId}-description`}
```
joshwooding, line 126: What happens if these aren't present?

packages/lab/src/dialog/Dialog.tsx
joshwooding, line 134: Is this cast needed?

packages/lab/src/dialog/DialogActions.tsx
```diff
@@ -27,7 +29,12 @@ export const DialogActions = forwardRef<HTMLDivElement, DialogActionsProps>(
     });
 
     return (
-      <div className={clsx(withBaseName(), className)} {...rest} ref={ref}>
+      <div
+        id={`${dialogId}-actions`}
```
joshwooding, line 33: Same question as above, what's the purpose of this id?

packages/lab/src/dialog/DialogContent.css
```diff
@@ -1,13 +1,14 @@
 .saltDialogContent {
   color: var(--salt-content-primary-foreground);
-  padding-left: var(--salt-spacing-300);
-  padding-right: var(--salt-spacing-300);
-  flex: 1 1 auto;
+  display: flex;
```
joshwooding, line 3: Do we really need to use flex here?

packages/lab/src/dialog/DialogContent.css
```diff
@@ -1,13 +1,14 @@
...
+  flex-direction: column;
```
joshwooding, line 4: I feel like this is too opinionated for something that should be similar to card.

packages/lab/src/dialog/DialogContent.css
```diff
@@ -1,13 +1,14 @@
...
+  background: var(--salt-container-primary-background);
 
-  font-family: var(--salt-text-fontFamily);
-  font-size: var(--salt-text-fontSize);
-  font-weight: var(--salt-text-fontWeight);
-  line-height: var(--salt-text-lineHeight);
+  margin-left: var(--salt-spacing-300);
+  margin-right: var(--salt-spacing-300);
+  padding-right: var(--salt-spacing-300);
+  padding-bottom: 10px;
```
joshwooding: Should this be hard coded?

packages/lab/src/dialog/DialogContent.css
```diff
@@ -1,13 +1,14 @@
...
+  border-bottom: 1px solid var(--salt-separable-tertiary-borderColor);
```
joshwooding: I'm assuming this should use the tokens from the theme?

```suggestion
  border-bottom: var(--salt-size-border) var(--salt-separable-borderStyle) var(--salt-separable-tertiary-borderColor);
```

packages/lab/src/dialog/DialogContent.tsx
```diff
@@ -29,14 +38,26 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
-        id={`${dialogId!}-description`}
-        className={clsx(withBaseName(), className)}
-        {...rest}
-        ref={ref}
-      >
-        {children}
-      </div>
+      <>
+        <div
+          style={{
```
joshwooding, line 44: This would be better using classes.

packages/lab/src/dialog/DialogContent.tsx
```diff
@@ -29,14 +38,26 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
+            border:
+              scrollTop > 0
+                ? "solid 1px var(--salt-separable-tertiary-borderColor)"
+                : "solid 1px transparent",
+          }}
+        />
+        <div
+          id={`${dialogId}-description`}
+          className={clsx(withBaseName(), className)}
+          onScroll={useScrollTop}
```
joshwooding, line 56: ?

packages/lab/src/dialog/DialogContext.tsx
```diff
@@ -3,8 +3,11 @@ import { ValidationStatus } from "@salt-ds/core";
 
 export const DialogContext = createContext<{
   dialogId?: string;
-  status?: ValidationStatus;
-}>({});
+  status?: ValidationStatus | "";
```
joshwooding, line 6: This doesn't make sense. We should use `undefined` to represent no status.

packages/lab/src/dialog/Dialog.tsx
lukeac123: switched to a conditional render

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +107,47 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
-                  {children}
-                </div>
-              </DialogContext.Provider>
-            </FloatingFocusManager>
-          </FloatingOverlay>
-        )}
-      </SaltProvider>
-    </FloatingPortal>
+    open && (
+      <Scrim id={`${dialogId}-scrim`}>
```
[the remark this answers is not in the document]
lukeac123: removed

packages/lab/src/dialog/DialogContext.tsx
lukeac123: assuming this would be the same for dialogId as well ?

packages/lab/src/dialog/Dialog.tsx
lukeac123: will look into it - thanks :)

joshwooding: had the same thought but this would work better for now with supporting custom break points.

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -90,48 +115,42 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
   );
 
   return (
-    <FloatingPortal>
-      {/* The provider is needed to support the use case where an app has nested modes. The element that is portalled needs to have the same style as the current scope */}
-      <SaltProvider>
+    <DialogContext.Provider value={contextValue}>
+      <FloatingComponent
+        id={`${dialogId}-FLoatigComponent`}
```
feliciackh, line 120:
```suggestion
        id={`${dialogId}-FloatingComponent`}
```

packages/lab/src/dialog/DialogContext.tsx
```diff
@@ -2,9 +2,12 @@ import { createContext, useContext } from "react";
 import { ValidationStatus } from "@salt-ds/core";
 
 export const DialogContext = createContext<{
-  dialogId?: string;
-  status?: ValidationStatus;
-}>({});
+  dialogId?: string | undefined;
+  status?: ValidationStatus | undefined;
```
feliciackh, line 6:
```suggestion
  dialogId?: string;
  status?: ValidationStatus;
```
they mean the same so could omit

packages/lab/stories/dialog/dialog.doc.stories.mdx
feliciackh, line 1: we can remove this file entirely as any `x.doc.stories.mdx` have been removed in recent PRs

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,27 +68,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
 
-  const { floating, context, getFloatingProps } = useDialog({
+  const { context, floating } = useFloatingUI({
     open,
     onOpenChange,
   });
 
+  const { getFloatingProps } = useInteractions([
+    useClick(context),
+    useRole(context),
```
feliciackh, line 98: do we need to add role `dialog` here?

feliciackh:
needs cs :)

packages/lab/src/dialog/DialogActions.tsx
```diff
@@ -3,6 +3,7 @@ import { clsx } from "clsx";
 import { makePrefixer } from "@salt-ds/core";
 import { useWindow } from "@salt-ds/window";
 import { useComponentCssInjection } from "@salt-ds/styles";
+import { useDialogContext } from "./DialogContext";
```
feliciackh, line 6: can rm as unused

lukeac123:
re-trigger tests

.changeset/gold-onions-press.md
```diff
@@ -0,0 +1,52 @@
...
+"@salt-ds/lab": minor
+---
+
+- Refactored `Dialog` to use floating-ui and Salt's `Scrim`.
+- Implemented desktop support.
+- Added optional `closeOnBlur` prop to prevent a click away dismissing the dialog.
+- Added a `size` prop which takes `small`, `medium` and `large`.
+
+```tsx
+const AlertDialogTemplate = () => {
```
joshwooding, line 11: This example should be updated to be more realistic right now it's of a template which seems weird.

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,27 +68,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
   const {
     children,
     className,
-    open = true,
+    open = false,
     onOpenChange,
     status,
-    initialFocus,
+    id,
+    closeOnBlur = true,
```
joshwooding, line 75: The dialog doesn't receive focus therefore doesn't get blurred, we should update this name

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -44,6 +47,16 @@ export interface DialogProps extends HTMLAttributes<HTMLDivElement> {
    * Default value is 0 (first tabbable element).
    * */
   initialFocus?: ComponentProps<typeof FloatingFocusManager>["initialFocus"];
+
+  id?: string;
```
joshwooding, line 51: This isn't needed.

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +114,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
     [dialogId, status]
   );
 
+  if (!open && !showComponent) return null;
```
joshwooding, line 117: Was the impact of this to desktop support checked?

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -89,49 +114,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
+            withBaseName(),
+            withBaseName(size, currentbreakpoint),
+            {
+              [withBaseName("enterAnimation")]: open,
+              [withBaseName("exitAnimation")]: !open,
+              [withBaseName(status as string)]: status,
+            },
+            className
+          )}
+          onAnimationEnd={() => {
```
joshwooding, line 139: Did you test this with animations turned off?

packages/lab/src/dialog/DialogContent.tsx
```diff
@@ -20,6 +20,11 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
   function DialogContent(props, ref) {
     const { children, className, ...rest } = props;
     const { dialogId } = useDialogContext();
+    const [scrollTop, setScrollTop] = useState(0);
+
+    const handleScroll = (e: React.UIEvent<HTMLElement>) => {
```
joshwooding, line 25: Can we use full names please.
```suggestion
    const handleScroll = (event: React.UIEvent<HTMLElement>) => {
```

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,27 +68,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
     onOpenChange,
     status,
-    initialFocus,
+    id,
+    closeOnBlur = true,
+    size = "medium",
     ...rest
   } = props;
-  const dialogId = useId() || "dialog";
+  const dialogId = useId(id) ?? "";
```
joshwooding, line 79: Asked before but not sure I missed a response. Do we need this nullish check?

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -37,18 +43,18 @@ export const DialogTitle = ({
...
           [withBaseName("withAccent")]: accent && !status,
-          [withBaseName(status!)]: !!status,
+          [withBaseName(status as ValidationStatus)]: !!status,
         },
         className
       )}
       {...rest}
     >
-      {status && <StatusIndicator status={status} />}
+      {status && <StatusIndicator status={status as ValidationStatus} />}
```
joshwooding, line 57: Is this cast needed?

site/src/examples/dialog/Sizes.tsx
```diff
@@ -0,0 +1,143 @@
...
+    setOpen(value);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  return (
+    <>
+      <Button data-testid="dialog-button" onClick={handleRequestOpen}>
```
joshwooding, line 33: Why do we need these test ids in the examples?

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -14,14 +14,20 @@ import dialogTitleCss from "./DialogTitle.css";
 const withBaseName = makePrefixer("saltDialogTitle");
 
 interface DialogTitleProps extends ComponentPropsWithoutRef<"h2"> {
+  /**
+   * Status indicator
```
joshwooding, line 18: Can we align this to the other status props JSDocs please.

packages/lab/src/dialog/Dialog.tsx
lukeac123: This is how it's done in overlay, was waiting for the feedback on overlay.
Looked into conditionally rendering just the Scrim and Children, so the FloatingComponent was always inserted into the DOM, but then not sure which element {...getFloatingProps} would have to be attached to.
If the Scrim is a child of the FloatingElement then using the inbuilt 'useDimiss' hook won't work

packages/lab/src/dialog/Dialog.tsx
lukeac123: change to `disableDismiss`

packages/lab/src/dialog/Dialog.tsx
lukeac123: With animations turned off the Dialog won't close. The alternative is to use the setTimeOut() function to close the dialog but from my reading this can be temperamental

packages/lab/src/dialog/Dialog.tsx
lukeac123: this pr is to get to RC enabling testing in desktop, any necessary changes will be make in next sprint before promoting to core

packages/lab/src/dialog/Dialog.tsx
lukeac123: To be resolved in next sprint when promoting Dialog to Core

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -55,83 +68,91 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
+  const { context, floating } = useFloatingUI({
     open,
     onOpenChange,
   });
 
+  const id = useId(idProp);
+
+  const { getFloatingProps } = useInteractions([
+    useClick(context),
+    useRole(context, { role: "dialog" }),
```
lukeac123, line 99: Should this be changed to render 'alertdialog' when alert dialog is passed as the role, or is it fine to have dialog and alert dialog both attached as roles ?
Could alternatively advise to pass down aria-role='alert dialog'
Any thoughts ?

joshwooding: This should be configurable

lukeac123: no we don't - because it comes as a feature of use interactions it felt easier to use their implementation.
Will change so it is configurable

.changeset/gold-onions-press.md
```diff
@@ -0,0 +1,54 @@
...
+        role="alertdialog"
+        status={"error"}
+        open={open}
+        onOpenChange={onOpenChange}
+        initialFocus={1}
+        aria-describedby={"dialog-heading"}
+      >
+        <DialogTitle>Delete Transaction</DialogTitle>
+        <DialogContent>
+          Are you sre you want to permenently delete this transaction
```
feliciackh, line 42: sure* permanently*

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -36,14 +38,22 @@ export interface DialogProps extends HTMLAttributes<HTMLDivElement> {
    */
   onOpenChange?: (open: boolean) => void;
   /**
-   * Status indicator
+   * The status of the Dialog
```
feliciackh, line 41: would be helpful to note that what the statuses are for users to  quickly grab and use

.changeset/gold-onions-press.md
```diff
@@ -0,0 +1,54 @@
...
+        Click to open dialog
+      </Button>
+      <Dialog
+        size={"small"}
+        role="alertdialog"
+        status={"error"}
+        open={open}
+        onOpenChange={onOpenChange}
+        initialFocus={1}
+        aria-describedby={"dialog-heading"}
```
feliciackh, line 38: is this still needed?

feliciackh (commented):
in examples on site, all of them seem to display a scrollbar on the right where content sits (although the content isn't overflowing the container) - is that expected?

lukeac123:
> in examples on site, all of them seem to display a scrollbar on the right where content sits (although the content isn't overflowing the container) - is that expected?

I'm not seeing this, are you able to provide a screenshot ?

packages/lab/src/dialog/Dialog.tsx
lukeac123: These are noted below where the prop is passed into the component where they are picked up by the JS Docs, so i think not sure they're needed here as well

.changeset/gold-onions-press.md
lukeac123: correct can be removed - ty

packages/lab/src/dialog/Dialog.tsx
lukeac123: The types are declared below when passed into the component, this is also where they're picked up by the JSDocs.
It's typed here as `Validation Status` i could break this down if you think it is more user friendly but I think it is fine

packages/lab/stories/dialog/dialog.qa.stories.tsx
Fercas123, line 1: NIT: can we change the title for "Lab/Dialog/Dialog QA"

Fercas123 (approved):

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -14,35 +14,40 @@ import dialogTitleCss from "./DialogTitle.css";
 const withBaseName = makePrefixer("saltDialogTitle");
 
 interface DialogTitleProps extends ComponentPropsWithoutRef<"h2"> {
-  status?: ValidationStatus;
-  accent?: boolean;
+  /**
+   * The status of the Dialog
+   * */
+  status?: ValidationStatus | undefined;
```
feliciackh, line 20: nit: no need for undefined here

packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -211,6 +206,55 @@ ErrorStatus.args = {
...
+      </Button>
+      <Dialog
+        size={"small"}
+        status={"error"}
+        role="alertdialog"
+        open={open}
+        onOpenChange={onOpenChange}
+        initialFocus={1}
+        disableDismiss
+        aria-describedby={`dialog-description`}
```
feliciackh, line 241: nit: can delete

feliciackh (approved):

packages/lab/stories/dialog/dialog.qa.stories.tsx
lukeac123: ty - will resolve in promotion to core

packages/lab/src/dialog/DialogTitle.tsx
lukeac123: ty - will resolve in promotion to core

packages/lab/stories/dialog/dialog.stories.tsx
lukeac123: ty - will resolve in promotion to core

PR #2969, opened 2024-01-31
Add resizing support to MultilineInput

joshwooding:
Adds resizing to MultilineInput. Resizing can be done two ways:

- As new lines are entered, the field's height increases.
- An end-user can drag the resize handle to resize the field vertically. After doing this, the field no longer adjusts based on the content of the field.


alycrys (approved):

Fercas123 (approved):

packages/core/stories/multiline-input/multiline-input.stories.tsx
Fercas123, line 1: nit: should we remove the mdx?

joshwooding: I'll do it later. I'm going to look through labs soon :)

Issue #2595, opened 2023-10-11
Add a secondary variant to Card component


navkaur76:
### Area

UI Components

### The problem

Work on the Analytical Dashboard pattern has surfaced the need for a secondary fill variant to the Card component/characteristics.

This would allow consumers to achieve better visual hierarchy across dashboard designs, since we're proposing the Card component as a key content module/widget type for conveying data within our pattern. 

### The solution

Split Card styles into primary and secondary. 

### Alternatives and examples

One possible solution is leveraging the container characteristics for a secondary fill, but this will need exploring further. 

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

PR #2986, opened 2024-02-06
Add `size` prop to Card and Interactable Card

lilyvc:
This PR adds size prop to Card and InteractableCard.

Addresses https://github.com/jpmorganchase/salt-ds/issues/2595


.changeset/flat-buses-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added `size` prop to `Card` and `InteractableCard`. The `size` prop changes the padding and can be `small`, `medium` or `large`.
```
lukeac123, line 5: should this not be paddingSize - size would suggest that the total height and width of the card are changing ?

feliciackh (approved):

lilyvc:
closing and raising a new PR after some further Card discussions

PR #2987, opened 2024-02-06
Update ag grid theme doc with additional css required

origami-z:
`ag-grid-community/dist/styles/ag-grid.css` is required to render correctly


origami-z:
Replaced with #3078

PR #2988, opened 2024-02-06
Multi-select CB

joshwooding:


origami-z (commented):
- Existing CB behaviour, when list is open, mouse down + up within CB but not on the button results the list to be close and open again 
- Click input, start typing, list item's focus ring does not appear, until up/down arrow
- list 'shakes' when item being deselected (e.g. backspace)
- focus input (with some items selected already), type in something 'ab', leave unselect, click away, focus back, focus is on pill rather than input text

joshwooding:
> * Existing CB behaviour, when list is open, mouse down + up within CB but not on the button results the list to be close and open again
> * Click input, start typing, list item's focus ring does not appear, until up/down arrow
> * list 'shakes' when item being deselected (e.g. backspace)
> * focus input (with some items selected already), type in something 'ab', leave unselect, click away, focus back, focus is on pill rather than input text

I can't re-create any of these locally, most of these might be mac-only issues. The last behaviour is expected

Fercas123 (approved):

Fercas123 (approved):

PR #2995, opened 2024-02-08
dialog to core

lukeac123:


packages/lab/src/dialog/Dialog.tsx
```diff
@@ -28,6 +28,20 @@ import { useComponentCssInjection } from "@salt-ds/styles";
 import dialogCss from "./Dialog.css";
 import { DialogContext } from "./DialogContext";
 
+interface ConditionalWrapperProps {
+  condition: boolean;
+  wrapper: unknown;
```
lukeac123, line 33: I'm a bit unsure about how to type the wrapper here - if anyone has any suggestions

packages/core/src/dialog/Dialog.tsx
```diff
@@ -54,8 +54,23 @@ export interface DialogProps extends HTMLAttributes<HTMLDivElement> {
...
+   * Prevent Scrim from rendering
+   * */
+  disableScrim?: boolean;
 }
 
+interface ConditionalScrimWrapperProps extends React.PropsWithChildren {
+  condition: boolean;
+}
+
+export const ConditionalScrimWrapper = ({
```
lukeac123, line 67: Extract function to utils folder so it can be re-used for Drawer

lukeac123:
@mikearildbrown Not sure if the Dialog examples / docs have had a content review ?? - would you be able to give a once over if not please ?

joshwooding (changes requested):
Can we split this PR in two please, there are changes to the component as well as a migration to core. This makes it hard to review and things can easily be missed.

PR #3008, opened 2024-02-12
Add disableScrim to Dialog

lukeac123:


joshwooding (approved):

packages/core/src/scrim/Scrim.tsx
```diff
@@ -25,6 +23,7 @@ export const Scrim = forwardRef<HTMLDivElement, ScrimProps>(function Scrim(
   { className, children, fixed = false, open = true, ...rest },
   ref
 ) {
+  const withBaseName = makePrefixer("saltScrim");
```
lukeac123, line 26: Had to move this to inside the Scrim component, otherwise returning 'cannot access makePrefixer before initialisation'

joshwooding: All of our other components work without this being inside

site/docs/components/dialog/examples.mdx
```diff
@@ -100,4 +100,11 @@ You can also render the `DialogCloseButton` as a child of the `Dialog` component
     - This should be used to force a user to action something within the dialog
     </LivePreview>
 
+      <LivePreview componentName="dialog" exampleName="DisableScrim" >
+    ### Disble Scrim
+
+    - Use the disableScrim prop to prevent the scrim from rendering
```
Fercas123, line 106: nit, we should probably match all the list items to close with `.`

```suggestion
    - Use the disableScrim prop to prevent the scrim from rendering.
```

joshwooding:
Can we target this PR to main please, putting it back on the original branch will still make reviewing harder than if it wasn't

lukeac123:
> Can we target this PR to main please, putting it back on the original branch will still make reviewing harder than if it wasn't

@joshwooding so move everything back to labs ?

Issue #3015, opened 2024-02-13
Multiline input controlled rows not resizing when input cleared. 


Fercas123:
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

1.17.0

### Description

the height of the input does not change even when a new `rows` value is passed. 

### Steps to reproduce

https://stackblitz.com/edit/salt-test-question-e8mvxj?file=package.json,index.tsx,App.tsx,override.css

### Expected behavior

on clear, the input should go back to have just one row

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

PR #3018, opened 2024-02-13
fix multiline collapse

Fercas123:
fixes https://github.com/jpmorganchase/salt-ds/issues/3015


packages/core/src/__tests__/__e2e__/multiline-input/MultilineInput.cy.tsx
```diff
@@ -122,4 +122,34 @@ describe("GIVEN an MultilineInput", () => {
...
+          .invoke("height")
+          .then((newHeight) => {
+            expect(newHeight ?? 0).to.be.greaterThan(defaultHeight ?? 0);
+          });
+      });
+    cy.findByRole("textbox")
+      .invoke("height")
+      .then((defaultHeight) => {
+        cy.findByRole("textbox").realClick();
+        cy.realPress("Escape");
```
joshwooding, line 145: I'm assuming this shouldn't be Escape?

joshwooding: Also this isn't testing the bug you found which is around setting the value prop directly. This would still call onChange and work without the code change below

packages/core/src/multiline-input/MultilineInput.tsx
```diff
@@ -150,8 +151,11 @@ export const MultilineInput = forwardRef<HTMLDivElement, MultilineInputProps>(
       const value = event.target.value;
       setValue(value);
       onChange?.(event);
-      const input = event.target;
+    };
 
+    useEffect(() => {
```
joshwooding, line 156: useEffect might be too slow here

packages/core/stories/multiline-input/multiline-input.stories.tsx
```diff
@@ -56,6 +56,37 @@ export const Controlled: StoryFn<typeof MultilineInput> = (args) => {
...
+
+  return (
+    <MultilineInput
+      {...args}
+      value={value}
+      onChange={handleChange}
+      style={{ maxWidth: "266px" }}
+      endAdornment={
+        <Button onClick={onClear} variant={"secondary"}>
+          <CloseIcon />
```
joshwooding, line 83:
```suggestion
          <CloseIcon aria-hidden />
```

packages/core/stories/multiline-input/multiline-input.stories.tsx
```diff
@@ -56,6 +56,37 @@ export const Controlled: StoryFn<typeof MultilineInput> = (args) => {
...
+  };
+
+  return (
+    <MultilineInput
+      {...args}
+      value={value}
+      onChange={handleChange}
+      style={{ maxWidth: "266px" }}
+      endAdornment={
+        <Button onClick={onClear} variant={"secondary"}>
```
joshwooding, line 82:
```suggestion
        <Button onClick={onClear} variant={"secondary"} aria-label="Clear input">
```

.changeset/clean-ravens-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed multiline input not collapsing when value is reduced
```
joshwooding, line 5: The actual issue is the multiline input not collapsing when it's controlled API is used.

packages/core/src/__tests__/__e2e__/multiline-input/MultilineInput.cy.tsx
```diff
@@ -122,4 +126,42 @@ describe("GIVEN an MultilineInput", () => {
...
+  it("should collapse back to fit content when value is reset", () => {
+    cy.mount(<ControlledWithAdornment rows={1} />);
+    cy.findByRole("textbox")
+      .invoke("height")
+      .then((defaultHeight) => {
+        cy.findByRole("textbox").realClick();
+        cy.realPress("Enter");
+        cy.realPress("Enter");
+        cy.realPress("Enter");
+        cy.findAllByRole("button").realClick();
```
joshwooding, line 159:
```suggestion
        cy.findByRole("button").realClick();
```

joshwooding:
I'm seeing a bug where by default the input has an extra row so when you hit enter it collapses

origami-z (commented):
I don't think the tests are fixing what #3015 describes, i.e., passing / controlling `rows` meaning the rows should always follow the prop?

Fercas123:
release-pr

Fercas123:
> I don't think the tests are fixing what #3015 describes, i.e., passing / controlling `rows` meaning the rows should always follow the prop?

I went for this approach to avoid making rows a controlled prop, but agree that if that is the case maybe the prop should be called `minRows` or something that doesn't suggest it can be changed... maybe something to clarify in the docs?

lukeac123 (approved):

origami-z:
> > I don't think the tests are fixing what #3015 describes, i.e., passing / controlling `rows` meaning the rows should always follow the prop?
> 
> I went for this approach to avoid making rows a controlled prop, but agree that if that is the case maybe the prop should be called `minRows` or something that doesn't suggest it can be changed... maybe something to clarify in the docs?

Let's edit the description of `rows` prop to express what's expected behaviour. Good to cross reference react.dev on the same prop, and test `textarea` native behaviour. If different, let's point out.

packages/core/src/multiline-input/MultilineInput.tsx
```diff
@@ -170,6 +167,17 @@ export const MultilineInput = forwardRef<HTMLDivElement, MultilineInputProps>(
...
+    const handleChange = (event: ChangeEvent<HTMLTextAreaElement>) => {
+      const value = event.target.value;
+      setValue(value);
+      onChange?.(event);
+      changeHeight();
+    };
+
+    useLayoutEffect(() => {
+      changeHeight();
+    }, [value]);
```
joshwooding, line 179: The lint error here should be fixed.

.changeset/clean-ravens-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed multiline input not collapsing when it's controlled API is used.
```
joshwooding, line 5:
```suggestion
Fixed multiline input not collapsing when its controlled API is used.
```

packages/core/src/multiline-input/MultilineInput.tsx
```diff
@@ -145,17 +147,13 @@ export const MultilineInput = forwardRef<HTMLDivElement, MultilineInputProps>(
...
-      onChange?.(event);
-      const input = event.target;
-
+    const changeHeight = useCallback(() => {
+      if (!input) return;
       const hasBeenManuallyResized =
         previousHeight.current !== undefined &&
-        input.style.height !== previousHeight.current;
-
+        input?.style.height !== previousHeight.current;
```
joshwooding, line 156: NIT: input isn't conditional here since the check above.

.changeset/clean-ravens-rest.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed multiline input not collapsing when its controlled API is used.
```
origami-z, line 5: "controlled" needs update

joshwooding (approved):

PR #3019, opened 2024-02-13
refactor dialog + disable scrim prop

lukeac123:
- Add disableScrim prop + examples 
- Fix useFloatingUI focus manager implementation

### DIFF
--- a/packages/lab/src/dialog/Dialog.tsx
+++ b/packages/lab/src/dialog/Dialog.tsx
@@ -135,7 +135,7 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
       <ConditionalScrimWrapper condition={open && !disableScrim}>
         <FloatingComponent
           open={showComponent}
-          role={"dialog"}
+          role="dialog"
           aria-modal="true"
           ref={floatingRef}
           width={elements.floating?.offsetWidth}
--- a/packages/lab/src/dialog/Dialog.tsx
+++ b/packages/lab/src/dialog/Dialog.tsx
@@ -5,6 +5,7 @@ import {
   useMemo,
   useState,
   ComponentProps,
+  ReactNode,
 } from "react";
 import { clsx } from "clsx";
 import {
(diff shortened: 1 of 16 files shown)


packages/lab/src/dialog/Dialog.tsx
```diff
@@ -15,19 +15,30 @@ import {
...
-  useId,
 } from "@salt-ds/core";
 import { useWindow } from "@salt-ds/window";
 import { useComponentCssInjection } from "@salt-ds/styles";
 import dialogCss from "./Dialog.css";
 import { DialogContext } from "./DialogContext";
 
+interface ConditionalScrimWrapperProps extends React.PropsWithChildren {
+  condition: boolean;
+}
```
joshwooding, line 33:
```suggestion
interface ConditionalScrimWrapperProps {
  children?: ReactNode
  condition: boolean;
}
```

joshwooding: More of a stylistic thing

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -15,19 +15,30 @@ import {
...
+
+export const ConditionalScrimWrapper = ({
+  condition,
+  children,
+}: ConditionalScrimWrapperProps) => {
+  return condition ? <Scrim fixed> {children} </Scrim> : <>{children} </>;
```
joshwooding, line 39:
```suggestion
  return condition ? <Scrim fixed>{children}</Scrim> : <>{children} </>;
```

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -119,39 +135,39 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
-            {...rest}
-          >
-            {children}
-          </FloatingComponent>
-        </Scrim>
-      )}
+      <ConditionalScrimWrapper condition={open && !disableScrim}>
+        <FloatingComponent
+          open={showComponent}
+          role={role}
```
joshwooding, line 141: role usage here can be simplified by hardcoding it here: role="dialog" and allowing role to be spread below by removing it from being destructured. Also we need to look at useRole again. I didn't realise floating-ui will change things depending on what value you provide it so need to double check it doesn't do anything special with dialog.

lukeac123: when role='alert dialog' is passed will both 'dialog' and 'alert dialog' roles then not be attached to the same element ?
This no longer uses the useRole hook from floating-ui, you thinking we should be added back in ?

packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -233,12 +238,12 @@ export const MandatoryAction: StoryFn<typeof Dialog> = ({
       <Dialog
         size={"small"}
         status={"error"}
+        id={"Mandatory Action"}
```
Fercas123, line 241: nit
```suggestion
        id="Mandatory Action"
```

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -119,39 +132,38 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
...
-            {...rest}
-          >
-            {children}
-          </FloatingComponent>
-        </Scrim>
-      )}
+      <ConditionalScrimWrapper condition={open && !disableScrim}>
+        <FloatingComponent
+          open={showComponent}
+          role={"dialog"}
```
Fercas123, line 138: nit
```suggestion
          role="dialog"
```

Fercas123 (dismissed):

Fercas123 (approved):

joshwooding (approved):

PR #3026, opened 2024-02-19
Promote Dialog to Core

lukeac123:


.changeset/empty-adults-allow.md
```diff
@@ -0,0 +1,53 @@
...
+    setOpen(value);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  return (
+    <>
+      <Button data-testid="dialog-button" onClick={handleRequestOpen}>
```
origami-z, line 27: Try remove `data-testid` for buttons? Shouldn't be needed in all tests and examples

PR #3035, opened 2024-02-21
Dialog Title Refactor

lukeac123:
- Title and subtitle now passed via props to ensure consistent styling for the Title and Subtitle
- id attached to `Dialog` and passed to aria-labelledby to read out title and subtitle when using a screen reader
- tests added for `title` / `subtitle` props and `disableDismiss`

### DIFF
--- a/.changeset/cuddly-hats-bake.md
+++ b/.changeset/cuddly-hats-bake.md
@@ -26,7 +26,7 @@ export const Default = (): ReactElement => {
 
   return (
     <>
-      <Button data-testid="dialog-button" onClick={handleRequestOpen}>
+      <Button onClick={handleRequestOpen}>
         Open default dialog
       </Button>
       <Dialog open={open} onOpenChange={onOpenChange} id={id}>
--- a/.changeset/cuddly-hats-bake.md
+++ b/.changeset/cuddly-hats-bake.md
@@ -5,3 +5,44 @@
 - Convert `Dialog Title` to accept props instead of a composable api
 - Optional Props `title` and `subtitle` added to `Dialog Title`
 - `Dialog Title` no longer accepts children
+- Optional `id` prop added to `Dialog` to announce the `title` and `subtitle` when using a screen reader
+
+```tsx
+export const Default = (): ReactElement => {
+  const [open, setOpen] = useState(false);
+  const id = useId();
+
+  const handleRequestOpen = () => {
+    setOpen(true);
+  };
+
+  const onOpenChange = (value: boolean) => {
+    setOpen(value);
+  };
+
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  return (
+    <>
+      <Button data-testid="dialog-button" onClick={handleRequestOpen}>
+        Open default dialog
+      </Button>
+      <Dialog open={open} onOpenChange={onOpenChange} id={id}>
+        <DialogTitle title="Terms and conditions" />
+        <DialogContent>
+          Dialog Content
+          </StackLayout>
+        </DialogContent>
+        <DialogActions>
+          <Button onClick={handleClose}>Cancel</Button>
+          <Button variant="cta" onClick={handleClose}>
+            Accept
+          </Button>
+        </DialogActions>
+      </Dialog>
+    </>
+  );
+};
+```
--- a/.changeset/cuddly-hats-bake.md
+++ b/.changeset/cuddly-hats-bake.md
@@ -1,0 +1,7 @@
+---
+"@salt-ds/lab": minor
+---
+
+- Convert `Dialog Title` to accept props instead of a composable api
+- Optional Props `title` and `subtitle` added to `Dialog Title`
+- `Dialog Title` no longer accepts children
--- a/packages/lab/src/dialog/Dialog.tsx
+++ b/packages/lab/src/dialog/Dialog.tsx
@@ -70,6 +70,11 @@ export interface DialogProps extends HTMLAttributes<HTMLDivElement> {
    * Prevent Scrim from rendering
    * */
   disableScrim?: boolean;
+  /**
+   * Optional id prop
+   * Used for accessibility purposes to announce the title and subtitle when using a screen reader
+   * */
+  id?: string;
 }
 
 const withBaseName = makePrefixer("saltDialog");
@@ -87,6 +92,7 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
     disableDismiss,
     size = "medium",
     disableScrim,
+    id,
     ...rest
   } = props;
   const targetWindow = useWindow();
@@ -127,7 +133,7 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
     }
   }, [open, showComponent, setShowComponent]);
 
-  const contextValue = useMemo(() => ({ status }), [status]);
+  const contextValue = useMemo(() => ({ status, id }), [status, id]);
 
   return (
     <DialogContext.Provider value={contextValue}>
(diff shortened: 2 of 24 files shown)


packages/lab/src/dialog/DialogContent.css
```diff
@@ -5,9 +5,10 @@
   background: var(--salt-container-primary-background);
   padding-bottom: var(--salt-spacing-50);
 
-  margin-left: var(--salt-spacing-300);
+  margin-left: var(--salt-spacing-200);
   margin-right: var(--salt-spacing-300);
   padding-right: var(--salt-spacing-300);
+  padding-left: var(--salt-spacing-100);
```
lukeac123, line 11: Margin was covering focus ring so using padding instead

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -13,20 +16,37 @@ import dialogTitleCss from "./DialogTitle.css";
...
+   * Displays the Dialog Subtitle in a Label component
+   * */
+  id?: string;
 }
 
 export const DialogTitle = ({
   children,
   className,
+  title,
+  subtitle,
```
lukeac123, line 49: Passing these down as props was chosen over the composable api to ensure the formatting for Dialog Title and subtitle would be consistent across all dialog components

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -42,19 +62,22 @@ export const DialogTitle = ({
...
           [withBaseName("withAccent")]: !disableAccent && !status,
-          [withBaseName(status)]: !!status,
+          [withBaseName(status)]: !status,
         },
         className
       )}
       {...rest}
     >
       {status && <StatusIndicator status={status} />}
-      {children}
```
lukeac123, line 57: children removed - if and additional component is needed e.g Step tracker this can be done by wrapping Dialog title in a Layout Component. See the Large Dialog within the size examples

ZarkoPernar: Why do you need to remove children, which is a breaking change, and not instead just introduce one prop `subtitle`?

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -13,20 +15,34 @@ import dialogTitleCss from "./DialogTitle.css";
 
 const withBaseName = makePrefixer("saltDialogTitle");
 
-interface DialogTitleProps extends ComponentPropsWithoutRef<"h2"> {
+interface DialogTitleProps {
```
joshwooding, line 18: These types no longer match now it doesn't extend anything. The change should have been
```suggestion
interface DialogTitleProps extends ComponentPropsWithoutRef<"div"> {
```

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -13,20 +15,34 @@ import dialogTitleCss from "./DialogTitle.css";
...
+   */
+  title: ReactNode;
+  /**
+   * Displays the Dialog Subtitle in a Label component
+   **/
+  subtitle?: ReactNode;
+  /**
+   * Displays the Dialog Subtitle in a Label component
+   * */
+  id?: string;
```
joshwooding, line 37: This is the wrong description

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -42,19 +58,22 @@ export const DialogTitle = ({
...
         {
           [withBaseName("withAccent")]: !disableAccent && !status,
-          [withBaseName(status)]: !!status,
+          [withBaseName(status)]: !status,
         },
         className
       )}
       {...rest}
     >
       {status && <StatusIndicator status={status} />}
```
joshwooding, line 72: Have you checked the status alignment?

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -42,19 +58,22 @@ export const DialogTitle = ({
...
-      {children}
-    </H2>
+      <StackLayout direction="column" gap={0}>
```
joshwooding, line 73: Could we use CSS instead of the layout helper

.changeset/cuddly-hats-bake.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/lab": minor
+---
+
+- Convert `Dialog Title` to accept props instead of a composable api
+- Optional Props `title` and `subtitle` added to `Dialog Title`
```
origami-z, line 6: Need some test (cy / chromatic) to cover this?

site/src/examples/dialog/Subtitle.tsx
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  return (
+    <>
+      <Button onClick={handleRequestOpen}>Open dialog with subtitle</Button>
+      <Dialog open={open} onOpenChange={onOpenChange} size="small">
+        <DialogTitle
+          title="Subscribe"
+          subtitle="Recieve emails about the latest updates"
+        />
```
lukeac123, line 34: Adding an aria-labelled by like in the other examples doesn't work here. The screen reader reads the subtitle first followed by the title.
Id needs to be passed to Dialog Title and then onto the H2 tag

origami-z: Yes, this is one of the advantage when choosing composition, i.e. user can have access to individual element. 

`id` should still be mapped to root element by convention, if needed, can add old approach of `TitleProps` and spread that to sub-components

lukeac123: i've passed added `id` as a prop to DialogTitle and then onto the H2 tag

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -13,36 +14,50 @@ import dialogTitleCss from "./DialogTitle.css";
...
+  const { status: statusContext, id } = useDialogContext();
   const targetWindow = useWindow();
   useComponentCssInjection({
     testId: "salt-dialog-title",
     css: dialogTitleCss,
     window: targetWindow,
   });
 
+  const subtitleId = `${id}-subtitle`;
+  const titleId = `${id}-title`;
```
lukeac123, line 55: subtitle and title ids passed down to their respective components, enables screen reader to announce the respective content from aria-labelelledby in `Dialog` root

packages/lab/src/dialog/Dialog.tsx
```diff
@@ -137,6 +142,7 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(function Dialog(
           open={showComponent}
           role="dialog"
           aria-modal="true"
+          aria-labelledby={`${id}-subtitle ${id}-title`}
```
lukeac123, line 145: Subtitle read out first in the cases where it is defined

packages/lab/src/dialog/DialogTitle.tsx
```diff
@@ -13,36 +14,50 @@ import dialogTitleCss from "./DialogTitle.css";
...
+   * Displays the accent bar in the Dialog Title */
   disableAccent?: boolean;
+  /**
+   * Displays the Dialog Title in a H2 component
+   */
+  title: ReactNode;
+  /**
+   * Displays the Dialog Subtitle in a Label component
+   **/
+  subtitle?: ReactNode;
```
lukeac123, line 33: titles naming convention pending design review

lilyvc (dismissed):
lgtm

Fercas123 (dismissed):

packages/lab/src/dialog/DialogTitle.tsx
joshwooding: Have you tried just applying the id to the root of the component?

lilyvc (approved):

Fercas123 (approved):

PR #3039, opened 2024-02-22
Add valueToString to DropdownNext and ComboBoxNext

joshwooding:


packages/lab/src/list-control/ListControlContext.tsx
```diff
@@ -46,6 +46,9 @@ export const ListControlContext = createContext<ListControlContextValue<any>>(
     },
     multiselect: false,
     focusVisibleState: false,
+    valueToString(item) {
```
origami-z, line 49: not sure whether should use default one here

lilyvc:
noticed that the cursor and state changes on hover of disabled options:
https://saltdesignsystem-git-add-value-to-string-to-cb-and-dd-fed-team.vercel.app/salt/components/dropdown/examples#disabled-option

same for combo box

origami-z (commented):
Looks ok. Maybe some additional tests needed for different `valueToString` and other fixes

origami-z (approved):

Fercas123 (approved):

PR #3072, opened 2024-03-01
Fix target _blank link icon alignment

### LINKED ISSUE #3048: CSS on Link (target=_blank) specificity issue
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

    "@salt-ds/core": "^1.17.0",
    "@salt-ds/icons": "^1.9.1",
    "@salt-ds/theme": "^1.11.1",

### Description

The "external" icon is too close to the link, because .saltIcon is taking precedence over .saltLink-icon - and margin is overridden.

Github is not letting me upload images, but I have screenshots of the broken css and the chrome inspector - please contact me internally.

### Steps to reproduce

```
const Repro: React.FunctionComponent = () => {
    return (
        <div>
            <Link target="_blank" href={"whatever"}>
                Whatever
            </Link>
        </div>
    );
};
```

### Expected behavior

I expect the margins to behave like they do in the saltdesignsystem.com docs, but they do not.

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

joshwooding:
Closes #3048

### DIFF
--- a/packages/core/src/link/Link.css
+++ b/packages/core/src/link/Link.css
@@ -15,7 +15,7 @@
 /* Main css class */
 .saltLink {
   color: var(--link-color);
-  letter-spacing: 0px;
+  letter-spacing: var(--salt-text-letterSpacing);
   text-decoration: var(--link-textDecoration);
   font-family: var(--link-fontFamily);
 }
@@ -31,8 +31,8 @@
 }
 
 /* External link's icon */
-.saltLink-icon {
-  margin-left: var(--salt-size-adornmentGap);
+.saltLink .saltLink-icon {
+  margin-left: var(--salt-spacing-75);
   margin-bottom: -2px;
 }
 
--- a/.changeset/angry-houses-bake.md
+++ b/.changeset/angry-houses-bake.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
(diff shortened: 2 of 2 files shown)


origami-z (commented):
The link icon is lower than i expected, especially in LD/TD in the QA grid?

joshwooding:
> The link icon is lower than i expected, especially in LD/TD in the QA grid?

Reverted the change, I did play around with a less hardcoded approach e.g. top: calc((var(--salt-text-lineHeight) - var(--icon-size) / 2). But it's quite ugly and for now -2px works across all densities anyway

packages/core/src/link/Link.css
```diff
@@ -31,8 +31,8 @@
 }
 
 /* External link's icon */
-.saltLink-icon {
-  margin-left: var(--salt-size-adornmentGap);
```
origami-z, line 35: Spec need to change?

origami-z (approved):

lilyvc (approved):

packages/core/src/link/Link.css
joshwooding: I don't think it's even in the spec

Issue #2781, opened 2023-11-23
Tabstrip Next padding breaks collapse behaviour


el-dav:
### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

@salt-ds/lab/1.0.0-alpha.22

### Description

Tabstrip has left/right padding added in the following commit:

https://github.com/jpmorganchase/salt-ds/commit/7e14397948a36ae1b6cd7ef382c3d1b678ead51d

This appears to break the collapse behaviour, as the Tabstrip doesn't begin collapsing at the right point.

This implicit padding also makes Tabstrip difficult to use as you can't easily control the spacing to adjacent components

### Steps to reproduce

Place a Tabstip with several tabs in a horizontal <StackLayout> next to some other components. Reduce the width of the browser using window resizing or the devtools panel. You'll notice that the Tabstrip doesn't start collapsing till after it has been forced off the edge of the screen.

### Expected behavior

The tabstrip should never be pushed of the edge of the screen. It should start collapsing at and earlier point.

If you disable the passing css using devtools the behaviour seems correct

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

origami-z:
Possibly related to #2726

Fercas123:
this seems to be resolved by https://github.com/jpmorganchase/salt-ds/pull/2773

origami-z:
Still broken on https://www.saltdesignsystem.com/salt/components/tabs/examples#inline?

Fercas123:
there seem to be a different issue with inline, to reproduce this one: https://stackblitz.com/edit/salt-test-question-rrs5d9?file=App.tsx,override.css

I'll have a look at the inline one

PR #3086, opened 2024-03-05
Fix tabstrip not collapsing when sharing parent with other items

Fercas123:
fixes https://github.com/jpmorganchase/salt-ds/issues/2781


joshwooding (changes requested):
Is there a way to test this?

origami-z (approved):

origami-z:
> There is a regression in touch QA, Looks like the padding left is making the tabstrip think it has enough space to fit the button, and overflow hidden now hides it. One more update to the PR coming up

Maybe QA needs to be updated to include both main type and inline type

origami-z:
TD of Badge QA should overflow or not?

origami-z (approved):

lukeac123 (approved):

PR #3092, opened 2024-03-07
Add salt icons css

Fercas123:


packages/icons/saltIcons.css
```diff
@@ -0,0 +1,1106 @@
+/** WARNING: This file was generated by a script. Do not modify it manually */
+[class*=" saltIcon-"],
+[class^="saltIcon-"] {
+  mask-size: cover;
```
joshwooding, line 4: Are you sure this is supported by all our supported browsers? From looking at caniuse.com it needs the webkit prefix.

packages/icons/saltIcons.css
```diff
@@ -0,0 +1,1106 @@
...
+  background-color: currentColor;
+  height: max(var(--salt-size-icon, 12px), 12px);
+  width: max(var(--salt-size-icon, 12px), 12px);
+}
+.saltIcon-AccessibleSolid {
+  mask-image: url("data:image/svg+xml,%3Csvg%20width%3D%2212%22%20height%3D%2212%22%20viewBox%3D%220%200%2012%2012%22%20%3E%0A%3Cpath%20d%3D%22M5%203.00006C5.82843%203.00006%206.5%202.32849%206.5%201.50006C6.5%200.671634%205.82843%206.10352e-05%205%206.10352e-05C4.17157%206.10352e-05%203.5%200.671634%203.5%201.50006C3.5%202.32849%204.17157%203.00006%205%203.00006Z%22%20%2F%3E%0A%3Cpath%20d%3D%22M4%204.00006H6V6.00006H7.99997L10.8284%208.82846L9.41421%2010.2427L7.1716%208.00006H4V4.00006Z%22%20%2F%3E%0A%3Cpath%20d%3D%22M3%204.1261C1.27477%204.57014%200%206.13624%200%208.00008C0%2010.2092%201.79086%2012.0001%204%2012.0001C5.57942%2012.0001%206.94504%2011.0847%207.59525%209.75548L6.83204%208.99228C6.42242%2010.1615%205.30913%2011.0001%204%2011.0001C2.34315%2011.0001%201%209.65694%201%208.00008C1%206.69386%201.83481%205.58262%203%205.17079V4.1261Z%22%20%2F%3E%0A%3C%2Fsvg%3E");
```
joshwooding, line 10: Same question here

packages/icons/saltIcons.css
Fercas123: good point, will add it in

lilyvc (dismissed):

packages/icons/package.json
```diff
@@ -39,6 +39,7 @@
     "provenance": true
   },
   "files": [
-    "css"
+    "css",
+    "saltIcons.css"
```
origami-z, line 43: Is it better to put this inside existing `css` folder?

And i realized that we may need some additional comment on top of `salt-icon.css` and this new `saltIcons.css` to make it clearly to the user, one is for no injection usage

site/src/examples/icon/CssBackgroundImage.tsx
```diff
@@ -0,0 +1,5 @@
+import "@salt-ds/icons/saltIcons.css";
```
joshwooding, line 1: This should use the new one and we should make this import not possible

packages/icons/saltIcons.css
```diff
@@ -0,0 +1,1473 @@
+/** WARNING: This file includes all icons in a css mask format and was generated by a script. Do not modify it manually */
+[class*=" saltIcon-"],
+[class^="saltIcon-"] {
```
libertymayc, line 3: this causes problems if you set the color to primary or secondary (to test, add color='primary' to all icons QA story)
I assume the use case for using the icon component directly vs the class name would mean this rarely occurs, someone wouldn't import the css if using the component directly, but something to consider

Fercas123: good spot!! will change the classname

.changeset/odd-cats-act.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/icons": minor
+---
+
+Added `saltIcons.css` with all icons SVG as background image.
+
+```js
+import "@salt-ds/icons/saltIcons.css";
```
libertymayc, line 8: checking since it's copied to salt-icons.css, is saltIcons.css the one consumers would use or would it be salt-icons?

Fercas123: i removed the copy for it to work how country symbols do for now

.changeset/odd-cats-act.md
```diff
@@ -0,0 +1,14 @@
...
+
+const Example = () => {
+  const iconName = `AddDocument`;
```
joshwooding, line 11:
```suggestion
  const iconName = "AddDocument";
```

.changeset/odd-cats-act.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/icons": minor
+---
+
+Added `saltIcons.css` with all icons SVG as background image.
```
joshwooding, line 5:
```suggestion
Added `saltIcons.css` with all icon SVGs as background images.
```

joshwooding (approved):

lukeac123 (approved):

PR #3110, opened 2024-03-11
Add box-sizing: border-box to more components

joshwooding:


origami-z (approved):
Anyway we can catch these in Chromatic?  maybe set it globally to some other value in storybook CSS?

These weren't caught in our Chromatic tests initially

joshwooding:
> Anyway we can catch these in Chromatic? maybe set it globally to some other value in storybook CSS?

These were wrong in Storybook already but weren't caught by our visual regression tests. Probably due to some quirk with the QA stories

shey-v (approved):

PR #3133, opened 2024-03-15
Interactable card keyboard controls

lilyvc:
adding keyboard controls to interactable card


packages/lab/stories/interactable-card/interactable-card.stories.tsx
```diff
@@ -0,0 +1,336 @@
...
+  const [selected, setSelected] = useState<InteractableCardValue>();
+
+  return (
+    <StackLayout>
+      <StackLayout gap={1}>
+        <H3>Payment method</H3>
+        <Text>Choose your payment method from the options below.</Text>
+      </StackLayout>
+      <InteractableCardGroup
+        onChange={(_event, value) => {
```
lilyvc, line 241: is this api okay? having to use event and value? I was thinking if we changed it to input then we might not need to do this

packages/lab/src/interactable-card/InteractableCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+      aria-checked={ariaChecked}
+      data-id={value}
+      aria-disabled={disabled}
+      data-value={value}
+      className={clsx(
+        withBaseName(),
+        withBaseName(variant),
+        {
+          [withBaseName("accent")]: accent,
+          [withBaseName(`accent${capitalize(accent || "")}`)]: accent,
```
Fercas123, line 166:
```suggestion
          [withBaseName(`accent${capitalize(accent ?? "")}`)]: accent,
```

packages/lab/src/interactable-card/InteractableCard.tsx
```diff
@@ -0,0 +1,181 @@
...
+    onBlur,
+    onClick,
+  });
+
+  return (
+    <div
+      {...cardProps}
+      role={role}
+      aria-checked={ariaChecked}
+      data-id={value}
```
Fercas123, line 158: do you need data-id?

packages/lab/src/interactable-card/InteractableCardGroup.tsx
```diff
@@ -0,0 +1,206 @@
...
+  useEffect(() => {
+    const childElements: HTMLElement[] = Array.from(
+      groupRef.current?.querySelectorAll(
+        ".saltInteractableCard:not([disabled])"
+      ) ?? []
+    );
+    setElements(childElements);
+  }, [children]);
+
+  const select = useCallback(
```
Fercas123, line 98: in the group multiselect, clicking twice seem to unselect a card, but key nav doesn't (it might still be in the tbd but thought of mentioning it still just in case)

lilyvc: where do you see this? In storybook the multiselect unselects with space, is it the focus ring making it less visible?

Fercas123: ah, i see what is happening, so keyboard selection happens with spacebar but not with enter, that confused me, but if that's the intentional behaviour its fine

packages/lab/src/interactable-card/InteractableCard.tsx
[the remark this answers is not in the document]
lilyvc: I'm not sure, it seems to be like that in UITK

packages/lab/src/interactable-card/InteractableCardGroup.tsx
lilyvc: yeah I think it should behave like a checkbox, so only space not enter

PR #3143, opened 2024-03-19
[WIP] json for theme 2.0

libertymayc:
Builds from color-palette-next branch to use salt theme 2.0


packages/theme/json/theme.json
```diff
@@ -985,826 +960,640 @@
...
+        "$value": "#ffffff00",
         "$type": "color"
       },
-      "color-white-fade-background-hover": {
-        "$value": {
-          "color": "{foundations.color.color-white}",
-          "opacity": "{foundations.opacity.opacity-8}"
-        },
+      "color-white-8a": {
+        "$value": "#ffffff00",
```
libertymayc, line 1596: unsure where opacity stands in theme 2.0, if we would link to an opacity foundation here, and have color and opacity as separate values like the JSON structure for theme 1 fade tokens @origami-z

packages/theme/json/parsers/validate.js
```diff
@@ -0,0 +1,40 @@
...
+        }
+      }
+      if (findValue && !findValue["$value"]) {
+        console.log(`Path found ${findValue} but no value defined`);
+      }
+    }
+  }
+}
+
+function validate() {
```
libertymayc, line 28: @origami-z this should check all paths are correct refs

origami-z:
Some quick thoughts for future myself while starting #4041, trying to think how we can best leverage tools out there, e.g. v4 of style-dictionary [supports DTCG](https://styledictionary.com/reference/utils/dtcg/) now.

- `packages/theme/json/theme.json`, need to think about tooling compatibility on top level mode/density/etc blocks, how can we indicate they are replacements
- To help use tools out there, trying to see how existing tools define variable "modes"
    - If there are format out there, test them 
    - If not, think about whether Salt format / tool should be used to produce individual `json` (in specific mode/density/stylingOptions) to use other tools

PR #3148, opened 2024-03-20
Add code to styleAs Text prop

### LINKED ISSUE #3124: Add code styleAs to Text.
### Area

UI Components

### The problem

Currently the only way to consume the --salt-typography-fontFamily-code style is via CSS token. It would be nice if we could consumer it via a component prop too.

### The solution

Add styleAs="code"

### Alternatives and examples

.

libertymayc:
closes #3124


packages/core/src/text/TextCode.tsx
```diff
@@ -0,0 +1,13 @@
+import { forwardRef } from "react";
+import { Text, TextProps } from "./Text";
+
+export const TextCode = forwardRef<
```
joshwooding, line 4: This will probably be better as "Code"

joshwooding (changes requested):
This should be done the same way as headers so the styles are applied to `code.saltText` and `.saltText-code`

packages/core/src/text/Code.tsx
```diff
@@ -0,0 +1,12 @@
+import { forwardRef } from "react";
+import { Text, TextProps } from "./Text";
+
+export const Code = forwardRef<HTMLSpanElement, Omit<TextProps<"span">, "as">>(
+  function TextAction({ children, className, ...rest }, ref) {
+    return (
+      <Text as="code" styleAs="code" ref={ref} {...rest}>
```
joshwooding, line 7:
```suggestion
      <Text as="code" ref={ref} {...rest}>
```

packages/core/src/text/Code.tsx
```diff
@@ -0,0 +1,12 @@
+import { forwardRef } from "react";
+import { Text, TextProps } from "./Text";
+
+export const Code = forwardRef<HTMLSpanElement, Omit<TextProps<"span">, "as">>(
```
joshwooding, line 4:
```suggestion
export const Code = forwardRef<HTMLSpanElement, Omit<TextProps<"code">, "as">>(
```
Is it possible to use code here?

.changeset/silly-forks-nail.md
```diff
@@ -0,0 +1,12 @@
...
+---
+
+Added "code" option to Text 'styleAs' prop
+
+```tsx
+<>
+  <Code>Code text</Code>
+  <Text styleAs="code">Text styled as code</Text>
+</>
+```
```
joshwooding, line 12: Realised we've had a changelog like this before so we should just copy it :)

```suggestion
`Text` has a new variant: `Code`. See [Text component](/salt/components/text) for more information.

```tsx
<>
  <Code>Code text</Code>
  <Text as="code">Text styled as code</Text>
  <Text styleAs="code">Text styled as code</Text>
</>
```
```

joshwooding (approved):

joshwooding (approved):

joshwooding (changes requested):
Actually one small thing, can you add the code style to the site under the "Styles" example please

joshwooding (approved):

shey-v (approved):

PR #3150, opened 2024-03-20
Fix Switch's alignment when it's used with a left/right aligned FormFieldLabel

### LINKED ISSUE #3109: Create an example showing FormField and Switch
We have had requests to have a label to the left of the switch. Our current recommendation is to do this with a FormField. We should add a code example to our site (https://www.saltdesignsystem.com/salt/components/switch/examples).

libertymayc:
closes #3109

### DIFF
--- a/.changeset/rare-turkeys-change.md
+++ b/.changeset/rare-turkeys-change.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix padding when Switch used with left aligned Form field label
--- a/.changeset/rare-turkeys-change.md
+++ b/.changeset/rare-turkeys-change.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fix padding when Switch used with left/right aligned FormFieldLabel
+Fixed Switch's alignment when it's used with a left/right aligned FormFieldLabel.
--- a/.changeset/rare-turkeys-change.md
+++ b/.changeset/rare-turkeys-change.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fix padding when Switch used with left aligned Form field label
+Fix padding when Switch used with left/right aligned FormFieldLabel
--- a/packages/core/src/switch/Switch.css
+++ b/packages/core/src/switch/Switch.css
@@ -30,6 +30,10 @@
   box-sizing: border-box;
 }
 
+.saltSwitch-label {
+  padding-top: var(--salt-spacing-25);
+}
+
 /* Styles applied to the icon component if focused */
 .saltSwitch-input:focus-visible + .saltSwitch-track {
   outline-offset: var(--salt-focused-outlineOffset);
--- a/packages/core/src/switch/Switch.css
+++ b/packages/core/src/switch/Switch.css
@@ -127,3 +127,7 @@
 .salt-density-high .saltSwitch-thumb {
   border: 0;
 }
+
+.saltFormField-labelLeft .saltSwitch {
+  padding-top: var(--salt-spacing-75);
+}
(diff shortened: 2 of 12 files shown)


joshwooding (changes requested):
I'm not a fan of these changes, we should find a way to do this without requiring users to add a class.

libertymayc:
> I'm not a fan of these changes, we should find a way to do this without requiring users to add a class.

me too. are you thinking via a prop? If we use the form field label it will need styling since it's different from the Switch label. as mentioned in the description it also wouldn't have the interaction, plus makes the component itself technically have two labels since Switch is wrapped in one itself

joshwooding:
> > I'm not a fan of these changes, we should find a way to do this without requiring users to add a class.
> 
> me too. are you thinking via a prop? If we use the form field label it will need styling since it's different from the Switch label. as mentioned in the description it also wouldn't have the interaction, plus it makes the component itself have two labels since Switch is wrapped in one itself 👎

I think the expectation is for it to look different anyway. It's more making sure Switch works with FormField

site/src/examples/switch/LeftAlignedLabel.tsx
```diff
@@ -0,0 +1,11 @@
+import { ReactElement } from "react";
+import { FormField, FormFieldLabel, Switch } from "@salt-ds/core";
+
+export const LeftAlignedLabel = (): ReactElement => (
+  <>
+    <FormField labelPlacement="left">
+      <FormFieldLabel>Active</FormFieldLabel>
+      <Switch />
+    </FormField>
+  </>
```
joshwooding, line 10:
```suggestion
  <div>
    <FormField labelPlacement="left">
      <FormFieldLabel>Active</FormFieldLabel>
      <Switch />
    </FormField>
  </div>
```

site/src/examples/form-field/WithControls.tsx
```diff
@@ -121,6 +121,10 @@ export const WithCheckboxAndRadioButton = (): ReactElement => {
           isCheckboxError ? "Must select at least one option. " : ""
         }`}</FormFieldHelperText>
       </FormField>
+      <FormField>
```
joshwooding, line 124: This example isn't loading?

libertymayc: fixed

packages/core/src/switch/Switch.css
```diff
@@ -30,10 +30,6 @@
   box-sizing: border-box;
 }
 
-.saltSwitch-label {
```
origami-z, line 33: Need this for default switch label

bhoppers2008 (dismissed):

packages/core/src/switch/Switch.css
```diff
@@ -119,3 +119,10 @@
 .salt-density-high .saltSwitch-thumb {
   border: 0;
 }
+
+/* Styles applied when switch is inside a form field with label left or right, ensuring alignment with base / input height */
+.saltFormField-labelRight .saltSwitch,
+.saltFormField-labelLeft .saltSwitch {
```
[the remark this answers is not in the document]
origami-z: added new

joshwooding (approved):

.changeset/rare-turkeys-change.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix padding when Switch used with left/right aligned FormFieldLabel
```
joshwooding, line 5:
```suggestion
Fixed padding when a Switch is used with a left/right aligned FormFieldLabel.
```

joshwooding (approved):

Fercas123 (approved):

PR #3168, opened 2024-03-22
Fix text inside Panel having the incorrect color when color-scheme is not set or doesn't change the default text color e.g. in Safari

### LINKED ISSUE #3013: Text color issue in a Panel within nested modes in Safari
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

```
"@salt-ds/core": "1.16.0"
"@salt-ds/theme": "1.11.0"
```

### Description

When text (`<p>`) is used in a Panel within nested `SaltProvider` with different modes, they appear in wrong color (only in Safari not Chrome)

Workaround: use Text components instead

wrong text color in Safari

Below shows what's in the inspect panel of `<p>`

inspect color property of p tag

### Steps to reproduce

https://stackblitz.com/edit/salt-template-nwlhdh?file=App.tsx

```
<SaltProvider>
  <Panel>
    <p>abc</p>
    <SaltProvider mode="dark">
      <Panel>
        <p>abc</p>
      </Panel>
    </SaltProvider>
  </Panel>
</SaltProvider>
```

### Expected behavior

_No response_

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [X] Safari
- [ ] Firefox
- [ ] Edge

libertymayc:
closes #3013

also adds font-family to Panel, as in Safari this isn't picked up within plain text in a Panel (unless Text component directly used)

### DIFF
--- a/packages/core/src/panel/Panel.css
+++ b/packages/core/src/panel/Panel.css
@@ -16,7 +16,7 @@
 /* Styles applied to the root element */
 .saltPanel {
   background: var(--saltPanel-background, var(--panel-background));
-  color: var(--saltPanel-color, initial);
+  color: var(--saltPanel-color, inherit);
   height: var(--saltPanel-height, 100%);
   overflow: auto;
   padding: var(--saltPanel-padding, var(--salt-size-container-spacing));
--- a/.changeset/plenty-pans-film.md
+++ b/.changeset/plenty-pans-film.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed text inside Panel having the incorrect color when color-scheme is not set or doesn't change the default text color e.g. in Safari.


packages/core/src/panel/Panel.css
```diff
@@ -11,7 +11,8 @@
 /* Styles applied to the root element */
 .saltPanel {
   background: var(--saltPanel-background, var(--panel-background));
-  color: var(--saltPanel-color, initial);
+  color: var(--saltPanel-color, inherit);
+  font-family: var(--saltPanel-fontFamily, var(--salt-text-fontFamily));
```
joshwooding, line 15: I believe this has already been tried and were previously removed https://github.com/jpmorganchase/salt-ds/pull/2952
cc @origami-z

libertymayc: issue seems to be related to font family but I think color fix would be suitable here? @origami-z

origami-z: If `color` change fixes the issue, could do. Need to do a few tests across browsers to understand impact

tomhazledine (approved):

joshwooding (approved):

PR #3172, opened 2024-03-25
Segmented Button Group to core

lukeac123:
- Promote SegmentedButtonGroup to core
- This PR does not include site docs, these will follow after


joshwooding (changes requested):
Site docs need to go in before as they are a pre-requisite to promoting to core.

PR #3174, opened 2024-03-26
Overlay close button

lukeac123:
- Add OverlayPanelCloseButton and OverlayPanelContent Components as children of OverlayPanel
- Separate OverlayPanelContent component is needed to make the close button sticky, it also aligns with Dialog
- Remove forward ref from the Overlay Parent Component, Overlay Parent component is just a context provider so don't think this is needed


packages/lab/src/overlay/OverlayPanelCloseButton.css
```diff
@@ -0,0 +1,15 @@
...
+  z-index: 1;
+}
+
+.saltOverlayPanelCloseButton-button.saltButton-secondary {
+  position: sticky;
+  right: 0;
+  top: 0;
+}
+
+.saltOverlayPanelContent {
```
lukeac123, line 13: This content spacing is only needed when the close button is present

packages/lab/src/overlay/OverlayTrigger.tsx
```diff
@@ -21,17 +21,10 @@ export function OverlayTrigger(props: OverlayTriggerProps) {
     return <>{children}</>;
   }
 
-  const getTriggerProps = () =>
-    getReferenceProps({
-      ref: reference,
-      id: `${id}-trigger`,
-      "aria-controls": `${id}-panel`,
```
lukeac123, line 28: Is aria-controls needed here ? 
Reference ref is already passed down in trigger ref and id is not needed

packages/lab/src/overlay/Overlay.tsx
```diff
@@ -75,51 +61,42 @@ export const Overlay = forwardRef<HTMLDivElement, OverlayProps>(
...
-          context,
-          arrowProps,
-          floating,
-          reference,
-          handleCloseButtonClick,
-          getFloatingProps,
-          getReferenceProps,
-        }}
-      >
-        <div ref={ref} {...rest}>
```
lukeac123, line 119: I don't think ref and {...rest} here are needed and therefore removed forward ref.
This component is just providing context, a ref and styles can still be passed down directly to the OverlayPanel or OverlayTrigger component - can't think of a scenario where it would need to be passed down to the Overlay Component directly

packages/lab/src/overlay/OverlayPanelCloseButton.css
lukeac123: using !important doesn't seem like the ideal solution to override the padding when the close button is being used - thoughts ?

packages/lab/src/overlay/OverlayPanelCloseButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-overlay-panel-close-button",
+    css: overlayPanelCloseButtonCss,
+    window: targetWindow,
+  });
+
+  return (
+    <div className={clsx(withBaseName(), className)}>
+      <Button
```
lukeac123, line 24: In our Button accessibility we say to add a tooltip to buttons with just an icon - is the close button is an exception or should this also have a tooltip ?

packages/lab/stories/overlay/overlay.stories.tsx
```diff
@@ -4,78 +4,108 @@ import {
...
   OverlayTrigger,
+  OverlayPanelCloseButton,
+  OverlayPanelContent,
 } from "@salt-ds/lab";
 
 import "./overlay.stories.css";
 
 export default {
   title: "Lab/Overlay",
-  component: Overlay,
```
lukeac123, line 22: Removed because this was always passing through an empty `onOpenChange` function in {..args } which was messing up the stories

packages/lab/src/overlay/OverlayPanelCloseButton.css
lilyvc: yeah this isn't ideal because it applies on the site examples without a button

lukeac123: updated to use the sibling selector instead, so will only apply to the specific overlay - but yeh `disableCloseButton` could be a prop, depends on what api we would prefer

packages/lab/src/overlay/OverlayPanelCloseButton.tsx
lilyvc: I imagine this isn't required for something as universally recognisable as a close button. The aria-label is there for accessibility. Maybe the button guidance is too prescriptive.

lukeac123: yeh agreed i don't think it is needed

lilyvc (approved):

Fercas123 (approved):

PR #3178, opened 2024-03-27
Overlay to core

lukeac123:


packages/lab/src/color-chooser/ColorChooser.tsx
```diff
@@ -144,7 +146,7 @@ export const ColorChooser = ({
     setActiveTab(index);
   };
 
-  const OverlayData = () => {
```
lukeac123, line 147: Did this renaming previously, but think OverlayContent makes more sense than OverlayData, same as seen in the contact details pattern

PR #3181, opened 2024-03-27
Vertical navigation updates

mikearildbrown:


site/docs/patterns/vertical-navigation.mdx
```diff
@@ -5,6 +5,7 @@ aliases:
   - /salt/patterns/vertical-navigation
 data:
   components: ["Navigation item"]
+  relatedPatterns: ["Navigation"]
   resources:
     [
       {
```
joshwooding, line 11: I think we should remove the resources here for now. Just so we're consistent with the other patterns. We can then add back relevant links to all the patterns holistically

joshwooding (changes requested):

joshwooding (approved):

Fercas123 (approved):

PR #3185, opened 2024-03-28
Sharp country flags

tomhazledine:
Adds "sharp" (a.k.a. rectangular) variant to `@salt-ds/countries` components.

* New assets are located in `packages/countries/src/SVG/sharp` and named by country-code (this is important for the flag generation script).
* The flag generation script (`packages/countries/scripts/generateCountrySymbol.mjs`) has been amended to generate the sharp svg elements and viewbox which are optimised and passed into the mustache template that scaffolds the flag components.
* The component template has been modified to use the sharp viewbox and svg elements when the `variant` prop is set to "sharp".
Screenshot 2024-03-28 at 13 56 10


packages/countries/scripts/generateCountrySymbol.mjs
```diff
@@ -149,71 +158,92 @@ const generateCountrySymbolComponents = ({
...
+    };
+
+    const optimizedSvg = optimize(svgString, optimizeOptions);
+    const optimizedSharpSvg = optimize(sharpSvgString, optimizeOptions);
 
     const getUidString = (countryCode, value) => {
       return `\$\{uid\}-${countryCode}-${value}`;
     };
 
-    const svgPaths = optimize(optimizedSvg.data, {
```
tomhazledine, line 158: This diff looks more chunky than it actually is.

To avoid excessive duplication, the `mapHTMLAttributesToReactProps` and `removeSvg` have been moved out of the `optimize()` call and into a separate object. This is because they are duplicated in `const svgPaths = optimize(...` and `const sharpSvgPaths = optimize(...` but the `find-viewBox` options are different for each instance *(and the order of the plugin declaration is important: `find-viewBox` must go between `mapHTMLAttributesToReactProps` and `removeSvg`)*

PR #3215, opened 2024-04-08
Sharp country symbols

tomhazledine:
Adds sharp (rectangular) symbol to `@salt-ds/countries`

All country symbol components (`AD`, `AE`, `AF`, etc...) now have a "sharp" equivalent (`AD_Sharp`, `AE_Sharp`, `AF_Sharp`, etc...).

Screenshot 2024-04-09 at 17 47 58

Screenshot 2024-04-09 at 17 27 31

---

`LK` (Sri Lanka) and `MX` (Mexico) also have updated SVGs for the default round variants (adding detail to flag motifs)


.changeset/modern-weeks-jog.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/countries": patch
```
joshwooding, line 2:
```suggestion
"@salt-ds/countries": minor
```
This is a minor change since it is adding something new

.changeset/modern-weeks-jog.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Sharp country flags
```
joshwooding, line 5: This should be an introduction to the change and preferrably contain a code example.

joshwooding (changes requested):
We need an update to the documentation too

joshwooding (changes requested):
I believe the country symbols are the wrong size, also the sharp ones have an incorrect width/height applied to them cc @pseys

pseys:
> I believe the country symbols are the wrong size, also the sharp ones have an incorrect width/height applied to them cc @pseys

Yes, the intention was that the height of the sharp country symbol was the same height as the corresponding round symbol. It would mean pushing other elements out horizontally (if applicable), but would not change the vertical height of a design.

tomhazledine:
@pseys @joshwooding 

> > I believe the country symbols are the wrong size, also the sharp ones have an incorrect width/height applied to them cc @pseys
> 
> Yes, the intention was that the height of the sharp country symbol was the same height as the corresponding round symbol. It would mean pushing other elements out horizontally (if applicable), but would not change the vertical height of a design.

Ah, gotcha. Was just following he sizing from Figma... So I'm sure what I'm changing, the end result in Storybook should look like this?

Screenshot 2024-04-09 at 16 56 40

.changeset/modern-weeks-jog.md
tomhazledine: Have fleshed out the changeset. Let me know if we need more detail

tomhazledine:
Have updated to set "sharp" flag heights to match those of the round icons (with widths adjusted by aspect ratio) and fleshed out the changeset.

(rebased and force-pushed to squash messy commit history)

joshwooding:
> > We need an update to the documentation too
> 
> @joshwooding The site pulls in all the county icons automatically (unlike the regular icons section). Is there any other part of docs that needs updating?
> 
> Otherwise think everything's ready for merging now 👍

We should have documentation on the site showing sharp country symbols are how to use them

tomhazledine:
> > > We need an update to the documentation too
> > 
> > 
> > @joshwooding The site pulls in all the county icons automatically (unlike the regular icons section). Is there any other part of docs that needs updating?
> > Otherwise think everything's ready for merging now 👍
> 
> We should have documentation on the site showing sharp country symbols are how to use them

@joshwooding Assumed (incorrectly) that the stories were pulled through to the site automatically. Have added a sharp example and updated docs.

packages/countries/scripts/generateCountrySymbol.mjs
```diff
@@ -251,20 +291,42 @@ const generateCountrySymbolComponents = ({
...
+      borderRadius: "50%",
+      aspectRatioMultiplier: "1",
+    });
+    const sharpFileContents = Mustache.render(template, {
+      svgElements: sharpSvgPaths.data,
+      componentName: countryCodeToComponentName(countryCode) + "_Sharp",
+      ariaLabel: countryName,
+      viewBox: sharpViewBox ?? "0 0 72 50",
+      borderRadius: "0",
+      aspectRatioMultiplier: "1.44",
```
joshwooding, line 303: Is there a way to make this a class rather than it being added like this?

tomhazledine: Yeah, could go the class route if you think that's better?

Was just following the precedent of `packages/countries/src/country-symbol/CountrySymbol.tsx` (which sets the `--saltCountrySymbol-size-multiplier` in the same manner)

tomhazledine: Made that change 👍 Moved these styles to the stylesheet keyed by a `saltSharpCountrySymbol` class

joshwooding: let's chat about this name, I I'm not sure if it should be saltCountrySymbol-sharp instead to fit with every other class

tomhazledine: Makes sense, and consistency always gets my vote 👍

joshwooding (approved):

Fercas123 (approved):

PR #3234, opened 2024-04-10
Date Picker

Fercas123:


origami-z:
Some feedback after using the date picker in storybook
1. when typing a valid date in input, should calendar be moved to that month?
1. when using mouse to open the calendar from the icon, should input continue to show in focus ring (compare to combobox)
1. when using keyboard, focus using keyboard, "Enter", focus ring is lost?
1. when using keyboard, focus using keyboard, is there any other way to open the calendar, apart from focus to next icon?
1. is there way to custom default date parsing (e.g. "01 Apr" resolves to this year instead of 2001)
1. when using mouse to open calendar, press "Esc", focus is lost (e.g. pressing number is not entering input)

API question: What's the plan of passing props to Calendar? e.g., disable days, customize year dropdown, change first day of week

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -112,37 +147,16 @@ export const DateInput = forwardRef<HTMLDivElement, DateInputProps>(
...
-      state: "value",
-    });
-
-    const getDateValidationStatus = (value: string) =>
-      isInvalidDate(value) ? "error" : undefined;
+    const isDisabled = disabled || formFieldDisabled;
 
-    const [dateStatus, setDateStatus] = useState<"error" | undefined>(
-      getDateValidationStatus(value as string)
+    const dateStatus = useMemo(
```
joshwooding, line 153: Do we need useMemo here?

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+      useFloatingUI({
+        open: open,
+        onOpenChange: setOpen,
+        placement: "bottom-start",
+        middleware: [
+          size({
+            apply({ rects, elements, availableHeight }) {
+              Object.assign(elements.floating.style, {
+                minWidth: `${rects.reference.width}px`,
+                maxHeight: `max(calc((var(--salt-size-base) + var(--salt-spacing-100)) * 5), calc(${availableHeight}px - var(--salt-spacing-100)))`,
```
joshwooding, line 82: We can simplify this, this was for the items in a list but since DatePicker doesn't have them it's fine. We can probably just remove this for now.

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+      controlled: startDateProp,
+      default: defaultStartDate,
+      name: "StartDate",
+      state: "value",
+    });
+    const [endDate, setEndDate] = useControlled({
+      controlled: endDateProp,
+      default: defaultEndDate,
+      name: "EndDateInput",
+      state: "value",
```
joshwooding, line 69: This will need updating

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+      endDate: endDateProp,
+      defaultStartDate,
+      defaultEndDate,
+      dateFormatter,
+      CalendarProps,
+      ...rest
+    },
+    ref
+  ) {
+    const [open, setOpen] = useState<boolean>(false);
```
joshwooding, line 57: More of a comment, do you think we need to make this controlled?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  describe("WHEN single datepicker is mounted", () => {
+    it("THEN it should mount with the specified defaultStartDate", () => {
+      cy.mount(<Default defaultStartDate={testDate} />);
+      cy.findByRole("textbox").should("have.value", "01 Feb 2000");
+    });
+    it("THEN should format a valid date with a different format on blur", () => {
+      const onChange = (event: ChangeEvent<HTMLInputElement>) => {
+        // React 16 backwards compatibility
+        event.persist();
```
joshwooding, line 22: Are we going to check a value in these handlers?

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+          endInputRef={endInputRef}
+          placeholder={placeholder}
+          dateFormatter={dateFormatter}
+          endAdornment={
+            <Button
+              variant="secondary"
+              onClick={handleCalendarButton}
+              disabled={disabled || formFieldReadOnly || formFieldDisabled}
+            >
+              <CalendarIcon />
```
joshwooding, line 153:
```suggestion
              <CalendarIcon aria-hidden />
```
We need to follow Icon button guidance

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+          {...getReferenceProps()}
+          startInputRef={startInputRef}
+          endInputRef={endInputRef}
+          placeholder={placeholder}
+          dateFormatter={dateFormatter}
+          endAdornment={
+            <Button
+              variant="secondary"
+              onClick={handleCalendarButton}
+              disabled={disabled || formFieldReadOnly || formFieldDisabled}
```
joshwooding, line 151: This will need an aria-label with something like Open Calendar as the label

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,167 @@
...
+      defaultStartDate,
+      selectionVariant,
+      context,
+      getPanelPosition,
+    };
+
+    return (
+      <DatePickerContext.Provider value={datePickerContextValue}>
+        <DateInput
+          className={clsx(withBaseName())}
```
joshwooding, line 140: className needs to be pulled out of the ...rest spread and merged

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -101,6 +119,23 @@ export const DateInput = forwardRef<HTMLDivElement, DateInputProps>(
...
+      selectionVariant,
+      setOpen,
+    } = useDatePickerContext();
+
+    const [focused, setFocused] = useState(false);
+    const [startDateStringValue, setStartDateStringValue] = useState<string>(
+      startDate?.toString() ?? ""
+    );
+    const [endDateStringValue, setEndDateStringValue] = useState<string>(
+      endDate?.toString ?? ""
```
joshwooding, line 136:
```suggestion
      endDate?.toString() ?? ""
```

ivan-calderon:
@Fercas123 & @joshwooding I have a question. When using the **mouse** and clicking the calendar button, shouldn't the focus ring be visible on the field on active state. Something like [the spectrum date picker](https://react-spectrum.adobe.com/react-spectrum/DatePicker.html)? 
I have also added an image for reference.

ivan-calderon:
@Fercas123 a few comments below: 

- There is a difference in the date format. For instance, the placeholder format is "dd mmm yyyy". When typing/selecting a date it returns this format "mmm dd, yyyy". 
- Keyboard navigation has still some challenges tbd. So far, keyboard navigation is not aligned to this [focus sequence](https://www.figma.com/file/HLw3o2SkRUrcrN0z7kcki8/UITK-Date-Picker-2.0?type=design&node-id=2372%3A369441&mode=design&t=Yq0LcWprbzGT6tf6-1). Let me know, if we need a quick call to chat about it.

origami-z:
- [Single select] Default closed state, tab to calendar icon, enter, focus move back to input (not to calendar?)
  - With above, tab x2, focus first land in day (no longer left month arrow?)
- [Single select] When some day is selected in the month, should tab move the focus to selected day (currently default to 1st day)?
- [Single select] default placeholder is "dd mmm yyyy", when i select some day from calendar, it's filled with different format "Apr 09, 2024" ?
- [Single select] when calendar is open, change input to a valid date, e.g. Apr 09, 2024 -> Apr 08, 2024, selected/active day in the calendar is not updated?
- design question, should today indicator/underline change color when it's in the previous month (out of range)?
- Out of range, should them be balanced? e.g. Feb 2015, have 1 week from prev month & 1 week from next month?
- [Range picker] focus ring only appear when focused on the start date input, not end date input?
- [Range picker] Similar to single one, when open calendar via keyboard, tab x2, focus lands on 2nd calendar's 1st day? instead of left arrow?
- [Range picker] mouse, open calendar, click right arrow of the 2nd calendar a couple of times (e.g., until May 2025), select 20 May 2025 left month is moved to the same month (expected i guess), right arrow of left calendar is disabled with "Future dates are out of range". Not sure why only left one is disabled ?
  - at this point, both calendar showing 20 May 2025 as selected date, which is the same in start date input, but i can't select any end date anymore
 - [Range picker] keyboard, when use keyboard to select a day from the calendar, focus is return to the input after selecting the first date, i'd assume the focus need to stay in calendar, so the end date can be selected using keyboard as well

Fercas123:
@origami-z I'll tackle range in an upcoming PR, for the single select: 

[fixed] Default closed state, tab to calendar icon, enter, focus move back to input (not to calendar?)
[fixed] default placeholder is "dd mmm yyyy", when i select some day from calendar, it's filled with different format "Apr 09, 2024" ?
[expected behaviour, change registered on blur/enter] when calendar is open, change input to a valid date, e.g. Apr 09, 2024 -> Apr 08, 2024, selected/active day in the calendar is not updated?
[expected behaviour, bug below] With above, tab x2, focus first land in day (no longer left month arrow?)
[#3340] When some day is selected in the month, should tab move the focus to selected day (currently default to 1st day)?
[#3341] design question, should today indicator/underline change color when it's in the previous month (out of range)?
[TBC] Out of range, should them be balanced? e.g. Feb 2015, have 1 week from prev month & 1 week from next month?
[Fixing] Blur/Enter should not delete the input when the date is incorrect

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -5,52 +5,76 @@ import {
...
-  const date = new Date(input);
+function getCalendarDate(inputDate: string) {
+  const date = createDate(inputDate);
+  if (!date) return undefined;
+  const isoDate = parseAbsoluteToLocal(date?.toISOString());
+  return new CalendarDate(isoDate.year, isoDate.month, isoDate.day);
+}
+
+const getDateValidationStatus = (value: string | undefined) =>
+  value && isInvalidDate(value) ? "error" : undefined;
```
lukeac123, line 50: why do we set to 'undefined' here if date is not valid, in the try / catch example we set to 'null' - there a reason for the difference ?

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -153,30 +181,78 @@ export const DateInput = forwardRef<HTMLDivElement, DateInputProps>(
       ...restDateInputProps
     } = inputProps;
 
+    // Effects. Update date strings when dates change
+    useEffect(() => {
```
lukeac123, line 185: would useCallback work here instead ?

packages/lab/src/date-picker/DatePickerPanel.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+const withBaseName = makePrefixer("saltDatePickerPanel");
+
+export const DatePickerPanel = forwardRef<HTMLDivElement, DatePickerPanelProps>(
+  function DatePickerPanel(props, ref) {
+    const { className, onSelect, CalendarProps, ...rest } = props;
+
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-date-input-panel",
```
lukeac123, line 42:
```suggestion
      testId: "salt-date-picker-panel",
```

packages/lab/src/date-input/DateInput.tsx
Fercas123: just the intention of each value, 
the null is set on the date where we want to indicate explicitly the date value is empty
this is the validationStatus that indicates if we need to add a error state to the input, if there is no "error" state, i treated it as "an error state hasn't been set" 
not sure if the explanation makes sense XD

packages/lab/src/date-input/DateInput.tsx
Fercas123: because `startDateStingValue` needs to change on every stroke in the input `onChange`, it can't just follow the calendar value

lukeac123 (approved):
lgtm - although does this need a cs?

site/src/examples/date-picker/Default.tsx
```diff
@@ -0,0 +1,6 @@
+import { ReactElement } from "react";
+import { DatePicker } from "@salt-ds/lab";
+
+export const Default = (): ReactElement => (
+  <DatePicker style={{ width: "256px" }} />
```
JoanaMMoreira, line 5: I wonder if we should try to avoid adding inline styles directly to the components in our examples, if we're doing it for others then I guess it's fine

Fercas123: looks like it is standard in other examples so it doesn't take 100%w ...  i guess we could also make a wrapper but for now i took the same size as the input examples

joshwooding:
Accessibility Notes:
- We need an aria-label on the calendar popover
- We should change the aria-label on the calendar day to be a formatted date
- We need to add event preventDefault to prevent scrolling when navigating in the calendar

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+    year: "numeric",
+  });
+};
+
+describe("GIVEN a DatePicker", () => {
+  checkAccessibility(composedStories);
+
+  describe("WHEN single datepicker is mounted", () => {
+    it("THEN it should mount with the specified defaultStartDate", () => {
+      cy.mount(<Default defaultStartDate={testDate} />);
```
origami-z, line 46: `defaultStartDate` feels uncomfortable to me as a single date picker's api. Is there a need to split 2 components?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findByRole("textbox").should("have.value", formatInput(testDate));
+    });
+    it("THEN should format a valid date with a different format on blur", () => {
+      const onChange = (event: ChangeEvent<HTMLInputElement>) => {
```
origami-z, line 50: Don't think correct `onChange` callback is tested anywhere?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      };
+      cy.mount(<Default defaultStartDate={testDate} onChange={onChange} />);
+      cy.findByRole("textbox").click().clear().type(testInput);
+      cy.findByRole("textbox").blur();
+      cy.findByRole("textbox").should(
+        "have.value",
+        formatInput(testDate.add({ days: 1 }))
+      );
+    });
+    it("THEN should error and not format invalid dates on blur", () => {
```
origami-z, line 62: The part of "show error" isn't tested?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findByRole("button", { name: "Open Calendar" }).realClick();
+      cy.findByRole("button", {
+        name: formatDay(testDate.add({ days: 11 })),
+      }).should("exist");
+      cy.findByRole("button", {
+        name: formatDay(testDate.add({ days: 11 })),
+      }).realClick();
+      cy.findByRole("application").should("not.exist");
+    });
+    it("THEN should not open the calendar when disabled", () => {
```
origami-z, line 84: Description doesn't really match the test? This feels more like "open button is disabled"

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findByRole("textbox").click().clear();
+      cy.findByRole("textbox").blur();
+      cy.findByRole("button", { name: "Open Calendar" }).realClick();
+      cy.findByRole("button", { name: formatDay(testDate) }).should(
+        "not.have.attr",
+        "aria-pressed",
+        "true"
+      );
+    });
+  });
```
origami-z, line 114: Is `readonly` tested?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findByRole("button", {
+        name: formatDay(testDate.add({ days: 11 })),
+      }).realClick();
+      cy.findByRole("application").should("not.exist");
+    });
+    it("THEN should not open the calendar when disabled", () => {
+      cy.mount(<Default defaultStartDate={testDate} disabled />);
+      cy.findByRole("button").should("be.disabled");
+    });
+    it("THEN it should update the selected month when changing selected date", () => {
```
origami-z, line 88: The test is not updating date or month? Should we test controlled date update?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findByRole("button", {
+        name: "Next Month",
+      }).focus();
+    });
+    it("THEN it should clear the date if an empty input is submitted", () => {
+      cy.mount(
+        <Default
+          defaultStartDate={testDate}
+          // @ts-ignore
+          CalendarProps={{ visibleMonth: testDate }}
```
origami-z, line 102: Is this a valid use case for our user? if so, `ts-ignore` may need to be fixed?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      );
+      cy.findAllByRole("textbox")
+        .eq(0)
+        .should("have.value", formatInput(testDate));
+      cy.findAllByRole("textbox")
+        .eq(1)
+        .should("have.value", formatInput(testDate.add({ months: 1 })));
+    });
+
+    it("THEN it should allow selecting start and end date trough input", () => {
```
origami-z, line 131: "through"

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.mount(<Range />);
+      cy.findAllByRole("textbox").eq(0).clear().click().type(testInput);
```
origami-z, line 133: This makes me wonder whether label on 2 inputs should be different, indicating one is start and one is end?

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.findAllByRole("textbox").eq(1).click().type(rangeTestInput);
+      cy.findByRole("button", { name: "Open Calendar" }).realClick();
+      cy.findAllByRole("button", {
+        pressed: true,
+      }).should("have.length", 30);
+    });
+    it("THEN it should error if on of the dates has the wrong format", () => {
```
origami-z, line 140: Not checking "error"

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -0,0 +1,205 @@
...
+      cy.mount(<Range />);
+      cy.findAllByRole("textbox").eq(0).click().type("date");
+      cy.findAllByRole("textbox").eq(0).blur();
+      cy.findAllByRole("textbox").eq(0).should("have.value", "date");
+    });
+    it("THEN it should not close the calendar when a start date is selected", () => {
```
origami-z, line 146: Should the test cover start date input is updated as well?

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -192,27 +308,70 @@ export const DateInput = forwardRef<HTMLDivElement, DateInputProps>(
...
+              onFocus={!isDisabled ? handleFocus : undefined}
+              placeholder={placeholder}
+              size={placeholder.length}
+              value={
+                isReadOnly && !endDateStringValue
+                  ? emptyReadOnlyMarker
+                  : endDateStringValue
+              }
+              {...restDateInputProps}
+              required={isRequired}
```
origami-z, line 363: A little random that `required` is after `rest`. If this is intentional, add some comment.

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,224 @@
...
+      dateFormatter,
+      CalendarProps,
+      className,
+      open: openProp,
+      onOpenChange: onOpenChangeProp,
+      ...rest
+    },
+    ref
+  ) {
+    const [open, setOpen] = useControlled({
```
origami-z, line 94: `open` is not covered in test?

packages/lab/src/date-picker/DatePicker.tsx
```diff
@@ -0,0 +1,224 @@
...
+      controlled: openProp,
+      default: false,
+      name: "openPanel",
+      state: "openPanel",
+    });
+    const [startDate, setStartDate] = useControlled({
```
origami-z, line 100: Controlled dates is not tested?

site/src/examples/date-picker/Range.tsx
```diff
@@ -0,0 +1,6 @@
+import { ReactElement } from "react";
+import { DatePicker } from "@salt-ds/lab";
+
+export const Range = (): ReactElement => (
+  <DatePicker selectionVariant="range" style={{ width: "250px" }} />
```
origami-z, line 5: This example's year dropdown is truncated?

origami-z:
Any QA stories needed?

origami-z (dismissed):
Tick, assuming notes to be addressed later. 
This would allow users to start using them.

site/docs/components/date-picker/examples.mdx
```diff
@@ -0,0 +1,38 @@
...
+The Date Range Picker enables selection of start and end dates, using the input fields or two calendars displayed. The range is only committed when the user clicks ‘Apply’, and can be reset if the user wants to start again.
+
+</LivePreview>
+<LivePreview componentName="date-picker" exampleName="RangeWithFormField">
+## Date range with form field
+
+This Date Picker has a form field label and is configured with validation logic. If an invalid date is entered, the error is detected and a tooltip advises the user of the correct format.
+
+</LivePreview>
+</LivePreviewControls>
```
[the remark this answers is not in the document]
Fercas123: noted

site/docs/components/date-picker/examples.mdx
```diff
@@ -0,0 +1,38 @@
...
+## Default
+
+The default Date Picker allows the user to pick a single date, which is committed when it has been typed into the field or selected from the calendar.
+
+</LivePreview>
+<LivePreview componentName="date-picker" exampleName="WithFormField">
+
+## With form field
+
+This Date Picker has a form field label and is configured with validation logic. If an invalid date is entered, the error is detected and a tooltip advises the user of the correct format.
```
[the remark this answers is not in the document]
Fercas123: noted

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
Fercas123: makes sense, it probably should follow calendar usage of selectedDate, noted

site/src/examples/date-picker/Range.tsx
Fercas123: i think that is a problem with how the site is rendering fonts, it happens in calendar too

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
Fercas123: 📝

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
Fercas123: 📝

packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
Fercas123: 📝

packages/lab/src/date-picker/DatePicker.tsx
Fercas123: 📝

packages/lab/src/date-picker/DatePicker.tsx
Fercas123: 📝

origami-z (approved):

site/docs/components/date-picker/usage.mdx
```diff
@@ -0,0 +1,31 @@
...
+
+## Using the component
+
+### When to use
+
+Use a Date Picker when enabling users to filter data that’s displayed on screen, or whenever a specific timeframe or period needs to be selected.
+
+### Best practices
+
+To improve consistency and simplify your messaging, consider recommending just one acceptable date format in the label and/or tooltip – ideally, the one most frequently used in your application.
```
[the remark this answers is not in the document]
Fercas123: the new approach will be to show the error in the panel header instead of tooltip, so will update this

https://www.figma.com/design/iVb153RCmU5R7uvZ3ySPoR/Date-Picker?node-id=1958%3A77880&t=cSQYF2zmLzvezqKK-1

ivan-calderon (approved):

PR #3240, opened 2024-04-11
Parent child layout to core

lukeac123:


packages/core/src/utils/useIsViewportLargerThanBreakpoint.ts
lilyvc, line 1: referencing core?

lukeac123: yup, changing now - ty

.changeset/rare-carrots-pretend.md
```diff
@@ -0,0 +1,16 @@
+---
+"@salt-ds/core": minor
+---
+
+Promoted `ParentChildLayout` from labs to core.
+Promoted `useIsViewportLargerThanBreakpoint` from labs to core.
```
joshwooding, line 6: Do we really want this as a public hook?

packages/core/src/parent-child-layout/useIsViewportLargerThanBreakpoint.ts
lukeac123, line 1: Copied this to core so that it can be used locally within parent child layout and not exported to the user.
Still used in the LayerLayout so couldn't delete from labs

joshwooding (approved):

DavieReid (approved):

PR #3134, opened 2024-03-18
Column template and auto row height specification added to `GridLayout`.

brianblakely:
Implements proposals in #3037.

PR #3242, opened 2024-04-11
Enhance responsive props

joshwooding:
Alternative approach to #3134


libertymayc:
looks good, needs changeset

Issue #3167, opened 2024-03-22
Calendar's spacing is broken on the site.


joshwooding:
### Package name(s)

Lab (@salt-ds/lab), Website / Documentation

### Description

Due to box-sizing border-box being set on the site the Calendar spacing is broken.

![image](https://github.com/jpmorganchase/salt-ds/assets/12938082/22b9317c-837c-469d-9c3d-18bbe0a24042)

### Steps to reproduce

Visit https://saltdesignsystem-git-clean-site-sidebar-fed-team.vercel.app/salt/components/calendar/examples

### Expected behavior

![image](https://github.com/jpmorganchase/salt-ds/assets/12938082/095a0fc9-f457-428e-b014-da89eb0ec9a0)

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [ ] I am an employee of JPMorgan Chase & Co.

PR #3249, opened 2024-04-15
Fix Calendar spacing is broken

jitendradav16:
Class Name was raised this issue of Lab css everywhere, I have changed class name for specific use only. No change in core component.


joshwooding:
Hi, thanks for the PR.

Could you provide more explanation around this change? This wouldn't affect #3167 as it uses a different Dropdown and "saltDropdown" is the correct class to use in this case.

PR #3256, opened 2024-04-16
Remove Data Grid alpha and export more internals

origami-z:
We need to exist alpha so existing user with "^" dependencies (but not specifically using `alpha` versions) can use up to date / matching version of `@salt-ds/styles` with rest of packages. Otherwise, it needs to be overridden to be `@salt-ds/styles: ^0.2.1`

Export some components to make it possible to produce custom cell editors (renamed internal `Cell` to `CellFrame`)


joshwooding (changes requested):
The patchfile also will need to be updated

joshwooding (dismissed):

Fercas123 (approved):

joshwooding (approved):

Issue #3279, opened 2024-04-23
Allow passing more props to NavigationItem's link element


manelmadeira:
### Area

UI Components

### The problem

We are using `@tanstack/react-router` to manage our app's routing and we want to be able to use Salt's `NavigationItem` with their `Link` attributes. 

They export a hook `useLinkProps` ([docs](https://tanstack.com/router/latest/docs/framework/react/api/router/useLinkPropsHook)) that allows us to send all the navigation related props to a custom anchor tag. 

We tried using it with `NavigationItem` like this

```jsx
const linkProps = useLinkProps({ to: '/apps/$appSlug', params: { appSlug: 'my-app' } });

return (
  <NavigationItem href={linkProps.href}>My Link</NavigationItem>
)
```

but that's not enough for `@tanstack/react-router` to navigate to the new route. After looking into their hook's [source code](https://github.com/TanStack/router/blob/main/packages/react-router/src/link.tsx#L715-L738), I realised we need to pass more than just `href` to the anchor tag for routing to work. 

### The solution

Is it possible to add an additional prop to `NavigationItem` to allow this (eg. `linkProps`)? 

This would allow to do something like:

```jsx
const linkProps = useLinkProps({ to: '/apps/$appSlug', params: { appSlug: 'my-app' } });

return (
  <NavigationItem linkProps={linkProps}>My Link</NavigationItem>
)
```

### Alternatives and examples

Alternative we could allow passing a `ReactElement` to replace the anchor element rendered by `NavigationItem` and add styling to it. 

Chakra UI allows to pass other elements to be rendered instead using an `as` property. https://v2.chakra-ui.com/community/recipes/as-prop

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

manelmadeira:
> We have similar issues with Next JS apps and `next/link`. We are going to provisionally pull this into our next sprint, at least for discussion and decide on the final approach.

I'd imagine that the approach used for this would be the same for the Lab's `BreadcrumbItem` component, right?

mark-tate:
to be released in next release

joshwooding:
Closed by #3709

PR #3280, opened 2024-04-23
Add linkProps prop to NavigationItem component

manelmadeira:
Solves #3279 .


mark-tate:
how about this ?
https://github.com/jpmorganchase/salt-ds/pull/3417

PR #3283, opened 2024-04-23
fix tooltip logic to display the number 0

lukeac123:


origami-z (commented):
Good to have a test for regression?

.changeset/violet-dingos-hammer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Adjust `Tooltip` logic return the number 0 as truthy and display in the tooltip
```
joshwooding, line 5:
```suggestion
Fixed `Tooltip` not displaying when the `content` prop value is falsy e.g. 0
```

joshwooding (changes requested):
See comments

lukeac123:
Added a cypress test to makes sure if the content is falsy then the tooltip is still shown

packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
```diff
@@ -200,6 +200,13 @@ describe("GIVEN a Tooltip", () => {
     });
   });
 
+  describe("WHEN content is falsy", () => {
+    it("then tooltip should still display", () => {
+      cy.mount(<Open content={0} />);
+      cy.findByRole("tooltip").should("exist");
```
joshwooding, line 206: We should check that 0 is displayed visually.

Also haven't checked but our tests need to cover:
- Falsy content
- Undefined content
- Empty string content
to make sure everything is covered

joshwooding (changes requested):

packages/core/src/__tests__/__e2e__/tooltip/Tooltip.cy.tsx
```diff
@@ -213,6 +213,20 @@ describe("GIVEN a Tooltip", () => {
     });
   });
 
+  describe("WHEN content is undefined", () => {
+    it("then tooltip doesn't display", () => {
```
joshwooding, line 217: Can you also test "" please

lukeac123: This test already exists - test above, line 208

joshwooding (approved):

PR #3299, opened 2024-04-24
Navigation pattern example

lilyvc:
Adds an example of Navigation layout to storybook

### DIFF
--- a/packages/core/stories/patterns/navigation/navigation.stories.tsx
+++ b/packages/core/stories/patterns/navigation/navigation.stories.tsx
@@ -1,0 +1,210 @@
+import { useEffect, useState } from "react";
+import {
+  Button,
+  FlexItem,
+  FlexLayout,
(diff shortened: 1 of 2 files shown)


origami-z (commented):
Need something to show best practice of small viewport? 
This could be a good place to think how to reuse other pattern code

lilyvc:
> Need something to show best practice of small viewport? This could be a good place to think how to reuse other pattern code

There's no guidance on the pattern page about small viewport, so I'm actually not sure of the approach if you have a sidebar and vertical nav and it's mobile. This could be follow up work, this is more just to catch up the examples to the patterns page.

origami-z:
> This could be follow up work, this is more just to catch up the examples to the patterns page.

OK, created #3344

lukeac123 (approved):

origami-z (approved):

Issue #2743, opened 2023-11-17
Lint CSS to make sure components not using deprecated tokens


origami-z:
Also, we should have lint rules to not allowing our own components using [deprecated tokens](https://github.com/jpmorganchase/salt-ds/tree/main/packages/theme/css/deprecated), so that repo-wide characteristics renaming (e.g. #2530) will be safer.

Also, we have [existing stylelint rule](https://github.com/jpmorganchase/salt-ds/blob/main/tooling/stylelint/correct-theme-token-usage/index.js) to use correct characteristics names, we should update it to not allow removed characteristics.

```[tasklist]
### Tasks
- [x] Implement lint rule with warning #3301
- [x] Dradaully remove deprecated tokens in core package
- [x] Make lint rule an error for core package, warning for lab
```

mark-tate:
Related to #1107

origami-z:
> Related to #1107

This isn't a breaking change, deprecated tokens shouldn't be used in our components

joshwooding:
Was closed by #4619

PR #3300, opened 2024-04-24
Fix usage of deprecated tokens in core and data-grid

libertymayc:
Noticed in Card that size-accent was being used which is deprecated; checked spec and it should be size-bar

This was the corresponding replacement noted in changelog - I changed all size-accent used in core components/data-grid to use size-bar

Also went ahead and changed size-unit (deprecated) to spacing-100 or corresponding multiplier


origami-z:
Great! Down to 14 tokens in core
#2743

origami-z (approved):

joshwooding (approved):

PR #3337, opened 2024-04-29
Fix disabled tooltips not being fully disabled

joshwooding:


Fercas123 (approved):

packages/core/src/tooltip/useTooltip.ts
```diff
@@ -85,12 +90,12 @@ export function useTooltip(props?: UseTooltipProps) {
         open: enterDelay,
         close: leaveDelay,
       },
-      enabled: !disableHoverListener,
+      enabled: !disableHoverListener || !disabled,
       handleClose: safePolygon(),
     }),
-    useFocus(context, { enabled: !disableFocusListener }),
+    useFocus(context, { enabled: !disableFocusListener || !disabled }),
```
origami-z, line 96: This logic may not be right? 

If either is true, `enabled` should be false?

origami-z (approved):

Fercas123 (approved):

PR #3349, opened 2024-05-01
Fix CSS Icons

joshwooding:
Turns out re-exporting all the icons had a downside... The plugin strips the xmlns info which is needed for the CSS icons.

This would have been caught but the qa story didn't have chromatic turned on.

I've improved the story by showing all icons and making sure that chromatic captures it.

I also made sure that saltIcons.css is in the package.json so it gets published.


lukeac123 (approved):

lilyvc (approved):

packages/countries/saltCountries.css
origami-z, line 1: Why there a change on this file? no build script change on countries package?

origami-z:
Plugin has a bug, it striped `xmlns` by mistake while trying to remove `fill`. I will update

packages/countries/saltCountries.css
joshwooding: The build script for icons and country symbols doesn't ignore newlines. So you get different build outputs on Windows. I've made a fix that strips all new lines when encoding the icon.

joshwooding:
Need to look at the different in the status indicator

pseys (approved):

origami-z (approved):

PR #3296, opened 2024-04-24
Refactor Slider

lukeac123:
Updates `Slider` component to new designs and behaviours.

Highlights:

### Integration with Input
Screenshot 2024-07-25 at 11 39 30

### Support for custom step amount
Screenshot 2024-07-25 at 11 39 36

### Full "range" support
Screenshot 2024-07-25 at 11 39 55

PR #3353, opened 2024-05-02
Slider Cypress / QA Tests

lukeac123:


joshwooding:
@lukeac123 Should this just be merged into #3296? Tests arguably should be delivered with the code

PR #3371, opened 2024-05-07
Add examples for options with descriptions for combobox and dropdown

ThusharaJ07:


packages/core/stories/combo-box/combo-box.stories.tsx
```diff
@@ -653,6 +655,54 @@ MultiplePillsTruncated.args = {
...
+      onChange={handleChange}
+      onSelectionChange={handleSelectionChange}
+      value={value}
+      valueToString={(person) => person.displayName}
+    >
+      {options.map((person) => (
+        <Option key={person.id} value={person}>
+          <StackLayout gap={0.25} align="start">
+            <Text variant="primary">{person.firstName}</Text>
+            <Label variant="secondary">{person.lastName}</Label>
```
joshwooding, line 698: Label isn't semantically correct here we could probably just use:

```suggestion
            <Text variant="secondary">{person.lastName}</Text>
```

But we should also use the color prop

packages/core/stories/combo-box/combo-box.stories.tsx
```diff
@@ -653,6 +655,54 @@ MultiplePillsTruncated.args = {
   truncate: true,
 };
 
+export const SecondaryLabel: StoryFn<ComboBoxProps<Person>> = (args) => {
```
joshwooding, line 658: The Storybook and site example should be the same

origami-z:
In #3431, gap between labels are `0.5`, maybe this should be updated as well?

ThusharaJ07:
> In #3431, gap between labels are `0.5`, maybe this should be updated as well?

updated the gap to 0.5

origami-z (dismissed):

joshwooding (approved):

joshwooding (approved):

shey-v (approved):

PR #3379, opened 2024-05-08
added free solo changes

ThusharaJ07:


joshwooding:
Will be replaced by individual features to CB.

PR #3382, opened 2024-05-08
Update server side CB example

joshwooding:


site/src/examples/combo-box/ServerSideData.tsx
```diff
@@ -42,25 +49,42 @@ export const ServerSideData = (): ReactElement => {
...
+        {!loading && hasResults
+          ? data?.map((state) => <Option value={state} key={state} />)
+          : null}
+        {!loading && !hasResults ? (
+          <div
+            className={styles.statusOption}
+            role="option"
+            aria-selected="false"
+          >
+            No results found
```
origami-z, line 72: Thinking about this around "min 2 char" message, should we show something different than no result when user first started typing?

joshwooding: Ah, I thought I handled that case :(

origami-z (approved):
It's better than before.

Not so sure whether showing empty list is the best solution, it feels something is rendered (arrow flipped, list shadow, some border) and also not.

joshwooding:
> It's better than before.
> 
> Not so sure whether showing empty list is the best solution, it feels something is rendered (arrow flipped, list shadow, some border) and also not.

I think that's more a bug in CB, I need to talk to @bhoppers2008 but I'm thinking CB shouldn't be in an open state when no items are in the list

shey-v (approved):

PR #3399, opened 2024-05-10
combobox select on blur

ThusharaJ07:


packages/core/src/combo-box/ComboBox.tsx
```diff
@@ -248,7 +249,7 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
 
         break;
       case "Tab":
-        if (!multiselect && activeState) {
+        if (tabToSelect && activeState) {
```
mark-tate, line 252: this is a breaking change as the single select menu used to select on TAB and now it only will, if you provide this prop

ThusharaJ07: @mark-tate @origami-z  yes, now with this prop only we will be able to select the typed item on tab key press. the behaviour is kept common for both single/multiselect combobox for consistency.
Let me know if we want to only apply this for multiselect only.

origami-z: From meeting: let's try default to different values in single / multi select

ThusharaJ07: Added the below change to not break the existing behaviour >>>
the default value for prop will be:
- true for single select
- false for multiselect

origami-z: Makes sense. Let's add some test for both

site/docs/components/combo-box/accessibility.mdx
```diff
@@ -16,6 +16,7 @@ data:
 - If the combo box is single-select and if focus is above the combo box, Tab sets focus on the component. When focused, combo box has a visible text cursor and the list is opened.
 - If the combo box is multi-select and if the focus is above the combo box, Tab sets focus on the pill list if pills are present. If no pills are present then focus is set on the input. 
 - If the list is open, Tab moves focus to the next focusable element in the tab order and selects the current active option.
```
origami-z, line 18: Could combine this with next point?

packages/core/src/combo-box/ComboBox.tsx
ThusharaJ07: @origami-z have added test.

bhoppers2008 (dismissed):

packages/core/src/combo-box/ComboBox.tsx
```diff
@@ -248,7 +249,11 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
 
         break;
       case "Tab":
-        if (!multiselect && activeState) {
+        if (
+          tabToSelect &&
+          activeState &&
+          !selectedState.includes(activeState?.value)
```
origami-z, line 255: This is not captured in test i guess? (not deselect)

ThusharaJ07: @origami-z  have added the test for the same

origami-z (dismissed):

joshwooding:
Moved to draft due to conflicts

site/docs/components/combo-box/accessibility.mdx
```diff
@@ -13,8 +13,8 @@ data:
 <KeyboardControls>
 <KeyboardControl keyOrCombos="Tab">
   
-- If the combo box is single-select and if focus is above the combo box, Tab sets focus on the component. When focused, combo box has a visible text cursor and the list is opened.
-- If the combo box is multi-select and if the focus is above the combo box, Tab sets focus on the pill list if pills are present. If no pills are present then focus is set on the input. 
+- If the combo box is single-select and if focus is above the combo box, Tab sets focus on the component. When focused, combo box has a visible text cursor and the list is opened. After typing a value, if tab is pressed, by default the value gets selected from the list. Pass 'tabToSelect' prop as false to not select a typed value on tab press.
```
joshwooding, line 16: This guidance needs to be updated. Before the sentences we small and easy to follow but we're added a couple of extra sentences to both which makes it harder to follow.

site/docs/components/combo-box/examples.mdx
```diff
@@ -177,4 +177,13 @@ When the `value` prop is an object, the `valueToString` prop must be provided to
 Use the `truncate` prop when you are limited on space and want to prevent the multi-select combobox from spanning multiple lines.
 
 </LivePreview>
+
+<LivePreview componentName="combo-box" exampleName="TabSelect" >
+
+## Tab Select
```
joshwooding, line 183: Selection on Tab ? This example name is very abstract.

joshwooding: Or at least tabToSelect

packages/core/src/list-control/ListControlState.ts
```diff
@@ -51,6 +51,10 @@ export type ListControlProps<Item> = {
    * Callback used to convert an option's `value` to a string. This is needed when the value is different to the display value or the value is not a string.
    */
   valueToString?: (item: Item) => string;
+  /**
+   * If true, options will be selected on tab key press.
+   */
+  tabToSelect?: boolean;
```
joshwooding, line 57: I'm wondering if this name is confusing? Maybe selectionOnTab is better

ThusharaJ07: will change this to selectionOnTab

site/docs/components/combo-box/accessibility.mdx
ThusharaJ07: Hi @joshwooding , I was just reading the previous version of this, and the behaviour is actually same. Do I actually need to add something here for the prop added?

packages/core/src/list-control/ListControlState.ts
joshwooding: Can you double check with design please

ThusharaJ07: changed it to `selectOnTab` as per discussion.

bhoppers2008 (approved):

origami-z (approved):
Last bug fixed and covered in test. Good to go.

site/docs/components/combo-box/examples.mdx
```diff
@@ -178,6 +178,14 @@ Use the `truncate` prop when you are limited on space and want to prevent the mu
 
 </LivePreview>
 
+<LivePreview componentName="combo-box" exampleName="SelectOnTab" >
+
+## Select on Tab
+
+Use the `selectOnTab` prop to select the typed item from the list on Tab key press. This prop can be applied to both multi/single select type of `ComboBox`.
```
joshwooding, line 185:
```suggestion
Use the `selectOnTab` prop to select the typed item from the list on Tab key press. 
```
This guidance is redundant

packages/core/src/list-control/ListControlState.ts
```diff
@@ -51,6 +51,10 @@ export type ListControlProps<Item> = {
    * Callback used to convert an option's `value` to a string. This is needed when the value is different to the display value or the value is not a string.
    */
   valueToString?: (item: Item) => string;
+  /**
+   * If true, options will be selected on tab key press.
+   */
+  selectOnTab?: boolean;
```
joshwooding, line 57: I don't think this is the correct place to add the type? Right now we only have the logic in ComboBox but this will a prop to Dropdown too.

image

joshwooding (changes requested):
Needs some minor updates

joshwooding:
Moving to draft temporarily, can be moved out after #3537

joshwooding (approved):

origami-z (approved):

Issue #3381, opened 2024-05-08
Split buttons created with segmented button group menu's are misaligned


joshwooding:
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

latest

### Description

Menu create's some DOM nodes around its trigger via floating-ui but they end up getting dividers because SBG targets any element. We should make SBG's selector more specific.

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [ ] I am an employee of JPMorgan Chase & Co.

PR #3401, opened 2024-05-10
Fix Segmented Button Group adding dividers to all DOM Nodes

lilyvc:
addresses #3381


site/src/examples/menu/WithIcon.tsx
joshwooding, line 1: This file needs to be reverted, the original demo was correct

packages/core/stories/segmented-button-group/segmented-button-group.stories.tsx
```diff
@@ -60,3 +72,33 @@ export const Icons: StoryFn<SegmentedButtonGroupProps> = () => {
     </StackLayout>
   );
 };
+
+export const WithMenu: StoryFn<SegmentedButtonGroupProps> = () => {
```
joshwooding, line 76: It might be quicker to create a short Split Button pattern example rather than place it here

packages/core/stories/segmented-button-group/segmented-button-group.stories.tsx
```diff
@@ -60,3 +72,33 @@ export const Icons: StoryFn<SegmentedButtonGroupProps> = () => {
...
+  return (
+    <SegmentedButtonGroup>
+      <Button>Button</Button>
+      <Button>Button</Button>
+      <Menu>
```
joshwooding, line 81: Actual feedback about the example, this Menu has the incorrect placement

site/src/examples/menu/WithIcon.tsx
lilyvc: ah my bad, I was testing using the site bc storybook wasn't working :/

joshwooding (approved):

Fercas123 (approved):

PR #3410, opened 2024-05-13
Export saltSharpCountries.css and remove countryCode_Sharp entries from the meta map

joshwooding:
`<countryCode>_Sharp` was accidentally added to the meta map in #3215


packages/countries/saltSharpCountries.css
joshwooding, line 1: Removed the newlines from the svgs

packages/countries/scripts/generateCountrySymbol.mjs
```diff
@@ -419,12 +422,7 @@ const generateCountryMetaMap = ({ countryMetaMap, basePath }) => {
       countryName: "${countryMeta.countryName}",
     },`;
 
-    const sharpEntryText = `"${code}_Sharp": {
```
joshwooding, line 422: Removed _Sharp entries

JoanaMMoreira (approved):

origami-z (approved):

PR #3411, opened 2024-05-14
Button Example of full width button

Vineet119:


origami-z (commented):
I believe Button will be full width when placed in a stack layout. Is there any preference on doing so? 
Worth a discussion with @lilyvc who added the button bar examples

Vineet119:
Agreed, modified to use StackLayout as wrapper

shey-v:
Examples look fine. However, on the site docs, we need to provide more guidance for this example.

Let's update the guidance to say:

'Use full width buttons on mobile devices or smaller viewports, the button will take up the full width of its container.'

site/docs/components/button/examples.mdx
```diff
@@ -71,5 +71,13 @@ Use the disabled state for a button that the user can’t press.
 
 A button with the prop `disabled={true}` will suppress all functionality. If you need to allow the user to place focus on a disabled button, you also need to pass `focusableWhenDisabled={true}`.
 
+</LivePreview>
+
+<LivePreview componentName="button" exampleName="FullWidth" >
+
+## FullWidth
```
origami-z, line 78:
```suggestion
## Full width
```

Vineet119: corrected

shey-v (commented):
'Full width' should be two words in the examples show, if we are displaying three examples and the button labels are displaying as 'primary full width button' 'secondary full width button' then I think the last example should be displayed as 'cta full width button' rather than what it is now. Otherwise, please make them all contextual.

origami-z (dismissed):

Vineet119:
> 'Full width' should be two words in the examples show, if we are displaying three examples and the button labels are displaying as 'primary full width button' 'secondary full width button' then I think the last example should be displayed as 'cta full width button' rather than what it is now. Otherwise, please make them all contextual.

origami-z (approved):

PR #3413, opened 2024-05-14
Fix ComboBox React 16 support

### LINKED ISSUE #3348: ComboBox controlled causes TypeErrors with React 16
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

 1.24.0

### Description

The ComboBox seems to be broken when used with React 16 (I haven't tested with React 18 as we need still to upgrade). Since we support React 16 (and need to for compatibility with UITK), this needs to be fixed.

(Note: reporting this on behalf of an internal consumer, so adding the "community" label)

### Steps to reproduce

1. Setup an app that uses React 16 and include some basic Controlled ComboBox code (the [Custom filtering from our StoryBook](https://www.saltdesignsystem.com/salt/components/combo-box/examples#custom-filtering). I've copied the code below for convenience)
2. Type anything in the ComboBox to filter the displayed options.

The following runtime error will occur:
`×
TypeError
Cannot read properties of null (reading 'value')
    at eval (https://w3k8s5.csb.app/node_modules/
salt-ds/core/dist-es/combo-box/ComboBox.js:266:24
This screen is visible only in development. It will not appear if the app crashes in production.
Open your browser’s developer console to further inspect this error.
This error overlay is powered by `react-error-overlay` used in `create-react-app`.`

Error reproduced in CodeSandBox:
https://codesandbox.io/p/sandbox/checkboxgroup-reproduction-forked-w3k8s5

### Expected behavior

You should be able to filter the available options in the ComboBox without crashing the app

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

joshwooding:
Closes #3348

Moves reading event.target.value outside of the microtask so event is read immediately.

### DIFF
--- a/.changeset/big-pumpkins-complain.md
+++ b/.changeset/big-pumpkins-complain.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed ComboBox logging event errors in React 16.
--- a/packages/core/src/combo-box/ComboBox.tsx
+++ b/packages/core/src/combo-box/ComboBox.tsx
@@ -289,17 +289,17 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
 
     setValueState(value);
 
-    if (value != "") {
-      // Wait for the filter to happen
-      queueMicrotask(() => {
+    // Wait for the filter to happen
+    queueMicrotask(() => {
+      if (value != "") {
         const newOption = getOptionAtIndex(0);
         if (newOption) {
           setActive(newOption);
         }
-      });
-    } else {
-      setActive(undefined);
-    }
+      } else {
+        setActive(undefined);
+      }
+    });
 
     onChange?.(event);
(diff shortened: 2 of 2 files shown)


origami-z (commented):
Is there any regression test needs to be added? Feels some area is not covered by existing test

joshwooding:
> Is there any regression test needs to be added? Feels some area is not covered by existing test

It's already tested but there might be a flaw in how the Cypress tests work in different React versions. Needs investigating

origami-z (approved):

joshwooding:
Blurring the field still caused event errors.

Fercas123 (approved):

origami-z (approved):

PR #3427, opened 2024-05-15
Remove padding in `DialogContent`

JoanaMMoreira:
Preview at https://14c812da.saltdesignsystem-storybook.pages.dev/?path=/story/core-dialog--long-content

### DIFF
--- a/.changeset/bright-dogs-beg.md
+++ b/.changeset/bright-dogs-beg.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix `DialogContent` spacing
--- a/.changeset/bright-dogs-beg.md
+++ b/.changeset/bright-dogs-beg.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fix `DialogContent` spacing
+Remove padding bottom and padding right from `DialogContent`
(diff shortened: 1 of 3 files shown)


joshwooding:
I don't think this is the correct way to solve it. DialogContent should have padding

joshwooding:
Before:
image
image

After:
image
image

Without scrolling looks fine, but once there's a scrollbar it looks a little tight? Might not be an issue though

origami-z:
> Without scrolling looks fine, but once there's a scrollbar it looks a little tight? Might not be an issue though

We probably need to live with this, unless there's a way to add padding when scroll bar appears

JoanaMMoreira:
> > Without scrolling looks fine, but once there's a scrollbar it looks a little tight? Might not be an issue though
> 
> We probably need to live with this, unless there's a way to add padding only when scroll bar appears

I don't think there's a nice way of adding padding for the scroll bar, there are some hacks like adding a transparent thicker border to it but it's not really ideal

origami-z (commented):
Can we update QA stories, so that right spacing change could be captured in Chromatic

.changeset/bright-dogs-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix `DialogContent` spacing
```
origami-z, line 5: Maybe be a little more specific? So when user reads this, they could know where to look for the diff in their own app.

origami-z (approved):

origami-z:
@navkaur76 does any of padding removal impact pref dialog pattern?

joshwooding (approved):

PR #3429, opened 2024-05-15
[WIP] Indeterminate linear progress bar

JoanaMMoreira:


packages/core/src/progress/LinearProgress/LinearProgress.css
```diff
@@ -56,3 +56,29 @@
...
+
+.saltLinearProgress-indeterminate.saltLinearProgress-bar {
+  position: absolute;
+  left: 0px;
+  bottom: 0px;
+  top: 0px;
+  transition: transform 0.2s linear 0s;
+  transform-origin: left center;
+  width: auto;
+  animation: 1.8s linear infinite indeterminate-progress-bar;
```
JoanaMMoreira, line 68: @navkaur76 @bhoppers2008 are we adding a new token for the 1.8 animation speed?

JoanaMMoreira: Closing this as we've agreed to not add the token for now

JoanaMMoreira:
> Do we need a site doc example update?
> 
> Otherwise looks good

I've updated the original example but I'm not sure if it should be separate? Will need some wording around it if that's the case

origami-z:
> I've updated the original example but I'm not sure if it should be separate? Will need some wording around it if that's the case

I think we need some extra design guidance on when to use and avoid. maybe MUI would have something to get us started

joshwooding (approved):

.changeset/four-bugs-reply.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added variant prop to `LinearProgress` component. It defaults to `determinate` and the new `indeterminate` variant will display a moving line to represent an unspecified wait time.
```
joshwooding, line 5: Can you add a code block here showing the new variant prop please

site/src/examples/progress/Linear.tsx
```diff
@@ -1,6 +1,9 @@
 import { ReactElement } from "react";
-import { LinearProgress } from "@salt-ds/core";
+import { LinearProgress, StackLayout } from "@salt-ds/core";
 
 export const Linear = (): ReactElement => (
-  <LinearProgress aria-label="Download" value={38} />
+  <StackLayout>
+    <LinearProgress aria-label="Download" value={38} />
+    <LinearProgress aria-label="Download" variant="indeterminate" />
```
joshwooding, line 7: Can we create a separate example for indeterminate

joshwooding (commented):
Left a few comments. but we should also show going from an indeterminate to a determinate progress bar.

JoanaMMoreira:
> Left a few comments. but we should also show going from an indeterminate to a determinate progress bar.

Do you mean having an example that toggles between those two states?

joshwooding:
> > Left a few comments. but we should also show going from an indeterminate to a determinate progress bar.
> 
> Do you mean having an example that toggles between those two states?

We need an example that shows a realistic use-case.

joshwooding (changes requested):
Just marking it as needs changes to make it clear

origami-z (dismissed):

origami-z (dismissed):

packages/core/src/__tests__/__e2e__/progress/LinearProgress.cy.tsx
```diff
@@ -15,7 +15,14 @@ describe("GIVEN a LinearProgress", () => {
     cy.mount(<Default min={20} max={40} value={35} />);
     cy.findByRole("progressbar").should("have.attr", "aria-valuemax", "40");
     cy.findByRole("progressbar").should("have.attr", "aria-valuemin", "20");
-    cy.findByRole("progressbar").contains("75 %");
+    cy.findByRole("progressbar").contains("75%");
```
origami-z, line 18: Is this a design update? If so, need to mention in changeset.

packages/core/stories/progress/progress.qa.stories.tsx
```diff
@@ -35,6 +35,12 @@ export const ExamplesGrid: StoryFn<QAContainerProps> = (props) => {
         hideLabel
       />
       <CircularProgress aria-label="Download" value={38} hideLabel />
+      <LinearProgress
+        aria-label="Download"
+        className={className}
+        style={{ padding: "50px" }}
+        variant="indeterminate"
```
origami-z, line 42: Will animation makes QA story always changing? Any possibility to override certain CSS to make it stable?

packages/core/src/__tests__/__e2e__/progress/LinearProgress.cy.tsx
JoanaMMoreira: @navkaur76 I think in the designs we didn't have a space between the number and the "%" in the label, should we double check?

packages/core/stories/progress/progress.qa.stories.tsx
JoanaMMoreira: Good point, I'll look at what we did for spinner

packages/core/src/__tests__/__e2e__/progress/LinearProgress.cy.tsx
origami-z: Nav will be OOO this week. Let's revert back and update later if needed.

origami-z (dismissed):

site/src/examples/progress/LinearIndeterminate.tsx
```diff
@@ -0,0 +1,76 @@
...
+      setToastStatus({
+        header: "Upload complete",
+        status: "success",
+        message: "File has successfully been uploaded to shared drive.",
+      });
+    }
+  }, [value]);
+
+  return (
+    <Toast status={toastStatus.status}>
```
JoanaMMoreira, line 55: @joshwooding I've added the Toast example to the site, did you also want to change it?

packages/core/src/progress/LinearProgress/LinearProgress.tsx
```diff
@@ -56,8 +62,11 @@ export const LinearProgress = forwardRef<HTMLDivElement, LinearProgressProps>(
       window: targetWindow,
     });
 
-    const progress = ((value - min) / (max - min)) * 100;
-    const buffer = ((bufferValue - min) / (max - min)) * 100;
+    const isIndeterminate = variant === "indeterminate";
+    const progress = isIndeterminate ? 66 : ((value - min) / (max - min)) * 100;
```
mark-tate, line 66: 66 is a magic number, I have to read through the code to understand what it means, assigning it to a variable would make the value clearer. 
 
 How did we arrive at 66 ?

packages/core/src/progress/LinearProgress/LinearProgress.tsx
```diff
@@ -71,19 +80,22 @@ export const LinearProgress = forwardRef<HTMLDivElement, LinearProgressProps>(
         role="progressbar"
         aria-valuemax={max}
         aria-valuemin={min}
-        aria-valuenow={Math.round(value)}
+        aria-valuenow={isIndeterminate ? undefined : Math.round(value)}
```
mark-tate, line 83: Correct me if I am wrong ?
if I am a determinate progress bar then the  aria-valuenow changes, so I get non-visual feedback of the state change. If I am indeterminate, then there is nothing here that gives us any non-visual feedback ?
The progress bar could be active (indeterminate) but the non-visual feedback would be the same.
The associated label does not use labelledBy so, it has no association.
This feels like an accessibility issue?

site/src/examples/progress/LinearIndeterminate.tsx
mark-tate: Is this a new pattern we are trying to create ?

When I look at this I think the Toast is part of the Progress Bar, where did my Progress bar go ?
Is this confusing to the product or design folks, who do not look at the code, they will say, that the final state of the indeterminate Progress bar replaces itself with something that looks like Toast.

I don't actually see the point of this example, you want to show it going from indeterminate to determinate but after that it's just the same as any other Progress bar example ?

packages/core/src/progress/LinearProgress/LinearProgress.tsx
JoanaMMoreira: This was a value agreed by design, I'll move it into a variable

packages/core/src/progress/LinearProgress/LinearProgress.tsx
JoanaMMoreira: I can change it to match checkbox but it will deviate from what we have in UITK, can I double check we are happy with that @mark-tate ?

site/src/examples/progress/LinearIndeterminate.tsx
JoanaMMoreira: I think we will need to discuss with design cc @navkaur76 @bhoppers2008

Fercas123 (dismissed):

packages/core/stories/progress/progress.qa.stories.tsx
```diff
@@ -35,6 +37,11 @@ export const ExamplesGrid: StoryFn<QAContainerProps> = (props) => {
         hideLabel
       />
       <CircularProgress aria-label="Download" value={38} hideLabel />
+      <LinearProgress
+        aria-label="Download"
+        className="noAnimation"
```
joshwooding, line 42: Is this actually needed? Chromatic should already pause animations https://www.chromatic.com/docs/animations/#css-animations

joshwooding (changes requested):
The API change broke the buffer example https://saltdesignsystem-git-indeterminate-linear-progress-bar-fed-team.vercel.app/salt/components/progress/examples#with-progressing-buffer-value. A quick fix would be indeterminate is if buffer and value is undefined but does this meet our expectations

Fercas123 (approved):

Fercas123 (dismissed):

.changeset/four-bugs-reply.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/core": minor
+---
+
+Updated `LinearProgress` to display a moving line to represent an unspecified wait time, when no `value` is set.
```
joshwooding, line 5:
```suggestion
Updated `LinearProgress` to display a moving line to represent an unspecified wait time, when no `value` is undefined.
```

? Because I'm getting value={undefined} will work

packages/core/src/progress/LinearProgress/LinearProgress.css
```diff
@@ -56,3 +56,29 @@
...
   padding-left: var(--salt-spacing-100);
 }
+
+.saltLinearProgress-indeterminate.saltLinearProgress-bar {
+  position: absolute;
+  left: 0px;
+  bottom: 0px;
+  top: 0px;
+  transition: transform 0.2s linear 0s;
+  transform-origin: left center;
```
joshwooding, line 66: Is this styling needed?

packages/core/src/progress/LinearProgress/LinearProgress.tsx
```diff
@@ -36,15 +36,17 @@ export interface LinearProgressProps extends ComponentPropsWithoutRef<"div"> {
...
 export const LinearProgress = forwardRef<HTMLDivElement, LinearProgressProps>(
   function LinearProgress(
     {
       className,
       hideLabel = false,
       max = 100,
       min = 0,
-      value = 0,
-      bufferValue = 0,
+      value,
```
joshwooding, line 48: Is changing the default values of these props relevant to document?

site/docs/components/progress/examples.mdx
```diff
@@ -26,45 +26,36 @@ Use linear or circular depending on the context, layout and functionality of an
...
 </LivePreview>
+<LivePreview componentName="progress" exampleName="WithBuffer" displayName="With value and buffer">
 
-<LivePreview componentName="progress" exampleName="WithMinVal" displayName="With minimum value">
-
-## With minimum value
+## With buffer
 
-Specify the minimum value of the progress indicator. The percentage progress will be calculated using this value. The default value is 0.
+Specify a buffer value to indicate loading state. The buffer does not have a label and will not affect the progress label.
```
joshwooding, line 48: I'm not sure what "the buffer does not have a label" means?

site/docs/components/progress/examples.mdx
```diff
@@ -26,45 +26,36 @@ Use linear or circular depending on the context, layout and functionality of an
...
-Use the `hideLabel` prop to display the progress without a label.
+Use an indeterminate linear progress to display an unspecified wait time such as a loading state.
 
 </LivePreview>
 
-<LivePreview componentName="progress" exampleName="WithMaxVal" displayName="With maximum value">
+<LivePreview componentName="progress" exampleName="WithMinVal" displayName="Custom min and max">
 
-## With maximum value
+## With minimum value
```
joshwooding, line 39: This is min and max now

site/docs/components/progress/examples.mdx
```diff
@@ -26,45 +26,36 @@ Use linear or circular depending on the context, layout and functionality of an
...
 </LivePreview>
 
-<LivePreview componentName="progress" exampleName="HiddenLabel" displayName="Hide label">
+<LivePreview componentName="progress" exampleName="LinearIndeterminate" displayName="Indeterminate linear progress">
 
-## Hide label
+## Indeterminate linear progress
 
-Use the `hideLabel` prop to display the progress without a label.
+Use an indeterminate linear progress to display an unspecified wait time such as a loading state.
```
joshwooding, line 33: @navkaur76 will this cause confusion between loading e.g. spinner and loading of a determinate progress state?

joshwooding: Is there any more guidance we can give.

packages/core/stories/progress/linear-progress.stories.tsx
joshwooding, line 1: Just a more general comment, is it better to have states in this example:

pending, in progress and completed 
then separately the value.

it might be easier to understand.

joshwooding: So after 4 seconds go from pending to in progress

site/docs/components/progress/examples.mdx
navkaur76: Based on the duration foundation, an indeterminate linear progress would fall between notable and cut-off, the same as spinner. However, the difference is a linear progress would eventually provide indication of how long a system may take to respond once it goes from indeterminate > determinate. Could we possibly amend it to something like "Use an indeterminate linear progress to indicate the system is processing or loading an action, with the intention of providing a determined wait time before the cut-off point." What do you think? @joshwooding

packages/core/stories/progress/linear-progress.stories.tsx
origami-z: Removed. 
Not feeling this is a good story for progress. If one day we have a loading related pattern, it would be a good one there.

site/docs/components/progress/examples.mdx
origami-z: Added additional line before Nav's comment, can combine if needed

packages/core/src/progress/LinearProgress/LinearProgress.css
origami-z: Updated to use transform

packages/core/src/progress/LinearProgress/LinearProgress.tsx
origami-z: Updated props table

site/docs/components/progress/examples.mdx
origami-z: Updated slightly

packages/core/stories/progress/progress.qa.stories.css
joshwooding, line 1: ?

site/docs/components/progress/examples.mdx
```diff
@@ -26,45 +26,37 @@ Use linear or circular depending on the context, layout and functionality of an
...
+<LivePreview componentName="progress" exampleName="WithMinVal" displayName="Custom min and max">
 
-## With maximum value
+## Min and max values
 
-Specify the maximum value of the progress indicator. The percentage progress will be calculated using this value. The default value is 100.
+Specify the `min` and `max` values of the progress indicator. The default range is 0 - 100.
 
 </LivePreview>
+<LivePreview componentName="progress" exampleName="WithBuffer" displayName="With value and buffer">
```
joshwooding, line 45:
```suggestion
<LivePreview componentName="progress" exampleName="WithBuffer" displayName="With buffer">
```

site/docs/components/progress/examples.mdx
```diff
@@ -26,45 +26,37 @@ Use linear or circular depending on the context, layout and functionality of an
...
-
-## With buffer
-
-Specify a buffer value to indicate loading state. The buffer does not have a label and will not affect the progress label.
-
-</LivePreview>
-<LivePreview componentName="progress" exampleName="WithProgBufferVal" displayName="With progressing buffer value">
+<LivePreview componentName="progress" exampleName="WithProgBufferVal" displayName="With progressing buffer">
 
 ## With progressing buffer value
```
joshwooding, line 61:
```suggestion
## With progressing buffer
```

.changeset/four-bugs-reply.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/core": minor
+---
+
+Updated `LinearProgress` to display a moving line to represent an unspecified wait time, when `value` is `undefined`.
+
+`<LinearProgress />`
```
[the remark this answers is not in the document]
origami-z: Not prettier format

packages/core/stories/progress/progress.qa.stories.tsx
origami-z: Doesn't work without

packages/core/stories/progress/progress.qa.stories.css
origami-z: Added back

site/docs/components/progress/examples.mdx
origami-z: @navkaur76 updated copy

.changeset/four-bugs-reply.md
```diff
@@ -0,0 +1,9 @@
...
+
+`value` and `bufferValue` are no longer default to `0`. Previously above code would render a 0% progress bar, which was not a good reflection of intent. You can still achieve it by passing in `value={0}`.
```
joshwooding, line 9:
```suggestion
*Note*: `value` and `bufferValue` are no longer default to `0`. Previously above code would render a 0% progress bar, which was not a good reflection of intent. You can still achieve it by passing in `value={0}`.
```
?

joshwooding (approved):

origami-z (approved):

packages/core/src/progress/LinearProgress/LinearProgress.tsx
```diff
@@ -43,8 +45,8 @@ export const LinearProgress = forwardRef<HTMLDivElement, LinearProgressProps>(
       hideLabel = false,
       max = 100,
       min = 0,
-      value = 0,
```
mark-tate, line 46: is this considered a breaking change ?
Should we  leave this as 0 and opt into indeterminate ?
At some point a breaking change release we could make this the default ?

PR #3447, opened 2024-05-21
Fixed Dialog border css

origami-z:
Order of the css was not correctly applied


joshwooding (approved):
It looked fine for me but, we should be consistent.

Fercas123 (approved):

PR #3475, opened 2024-05-29
Fix combo box can be focused when disabled

### LINKED ISSUE #3369: Disabled combo box stays focused
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

"@salt-ds/core": "^1.15.0",

### Description

For our use case the combobox is disabled after the user selects an option. The issue is that the combo box stays in focus (shown by the dotted blue lines around the combobox). Even when a user clicks on something else the diabled combobox stays focused.

### Steps to reproduce

set disabled set to true on onSelectionChange if a option is selected 

### Expected behavior

Once the combo box disables and the user click on another part of the screen then the combobox should lose focus. 

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

origami-z:
Try fix #3369

### DIFF
--- a/packages/core/src/__tests__/__e2e__/combo-box/ComboBox.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/combo-box/ComboBox.cy.tsx
@@ -316,7 +316,7 @@ describe("Given a ComboBox", () => {
     cy.findByRole("combobox").should("be.disabled");
   });
 
-  it.only("should not stay focus if disabled after option selection", () => {
+  it("should not stay focus if disabled after option selection", () => {
     // Regression - #3369
     const DisabledAfterSelection = () => {
       const [disabled, setDisabled] = useState(false);
--- a/packages/core/src/__tests__/__e2e__/combo-box/ComboBox.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/combo-box/ComboBox.cy.tsx
@@ -3,6 +3,7 @@ import * as comboBoxStories from "@stories/combo-box/combo-box.stories";
 import { ComboBox, Option } from "@salt-ds/core";
 
 import { CustomFloatingComponentProvider, FLOATING_TEST_ID } from "../common";
+import { useState } from "react";
 
 const {
   Default,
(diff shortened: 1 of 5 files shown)


packages/core/src/combo-box/ComboBox.tsx
```diff
@@ -134,6 +134,12 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
     removePill,
   } = listControl;
 
+  useEffect(() => {
```
origami-z, line 137: @joshwooding  i don't quite like this, but can't think of another way

joshwooding:
Hmm, we might need to check disabled onFocus again

joshwooding (approved):

packages/core/src/__tests__/__e2e__/combo-box/ComboBox.cy.tsx
```diff
@@ -306,6 +308,38 @@ describe("Given a ComboBox", () => {
     cy.findByRole("button", { name: "end" }).should("be.focused");
   });
 
+  it("should not receive focus via mouse click if it is disabled", () => {
+    cy.mount(<Disabled />);
+    cy.findByRole("combobox").realClick();
+    cy.get(".saltComboBox").should("not.have.class", "saltComboBox-focused");
```
joshwooding, line 314: Should we test actual focus too? e,g, be.focused

packages/core/src/__tests__/__e2e__/dropdown/Dropdown.cy.tsx
```diff
@@ -240,6 +241,36 @@ describe("Given a Dropdown", () => {
     cy.findByRole("button", { name: "end" }).should("be.focused");
   });
 
+  it("should not receive focus via mouse click if it is disabled", () => {
+    cy.mount(<Disabled />);
+    cy.findByRole("combobox").realClick();
+
+    cy.findByRole("combobox").should("be.disabled");
```
joshwooding, line 248: This technically doesn't test focus only if it's disabled and the list doesn't open

joshwooding (approved):

Fercas123 (approved):

PR #3478, opened 2024-05-29
Add select all example for Dropdown

ThusharaJ07:


packages/core/stories/dropdown/dropdown.stories.tsx
```diff
@@ -341,3 +341,80 @@ export const ObjectValue: StoryFn<DropdownProps<Person>> = () => {
...
+      onSelectionChange={handleSelectionChange}
+      multiselect
+    >
+      <div
+        style={{
+          position: "sticky",
+          top: 0,
+          zIndex: 1,
+          background: !selected.includes(allSelectedOptionValue)
+            ? "var(--salt-color-white)"
```
mark-tate, line 395: using a specific color here, rather than a characteristic token, seems fragile ?
What happens when you user switches between light/dark ?

ThusharaJ07: Hi @mark-tate , I have updated the this to use --salt-container-primary-background.

packages/core/stories/dropdown/dropdown.stories.tsx
```diff
@@ -341,3 +341,80 @@ export const ObjectValue: StoryFn<DropdownProps<Person>> = () => {
...
+            borderBottom: "solid",
+            borderWidth: "1px",
+            borderColor:
+              selected.includes(usStates[0]) ||
+              selected.includes(allSelectedOptionValue)
+                ? "transparent"
+                : "var(--salt-separable-tertiary-borderColor)",
+          }}
+          value={allSelectedOptionValue}
+          key={allSelectedOptionValue}
```
origami-z, line 410: Is this `key` really needed?
Do we know why focus will shift after selecting it? @joshwooding

ThusharaJ07: @origami-z , I have removed the key from the example for select all.

packages/core/stories/dropdown/dropdown.stories.tsx
```diff
@@ -341,3 +341,70 @@ export const ObjectValue: StoryFn<DropdownProps<Person>> = () => {
...
+        selected.length < 2
+          ? selected[0]
+          : selected.includes("all")
+          ? "All Selected"
+          : `${selected.length} items selected`
+      }
+      onSelectionChange={handleSelectionChange}
+      multiselect
+    >
+      <div>
```
origami-z, line 389: `div` is no longer needed ..?

packages/core/stories/dropdown/dropdown.stories.tsx
```diff
@@ -341,3 +341,70 @@ export const ObjectValue: StoryFn<DropdownProps<Person>> = () => {
...
+          (el) => el !== allSelectedOptionValue
+        );
+      } else {
+        newOptionsSelected = [];
+      }
+    } else if (
+      isAllSelected ||
+      (!isAllSelected && newOptionsSelected.length === usStates.length)
+    ) {
+      newOptionsSelected = [...usStates, allSelectedOptionValue];
```
origami-z, line 368: I think order of `allSelectedOptionValue` is the reason of losing focus on selection. `allSelectedOptionValue` in front resolves it

packages/core/stories/dropdown/dropdown.stories.tsx
ThusharaJ07: div is added to handle not showing double border when select all option is selected since pseudo elements are applied to add borders for selection

packages/core/stories/dropdown/dropdown.stories.tsx
joshwooding: I'm not seeing the focus change?

origami-z (approved):

shey-v:
The prop 'handleSelectionChange' needs to be styled in the code font.

in terms of guidance, do you think it's worth adding: 

Best Practices: 

- Select All is beneficial in situations where users handle extensive data sets or need to execute bulk actions involving the selection of multiple items at once. 
- Select All ensures users can easily manage their selections and enhances the overall user experience by reducing the number of required clicks and providing an easy reversal option if selections are made accidentally.

origami-z:
> The prop 'handleSelectionChange' needs to be styled in the code font.

Good point, it actually should be `onSelectionChange` prop

origami-z (approved):

origami-z (dismissed):

joshwooding (dismissed):

joshwooding (dismissed):

origami-z (dismissed):

joshwooding (approved):

mark-tate (approved):

PR #3485, opened 2024-05-30
Stepped Tracker: compact variant

tomhazledine:
Adds a `compact` prop to `<SteppedTracker />`

Screenshot 2024-05-30 at 11 39 06

---

Currently WIP because of question around variant prop naming (changeset and docs to follow once approach is finalised):

**Qs.** The "compact" styles are only applicable to the horizontal variant of the Stepped Tracker. This means there's some logic to only apply the relevant className when both the `component` prop is `true` and `orientation` prop is `"horizontal"`.

1. Would it be cleaner to have just the single "orientation" prop, and add `horizontal-compact` as a option there?
2. Higher-level question: is a `compact` prop a pattern we want to reuse in other components? AFICT the only other places where there is a compact version are Paginator and Input, which each have discrete compact components to export (i.e. `<CompactPaginator />`)

### DIFF
--- a/.changeset/strong-ravens-share.md
+++ b/.changeset/strong-ravens-share.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+horizontal-compact variant for Stepped Tracker
--- a/.changeset/strong-ravens-share.md
+++ b/.changeset/strong-ravens-share.md
@@ -2,4 +2,4 @@
 "@salt-ds/lab": patch
 ---
 
-horizontal-compact variant for Stepped Tracker
+Added "horizontal-compact" variant for Stepped Tracker
--- a/packages/lab/src/stepped-tracker/SteppedTracker.tsx
+++ b/packages/lab/src/stepped-tracker/SteppedTracker.tsx
@@ -33,11 +33,7 @@ export interface SteppedTrackerProps extends ComponentPropsWithoutRef<"ul"> {
   /**
    * The orientation of the SteppedTracker. Defaults to `horizontal`
    */
-  orientation?: "horizontal" | "vertical";
-  /**
-   * Render in visually compact mode.
-   */
-  compact?: boolean;
+  orientation?: "horizontal" | "horizontal-compact" | "vertical";
 }
 
 const useCheckInvalidChildren = (children: ReactNode) => {
@@ -66,7 +62,6 @@ export const SteppedTracker = forwardRef<HTMLUListElement, SteppedTrackerProps>(
       className,
       activeStep,
       orientation = "horizontal",
-      compact = false,
       ...restProps
     },
     ref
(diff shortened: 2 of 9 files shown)


packages/lab/src/stepped-tracker/SteppedTracker.tsx
```diff
@@ -34,6 +34,10 @@ export interface SteppedTrackerProps extends ComponentPropsWithoutRef<"ul"> {
    * The orientation of the SteppedTracker. Defaults to `horizontal`
    */
   orientation?: "horizontal" | "vertical";
+  /**
+   * Render in visually compact mode.
+   */
+  compact?: boolean;
```
lilyvc, line 40: just to respond to your discussion points, it doesn't seem ideal to have a compact prop that only works with one orientation
if it's not going to be a separate component then I think 'horizontal-compact' orientation prop makes sense

ivan-calderon:
Hey Tom! Thanks for sharing + good progress here. I found easier to put a few comments together to avoid long lines of confusing text here.  Please see [Compact stepped tracker comments](https://www.figma.com/design/pzgdSYa8O1Vgdlor3XmKzl/Salt-%2F-Story-Parity?m=auto&node-id=4325-4137&t=GBxTbmmjk6eHoEGW-1).

About your higher-level question I am not completely sure how in the tech side will work. The overall goal was to provide examples (storybook) supporting something similar to [Pepper progress stepper](https://www.figma.com/design/WtUyH0EIjgrrA6LSgnwCVI/StoryDS_Salt_Gap_Analysis?node-id=244-9798&t=90GH2hSERemwRihV-1). But that won't be part of the Salt component itself (No docs).

site/docs/components/stepped-tracker/examples.mdx
```diff
@@ -20,6 +20,20 @@ You can add a label to the `TrackerStep` using the `StepLabel` component.
 
   </LivePreview>
 
+  <LivePreview componentName="stepped-tracker" exampleName="Compact" >
```
origami-z, line 23: The example doesn't load from storybook story but a duplicate in `site/src/examples`

.changeset/strong-ravens-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+horizontal-compact variant for Stepped Tracker
```
origami-z, line 5:
```suggestion
Added "horizontal-compact" variant for Stepped Tracker
```

site/docs/components/stepped-tracker/examples.mdx
origami-z: May need adjustment, so it will sit in the middle of example container.

origami-z (approved):
Need to add QA stories for the new variant.

Happy to be merged in first.

tomhazledine:
Replaced with slightly different approach here: #3498

PR #3491, opened 2024-05-31
loading-button

Vineet119:


site/docs/components/button/examples.mdx
```diff
@@ -79,5 +79,34 @@ A button with the prop `disabled={true}` will suppress all functionality. If you
 
 Use full width buttons on mobile devices or smaller viewport, the button will take up the full width of its container.
 
+</LivePreview>
+
+<LivePreview componentName="button" exampleName="LoadingButtons" >
+
+## Loading Buttons
+
+The buttons will show a loading spinner when setting the isLoading prop to true. You should pass a message using the loadingText prop for screenreader to announce while loading state is true. It defaults to "Loading".
```
bhoppers2008, line 88: Prop names and values should be marked up i believe: `loadingText` `true` etc

packages/core/src/__tests__/__e2e__/button/Button.cy.tsx
```diff
@@ -6,58 +6,70 @@ const composedStories = composeStories(buttonStories);
 const { FeatureButton, FocusableWhenDisabled } = composedStories;
 
 describe("Given a Button", () => {
-  checkAccessibility(composedStories);
+  // checkAccessibility(composedStories);
```
origami-z, line 9: remove comments

packages/core/src/button/Button.css
```diff
@@ -91,23 +91,23 @@
...
+.saltButton.saltButton-active:focus-visible:not(.saltButton-loading-primary, .saltButton-loading-secondary, .saltButton-loading-cta),
+.saltButton:focus-visible(:active):not(.saltButton-loading-primary, .saltButton-loading-secondary, .saltButton-loading-cta) {
   background: var(--saltButton-background-active-hover, var(--button-background));
   color: var(--saltButton-text-color-active-hover, var(--button-text-color));
   border-color: var(--saltButton-borderColor-hover, var(--button-borderColor-hover));
 }
 
 /* Pseudo-class applied to the root element on hover when Button is not active or disabled */
-.saltButton:hover {
+.saltButton:hover:not(.saltButton-loading-primary, .saltButton-loading-secondary, .saltButton-loading-cta) {
```
origami-z, line 102: Do we really need these increased specificity? Can we do it another way by using `--button-x` variables?

packages/core/src/button/Button.tsx
```diff
@@ -26,6 +27,19 @@ export interface ButtonProps extends ComponentPropsWithoutRef<"button"> {
...
   variant?: ButtonVariant;
+  /**
+   * To show a loading spinner.
+   */
+  isLoading?: boolean;
+  /**
+   * For the screen reader to announce while button is in loading state
+   * 'Loading' is the default value.
+   */
+  loadingText?: string;
```
origami-z, line 38: Do we really need a new prop to deal with announcement? 
cc @joshwooding

packages/core/src/button/Button.css
Vineet119: Done

packages/core/src/button/Button.tsx
joshwooding: Without clear accessibility guidance I don't want any of these added apart from the isLoading prop

packages/core/src/button/Button.tsx
```diff
@@ -26,6 +27,19 @@ export interface ButtonProps extends ComponentPropsWithoutRef<"button"> {
    * 'primary' is the default value.
    */
   variant?: ButtonVariant;
+  /**
+   * To show a loading spinner.
+   */
+  isLoading?: boolean;
```
joshwooding, line 33: This does not follow our standard prop naming, "is" shouldn't be the prefix

packages/core/src/button/Button.tsx
```diff
@@ -72,15 +89,47 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
       >
-        {children}
+        {isLoading ? (
+          <>
+            <span className={clsx(withBaseName("loading-overlay"), className)}>
+              <Spinner
+                size="small"
+                className={clsx(withBaseName("loading-spinner"), className)}
+              />
+              <span
```
joshwooding, line 110: Even if this is the right handling of accessibility, why is this used over the built-in controls for Spinner?

packages/core/stories/button/button.stories.tsx
```diff
@@ -168,3 +172,111 @@ export const FullWidth: StoryFn<typeof Button> = () => {
...
+        {secondaryButtonLabel}
+        {withRightArrowIcon && <ChevronRightIcon aria-hidden />}
+      </Button>
+    </div>
+  );
+};
+
+export const LoadingButtons: StoryFn<typeof Button> = () => {
+  return (
+    <LoadingButtonGrid
```
joshwooding, line 262: I'm not a big fan of these examples. This looks like it's similar to a template but it's not using the storybook method

joshwooding: I'm also not a fan of changing labels via props either

Vineet119: done

packages/core/src/button/Button.tsx
Vineet119: removed

docs/css/salt-countries.css
```diff
@@ -0,0 +1,26 @@
+/* Style applied to the root element */
```
joshwooding, line 1: This shouldn't be committed

packages/core/src/button/Button.tsx
```diff
@@ -72,15 +89,41 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           withBaseName(variant),
           {
             [withBaseName("disabled")]: disabled,
+            [withBaseName(`loading-${variant}`)]: loading,
             [withBaseName("active")]: active,
           },
           className
         )}
+        aria-disabled={disabled}
```
joshwooding, line 97: aria-disabled is handled in useButton, this overrides it?

packages/core/src/button/Button.tsx
```diff
@@ -72,15 +89,41 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
           withBaseName(variant),
           {
             [withBaseName("disabled")]: disabled,
+            [withBaseName(`loading-${variant}`)]: loading,
```
joshwooding, line 92: We already have a variant class, why do we need to add loading-(variant) classes?

Vineet119: we need it, for handling css part, of different variant, like border for different variants, while loiading

packages/core/src/button/Button.tsx
Vineet119: removed

docs/css/salt-countries.css
Vineet119: removed

packages/core/src/button/Button.tsx
mark-tate: Now we see this in code ...Should this move to it's own component or are we comfortable having one API inside another ? On other component library's they just have one prop which drives the state.

packages/core/src/button/Button.tsx
```diff
@@ -72,15 +89,40 @@ export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
...
             [withBaseName("active")]: active,
           },
           className
         )}
+        aria-live={loading !== undefined ? "assertive" : undefined}
```
joshwooding, line 97: I can't remember the outcome but what was decided on the usage of assertive here?

packages/core/stories/button/button.stories.tsx
```diff
@@ -1,9 +1,29 @@
...
+  useId,
+  Dialog,
+  DialogHeader,
+  DialogActions,
+  DialogContent,
+  FormField,
+  FormFieldLabel,
+  Input,
+} from "@salt-ds/core";
+import { ChevronRightIcon, DoubleChevronRightIcon } from "@salt-ds/icons";
```
joshwooding, line 19: This important should be merged in the one below.

packages/core/stories/button/button.stories.tsx
```diff
@@ -168,3 +188,232 @@ export const FullWidth: StoryFn<typeof Button> = () => {
...
+              This page has updates. Refresh page if you want to see recent
+              changes.
+            </Text>
+            <FlowLayout gap={1} justify="end">
+              <Button
+                variant="cta"
+                loading={loadingState}
+                showLoadingText
+                loadingText="Refreshing"
+                aria-describedby=""
```
joshwooding, line 347: Is this needed?

packages/core/stories/button/button.stories.tsx
Vineet119: done

packages/core/stories/button/button.stories.tsx
Vineet119: removed

joshwooding:
Succeded by #3887

PR #3477, opened 2024-05-29
Add categorical colors

origami-z:
Adds foundation colors.
Updated foundation page to show colors.

#2523

PR #3497, opened 2024-05-31
Update citrine 900 lime 900 color

origami-z:
No need for a changeset, correct values merged in #3477 (design updates)


joshwooding (approved):

dplsek (commented):
new RGB values are correct for citrine 900 and lime 900

Issue #3456, opened 2024-05-22
Bordered form controls


joshwooding:
Bordered variants are needed for:

- [x] Input #3537 
- [ ] Dropdown
- [ ] ComboBox

They should align to MultilineInput

Spec for input
https://www.figma.com/design/C4uM2dzDWMcAFqP3PYFAy1/Bordered-Input?node-id=3-73&t=mNPDy8G82XQIkSpM-1

ThusharaJ07:
Added bordered prop to combobox - #3594 (merged)
Added bordered prop to dropdown - #3551  (merged)

origami-z:
Will be released in #3644

PR #3499, opened 2024-05-31
input - added bordered input variant

ThusharaJ07:
Part of #3456


packages/core/src/multiline-input/MultilineInput.css
```diff
@@ -69,6 +69,7 @@
   --multilineInput-background-hover: var(--salt-status-error-background);
   --multilineInput-borderColor: var(--salt-status-error-borderColor);
   --multilineInput-outlineColor: var(--salt-status-error-borderColor);
+  --multilineInput-background-readonly: var(--salt-status-error-background);
```
origami-z, line 72: Assuming this is fixing code not implementing design correctly before. reminder to add changeset.

site/src/examples/input/Bordered.tsx
```diff
@@ -0,0 +1,9 @@
+import { ReactElement } from "react";
+import { Input, FlowLayout } from "@salt-ds/core";
+
+export const Bordered = (): ReactElement => (
+  <FlowLayout style={{ maxWidth: "256px" }}>
+    <Input bordered defaultValue="Value" />
```
origami-z, line 6: When status is set (e.g. error)
- hover hides the activation indicator. Feels a bit weird, worth design confirmation.
- full border should turn into status color? (shown as gray right now)

packages/core/src/input/Input.css
```diff
@@ -178,6 +184,29 @@
   border-bottom: var(--input-borderWidth) var(--input-borderStyle) var(--input-borderColor);
 }
 
+/* Style applied if `bordered={true}` */
+.saltInput.saltInput-bordered {
+  --input-border: var(--salt-size-border) var(--salt-container-borderStyle) var(--input-borderColor);
+  --input-borderWidth: 0;
```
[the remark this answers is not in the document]
origami-z: `input-border` vs `input-borderWidth`.

First one has border width defined as well? I'd assume one of them is for activation indicator?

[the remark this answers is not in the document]
ThusharaJ07: yes. and the second one is set to 0 to negate the  border-bottom being applied from default case. instead of that i can just do `border-bottom: none; ` to avoid confusion

packages/core/src/multiline-input/MultilineInput.css
```diff
@@ -182,12 +185,17 @@
...
-.saltMultilineInput-bordered:active,
+/* Style applied if focused when `bordered={true}` */
 .saltMultilineInput-bordered.saltMultilineInput-focused {
   --multilineInput-activationIndicator-borderWidth: var(--salt-editable-borderWidth-active);
 }
 
+/* Style applied if active when `bordered={true}` */
+.saltMultilineInput-bordered:active {
+  --multilineInput-activationIndicator-borderWidth: var(--salt-editable-borderWidth-active);
+  outline: none;
```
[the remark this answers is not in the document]
ThusharaJ07: removed as discussed in meeting.

packages/core/src/multiline-input/MultilineInput.css
ThusharaJ07: @origami-z , can you help me on how to add it to changeset? I have also enabled border radius for multiline input.

site/src/examples/input/Bordered.tsx
ThusharaJ07: fixed this

packages/core/src/multiline-input/MultilineInput.css
origami-z: Added one (under `/.changeset`). You can add yourself as well. It will become the changelog / release note when released.

origami-z: Also given adding rounded corner support, let's also enable visual regression for it. See https://github.com/jpmorganchase/salt-ds/blob/c013b93ca7147aed7ee2055da6532599a743b238/packages/core/stories/text/text.qa.stories.tsx#L93-L108

ThusharaJ07: Added the above

packages/core/stories/multiline-input/multiline-input.qa.stories.tsx
```diff
@@ -138,16 +146,48 @@ export const AllVariantsGrid: StoryFn<QAContainerProps> = (props) => (
...
 AllVariantsGrid.parameters = {
-  chromatic: { disableSnapshot: false },
+  chromatic: {
+    disableSnapshot: false,
+    modes: {
+      theme: {
+        themeNext: "disabled",
+      },
+      themeNext: {
+        themeNext: "enabled",
```
origami-z, line 184:
```suggestion
        themeNext: "enable",
```

packages/core/src/input/Input.css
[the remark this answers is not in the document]
origami-z: Although not ideal, this is the same approach as existing multiline input, so will treat it the same way.

origami-z (dismissed):
Checked preview, aligned with agreed modification from first design review.

(mentioned in one of the conversation) variable name is not ideal, but it's aligned with multiline input, which we want to cause minimal change (even with private css var token)

origami-z (approved):

bhoppers2008:
@ThusharaJ07 there's an issue with the activation line when Corner = Round both for full and bottom bordered components (for both Input and Multiline). At the moment it doesn't follow the rounded corner.

origami-z:
Replaced by #3537

PR #3523, opened 2024-06-05
Fix SegmentedButtonGroup with corner rounded

joshwooding:
Was a quick fix.

 @origami-z how are we handling changesets for stuff like this?

### DIFF
--- a/packages/core/stories/segmented-button-group/segmented-button-group.qa.stories.tsx
+++ b/packages/core/stories/segmented-button-group/segmented-button-group.qa.stories.tsx
@@ -38,7 +38,20 @@ export const Default: StoryFn<SegmentedButtonGroupProps> = () => {
 };
 
 Default.parameters = {
-  chromatic: { disableSnapshot: false },
+  chromatic: {
+    disableSnapshot: false,
+    modes: {
+      theme: {
+        themeNext: "disabled",
+      },
+      themeNext: {
+        themeNext: "enabled",
+        corner: "rounded",
+        accent: "teal",
+        // Ignore headingFont given font is not loaded
+      },
+    },
+  },
 };
 
 export const Icons: StoryFn<SegmentedButtonGroupProps> = () => {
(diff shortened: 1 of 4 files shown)


origami-z:
> @origami-z how are we handling changesets for stuff like this?

I think we treat it as bug fix, but specially calling it only impacting theme next / rounded corner

origami-z (dismissed):
Do we want to snapshot this iin theme next if fixed?

joshwooding:
> Do we want to snapshot this iin theme next if fixed?

I'm not sure why it's not getting picked up, I guess because the kitchen sink imports other stories

origami-z:
Kitchen sink is not included in snapshot.

I mean this: https://github.com/jpmorganchase/salt-ds/blob/c36e8a4466bbfd15e3ad152a37056b1f4cb8e522/packages/core/stories/text/text.qa.stories.tsx#L100

origami-z (approved):

bhoppers2008 (dismissed):

packages/core/stories/segmented-button-group/segmented-button-group.qa.stories.tsx
```diff
@@ -38,7 +38,20 @@ export const Default: StoryFn<SegmentedButtonGroupProps> = () => {
...
 
 Default.parameters = {
-  chromatic: { disableSnapshot: false },
+  chromatic: {
+    disableSnapshot: false,
+    modes: {
+      theme: {
+        themeNext: "disabled",
+      },
+      themeNext: {
```
origami-z, line 47: Now this is more obvious, I'll need to look into why this is not taken into effect

joshwooding: Fixed it

origami-z (approved):

Fercas123 (approved):

PR #3527, opened 2024-06-06
Stepper Input `onChange` update

lilyvc:
Updated `StepperInput`'s `onChange` so that is passes the event as well as the value.
Also changed block keys to shift + up/down, and block value to be explicit not a multiplier.


packages/lab/src/stepper-input/StepperInput.tsx
```diff
@@ -43,7 +43,10 @@ export interface StepperInputProps
   /**
    * Callback when stepper input value is changed.
    */
-  onChange?: (changedValue: number | string) => void;
+  onValueChange?: (
+    event: ChangeEvent | MouseEvent | KeyboardEvent | FocusEvent,
+    changedValue: number
+  ) => void;
```
lilyvc, line 49: @joshwooding does this make more sense?

packages/lab/src/stepper-input/StepperInput.tsx
```diff
@@ -43,7 +51,10 @@ export interface StepperInputProps
   /**
    * Callback when stepper input value is changed.
    */
-  onChange?: (changedValue: number | string) => void;
+  onValueChange?: (
+    event: ChangeEvent | MouseEvent | KeyboardEvent | FocusEvent,
```
joshwooding, line 55:
```suggestion
    event: SyntheticEvent,
```

packages/lab/src/stepper-input/StepperInput.tsx
```diff
@@ -11,7 +19,7 @@ import stepperInputCss from "./StepperInput.css";
 const withBaseName = makePrefixer("saltStepperInput");
 
 export interface StepperInputProps
-  extends Omit<InputProps, "onChange" | "emptyReadOnlyMarker"> {
+  extends Omit<InputProps, "emptyReadOnlyMarker"> {
```
origami-z, line 22: What will be the `onChange` event if users uses it? Will they be confused with `onValueChange`?

lilyvc: It would be passed to the input, it could be simpler to exclude it though

packages/lab/src/__tests__/__e2e__/stepper-input/StepperInput.cy.tsx
origami-z, line 1: Will all input value change trigger an number event here? e.g. "1" -> "1."? if no, need to add test

Also, is copy / paste values covered in test?

packages/lab/src/stepper-input/StepperInput.tsx
```diff
@@ -59,13 +59,14 @@ export interface StepperInputProps
...
    */
-  value?: number | string;
+  value?: number;
 }
 
 export const StepperInput = forwardRef<HTMLDivElement, StepperInputProps>(
   function StepperInput(props, ref) {
     const {
       className,
+      decimalPlaces,
```
origami-z, line 69: destruct but not other change? why?

packages/lab/src/stepper-input/useStepperInput.ts
```diff
@@ -36,6 +40,10 @@ const toFloat = (inputValue: number | string) => {
 const sanitizedInput = (numberString: string) =>
   (numberString.match(ACCEPT_INPUT) || []).join("");
 
+const isIntermediateInput = (numberString: string) => {
```
origami-z, line 43: Is this covered in test?

packages/lab/src/stepper-input/useStepperInput.ts
```diff
@@ -54,9 +62,20 @@ export const useStepperInput = (
...
-    default: toFixedDecimalPlaces(defaultValue, decimalPlaces),
+    default: defaultValue,
     name: "stepper-input",
   });
+
+  const [inputValue, setInputValue] = useState<number | string>(
+    toFixedDecimalPlaces(defaultValue, decimalPlaces)
+  );
+
+  useEffect(() => {
```
origami-z, line 73: Why is this needed? Will this cause infinite loop? Edge case for `useControlled` ?

packages/lab/src/stepper-input/useStepperInput.ts
```diff
@@ -66,59 +85,43 @@ export const useStepperInput = (
...
+  const decrement = (event?: SyntheticEvent) => {
     if (currentValue === undefined || isAtMin()) return;
-    let nextValue = currentValue === "" ? -step : toFloat(currentValue) - step;
-
-    // Set value to `max` if it's currently out of range
-    if (max !== undefined && isOutOfRange()) nextValue = max;
-
-    setNextValue(nextValue);
+    let nextValue = toFloat(currentValue) - step;
+    if (nextValue < min) nextValue = min;
```
origami-z, line 99: Is this covered in test?

packages/lab/src/stepper-input/StepperInput.tsx
lilyvc: there was an error in console when it was passed to the input

packages/lab/src/stepper-input/useStepperInput.ts
lilyvc: because the input value is separate from the stepper input value so it needs to stay up to date. I tried to use useControlled first

packages/lab/src/stepper-input/StepperInput.tsx
origami-z: Is this prop expect to work with the component? 
Currently this is swallowed and take no effect?

If this meant to work, some of the props may need to go to `inputProps` rather than `Input` top level div.

packages/lab/src/stepper-input/useStepperInput.ts
origami-z: state update should be in sync with events that triggered them, not via `useEffect` in another render cycle

packages/lab/src/stepper-input/internal/useSpinner.ts
```diff
@@ -1,10 +1,12 @@
-import { useEffect, useState } from "react";
-
+import { useEffect, useState, KeyboardEvent, SyntheticEvent } from "react";
```
Fercas123, line 1: KeyboardEvent doesnt seem to be used
```suggestion
import { useEffect, useState, SyntheticEvent } from "react";
```

packages/lab/src/stepper-input/useStepperInput.ts
lilyvc: added test "does not call the `onChange` callback when the value is intermediate"

packages/lab/src/stepper-input/useStepperInput.ts
lilyvc: added:
"does not decrement below the minimum value"
and
"does not increment above the maximum value"

packages/lab/src/stepper-input/StepperInput.tsx
lilyvc: it does work as expected: https://9b4b76d8.saltdesignsystem-storybook.pages.dev/?path=/story/lab-stepper-input--decimal-places&args=decimalPlaces:4

lilyvc: it's still used in usestepperinput hook

packages/lab/src/stepper-input/useStepperInput.ts
lilyvc: you're right, I think I've managed to remove it and can't find any issues

packages/lab/src/__tests__/__e2e__/stepper-input/StepperInput.cy.tsx
lilyvc: added the intermediate value test
for copy and paste, I am not sure how to do this in cypress and can't find any examples in our code

packages/lab/src/__tests__/__e2e__/stepper-input/StepperInput.cy.tsx
```diff
@@ -214,6 +246,24 @@ describe("Stepper Input", () => {
...
   });
 
+  it("does not decrement below the minimum value", () => {
+    cy.mount(<StepperInput defaultValue={-1} min={-1} />);
+
+    cy.findByRole("spinbutton").should("have.value", -1);
+
+    cy.findByLabelText("decrement value").realClick();
+    cy.findByRole("spinbutton").should("have.value", -1);
+  });
```
origami-z, line 256: This doesn't check whether `onChange` should be called.

packages/lab/src/stepper-input/StepperInput.tsx
origami-z: This should trigger unused variable warning? Need to refactor how props are passed around. `decimalPlaces` warning likely means other props will have similar issue.

origami-z (commented):
need to refactor props destructing

packages/lab/src/__tests__/__e2e__/stepper-input/StepperInput.cy.tsx
lilyvc: updated

packages/lab/src/stepper-input/useStepperInput.ts
lilyvc: oh actually the same issue is there, if the value is controlled it doesn't update without this. See this example: https://628e27029fcf95004a62cafd-sipdhyjpaz.chromatic.com/?path=/story/lab-stepper-input--refresh-adornment

packages/lab/src/stepper-input/StepperInput.tsx
lilyvc: I've updated this to destruct more props so only input props get passed on input

packages/lab/src/stepper-input/internal/useSpinner.ts
```diff
@@ -22,12 +24,12 @@ function useSpinner(activationFn: () => void, isAtLimit: boolean) {
...
 
-  const activate = () => {
-    activationFn();
+  const activate = (event: SyntheticEvent) => {
+    activationFn(event);
     setButtonDown(true);
   };
 
-  useInterval(activationFn, buttonDown ? INTERVAL_DELAY : null);
+  useInterval(() => activationFn(), buttonDown ? INTERVAL_DELAY : null);
```
origami-z, line 32: Have you tried memorizing `event`?

If it doesn't work, need to update `onChange` event to be optional

packages/lab/src/stepper-input/useStepperInput.ts
```diff
@@ -131,24 +131,24 @@ export const useStepperInput = (
       setCurrentValue(roundedValue);
     }
 
-    if (onChange) {
-      onChange(roundedValue);
+    if (onChange && event) {
+      onChange(event, roundedValue);
```
origami-z, line 135: Does this mean button long press will not trigger the event? 

is this captured in test?

packages/lab/src/stepper-input/internal/useSpinner.ts
```diff
@@ -1,33 +1,45 @@
...
+      setButtonDown(true);
+      setEvent(event);
+    }
   };
-
-  useInterval(activationFn, buttonDown ? INTERVAL_DELAY : null);
+  useInterval(
+    () => {
+      if (buttonDown && event) {
+        activationFn(event);
```
origami-z, line 38: Have you tested whether storing `event` actually worked? e.g. whether accessing event target gives you something? What about event bubbling implications? Can you do some research?

packages/lab/src/stepper-input/internal/useSpinner.ts
```diff
@@ -1,33 +1,49 @@
-import { useEffect, useState } from "react";
-
+import { useEffect, useState, SyntheticEvent, useCallback } from "react";
```
Fercas123, line 1:
```suggestion
import { useEffect, useState, SyntheticEvent } from "react";
```

packages/lab/src/stepper-input/internal/useSpinner.ts
```diff
@@ -1,33 +1,49 @@
...
-function useSpinner(activationFn: () => void, isAtLimit: boolean) {
+function useSpinner(
+  activationFn: (event?: SyntheticEvent) => void,
+  isAtLimit: boolean
+) {
   const [buttonDown, setButtonDown] = useState(false);
+  const [delay, setDelay] = useState(INITIAL_DELAY);
 
-  const cancelInterval = () => setButtonDown(false);
+  const cancelInterval = () => {
```
Fercas123, line 14: should this be a callback?

packages/lab/src/stepper-input/StepperInput.tsx
```diff
@@ -32,18 +36,14 @@ export interface StepperInputProps
...
-  hideButtons?: boolean;
   /**
    * Callback when stepper input loses focus.
    */
   onBlur?: FocusEventHandler<HTMLInputElement>;
   /**
    * Callback when stepper input value is changed.
    */
-  onChange?: (changedValue: number | string) => void;
+  onChange?: (value: number | string, event?: SyntheticEvent) => void;
```
Fercas123, line 46: any reason why the value is not after the event? i think whenever we have onChanges with an event and a value we tend to have them on the opposite order

lilyvc: it's because the event is optional, it is a typescript error the other way round

packages/lab/src/stepper-input/internal/useSpinner.ts
lilyvc: honestly not sure when things should be a callback

packages/lab/src/stepper-input/StepperInput.tsx
Fercas123: ah i missed the optional, fair enough

Fercas123: ...just curious, why does it need to be optional?

packages/lab/src/stepper-input/internal/useSpinner.ts
Fercas123: ESLint says: The 'cancelInterval' function makes the dependencies of useEffect Hook (at line 28) change on every render. To fix this, wrap the definition of 'cancelInterval' in its own useCallback() Hook. (react-hooks/exhaustive-deps)

packages/lab/src/stepper-input/StepperInput.tsx
lilyvc: haha this was the whole problem, basically if you click and hold the button there is no event

Fercas123 (approved):

packages/lab/src/stepper-input/StepperInput.tsx
origami-z: This will result in a different API with others.
Have you tried any different optional approach? Maybe not set the arg being optional, but the actual type being event or undefined? Or make it null?

lilyvc: okay yeah I think undefined makes sense because it is actually passed as undefined

.changeset/polite-bags-act.md
```diff
@@ -0,0 +1,17 @@
+---
+"@salt-ds/lab": minor
+---
+
+- Changed `StepperInput`'s `onChange` to return a value and an optional event.
+- Replaced functionality of Page up and Page down keys in `StepperInput` with Shift + Up arrow and Shift + Down arrow. These keyboard controls change the value by the `block` amount.
+- The `block` amount now explicitly defines the value that is increment or decrement, not a multiplier of `step`.
```
lilyvc, line 7: There was conflicting docs in uitk about this, and I think this way gives more control as the block doesn't need to be an exact multiple of the step.

Fercas123 (approved):

.changeset/polite-bags-act.md
joshwooding: I believe the intent was more around a best practice. If we're removing it from the jsdoc can you make sure it's on the site docs cc @navkaur76

lilyvc: yeah I can update in the docs PR

Issue #2970, opened 2024-01-31
Dialog close button focus ring visibility and overlap with content


origami-z:
### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

1.0.0-alpha.29

### Description

1. Dialog close button focus ring is not fully visible (top right) see screenshot
2. Close button is overlapping with content (1-2px). This matters when content is filled with an image.

![Dialog close button overlaps with content](https://github.com/jpmorganchase/salt-ds/assets/5257855/f2a74315-1532-4b4e-87e1-30a8c3d159e8)

### Steps to reproduce

https://stackblitz.com/edit/salt-template-gu1jry?file=App.tsx

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

hyan83:
https://stackblitz.com/edit/salt-template-f83pqv?file=App.tsx

Adding margin: '5px' should fix this issue.

origami-z:
Extend dates to Xion, 1 sprint behind

mark-tate:
Goal: XS(1)
Similar to Overlay issue

tomhazledine:
Change done. Waiting for #3532 to be merged before raising PR

tomhazledine:
Still blocked waiting for #3532 - fix for this PR needs to be applied to Overlay too.

tomhazledine:
Change of approach following discussion with design team. New approach will address this and #3532 together.

New issue: #3839

PR #3533, opened 2024-06-07
Density-aware overlay close-button spacing

tomhazledine:
Adjusts the right-hand spacing of `<OverlayPanelContent>` to match the width of the close button in all densities.

### Touch density

Screenshot 2024-06-07 at 11 57 08

### Medium density

Screenshot 2024-06-07 at 12 49 29


packages/core/src/overlay/OverlayPanelCloseButton.css
```diff
@@ -11,5 +11,10 @@
 }
 
 .saltOverlayPanelCloseButton ~ .saltOverlayPanelContent {
-  padding-right: var(--salt-spacing-400);
+  /* (button-padding * 2) + button-icon-size + (button-borderWidth * 2) */
+  --saltOverlayCloseButton-width: calc(
```
tomhazledine, line 15: This feels like a pretty obtuse `calc`. Let me know if there's a better way to get the exact button width.

origami-z: Haven't looked existing code. if close button will be on its own space, shall we consider using `display:grid` with the button on its own column

tomhazledine: I did try a version that used the button's intrinsic size (I used flex rather than grid, but same concept). The problem was that it pushed the content area away from the right border (i.e. inserting a column for the button) so the scrollbar for long content ended up in the middle of the overlay rather than up against the side.

Keeping the scroll bar anchored on the side of the overlay was deemed better (discussed this with Ben), so the complicated calc is actually the simpler of the options.

origami-z: When i look at [example](https://cf73ea00.saltdesignsystem-storybook.pages.dev/?path=/story/core-overlay--close-button) with a lot of content and without title, it's shown "content | scrollbar | close"?

joshwooding:
Converted to draft due to the scrollbar overlapping the close button

tomhazledine:
> Converted to draft due to the scrollbar overlapping the close button

Talked to @bhoppers2008 about this while dev'ing, and we decided that the scrollbar/close-button overlap was the least bad of all the current options.

I guess our options are:
1. keep the overlap
2. make the scrollbar always visible
3. bump the margins so that the close button always leaves enough room for a scrollbar (whether needed or not)
4. work out if there is a scroll bar required, and only reposition the close button when there is a scroll bar

My view is that the happy-path is when all the content fits inside the overlay panel (thus no scrolling), so option 4 wouldn't be a sensible ROI 🤷 

Thoughts?

origami-z:
> Talked to @bhoppers2008 about this while dev'ing, and we decided that the scrollbar/close-button overlap was the least bad of all the current options.

Given long content isn't a common scenario, and overlapping with close button looks particularly bad on Windows, can we do

- When not long, make it content | close
- When long, make it content | scrollbar | close

I think `overflow: auto` is enough to achieve this?

origami-z:
Is this ready for review? Design decision made yet?

tomhazledine:
After talking to @bhoppers2008 we're going to revisit the scrollbar position in the future, but for now we're going with `content | scrollbar | closebutton` (which means the scrollbar never overlaps the close button). The main focus for this PR is the fix to spacing (which previously did not respect changes in density).

packages/core/src/overlay/OverlayPanelCloseButton.css
tomhazledine, line 1: Thought about removing this file completely, but a future PR will be adding some styles back (when we sort https://github.com/jpmorganchase/salt-ds/issues/2970 after this PR is merged)

packages/core/src/overlay/OverlayPanel.css
```diff
@@ -18,3 +18,8 @@
   z-index: var(--saltOverlay-zIndex, var(--salt-zIndex-flyover));
   width: max-content;
 }
+
+.saltOverlayPanel-inner {
+  display: flex;
+  flex-direction: row-reverse;
```
origami-z, line 24: Is this solution based on the fact that `OverlayTrigger` in examples is added before content?

tomhazledine: Ah, not OverlayTrigger, but OverlayCloseButton... For some reason I thought it was built into OverlayPanel (so always in the same source order). Might switch back to absolute positioning now...

tomhazledine: Reverted back to the absolutely positioned option, which puts `OverlayCloseButton` in the top-right regardless of source order but still keeps `content | scrollbar | closebutton` alignment.

packages/core/src/overlay/OverlayPanelCloseButton.css
```diff
@@ -4,12 +4,12 @@
...
 
-.saltOverlayPanelCloseButton-button.saltButton-secondary {
-  position: sticky;
-  right: 0;
-  top: 0;
-}
-
-.saltOverlayPanelCloseButton ~ .saltOverlayPanelContent {
-  padding-right: var(--salt-spacing-400);
+.saltOverlayPanel-inner:has(.saltOverlayPanelCloseButton) {
```
origami-z, line 7: `:has` isn't available in [supported browser](https://www.saltdesignsystem.com/salt/about/supported-platforms)?

origami-z: i don't know whether form field's grid slot approach would be easier?

tomhazledine: Oh wow - didn't realise we went back that far! 😱 Will re-visit with a grid option 👍

tomhazledine: Switched to grid approach that works:
* If close button comes before panel content in source order
* If close button comes after panel content
* If close button is not present

origami-z (approved):
Tested in chrome/safari/firefox on Mac

tomhazledine:
Closing now that we have a new approach from design: #3839

origami-z:
> Closing now that we have a new approach from design: #3839

That does not address our user needs. Also don't know yet whether #3839 would be a viable non-breaking change

PR #3538, opened 2024-06-10
Picker updates

Fercas123:
change picker to use generic types and add controlled stories 

input changes: 

- exposed start and end date in onChange

picker

- removed `startDate`, `defaultStartDate`, `setStartDate`, `endDate`, `defaultEndDate`, `setEndDate`
- added `selectedDate`, `defaultSelectedDate`, `setSelectedDate`


.changeset/mean-worms-brake.md
```diff
@@ -0,0 +1,14 @@
...
+- Exposed start and end input values in date input `onChange` picker.
+
+  ```tsx
+  <DatePicker
+    onChange={(event, start: string | undefined, end: string | undefined) => {}}
+  />
+  ```
+
+- Removed `startDate`, `defaultStartDate`, `setStartDate`, `endDate`, `defaultEndDate` and `setEndDate` in DatePicker.
+- Added `selectedDate`, `defaultSelectedDate`, `setSelectedDate` in DatePicker.
```
origami-z, line 14: These are in the context? first reading, this feels `setSelectedDate` is a prop

packages/lab/src/__tests__/__e2e__/date-input/DateInput.cy.tsx
```diff
@@ -33,5 +33,21 @@ describe("GIVEN a DateInput", () => {
...
+        const changeSpy = cy.stub().as("changeSpy");
+        const onChange = (event: ChangeEvent<HTMLInputElement>) => {
+          // React 16 backwards compatibility
+          event.persist();
+          changeSpy(event);
+        };
+        cy.mount(<Range onChange={onChange} />);
+        cy.findAllByRole("textbox").eq(0).click().clear().type("02-feb-2000");
+        cy.findAllByRole("textbox").eq(1).click().clear().type("05-feb-2000");
+        cy.get("@changeSpy").should("have.been.calledWithMatch", {
```
origami-z, line 47: Does this needs to check against both values?

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -115,272 +119,291 @@ export interface DateInputProps
...
+   * Callback fired when the selected date change.
    */
   onSelectionChange?: (
     event: SyntheticEvent,
-    selectedDate?: DateValue | { startDate?: DateValue; endDate?: DateValue }
+    selectedDate?: SelectionVariantType
+  ) => void;
+  onChange?: (
+    event: ChangeEvent<HTMLInputElement>,
+    startDateInputValue?: string,
```
origami-z, line 130: Given the other change, should input change be typed similarly as well?

Fercas123: do you mean with a generic? 
if so, i added the 2 strings separately because that way the first value you get is your single/start date and the next (optional) would be the range, else if i send an object with 2 strings the user would have to destructure it,  but happy to change it

origami-z: Was thinking on single date input, what will be "start" / "end" in that use case

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -115,272 +119,285 @@ export interface DateInputProps
...
   /**
-   * Callback fired when the selected daate change.
+   * Callback fired when the selected date change.
    */
   onSelectionChange?: (
     event: SyntheticEvent,
-    selectedDate?: DateValue | { startDate?: DateValue; endDate?: DateValue }
+    selectedDate?: SelectionVariantType
+  ) => void;
+  onChange?: (
```
origami-z, line 128: Is it possible to type this so that when single, only 1 value?
e.g. using conditional type depends on `SelectionVariantType`

origami-z (dismissed):
Can we double check whether new props are all captured in the changelog?

origami-z (approved):

lilyvc (approved):

PR #3558, opened 2024-06-13
Try StackBlitz SDK on site

origami-z:
Works in examples without additional dependencies. Couple of notes

- Where to put it the best (design wise), currently copy button is separate implementation from Mosaic
- White space of source issue
- Dependency issue (css, data, hook), will likely need to change implementation to include those. Or enforce 0 dependency on site example.
- React missing in StackBlitz, thus needing `import * as React from 'react'`

https://developer.stackblitz.com/platform/api/javascript-sdk#generate-and-embed-new-projects

### DIFF
--- a/site/src/components/components/LivePreview.tsx
+++ b/site/src/components/components/LivePreview.tsx
@@ -8,11 +8,12 @@ import {
   ElementType,
 } from "react";
 import clsx from "clsx";
-import { Switch } from "@salt-ds/core";
+import { Button, Switch } from "@salt-ds/core";
 import { SaltProvider } from "@salt-ds/core";
 import { Pre } from "../mdx/pre";
 import { useLivePreviewControls } from "./useLivePreviewControls";
 import useIsMobileView from "../../utils/useIsMobileView";
+import StackBlitzSDK from "@stackblitz/sdk";
 
 import styles from "./LivePreview.module.css";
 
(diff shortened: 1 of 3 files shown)


site/src/components/components/LivePreview.tsx
```diff
@@ -92,6 +93,98 @@ export const LivePreview: FC<LivePreviewProps> = ({
...
+          root.render(
+            <StrictMode>
+              <SaltProvider>
+                <App />
+              </SaltProvider>
+            </StrictMode>
+          );
+          `,
+          "App.tsx":
+            `import * as React from 'react';
```
lilyvc, line 165: If I'm understanding your whitespace issue correctly, could you use `prettier.format` on the code example so it's formatted in the stackblitz?

lilyvc: I tried like:
   
   const formatCode = (code) => {
  return prettier.format(code, {
    parser: "typescript",
    plugins: [parserTypeScript],
  });
};

......
   
     "App.tsx": formatCode(
            `import * as React from 'react';
` +
              ComponentExample.sourceCode +
              `
          
          export default ${exampleName};
          `
          ),

site/src/components/components/LivePreview.tsx
```diff
@@ -122,9 +215,12 @@ export const LivePreview: FC<LivePreviewProps> = ({
...
 
         {showCode && (
-          <Pre className={styles.codePreview}>
-            <div className="language-tsx">{ComponentExample.sourceCode}</div>
-          </Pre>
+          <>
+            <Pre className={styles.codePreview}>
+              <div className="language-tsx">{ComponentExample.sourceCode}</div>
+            </Pre>
+            <Button onClick={editInStackBlitz}>Edit in StackBlitz</Button>
```
lilyvc, line 222: You could align it right, and add a tear out icon.
Maybe also make it look like it is over the codeblock by adding background colour.

PR #3580, opened 2024-06-17
Date picker updates from feedback

Fercas123:
The following changes from feedback are addressed in this PR: 

- [x] Add docs paragraph about date formatter, if non-stable will reset state
- [x] Fix cursor position when selection happens
~~- [ ] need an example for a custom placeholder~~
- [x] Fix validation example, where retyping a valid date doesn't clear
- [x] check docs (props) inputAriaLabel removed

### DIFF
--- a/packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
+++ b/packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
@@ -57,9 +57,15 @@ describe("GIVEN a DatePicker", () => {
     });
     it("THEN should not format invalid dates on blur", () => {
       cy.mount(<Default defaultSelectedDate={testDate} />);
-      cy.findByRole("textbox").click().clear().type("date");
+      cy.findByRole("textbox").click().clear().type("2 fev");
       cy.findByRole("textbox").blur();
-      cy.findByRole("textbox").should("have.value", "date");
+      cy.findByRole("textbox").should("have.value", "2 fev");
+    });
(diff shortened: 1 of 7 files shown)


packages/lab/src/__tests__/__e2e__/date-picker/DatePicker.cy.tsx
```diff
@@ -300,35 +305,46 @@ describe("GIVEN a DatePicker", () => {
...
-        cy.findAllByRole("textbox").eq(1).click().clear().type(rangeTestInput);
-        cy.get("@changeSpy").should(
-          "have.been.calledWith",
-          testInput,
-          rangeTestInput
-        );
-      });
+      cy.mount(<ControlledRangePicker />);
+      cy.findAllByRole("textbox").eq(0).click().clear().type(testInput);
+      cy.findAllByRole("textbox").eq(1).click().clear().type(rangeTestInput);
```
origami-z, line 342: may check spy before 2nd input?

packages/lab/src/date-input/DateInput.tsx
```diff
@@ -239,8 +239,10 @@ export const DateInput = forwardRef<
   // Update date strings when dates change
   useEffect(() => {
```
origami-z, line 240: This only occurs to me now, should any of the string value update in the same render cycle?

origami-z (commented):
changeset?

origami-z (approved):

lilyvc (approved):

Issue #3583, opened 2024-06-17
Salt Icon CSS background image doesn't not work with additional classname in front


origami-z:
### Package name(s)

Icons (@salt-ds/icons)

### Package version(s)

v1.12.0

### Description

When another classname is used before Salt icon css, icon is broken

https://github.com/jpmorganchase/salt-ds/blob/a5e47e3ac958a05d1a7afe6f43461fbc5d78675d/packages/icons/scripts/generateIcons.mjs#L64C21-L64C39

### Steps to reproduce

https://stackblitz.com/edit/salt-template-xzwkid?file=App.tsx

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

origami-z:
Fixed in [@salt-ds/icons@1.12.1](https://github.com/jpmorganchase/salt-ds/releases/tag/%40salt-ds%2Ficons%401.12.1)

PR #3589, opened 2024-06-17
Fix Icon as CSS background not working when another className is in front

origami-z:
#3583


packages/icons/stories/icon.qa.stories.tsx
```diff
@@ -88,7 +88,10 @@ export const CssBackground: StoryFn = () => {
           iconName ? (
             <div
               key={iconName}
-              className={`saltIcons-${iconName.replace("Icon", "")}`}
+              className={`should-not-impact saltIcons-${iconName.replace(
```
origami-z, line 91: This is the proof code change worked

yunjungyeh (approved):

Fercas123 (approved):

PR #3602, opened 2024-06-19
Fix menu group label spacing

joshwooding:


packages/core/stories/menu/menu.qa.stories.css
```diff
@@ -0,0 +1,7 @@
+.singleLevel .background-block {
+  height: 500px;
+}
+
+.saltQAContainer.singleLevel {
+  --qaContainer-item-width: 160px !important;
```
mark-tate, line 6: At the next opportunity, worth looking at QA container to avoid having to publicly use !important.

yunjungyeh (approved):

yunjungyeh (approved):

Fercas123 (approved):

PR #3624, opened 2024-06-24
Tag site docs updates

Fercas123:


site/docs/components/tag/index.mdx
origami-z, line 1: Want to add `initialVersion`?

site/src/examples/tag/TagInCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const TagInCard = (): ReactElement => (
+  <Card
+    className={styles.category3}
+    style={{
+      width: 315,
+    }}
+    accent="top"
+  >
+    {/* in your css:
```
origami-z, line 13: May need whitespace better here?

Not sure whether we should actively promote override component tokens

site/src/examples/tag/TagInCard.tsx
```diff
@@ -0,0 +1,35 @@
...
+            .category3 {
+              --saltCard-accent-color: rgb(70, 118, 191);
+            }*/}
```
Fercas123, line 16:
```suggestion
```

site/src/examples/tag/TagInCard.tsx
Fercas123: I'll remove the comment since it is to do with card and not with tag, i think if we were to add category to card then we could use category here instead of the styling, but given it is already on the site, i would be inclined to keep it as it is (css hidden) and refactor it once we remove/update the card one

origami-z (approved):

shey-v (approved):

PR #3668, opened 2024-06-27
Fix invalid HTML structure in AccordionHeader

joshwooding:
Fixes the invalid HTML button > div vs button > span.
Updates a CSS selector to make it easier to override.
Adds a display name to a component to make it easier to debug.

### DIFF
--- a/packages/core/src/accordion/AccordionHeader.tsx
+++ b/packages/core/src/accordion/AccordionHeader.tsx
@@ -25,15 +25,14 @@ export interface AccordionHeaderProps
 
 const withBaseName = makePrefixer("saltAccordionHeader");
 
-const ExpansionIcon = ({ expanded }: { expanded: boolean }) =>
-  expanded ? (
-    <ChevronUpIcon aria-hidden="true" className={clsx(withBaseName("icon"))} />
-  ) : (
-    <ChevronDownIcon
-      aria-hidden="true"
-      className={clsx(withBaseName("icon"))}
-    />
-  );
+function ExpansionIcon({ expanded }: { expanded: boolean }) {
+  if (expanded) {
+    return <ChevronUpIcon aria-hidden className={withBaseName("icon")} />;
+  }
+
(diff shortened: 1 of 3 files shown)


packages/core/src/accordion/AccordionHeader.tsx
```diff
@@ -25,15 +25,14 @@ export interface AccordionHeaderProps
...
-const ExpansionIcon = ({ expanded }: { expanded: boolean }) =>
-  expanded ? (
-    <ChevronUpIcon aria-hidden="true" className={clsx(withBaseName("icon"))} />
-  ) : (
-    <ChevronDownIcon
-      aria-hidden="true"
-      className={clsx(withBaseName("icon"))}
-    />
-  );
+function ExpansionIcon({ expanded }: { expanded: boolean }) {
```
joshwooding, line 28: @Fercas123 I updated this so it has a display name

packages/core/src/accordion/AccordionHeader.tsx
```diff
@@ -72,7 +71,7 @@ export const AccordionHeader = forwardRef<
       {...rest}
     >
       {indicatorSide === "left" && <ExpansionIcon expanded={expanded} />}
-      <div className={withBaseName("content")}>{children}</div>
+      <span className={withBaseName("content")}>{children}</span>
```
joshwooding, line 74: This is the actual fix

packages/core/src/accordion/AccordionPanel.css
```diff
@@ -26,7 +26,7 @@
   padding-bottom: var(--salt-spacing-300);
 }
 
-.saltAccordionPanel-content.saltAccordionPanel-indentedContent {
+.saltAccordionPanel-indentedContent {
```
joshwooding, line 29: @Fercas123 I'm pretty sure we can lower the specificity of this.

Fercas123 (approved):

shey-v (approved):

PR #3716, opened 2024-07-04
DatePicker updates

mark-tate:
- `DatePicker` is now a wrapper around `DateInput` and does not contain any panels
- support `children`, so customisable panels can be provided by `DateInput`
- `DatePickerContext` now simplified and moved upto `DatePicker`, panels can define their own configuration whilst subscribing to the state from `DatePickerContext`
- `DatePickerSinglePanel` and `DatePickerRangePanel` provided to support basic single/range dates
- switch to conditional Typescript types to reduce complexity for end-user
- rename `selectionVariant value from `default` to `single`


packages/lab/src/date-input/DateInput.tsx
```diff
@@ -138,276 +159,328 @@ export interface DateInputProps<SelectionVariantType>
...
+    } = props;
+    const wrapperRef = useRef(null);
+    const inputRef = useForkRef<HTMLDivElement>(ref, wrapperRef);
+    const inputStringFormatter = (input: string): string => {
+      const date = getCalendarDate(input);
+      return !input || !date ? input : dateFormatter(date);
+    };
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-dateInput",
```
joshwooding, line 235:
```suggestion
      testId: "salt-date-input",
```

mark-tate:
YTGs

- Validation not working, keeps state in calendar upon error
- Open on enter example, does not close after selection, or open on click
- Focus behaviour not clear
- Use of null not clear
- useEffect review
- Performance check
- Date clamp

mark-tate:
This PR is now ready for review, folks who want a datepicker, does this work for you ?

To get this to Core, we need feedback from users.
The next step for this PR will be to merge to Labs so we  can get feedback on it.

packages/lab/stories/date-picker/date-picker.stories.tsx
```diff
@@ -1,209 +1,1099 @@
...
+                offsetDate.year,
+                offsetDate.month,
+                offsetDate.day,
+              );
+            }
+            return parseCalendarDate(parsedDate);
+          }}
+          onDateValueChange={(newDateValue: string) => {
+            const validationStatus =
+              isValidShortDate(newDateValue) ||
```
origami-z, line 729: `validateShortDate` is making it error again after "+7" correctly parsed

packages/lab/stories/date-picker/date-picker.stories.tsx
```diff
@@ -1,209 +1,1099 @@
...
+      selectionVariant="single"
+      onSelectedDateChange={(newSelectedDate) => {
+        console.log(`Selected date: ${formatDate(newSelectedDate)}`);
+        args?.onSelectedDateChange?.(newSelectedDate);
+      }}
+    >
+      <DatePickerSingleInput bordered />
+      <DatePickerOverlay>
+        <DatePickerSinglePanel
+          NavigationProps={{
```
origami-z, line 931: Not 100% sure whether the props name here would make it confusing, v.s. `CalendarNavigationProps`

packages/lab/stories/date-picker/date-picker.stories.tsx
```diff
@@ -1,209 +1,1099 @@
...
+    </DatePicker>
   );
 };
 
-export const ControlledOpenOnEnter: StoryFn<
-  DatePickerProps<SingleSelectionValueType | RangeSelectionValueType>
-> = (args) => {
-  // state
-  const [open, setOpen] = useState(false);
+const DatePickerTimeInput: React.FC = () => {
```
origami-z, line 968: Not sure what's intended example here. If I change time input from using mouse via popup, i can see selection change fired, but not seeing time input change (still `--:--`) nor seeing date range pick change (are we expecting time shown on it?)

packages/lab/src/date-picker/DatePickerOverlay.tsx
```diff
@@ -0,0 +1,74 @@
...
+  });
+
+  const { Component: FloatingComponent } = useFloatingComponent();
+
+  const { a11yProps } = useFormFieldProps();
+
+  const {
+    state: { open, floatingUIResult },
+    helpers: { getFloatingProps },
+  } = useDatePickerOverlay();
```
origami-z, line 43: Not to be fixed in this PR, we need a univeral solution for the user, if they want to explicitly move the position of overlay.

packages/lab/src/calendar/Calendar.tsx
```diff
@@ -1,46 +1,72 @@
...
 import { useWindow } from "@salt-ds/window";
 
-import { DateFormatter, getLocalTimeZone } from "@internationalized/date";
+import { DateFormatter } from "@internationalized/date";
 import calendarCss from "./Calendar.css";
-import { getCurrentLocale } from "./internal/utils";
 
-export type CalendarProps = useCalendarProps & {
+export interface CalendarBaseProps extends ComponentPropsWithoutRef<"div"> {
+  children?: ReactNode;
```
origami-z, line 30: I only realize this is inserted before the content when I dig into the code, and `<CalendarNavigation/>` is being set in the story. Not 100% certain whether there's some alternative way to not make this a little confusing, e.g. `heading` & `foot`, `position` (north/west/east/south/center) prop from `BorderItem`

origami-z (approved):

packages/lab/src/date-picker/DatePickerSingleInput.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+  const [value, setValue] = useControlled({
+    controlled: valueProp,
+    default: defaultValue,
+    name: "DatePickerSingleInput",
+    state: "value",
+  });
+
+  const handleCalendarButton = useCallback(() => {
+    setOpen(!open);
```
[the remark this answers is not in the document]
mark-tate: Jacob / @ivan-calderon  can we clarify, generally feeling is that these are two different types of content and so the behaviour can differ. 

According to the WCAG 2.1 AA standards, when a dialog is opened, the focus should generally move to the first focusable element within the dialog. This is to ensure that keyboard users can immediately interact with the dialog content without having to navigate away from the control that opened it.

Moving the focus to the "today" element (or the currently selected date) within the calendar dialog can be a good practice. This approach allows users to immediately use the arrow keys to navigate and select a new date, enhancing the user experience for keyboard users.

packages/lab/src/calendar/CalendarNavigation.tsx
```diff
@@ -232,10 +315,9 @@ export const CalendarNavigation = forwardRef<
         >
           <ChevronLeftIcon aria-label="Previous Month" />
         </Button>
-      </Tooltip>
-      <div className={withBaseName("dropdowns")}>
+      </ConditionalTooltip>
```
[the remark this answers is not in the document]
mark-tate: fixed by changing grid-gap 

```
.saltCalendarNavigation {
  display: grid;
  grid-template-columns: min-content auto min-content;
  padding-bottom: var(--salt-spacing-100);
  grid-gap: var(--salt-spacing-25);
  align-items: center;
  box-sizing: border-box;
}

```

@ivan-calderon  can you review.. this fixes the issue but then we still have an issue with width, try the Zh-CN locale example and set one of the 3 character dropdown widths, you get an ellipsis. 

Folks could set a width for the dropDown but then we lose the theming support, so worth seeing if by bumping the width a little, we can support short dates in multiple locales

packages/lab/src/date-picker/useDatePicker.ts
```diff
@@ -0,0 +1,199 @@
...
+    selectionVariant,
+    defaultSelectedDate = null,
+    selectedDate: selectedDateProp,
+    onSelectedDateChange,
+    defaultOpen = false,
+    open: openProp,
+    minDate: minDateProp,
+    maxDate: maxDateProp,
+    timeZone = getLocalTimeZone(),
+    locale = getCurrentLocale(),
```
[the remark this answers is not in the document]
mark-tate: @ivan-calderon can you confirm ?

DatePicker can support any form, for instance, the locale examples use 2 digit dates DD/MM/YYYY as well as MM/DD/YYYY

packages/lab/src/calendar/CalendarNavigation.tsx
```diff
@@ -173,38 +241,53 @@ export const CalendarNavigation = forwardRef<
...
+      moveToMonth(event, year[0]);
+    },
+    [moveToMonth],
+  );
 
   const formatMonth = (date?: DateValue) => {
     return !date
       ? ""
-      : formatDate(date, { month: hideYearDropdown ? "long" : "short" });
+      : formatDate(date, locale, {
```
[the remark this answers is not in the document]
mark-tate: Added your Zh-CN example to show how this can be done. Although this highlighted a width issue with the dropdown @ivan-calderon

packages/lab/stories/date-picker/date-picker.stories.tsx
mark-tate: fixed now, was a short-coming in the example code

packages/lab/stories/date-picker/date-picker.stories.tsx
mark-tate: agreed, renamed

packages/lab/src/calendar/Calendar.tsx
mark-tate: The reason it was split out, was so we could re-use Calendars together and they don't all have their own navigations.

CalendarNavigation is a component that can be composed within the DatePicker children, it appears in DatePickerRangePanel and DatePickerSinglePanel, as these are pre-canned patterns which contain the Calendars.
The intention is that folks can compose  DatePickerRangePanel/DatePickerSinglePanel with their own content (like we do in the custom panel example). Alternatively, folks can roll their own.

It's theoretically possible to create a calendar containing 12 calendars (or less) and then a Calendar Navigation with a `step` of 12 that advances the whole view by a calendar year.

origami-z: Yes, i understand the need for composability. It's more about API structure. Everyone would have different idea for `children` of a Calendar ? With the current structure, should it be something like `header`? Or should it not be part of Calendar all together, but be a completely separate piece?

ivan-calderon:
@mark-tate, @joshwooding all the progress you have made on date picker is amazing. I re-ran Date picker design review. Please [see Date Picker / Build QA](https://www.figma.com/design/iVb153RCmU5R7uvZ3ySPoR/Date-Picker?node-id=1970-15511&t=zv6uG1NVJEu6WvFJ-1) for more details.

joshwooding (approved):

packages/lab/src/date-picker/useDatePicker.ts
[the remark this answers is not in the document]
mark-tate: The comma format of the US date is quirky
@ivan-calderon  @bhoppers2008 
I can make it so we support comma and non-comma and it always defaults in the US to with a comma?
We can take an issue for this

packages/lab/src/date-picker/DatePickerOverlay.tsx
mark-tate: @joshwooding  FYI

packages/lab/src/date-picker/DatePickerSingleInput.tsx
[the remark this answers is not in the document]
mark-tate: This is still a on-going concern.. we will pick up in a new issue. @ivan-calderon  to feedback.

packages/lab/src/calendar/internal/utils.ts
```diff
@@ -61,10 +57,9 @@ function mapDate(currentDate: DateValue, currentMonth: DateValue) {
   };
 }
 
-export function generateVisibleDays(currentMonth: DateValue) {
+export function generateVisibleDays(currentMonth: DateValue, locale: string) {
```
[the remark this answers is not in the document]
mark-tate: Saw it too, bumped @internationalized/date as part of Labs release.

packages/lab/src/calendar/internal/CalendarDay.tsx
```diff
@@ -34,8 +34,15 @@ const withBaseName = makePrefixer("saltCalendarDay");
...
   function CalendarDay(props, ref) {
-    const { className, day, renderDayContents, month, TooltipProps, ...rest } =
-      props;
+    const {
+      className,
+      day,
+      renderDayContents,
+      month,
+      TooltipProps,
+      formatDate: formatDateProp = defaultFormatDate,
```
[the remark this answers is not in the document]
mark-tate: Revisited formatting after this and created a zh-CN example/test to show it working

packages/lab/stories/date-picker/date-picker.stories.tsx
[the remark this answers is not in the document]
mark-tate: Can see us creating alot more examples. Some of these exist in Calendar and will iterate on remaining based on further feedback.

packages/lab/src/date-picker/DatePickerSinglePanel.tsx
```diff
@@ -0,0 +1,175 @@
...
+  );
+
+  const baseCalendarProps: Partial<CalendarSingleProps> = {
+    selectionVariant: "single",
+    visibleMonth,
+    hoveredDate,
+    onHoveredDateChange: handleHoveredDateChange,
+    onVisibleMonthChange: handleVisibleMonthChange,
+    onSelectedDateChange: handleSelectedDateChange,
+    hideOutOfRangeDates: true,
```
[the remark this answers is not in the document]
mark-tate: I think this is so, the content below the calendar does not move position between months.
@ivan-calderon  to opine
We could speak to this in design documentation.

packages/lab/src/calendar/Calendar.tsx
mark-tate: We had opinion built into this component before by embedding one navigation to rule them all and we had props to control that component.

This approach removes those props and gives folks choice over the navigation element.

I can 't see any other use for children and don't think we should push any other elements to Calendar.
If folks can highlight a use-case then we could refactor our API around that as any change would just see us deprecate children and use a prop instead. 

However, what we would then use children for ? Perhaps there is a known use-case we have not yet heard about?

PR #3727, opened 2024-07-05
Fix ref warning on React 19

joshwooding:
Bumps React to 18.3.3
Adds getRefFromChildren to read refs from children in a way that supports all version of react. Also reduces the amount of ts-ignores/expect errors.
Updated Accordion, Menu, Overlay, Tooltip, CascadingMenu and Portal.


packages/core/src/utils/getRefFromChildren.ts
```diff
@@ -0,0 +1,10 @@
+import { isValidElement, ReactNode } from "react";
+
+export function getRefFromChildren(child: ReactNode) {
```
DavieReid, line 3: Does this need a test?

packages/lab/src/cascading-menu/CascadingMenu.tsx
```diff
@@ -195,9 +194,7 @@ export const CascadingMenu = forwardRef<HTMLDivElement, CascadingMenuProps>(
       childrenRef.current = node;
     }, []);
     const handleRef = useForkRef(
-      //TODO get to bottom of this
-      // @ts-ignore
-      isValidElement(children) ? children.ref : noop,
```
DavieReid, line 200: This one returned `noop` not `null` like in the other usages.  

The update looks Ok but just pointing out the diff

DavieReid (approved):

packages/lab/src/cascading-menu/CascadingMenu.tsx
joshwooding: Yeah, i think this was just an old way of doing it

packages/core/src/utils/getRefFromChildren.ts
joshwooding: Added :)

mark-tate (approved):

DavieReid (approved):

Issue #3333, opened 2024-04-29
Way to limit the max height of a combobox or dropdown list


joshwooding:
### Area

UI Components

### The problem

There is no easy way to limit the max height of a combobox or dropdown list.

The current way is to target the "saltOptionList" class and apply a max-height directly.

### The solution

.

### Alternatives and examples

.

### Are you a JPMorgan Chase & Co. employee?

- [ ] I am an employee of JPMorgan Chase & Co.

origami-z:
Extend dates to Xion, 1 sprint behind

mark-tate:
Zazu Goal: S(3)
- PR to be raised , docs only change

Fercas123:
notes from @jpmorganchase/salt-developer  conversation on 01 July 24: 

- there is no current way of passing `popperProps` in any of the components that use portal
- the preferred way wold be to increase composition to avoid adding props that pass down to portal.
- current solution could be to document how to target the list in global so users can override, downside is that it will be a change affecting all lists. 

TBD: More research about different approaches and follow up.

Fercas123:
documentation change addressed as short term solution, needs long term solution.

joshwooding:
I think we should create a new issue for a long-term solution. There is a way and it's documented.

efriandika:
Hi @Fercas123 
Which documentation did you refer to? I don't see any information related in this page: https://www.saltdesignsystem.com/salt/components/combo-box/usage

We also use this combobox in our application. 
Really interested in this solution.

origami-z:
Short term recommendation changed in #3728.

Long term solution to be investigated in #3599

PR #3728, opened 2024-07-05
Combobox and Dropdown documentation about list height

Fercas123:
documentation to address https://github.com/jpmorganchase/salt-ds/issues/3333


lilyvc (approved):

origami-z (approved):
We should leave a comment on the issue on working on long term solution

PR #3768, opened 2024-07-08
Dialog close button focus ring visibility and overlap with content

tomhazledine:
Issue: https://github.com/jpmorganchase/salt-ds/issues/2970

Fixes two issues:
1. Close-button focus ring hidden by Dialog `overflow: hidden` restriction.
2. Close-button overlaps Dialog content

Screenshot 2024-07-08 at 11 02 21

---

Two notable changes:

Close-button now has visible padding (when hovered) between itself and the container and content:

Screenshot 2024-07-08 at 11 00 37

Screenshot 2024-07-08 at 11 00 42

---

The Dialog header now gets it's height from the close-button (was previously hard-coded as `spacing-300`

Screenshot 2024-07-08 at 11 00 51

### DIFF
--- a/packages/core/src/dialog/Dialog.css
+++ b/packages/core/src/dialog/Dialog.css
@@ -8,7 +8,7 @@
   margin: auto;
   display: flex;
   flex-direction: column;
-  padding-top: var(--salt-spacing-300);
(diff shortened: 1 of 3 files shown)


tomhazledine:
Closing now that we have a new approach from design: #3839

PR #3824, opened 2024-07-16
Switch Chromatic to run based on labels

joshwooding:


.github/workflows/chromatic.yml
```diff
@@ -60,20 +66,20 @@ jobs:
...
           projectToken: ${{ secrets.CHROMATIC_PROJECT_TOKEN }}
           exitOnceUploaded: true
           # Turbosnap if this is not the main branch
-          onlyChanged: ${{ steps.chromatic_branch.outputs.branchName != 'main' }}
-          externals: ${{ steps.chromatic_branch.outputs.branchName != 'main' && 'packages/(theme)/**' || '' }}
+          onlyChanged: ${{ steps.chromatic_branch.outputs.branchName != 'main' && steps.chromatic_branch.outputs.noTurbosnap != 'true' }}
+          externals: ${{ steps.chromatic_branch.outputs.branchName != 'main' && steps.chromatic_branch.outputs.noTurbosnap != 'true' && 'packages/(theme)/**' || '' }}
           skip: "changeset-release/*"
-          autoAcceptChanges: ${{ steps.chromatic_branch.outputs.branchName == 'main' }}
+          autoAcceptChanges: ${{ steps.chromatic_branch.outputs.branchName == 'main' && steps.chromatic_branch.outputs.noTurbosnap != 'true' }}
```
origami-z, line 80: Only main is not enough?

origami-z (approved):
If we want to reduce run, should the workflow take off Chromatic label automatically?

.github/workflows/chromatic.yml
joshwooding: Storybook with throw an error if an invalid config is passed where autoAcceptChanges is turned on but onlyChanged is false :(

joshwooding:
I'll try remove the label too

shey-v (approved):

mark-tate (approved):

PR #3886, opened 2024-07-24
Fix header text color when column menu is open

origami-z:
Go to any ag grid story, open one of column menu by clicking triple dot on header, hover over the cell to see no header cell color flicking


origami-z:
Cortado Goal:
To be fixed and released

origami-z:
> This breaks the checkbox in the first column

Should be fixed now

joshwooding (approved):

dplsek (approved):

PR #3887, opened 2024-07-24
Add `loading` prop to Button

joshwooding:


packages/core/src/button/Button.tsx
```diff
@@ -43,6 +43,10 @@ export interface ButtonProps extends ComponentPropsWithoutRef<"button"> {
    * The color of the button. Options are 'accent' and 'neutral'.
    */
   color?: ButtonColor;
+  /**
+   * To show a loading spinner.
```
mark-tate, line 47: Note; it does not show a spinner, it enables you to add a spinner (or similar)

mark-tate (approved):

Issue #3738, opened 2024-07-05
Enable theme next features on site


origami-z:
With #3714, we need to think about how to surface those additional styling options on the main site. It could be additional a toggle on the main site, or deploy a copy specifically for theme next, or isolate theme next just to example area.

origami-z:
25 Jul - have some basic idea, going to prototype something and show others

origami-z:
Aug 12 - PR prototype in #3897, needs discussion

origami-z:
Aug 30 - will rearrange toggles from [Figma mock](https://www.figma.com/design/K3i9i0iYoJbdQcchhYXY63?node-id=393-19322#928001921)

mark-tate:
Espresso Goal: PR ready, could be merged once reviewed

link to be shared to team

origami-z:
Code merged in #4082, will be promoted to production during next release

PR #3897, opened 2024-07-25
[Prototype] Enable theme next for site component examples

origami-z:
Adds a few theme next controls to component examples, just like mode and density

There should be some priorities  for the new styling options (accent/corner/headingFont/actionFont) for the user to try out. My first thought is two font options is not high priority (also we don't have font file available on the site anyway)

- [ ] Decide which options to show
- [ ] Decide which control to use, same switch / toggle button 
- [ ] Work on page responsiveness 

#3738


origami-z:
@bhoppers2008 @pseys @dplsek a few outstanding questions to be figured out. let me know your thoughts

origami-z:
Replaced by #4082

Issue #3848, opened 2024-07-19
Update Form Field 'gap' to --spacing-75


bhoppers2008:
### Package name(s)

Core (@salt-ds/core)

### Package version(s)

Gap between Form Label 

### Description

Gap between Form Label and Form Control, and Control and Help/ Status Text in Form Field with top-aligned label is currently --spacing-100 in code. Update to --spacing-75 to match spec.

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #3901, opened 2024-07-26
Update gap for top aligned FormField

Fercas123:
fixes https://github.com/jpmorganchase/salt-ds/issues/3848

### DIFF
--- a/packages/core/src/form-field/FormField.css
+++ b/packages/core/src/form-field/FormField.css
@@ -12,6 +12,7 @@
 }
 
 .saltFormField-labelTop {
+  gap: var(--salt-spacing-75);
   grid-template-areas:
     "label"
     "controls";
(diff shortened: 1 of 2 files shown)


packages/core/src/form-field/FormField.css
```diff
@@ -12,6 +12,7 @@
 }
 
 .saltFormField-labelTop {
+  gap: var(--salt-spacing-75);
```
joshwooding, line 15: Can we move the other gap to labelLeft?

joshwooding (changes requested):
See comment

packages/core/src/form-field/FormField.css
Fercas123: to ```.saltFormField-labelLeft, .saltFormField-labelRight {...}``` ?  shouldn't it be the default?

joshwooding (approved):

yunjungyeh (approved):

PR #3907, opened 2024-07-29
Fix Dialog background scroll

### LINKED ISSUE #3906: Disable background scroll when Dialog open
### Area

UI Components

### The problem

The current behavior of the Dialog component allows for background scrolling when the Dialog is open. This can sometimes lead to unintentionally scrolling the background content while interacting with the Dialog. It would be nice to have an option to disable this. 

### The solution

Add a new prop to Dialog which disables background scrolling. eg `disableBackgroundScrolling: boolean`

### Alternatives and examples

.

liamsms:
Closes #3906

Added `disableScroll` prop to FloatingComponent, which sets the document body overflow to `hidden` to prevent background scrolling.

### DIFF
--- a/packages/core/src/utils/useFloatingUI/useFloatingUI.tsx
+++ b/packages/core/src/utils/useFloatingUI/useFloatingUI.tsx
@@ -22,6 +22,7 @@ import {
   useMemo,
 } from "react";
 import { SaltProvider, SaltProviderNext, useTheme } from "../../salt-provider";
+import { usePreventScroll } from "../usePreventScroll";
 
 export interface FloatingComponentProps
   extends ComponentPropsWithoutRef<"div"> {
@@ -45,6 +46,10 @@ export interface FloatingComponentProps
   width?: number;
   height?: number;
   position?: Strategy;
+  /**
+   * Makes the page unscrollable when the floating component is open.
+   */
+  lockScroll?: boolean;
 }
 
 const DefaultFloatingComponent = forwardRef<
@@ -59,6 +64,7 @@ const DefaultFloatingComponent = forwardRef<
     width,
     height,
     focusManagerProps,
+    lockScroll,
     ...rest
   } = props;
   const style = {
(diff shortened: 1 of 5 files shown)


liamsms:
May require additional testing on a Mac due to potential differences in scroll behavior, eg. shifts in positioning when the scrollbar appears/disappears.

Fercas123:
this change should probably be built as a prop in `FloatingComponent` rather than in dialog, since it could be reused in other implementations like drawer

packages/core/src/utils/useFloatingUI/useFloatingUI.tsx
```diff
@@ -71,6 +77,20 @@ const DefaultFloatingComponent = forwardRef<
 
   const ChosenSaltProvider = themeNext ? SaltProviderNext : SaltProvider;
 
+  useEffect(() => {
+    if (!disableScroll) {
+      return;
+    }
+    if (open) {
+      document.documentElement.style.overflow = "hidden";
```
Fercas123, line 85: we should probably use `targetWindow`

liamsms: Hi - I've updated it to use `targetWindow`.

Fercas123 (approved):

tayyabataimur (approved):

joshwooding (changes requested):
Will follow up with feedback

packages/core/src/utils/useFloatingUI/useFloatingUI.tsx
```diff
@@ -45,6 +46,10 @@ export interface FloatingComponentProps
   width?: number;
   height?: number;
   position?: Strategy;
+  /**
+   * Disables scrolling the document body while the floating component is open.
+   */
+  disableScroll?: boolean;
```
joshwooding, line 52: We should check if we're happy with the naming here

mark-tate: What about 
https://floating-ui.com/docs/floatingoverlay#lockscroll

joshwooding: I think the choices are: lockScroll, preventScroll or disableScroll

tayyabataimur: `lockscroll` is consistent with Floating UI's terminology but will it be intuitive for our users, and be consistent in naming with other components in Salt?

Fercas123 (approved):

PR #3910, opened 2024-07-31
Remove ComboBox's button when it has nothing to show

### LINKED ISSUE #3637: Combo Box would be good to not show empty list
### Area

UI Components

### The problem

When filter combo box into an empty list, top and bottom bottom would still be shown and look very weird (especially when using rounded corner theme next options)

https://storybook.saltdesignsystem.com/?path=/story/core-combo-box--default&globals=themeNext:enable;corner:rounded;accent:teal

### The solution

Not show list when empty

### Alternatives and examples

- 

joshwooding:
Closes #3637

Updated ComboBox to hide it's trigger button when no content is available to show. This covers two main scenarios:

- When a filter has been applied, and nothing matches that filter.
- When there is no initial data present, but it will be fetched after typing.

# To-do
- [x] Fix / Add tests

### DIFF
--- a/.changeset/pink-bears-refuse.md
+++ b/.changeset/pink-bears-refuse.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed ComboBox having a button when it has nothing to show.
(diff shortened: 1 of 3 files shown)


.changeset/pink-bears-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed ComboBox having a button when it has nothing to show.
```
origami-z, line 5: "a button" can be ambiguous ?

origami-z (approved):

.changeset/pink-bears-refuse.md
joshwooding: Any recommendation? "Trigger button", "button adornment"

PR #3970, opened 2024-08-08
Investigate adding an utils package

joshwooding:
Investigate adding a utils package to remove a dependency on "core" from other packages.


origami-z:
like the idea in principle, need to test whether this will solve the problem for our stakeholders. As we understand today, it's dedupe limitation of npm/yarn, may not be solved by utils package

PR #3999, opened 2024-08-16
Fixed the status adornment position in ComboBox where it would always be vertically centered when the field grows

joshwooding:


origami-z (approved):

yunjungyeh (approved):
looks lovely!

Issue #3839, opened 2024-07-17
Header Block component


bhoppers2008:
This issue will standardize a Header Block which can be composed into container components (Dialog, Drawer, Overlay) and provide a Header Block component.

**Requirements**

The Header Block will include the following attributes:

- Default: Accent bar, Title text, Close Button (all optional)
- Supports statuses (iRAG)
-- Header Block won't be opinionated about which components can display its status states. Salt will provide relevant examples of use per component it is used in.
- Padding-top/ -bottom at `--spacing-300`.

**Component updates**

The following components are updated to use the Header Block:

- Dialog
-- This will be the only example that shows status headers.
- Drawer
- Overlay
-- Padding-top/-bottom will be overwritten to `--spacing-200` to match the internal padding of Overlay.

In all cases, if the container element needs to be closed via Close Button, the Header Block close button should be used.

_Consideration: Should a Close Button component be created as it's own individual component?_

**Enhancement**
An overflow drop-shadow will be added at a later point.

Closes #3532
Closes #2970

origami-z:
Need to think through how to play nicely with container's padding. e.g. Dialog has only top/bottom padding to make accent bar work. Drawer & Overlay have full paddings for the content (different spacing as well).

origami-z:
Extend to 2 sprints, given multiple components, and user validation needed

tomhazledine:
Figma spec complete. Ready for dev work to begin.

mark-tate:
Cortado Goal: RC by EOS

dplsek:
@tomhazledine link to Figma [spec](https://www.figma.com/design/8msOmbLvIsTtaIcNcbAxiW/Header-block?node-id=2038-1016&t=xNOcioTQrT2WojUa-1)

origami-z:
Aug 12 - starting dev tomorrow (one minor design update remaining)

dplsek:
Aug 13th - had header block dev kick-off with @tomhazledine @navkaur76 @pseys 

Open questions:
1. if header block were used in overlay, how is it expected to work compared to how it currently does? (close button only vs requiring a header)
2. will header block also be used in status dialogs?
3. need to settle on a concept/name that works within the system for the different options (currently 200 and 300)

@bhoppers2008 @joshwooding would like to discuss more on these

navkaur76:
Meeting w/ @bhoppers2008 on Aug 14

**Answers to Qs in Darrin's comment above:** 

[...]

PR #4026, opened 2024-08-22
Header block

tomhazledine:
Adds a `HeaderBlock` component: issue #3839

### DIFF
--- a/packages/lab/src/header-block/HeaderBlock.css
+++ b/packages/lab/src/header-block/HeaderBlock.css
@@ -18,10 +18,6 @@
   gap: var(--salt-spacing-50);
 }
 
-.saltHeaderBlock .saltStatusIndicator.saltIcon {
-  --icon-size: var(--salt-text-h2-lineHeight);
-}
-
 .saltHeaderBlock-withAccent {
   position: relative;
 }
--- a/packages/lab/src/header-block/HeaderBlock.css
+++ b/packages/lab/src/header-block/HeaderBlock.css
@@ -1,5 +1,6 @@
 .saltHeaderBlock {
-  padding: var(--salt-spacing-300);
+  --headerBlock-padding: var(--salt-spacing-300);
+  padding: var(--headerBlock-padding);
   width: 100%;
   align-items: center;
   display: flex;
@@ -28,9 +29,9 @@
 .saltHeaderBlock-withAccent::before {
   content: "";
   position: absolute;
-  top: var(--salt-spacing-300);
+  top: var(--headerBlock-padding);
   left: 0;
-  bottom: var(--salt-spacing-300);
+  bottom: var(--headerBlock-padding);
   width: var(--salt-size-bar);
   background: var(--salt-accent-background);
 }
@@ -49,4 +50,20 @@
 
 .saltHeaderBlock-warning::before {
   background: var(--salt-status-warning-borderColor);
+}
+
+/* Overrides */
+
+.saltDialog .saltHeaderBlock {
+  margin-top: calc(0px - var(--salt-spacing-300));
+}
+.saltDrawer .saltHeaderBlock {
+  margin-top: calc(0px - var(--salt-spacing-300));
+  margin-left: calc(0px - var(--salt-spacing-300));
+  margin-right: calc(0px - var(--salt-spacing-300));
+  width: 200px;
+}
+
+.saltOverlay .saltHeaderBlock + .saltOverlayPanelContent {
+  padding-top: 0;
 }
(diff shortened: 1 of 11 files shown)


packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+   **/
+  preheader?: string;
+  /**
+   * Description text is displayed just below the header
+   **/
+  description?: string;
+  /**
+   * Padding of the header block. Default is "300"
+   **/
+  padding?: HeaderBlockPadding;
```
origami-z, line 45: This is the first introduction of spacing to any component. Worth a wider discussion around options / syntax (e.g., v.s. `size` in Icon / `gap` in layout component) being multiplier.

tomhazledine: IIRC there was a lot of discussion about this with @navkaur76 and @dplsek, so they might be able to provide more context.

To take the `gap` example as a pattern, this could be switched to `padding: 1 | 2 | 3`?

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+            <Text className={withBaseName("preheader")}>{preheader}</Text>
+          )}
+          <div>{header}</div>
+          {description && !status && (
+            <Text color="secondary" className={withBaseName("description")}>
+              {description}
+            </Text>
+          )}
+        </Heading>
+        {onClose && <HeaderBlockCloseButton onClick={onClose} />}
```
origami-z, line 120: This API feels against the composition approach?

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+   * The status of the Dialog
+   */
+  status?: ValidationStatus | undefined;
+  /**
+   * Displays the accent bar in the Dialog Title */
+  accent?: boolean;
+  /**
+   * Header text
+   */
+  header?: string;
```
origami-z, line 33: optional here? L113 is not optional?

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+    const headerBlockStyles = {
+      ...style,
+      "--headerBlock-padding": `var(--salt-spacing-${padding})`,
+    };
+
+    const status = statusProp ?? statusContext;
+
+    const Heading = padding === "300" ? H2 : H3;
+    const iconSize = padding === "300" ? 1.5 : 1;
```
origami-z, line 84: With the magic logic here, I think we need think about whether we need multiple header implementation

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+export type HeaderBlockPadding = "100" | "200" | "300";
+
+const withBaseName = makePrefixer("saltHeaderBlock");
+
+export interface HeaderBlockProps extends ComponentPropsWithoutRef<"div"> {
+  /**
+   * The status of the Dialog
+   */
+  status?: ValidationStatus | undefined;
```
joshwooding, line 26:
```suggestion
  status?: ValidationStatus;
```

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+   * Displays the accent bar in the Dialog Title */
+  accent?: boolean;
+  /**
+   * Header text
+   */
+  header?: string;
+  /**
+   * Preheader text is displayed just above the header
+   **/
+  preheader?: string;
```
joshwooding, line 37: Why would we only limit this to strings?

packages/lab/src/header-block/HeaderBlock.tsx
```diff
@@ -0,0 +1,124 @@
...
+  /**
+   * Description text is displayed just below the header
+   **/
+  description?: string;
```
joshwooding, line 41: Why would we only limit this to strings?

packages/lab/src/header-block/HeaderBlock.css
```diff
@@ -0,0 +1,62 @@
...
+  width: 100%;
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: stretch;
+  gap: var(--salt-spacing-100);
+  box-sizing: border-box;
+}
+
+.saltHeaderBlock-body {
```
joshwooding, line 13: Is "body" the correct way to refer to this?

joshwooding:
@tomhazledine I guess this can be closed now?

PR #4038, opened 2024-08-28
Fix Dialog to have `--salt-zIndex-modal`

origami-z:
Dialog to have modal z-index.
Scrim remains as Drawer zindex - 1, given Drawer also uses it.


site/propsGen.js
```diff
@@ -1,6 +1,8 @@
 const docgen = require("react-docgen-typescript");
```
joshwooding, line 1: Were these changes intentional?

origami-z: nope... i'll remove

joshwooding (approved):

Issue #3922, opened 2024-08-01
Ag grid theme copy cell background color is flipped


origami-z:
### Package name(s)

AG Grid Theme (@salt-ds/ag-grid-theme)

### Package version(s)

v2.0.1

### Description

When copy cell in dark mode, background flips to white (the blink), was not the case before v2. 

In light mode, it's flipped to dark.

### Steps to reproduce

1. Go to https://www.saltdesignsystem.com/salt/components/ag-grid-theme/examples#variants
2. Set to dark mode
3. Copy any cell (ctrl + c)
4. Observe the blinking color is white

### Expected behavior

Possible solution: `--ag-range-selection-highlight-color: var(--salt-overlayable-rangeSelection);`, or create a new token to be more prominent 

https://www.ag-grid.com/react-data-grid/global-style-customisation-selections/#cell-selections

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Edge

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

origami-z:
Cortado Goal:
To be fixed and released

origami-z:
Aug 12 - found potentail color, @origami-z  to work with @dplsek

origami-z:
Prototype stackblitz - https://stackblitz.com/edit/salt-ag-grid-v31-theme-v2-lkzmse?file=style.css,App.jsx

joshwooding:
Aug 20 - Work to resume after button work is done

origami-z:
New prototypes
- Salt current: https://stackblitz.com/edit/salt-ag-grid-v31-theme-v2-rx4qkp?file=App.jsx,style.css 
- Salt next: https://stackblitz.com/edit/salt-ag-grid-v31-theme-v2-lwyexe?file=style.css,App.jsx

dplsek:
We will be using blue 500/50 in Salt CURRENT and blue/teal 700/300 in Salt NEXT for the flashing copy state. Links included above to the coded examples and the corresponding Figma file with contrast ratios for text can be found [here](https://www.figma.com/design/n2I0jXlzOYZWjxTxkWkNPk/AG-Grid-theme-copy-cell-background-color?node-id=0-1&t=exTdKo9B6Mo2OYuC-1)

[...]

PR #4073, opened 2024-09-10
Add ag grid copy highlight color

origami-z:
Copy cells to see highlight color

Before release, double check `peerDependencies` is bumped in `@salt-ds/ag-grid-theme`

#3922


pseys (approved):

joshwooding (approved):
I can't tell if the token naming makes sense but the visual issue is fixed

PR #4091, opened 2024-09-13
Fix standalone ToggleButton corner radius

origami-z:
When standalone toggle button is used, it should have same corner radius as button. Only when it is used within toggle button group, it should change to match "nested" corner.

[#4140](https://github.com/jpmorganchase/salt-ds/issues/4140)

### DIFF
--- a/.changeset/heavy-mice-repeat.md
+++ b/.changeset/heavy-mice-repeat.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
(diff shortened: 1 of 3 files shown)


origami-z:
Take out from sprint, given design changes is also needed. Link to [#4140](https://github.com/jpmorganchase/salt-ds/issues/4140) for work to be scheduled together

joshwooding (approved):

dplsek (approved):

PR #4107, opened 2024-09-16
autoHideDuration

duke79:


joshwooding:
Hi @duke79, thanks for the contribution.

Could you create a feature request issue to document why this is needed and what usecases this will support?
I'm not sure Toast should be in charge of showing/hiding itself but admit this talks about a gap we have in our offering.

I'll close this PR for now until we have that discussion.

Thanks,
Josh

Issue #4041, opened 2024-08-28
JSON representation for theme


origami-z:
Catch all issue

```[tasklist]
### Tasks
- [ ] Create JSON representation 
- [ ] Generate CSS to replace most of existing theme package content
- [ ] Publish JSON as part of npm release
- [ ] Write documentation how to integrate with 3rd party libraries (e.g. vanilla-extract for Mosaic, theme visualization)
- [ ] Lint tool to check against Figma library (variables)
```

Some considerations 
- Theme vs Theme next
- Potential different folder / file structure ?
- Handle deprecation
- Token naming difference between Figma and CSS

Prototype in #3143

mark-tate:
Latte Goal: review current state, kickoff to understand requirements at EOS

origami-z:
Oct 17: start in w2, after ag grid work

origami-z:
Oct 21: scheduled a scope meeting on 28 Oct, start looking at dependencies / feasibility this week (e.g. styled dictionary v4 release)

mark-tate:
Lungo Goal: 
document what the JSON representation will NOT do

origami-z:
Met w/ @mark-tate @joshwooding @pseys @dplsek John on Monday

Primary goal for the JSON exercise is for safety and quality control, which would lay the foundation for further expansion for other BYOT requirements
-	For Salt, we can check between synchronization Figma and Code. e.g., See “Would have avoided” in GH issue description above. 
    - We don't want to change existing process (e.g., editing Figma variables via Figma UI, Figma only variables) until risk and new process of syncing between Figma/Code can be fully understood. So the first step would be creating some "lint" tool to warn the user if anything is different is detected, e.g., as a Figma plugin, custom CSS stylelint
    - We should be able to produce CSS code via JSON, instead of manually construct them like now 
    - Further usage could include theme configuration or virtualization tools (e.g., [prototype using spectrum tokens](https://origami-z.github.io/salt-ds-tokens/?filter=light%2Cmedium&remoteJsonUrl=https%253A%252F%252Fraw.githubusercontent.com%252Fjpmorganchase%252Fsalt-ds%252Fcae97dfe8b05fb91baef578f8ca04b25315fdc5e%252Fpackages%252Ftheme%252Fjson%252Ftheme.json&token=palette%2Cneutral))
[...]

PR #4133, opened 2024-09-18
Fix foreground secondary color token in Salt theme next

origami-z:
Code and Figma got out of sync during early salt next design phase.

To fully prevent this, #4041 is needed.


joshwooding (approved):

origami-z:
Some secondary text is not captured in Chromatic, captured in #4134

dplsek (approved):

PR #4135, opened 2024-09-18
Site quick responsiveness improvement

origami-z:
- Mobile drawer menu should be better now with global density change + [mosaic#663](https://github.com/jpmorganchase/mosaic/pull/663)
- Component page body & TOC spacing should be improved for viewport just above tablet
- Fixed component nav bar, so scrolling down would see last entry "Tooltip", before scrolling the whole page down

App header is constrained by Mosaic `LayoutBase`, with existing content, we'll need 5 specific breakpoints
- 1478+, full logo + menu items + search + buttons
- 1380+, mobile logo
- 1280+, remove buttons
- remaining, replace menu items with drawer

### DIFF
--- a/site/src/pages/_app.tsx
+++ b/site/src/pages/_app.tsx
@@ -20,8 +20,9 @@ import { StoreProvider, useCreateStore } from "@jpmorganchase/mosaic-store";
 import { ssrClassName } from "@jpmorganchase/mosaic-theme";
 import { themeClassName } from "@jpmorganchase/mosaic-theme";
 import {
+  type Density,
   SaltProvider,
-  UNSTABLE_SaltProviderNext,
+  SaltProviderNext,
   useCurrentBreakpoint,
 } from "@salt-ds/core";
 import clsx from "clsx";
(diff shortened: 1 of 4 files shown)


site/src/pages/_app.tsx
```diff
@@ -107,6 +101,61 @@ const amplitude = localFont({
...
+}: ThemeProviderProps) {
+  const hasHydrated = useHasHydrated();
+  const colorMode = useColorMode();
+
+  const ssrClassname = hasHydrated ? undefined : ssrClassName;
+
+  const ChosenSaltProvider = themeNext
+    ? UNSTABLE_SaltProviderNext
+    : SaltProvider;
+  const viewport = useCurrentBreakpoint();
```
joshwooding, line 139: I don't think this is working, the default site always returns touch

origami-z: Good catch, it's to do with hook on the same level of provider. Fixed with a workaround.

joshwooding (approved):

dplsek (approved):

PR #4136, opened 2024-09-18
Optional add a nvmrc file for deeper shell intergration

6d70:
I use [NVM with deeper shell integration](https://github.com/nvm-sh/nvm), which automatically switches the Node.js version when the nvmrc file is detected. This feature is particularly useful when working on multiple projects or exploring various forks, such as this one.

However, the documentation does not specify the required Node.js version. Based on the package.json file, which mentions "@types/node": "^20.0.0", I assumed Node.js version 20 is needed.

Feel free to close this PR if you do not like. 

image


.nvmrc
```diff
@@ -0,0 +1 @@
+20
```
6d70, line 1: We can fix this but not fixed in pacakge json so just matching that. 

e.g. 
```
20.16
```
or
```
20.16.0
```
or

joshwooding:
Hi @6d70 thanks for the contribution, I don't think we need to pick a specific version of node for the project. We tend to support all active LTS versions of Node e.g. 18, 20, 22 but maybe we'll revisit this in the future.

I'll close this PR for now but thanks again.

PR #4138, opened 2024-09-19
New Header for Overlay and Dialog

tomhazledine:
* Adds new `OverlayHeader` component to replace deprecated `OverlayPanelCloseButton`
* Updates `DialogHeader` to match updated Figma spec and deprecate `DialogCloseButton`

### DIFF
--- a/packages/core/src/overlay/OverlayHeader.css
+++ b/packages/core/src/overlay/OverlayHeader.css
@@ -1,0 +1,31 @@
+.saltOverlayHeader {
+  padding: var(--salt-spacing-100);
+  width: 100%;
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: stretch;
+  gap: var(--salt-spacing-100);
+  box-sizing: border-box;
+}
+
+.saltOverlayHeader-body {
+  flex-grow: 1;
+  margin: 0;
+  display: flex;
+  flex-direction: column;
+  gap: var(--salt-spacing-50);
+}
+
+.saltOverlayHeader-header {
+  margin: 0;
+}
+
+.saltOverlayHeader-endAdornmentContainer {
+  align-self: flex-start;
+}
+
+/* Overrides */
+.saltOverlayHeader ~ .saltOverlayPanelContent {
+  padding-top: 0;
+}
--- a/packages/core/src/overlay/OverlayHeader.css
+++ b/packages/core/src/overlay/OverlayHeader.css
@@ -9,7 +9,7 @@
   box-sizing: border-box;
 }
 
-.saltOverlayHeader-body {
+.saltOverlayHeader-container {
   flex-grow: 1;
   margin: 0;
   display: flex;
@@ -21,7 +21,7 @@
   margin: 0;
 }
 
-.saltOverlayHeader-endAdornmentContainer {
+.saltOverlayHeader-actionsContainer {
   align-self: flex-start;
 }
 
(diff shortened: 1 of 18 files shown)


packages/core/src/overlay/OverlayHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+import overlayHeaderCss from "./OverlayHeader.css";
+
+const withBaseName = makePrefixer("saltOverlayHeader");
+
+export interface OverlayPanelContentProps
+  extends ComponentPropsWithoutRef<"div"> {
+  /**
+   * Description text is displayed just below the header
+   **/
+  description?: string;
```
tomhazledine, line 20: There was a question on the previous PR about why `description|header|preheader` were limited to strings. (same applies to DialogHeader)

IMO string is the right choice here. The design is very deliberate about only allowing text content in these sections, and I don't think there's a scenario where we would want users to be able to inject any component they like here.

@joshwooding @origami-z @navkaur76 thoughts?

packages/core/src/overlay/OverlayHeader.css
```diff
@@ -0,0 +1,31 @@
...
+  width: 100%;
+  align-items: center;
+  display: flex;
+  flex-direction: row;
+  justify-content: stretch;
+  gap: var(--salt-spacing-100);
+  box-sizing: border-box;
+}
+
+.saltOverlayHeader-body {
```
tomhazledine, line 12: @joshwooding following up on question from previous PR: *"Is "body" the correct way to refer to this?"*

I think it is, as it's the body of the header (multiple text elements that are wrapped together, discrete from the end adornment). Not a strongly-help opinion, though, so happy to hear counter arguments 😄

packages/core/src/overlay/OverlayHeader.tsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * End adornment component
+   */
+  endAdornment?: ReactNode;
```
tomhazledine, line 24: Using `endAdornment` to be consistent with other components. Are there any strong opinions about the name? @joshwooding @origami-z

packages/core/src/overlay/OverlayHeader.tsx
origami-z: In most cases, pure string would be provided anyway. However, in case teams want to put additional layout, icon or other relevant stuff, we could try to accommodate? A few other components does this

joshwooding: I would rather use ReactNode or something similar. It's not worth being restrictive

packages/core/src/overlay/OverlayHeader.css
joshwooding: It needs to be aligned with design @navkaur76

packages/core/src/overlay/OverlayHeader.tsx
joshwooding: adornment is really only form controls previously... how does design refer to this? @navkaur76

packages/core/src/overlay/OverlayHeader.css
navkaur76: In figma, the frame containing the main heading, pre-header and description is called "Block", if that helps? The accent bar sits outside of this frame.

origami-z: Closest block in the system is in stepper input?

packages/core/src/overlay/OverlayHeader.tsx
origami-z: Adornment is also used in grid corner triangle? @bhoppers2008

joshwooding:
Replaced by other PRs.

PR #4143, opened 2024-09-21
DatePicker and Calendar updates

mark-tate:
- `Calendar` now supports children, enabling you to compose Calendars from `CalendarNavigation`, `CalendarWeekHeader` and `CalendarDateGrid`
- renamed `CalendarCarousel` to `CalendarDateGrid`
- better support for rest so that it passes to the root element in each component
- hooks are only passed values they support
- switched date format in examples to all use DD MMM YYYY
- DatePicker examples with US locale, can now parse dates with or without a comma
- updated docs with API changes


origami-z (commented):
In Single Controlled story, if I enter "22", `onSelectedDateChange` is returning `null, false`, but Input is changed to a date "31 Dec 21"

origami-z (approved):
Another observation from the support query, when invalid date is entered twice to input, `onSelectedDateChange` is called only once

mark-tate:
> In Single Controlled story, if I enter "22", `onSelectedDateChange` is returning `null, false`, but Input is changed to a date "31 Dec 21" (this doesn't happen in Chrome, but happening in Safari), this is due to `parseCalendarDate`'s usage of `new Date(inputDate)`. I felt this is the breaking point of `@internationalized/date` not supporting out of box parsing

i think we  change the default parser to only accept iso date format and not numbers. Like the idea of adapter API longer term and removing any dependency on third party library

joshwooding (approved):

mark-tate:
to keep the PR small, a following PR will address
- added a fix for CalendarWeekHeader which resolves locale issues
- new default parser that would see numbers as an error
- onSelectedDateChange becomes onSelectionChange
- CalendarDateGrid renamed to CalendarGrid
will look to invalid selection on this following PR as well

joshwooding (approved):

PR #4241, opened 2024-10-06
`onDateChange` on consecutive invalid dates

origami-z:
From support email, this is a valid use case (it's "different" invalid dates) before asking user to listen to low level input change event themselves


mark-tate:
superseded by https://github.com/jpmorganchase/salt-ds/pull/4291

PR #4294, opened 2024-10-14
Add scroll detection to OverlayPanelContent and DialogContent

Fercas123:


packages/core/src/dialog/DialogContent.tsx
```diff
@@ -37,14 +42,35 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
+    useEffect(() => {
+      const checkOverflow = () => {
+        if (!divRef.current) return;
+        setIsOverflowing(
+          divRef.current.scrollHeight > divRef.current.clientHeight,
+        );
+      };
+      checkOverflow();
+
+      window.addEventListener("resize", checkOverflow);
```
origami-z, line 54: `targetWindow`?

packages/core/src/dialog/DialogContent.tsx
```diff
@@ -37,14 +42,35 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
       window: targetWindow,
     });
 
+    useEffect(() => {
+      const checkOverflow = () => {
+        if (!divRef.current) return;
+        setIsOverflowing(
+          divRef.current.scrollHeight > divRef.current.clientHeight,
```
origami-z, line 49: Haven't look too deep, worth to confirm, `clientHeight` vs `offsetHeight`

packages/core/src/dialog/DialogContent.tsx
```diff
@@ -37,14 +42,35 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
       window: targetWindow,
     });
 
+    useEffect(() => {
```
origami-z, line 45: Is this a candidate for layout effect for DOM measuring?

packages/core/src/dialog/DialogContent.tsx
```diff
@@ -37,14 +42,35 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
+
     return (
       <>
         <div className={clsx({ [withBaseName("scroll")]: scrollTop > 0 })} />
         <div
-          className={clsx(withBaseName(), className)}
+          className={clsx(
+            withBaseName(),
+            {
+              [withBaseName("overflow")]: isOverflowing,
```
origami-z, line 67: Does `isOverflowing` serves the same condition as `scrollTop > 0` above ?

origami-z (commented):
Needing test and changeset

packages/core/src/dialog/DialogContent.tsx
Fercas123: not quite, isOverflowing checks if the content wraps into overflow since loading, scrollTop checks where the scroll is to add the top shadow... but I'll see if i can do without it

Fercas123: couldnt get rid of, but I optimised the  scrolled check

packages/core/src/__tests__/__e2e__/dialog/Dialog.cy.tsx
```diff
@@ -184,4 +184,13 @@ describe("GIVEN a Dialog", () => {
       cy.findByRole("button", { name: "Next" }).should("be.focused");
     });
   });
+  describe("WHEN overflowing content is detected", () => {
+    it("THEN it should add padding to the right of the scroll bar", () => {
```
joshwooding, line 188: Is this better shown as a visual snapshot?

packages/core/src/dialog/DialogContent.tsx
```diff
@@ -24,11 +31,15 @@ export interface DialogContentProps extends HTMLAttributes<HTMLDivElement> {
...
 
-    const handleScroll = (event: UIEvent<HTMLElement>) => {
-      setScrollTop(event.currentTarget.scrollTop);
-    };
+    const divRef = useRef<HTMLDivElement>(null);
+    const containerRef = useForkRef(divRef, ref);
+    const handleScroll = debounce(() => {
+      if (!divRef.current) return;
+      setScrolled(divRef.current.scrollTop > 0);
+    }, 50);
```
joshwooding, line 42: 50 is quite fast, was this chosen specifically?

packages/core/src/overlay/OverlayPanelContent.tsx
```diff
@@ -31,10 +40,48 @@ export const OverlayPanelContent = forwardRef<
...
-    <div className={clsx(withBaseName(), className)} {...rest} ref={ref}>
-      {children}
-    </div>
+    <>
+      <div
+        className={clsx(withBaseName("separator"), {
+          [withBaseName("scroll")]: scrolled,
+        })}
+      />
+      <div className={clsx(withBaseName("container"))}>
```
joshwooding, line 69: Unsure about the change in the order of classes... Typically withBaseName() is called on the root element but not sure if we already do this elsewhere

joshwooding, line 69:
```suggestion
      <div className={withBaseName("container")}>
```

packages/core/stories/dialog/dialog.stories.tsx
```diff
@@ -84,11 +84,11 @@ const DialogTemplate: StoryFn<
           <UnmountLogger />
         </DialogContent>
         <DialogActions>
-          <Button variant="secondary" onClick={handleClose}>
+          <Button appearance="transparent" onClick={handleClose}>
```
joshwooding, line 87: I think the PR needs to be rebased? Surprised these changes are here

.changeset/silent-ties-sing.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Added right padding in `OverlayPanelContent` between content and scrollbar when overflow
```
joshwooding, line 5:
```suggestion
Added padding between the content and scrollbar when `OverlayPanelContent` is scrollable.
```
?

packages/core/src/dialog/DialogContent.tsx
Fercas123: i was using the default but it is looks very laggy

packages/core/src/__tests__/__e2e__/dialog/Dialog.cy.tsx
Fercas123: the snapshot surfaced the change but the scrollbar doesn't show so it is not super evident, but i can remove this ones if you think they are redundant

packages/core/src/dialog/DialogContent.tsx
joshwooding: Have you tried without debouncing? Throttling may be more valid.

packages/core/src/__tests__/__e2e__/dialog/Dialog.cy.tsx
joshwooding: Happy for this to still be here then

joshwooding (dismissed):

packages/core/src/overlay/OverlayPanelContent.css
```diff
@@ -1,6 +1,28 @@
...
-  overflow: auto;
+  overflow: hidden;
   padding: var(--saltOverlay-content-padding, var(--salt-spacing-100));
 }
+
+.saltOverlayPanelContent-container {
+  overflow-y: auto;
+}
+
+.saltOverlayPanelContent-separator {
```
Fercas123, line 12:
```suggestion
.saltOverlayHeader ~ .saltOverlayPanelContent-separator {
```

joshwooding (approved):

packages/core/stories/overlay/overlay.stories.tsx
```diff
@@ -110,12 +110,10 @@ export const LongContent = () => {
       <OverlayPanel
         style={{
           width: 300,
-          height: 200,
```
origami-z, line 113: This feels a worth callout?

origami-z (approved):

packages/core/stories/overlay/overlay.stories.tsx
joshwooding: I think this would still work?

origami-z: i don't know, didn't try. Just thinking if that's the code change needed to be done, then we need to mention it

Fercas123: it is not changing the behaviour, but if the user sets height here then the paddings are not respected, so i just moved it to the content so the overflow is in the content, not in the panel and improve the example... but this would still work as before

PR #4298, opened 2024-10-15
Add DialogHeader to lab

Fercas123:


.changeset/plenty-monkeys-give.md
```diff
@@ -0,0 +1,16 @@
+---
+"@salt-ds/lab": minor
+---
+
+Added `DialogHeader` component to lab.
+
+```typescript
+  <Dialog open={open} onOpenChange={onOpenChange} id={id}>
+    <DialogHeader header="Terms and conditions" actions={<CloseButton />}/>
```
origami-z, line 9: Is this CloseButton existing core component or should be a generic button with close icon

packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -0,0 +1,433 @@
...
+        <DialogActions>
+          <Button>Cancel</Button>
+          <Button sentiment="accented">Ok</Button>
+        </DialogActions>
+      </FakeWindow>
+    </StackLayout>
+  );
+};
+
+export const StickyFooter: StoryFn<typeof Dialog> = ({
```
origami-z, line 378: Not too sure what's special for "Sticky" in this example

packages/lab/src/dialog/DialogHeader.css
```diff
@@ -0,0 +1,63 @@
...
+  content: "";
+  position: absolute;
+  top: 0;
+  left: 0;
+  bottom: var(--salt-spacing-300);
+  width: var(--salt-size-bar);
+  background: var(--salt-accent-background);
+}
+
+.saltDialogHeader-error::before {
```
origami-z, line 49: Not too sure what `::before` is setting, i don't see the accent bar when validation status is set?

packages/lab/src/dialog/DialogHeader.tsx
```diff
@@ -0,0 +1,104 @@
...
+        ref={ref}
+        {...rest}
+      >
+        {status && <StatusIndicator status={status} />}
+        <div className={withBaseName("container")}>
+          <div>
+            {preheader && (
+              <Text className={withBaseName("preheader")}>{preheader}</Text>
+            )}
+            <H2 className={withBaseName("header")}>{header}</H2>
```
origami-z, line 90: Similar heading hierarchy question

origami-z (commented):
Needing QA story..?

packages/lab/stories/dialog/dialog.qa.stories.tsx
```diff
@@ -0,0 +1,59 @@
+import { Dialog } from "@salt-ds/core";
+import { DialogHeader } from "@salt-ds/lab";
+import type { Meta, StoryFn } from "@storybook/react";
+import { QAContainer, type QAContainerProps } from "docs/components";
+import "./dialog.stories.css";
```
Fercas123, line 5:
```suggestion
```

packages/lab/src/dialog/DialogHeader.tsx
Fercas123: updated the way we pass headers for both dialog and overlay headers so we don't constrain the components by wrapping a node in a header tag

packages/lab/stories/dialog/dialog.stories.tsx
```diff
@@ -0,0 +1,222 @@
...
+  type ReactNode,
+  useEffect,
+  useState,
+} from "react";
+
+export default {
+  title: "Lab /Dialog Header",
+  component: Dialog,
+  args: {
+    preheader: <h2>Settlements</h2>,
```
Fercas123, line 25:
```suggestion
    preheader: "Settlements",
```

.changeset/plenty-monkeys-give.md
```diff
@@ -0,0 +1,28 @@
+---
+"@salt-ds/lab": minor
+---
+
+Added `DialogHeader` component to lab.
```
joshwooding, line 5: Can you add a description why it exists please

packages/lab/src/dialog/DialogHeader.css
```diff
@@ -0,0 +1,47 @@
+/* Styles applied to the root element */
+.saltDialogHeader {
```
joshwooding, line 2: If there are multiple dialog headers mounted this will cause an issue. Should we worry?

site/docs/components/dialog/examples.mdx
```diff
@@ -134,5 +134,11 @@ This forces a user to action something within the dialog.
 
 Use the `disableScrim` prop to prevent the scrim from rendering.
 
+</LivePreview>
+<LivePreview componentName="dialog" exampleName="WithHeader" >
+  ##  🚧 With header
```
joshwooding, line 139: Other examples have a header too, we should add why this exists.

packages/lab/src/dialog/DialogHeader.css
Fercas123: :S -next?

Fercas123: it would change bottom padding for the old ones from 100 to 300

amritadesmet (dismissed):

origami-z (commented):
Code looks fine. Will tick once Nav reviewed this. 

I think we need to talk about migration path (ad code diff vs current approach) for the user around it. Is it true that the only/major change is the additionally `actions` prop? If so, when this works, I guess we're not moving this lab header to core, but rather deprecate `DialogActions` in core and add `actions` prop to the existing Dialog header?

navkaur76 (approved):

origami-z (approved):

PR #4331, opened 2024-10-22
Biome updates

joshwooding:
Updates Biome and turns on noUnusedImports


mark-tate (approved):

origami-z (approved):
I often find Biome rules slows down development given format on save feature doesn't work well with auto fix. hopefully it can be improved over time

joshwooding:
> I often find Biome rules slows down development given format on save feature doesn't work well with auto fix. hopefully it can be improved over time

This release includes some fixes there, could you check if it improves?

PR #4276, opened 2024-10-08
Add ag grid v32 support

origami-z:
On top of #4257

Upgrade to v32 to support 

- New active filter icon style
- New default menu style (not tabbed menu), see [doc](https://www.ag-grid.com/react-data-grid/column-menu/#legacy-tabbed-column-menu)

Todos

- [x] filter icon active state (both in floating filter and column header cell), when filter is active, need to show some visual indicator (e.g. [ag grid examples](https://www.ag-grid.com/example/) shows a light blue background and dot badge). In this PR, it's implemented match closely to Salt toggle button active state.  #4275
    - [x] MTK feedback on different filter icon used (clear vs filled). ~~Preview available in #4335~~
- [x] filter panel apply (blue) / reset (gray) button, changed from `ref` attribute to `data-ref`. 
- [x] header sort icon should be next to filter / menu icon
    - [x] double check v31 regression
- [x] checkbox column (e.g. context menu story) shouldn't have filter
- [x] context menu size larger due to `ag-size` change
- [x] ~~mention `suppressMenuHide` [Customising the Column Menu](https://www.ag-grid.com/javascript-data-grid/column-menu/#customising-the-column-menu)~~ added to one site example
---

Snapshot test on StackBlitz

- v31 backwards compatibility: https://stackblitz.com/edit/salt-ag-grid-v31-theme-v2-oxpnan?file=package.json,App.jsx
- v32 new upgrade: https://stackblitz.com/edit/salt-ag-grid-v32-theme-v2-5m5lcf?file=package.json,App.jsx

PR #4335, opened 2024-10-23
Change filter icon to clear vs filled

origami-z:
Based on #4276

Test feedback on filter icon change from MTK


origami-z:
Will continue in #4276

Issue #4029, opened 2024-08-23
Document relationship between status and readonly / disabled


origami-z:
There are a few back and forth on when should status (icon) be visible when a component is readonly and disabled. We landed on 

- disabled: don't show status
- readonly: show status

This can go under About > Glossary before finding a better home

Some relevant wording

> Rationale is based on accessibility. If something is disabled there is no need for it to be 'visible' in the UI–it's irrelevant to a workflow. That contradicts 'status' which is very relevant to a workflow. In contrast, read-only is relevant information/data that i can't control but which is applicable to the flow. It needs to meet accessibility requirements. It's appropriate to therefore allow that state to communicate status if required as well.

mark-tate:
Galao Goal: to be defined/published

mark-tate:
Latte Goal: Ben/Amrita to draft a definition

dplsek:
Extending into Lungo pending review with @bhoppers2008 @amritadesmet

amritadesmet:
**Update:** Created PR #4348 

If this definition needs edits or anchor links to a different part of the site, let me know. @bhoppers2008 @origami-z 

**Status icon visibility**

For disabled components, status icons are not displayed. This approach is based on accessibility principles, as disabled elements don't require visibility in the UI.

In contrast, read-only components display status icons. These elements contain relevant information applicable to the workflow. Displaying status icons ensures accessibility and communicates necessary information.

mark-tate:
Lungo Goal: published this sprint

bhoppers2008:
@mark-tate Comment added to #4348 for @amritadesmet @jake-costa to update.

amritadesmet:
**Update**: Had call with @jake-costa to summarize @bhoppers2008 's new descriptions. 

**Next step:** Review + merge

amritadesmet:
Squashed & merged

PR #4348, opened 2024-10-28
Update glossary with status/readonly definition.mdx

amritadesmet:
Following request in GH issue #4029


site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,10 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+
+## Status icon visibility
```
joshwooding, line 41: I thought the docs were about status as a whole? Not just status icons?

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,10 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+
+## Status visibility
+
+For disabled components, status information is not displayed. This approach is based on accessibility principles, as disabled elements don't require visibility in the UI.
+
+In contrast, read-only components display status information. These elements contain relevant information applicable to the workflow. Displaying status ensures accessibility and communicates necessary information.
```
joshwooding, line 45: Not entirely sure accessibility fits in here... maybe another one for @jake-costa to comment on.

bhoppers2008:
@amritadesmet @joshwooding @jake-costa 

Let's make the guidance higher level, and then describe in context.

**Disabled state**

- A component in a disabled state is not considered relevant to the users current workflow. 
- User doesn't need to know the component is there.
- Doesn't need to meet min-contrast requirements.
- Better UI pattern could be progressive reveal.
- [Button](https://www.saltdesignsystem.com/salt/components/button/accessibility#focusable-disabled-button) can make exception with `focusableWhenDisabled` prop
- No system support for status messages in disabled state–disabled is not relevant to workflow vs RAG states that need to be compliant/ are relevant to a current flow and need recognition/ further action.
- If relevant to workflow consider read-only is more appropriate.

**Read-only state**

- A component in a read-only state is considered relevant to the users workflow. 
- A data point that that influences the flow but user cannot edit.
- Restricted to user or system permissions.
- User needs to know the component is there.
- Does need to meet accessibility requirements (min-contrast)
- System support for status messages in read-only state–a RAG state (error) may be driven by data manipulation in current/peripheral flow so needs to be visible for recognition/ further action.
- If not relevant to workflow consider disabled is more appropriate.

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,12 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+
+## Disabled state 
+
+A component in a disabled state is not considered relevant to the user’s workflow, as the user doesn’t need to know the component is there. It doesn't need to meet minimum contrast requirements, and a progressive reveal might be a better UI pattern. While buttons can use the focusableWhenDisabled prop, the system doesn't support status messages for disabled components. If a component is relevant, consider using a read-only state instead.
```
joshwooding, line 43: We typically prefer short hand e.g. isn't

joshwooding, line 43:
```suggestion
A component in a disabled state isn't considered relevant to the user’s workflow, as the user doesn’t need to know the component is there. It doesn't need to meet minimum contrast requirements, and a progressive reveal might be a better UI pattern. While buttons can use the focusableWhenDisabled prop, the system doesn't support status messages for disabled components. If a component is relevant, consider using a read-only state instead.
```

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,12 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+
+## Disabled state 
+
+A component in a disabled state isn't considered relevant to the user’s workflow, as the user doesn’t need to know the component is there. It doesn't need to meet minimum contrast requirements, and a progressive reveal might be a better UI pattern. While buttons can use the focusableWhenDisabled prop, the system doesn't support status messages for disabled components. If a component is relevant, consider using a read-only state instead.
```
origami-z, line 43:
```suggestion
A component in a disabled state isn't considered relevant to the user’s workflow, as the user doesn’t need to know the component is there. It doesn't need to meet minimum contrast requirements, and a progressive reveal might be a better UI pattern. While buttons can use the `focusableWhenDisabled` prop, the system doesn't support status messages for disabled components. If a component is relevant, consider using a read-only state instead.
```

origami-z (dismissed):

origami-z (approved):

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,11 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+## Disabled state
+
+A component in a disabled state isn't considered relevant to the user’s workflow, as the user doesn’t need to know the component is there. It doesn't need to meet minimum contrast requirements, and a progressive reveal might be a better UI pattern. While buttons can use the `focusableWhenDisabled` prop, the system doesn't support status messages for disabled components. If a component is relevant, consider using a read-only state instead.
```
bhoppers2008, line 42: Per @joshwooding feedback, have suggested alternative wording. Please alter as needed. Might want to revisit the read-only text as well.

A component in a disabled state isn't considered relevant to the user’s workflow _at that time_. As the function is not available to the user, the component doesn't need to meet minimum contrast requirements. Consider if a disabled state is the right UI pattern, or if a progressive reveal might be more appropriate. Salt doesn't provide error, warning or success states for disabled components–the status messaging implies a user action is required and needs to be addressed. In this case consider using a read-only state instead. Some components may make individual exceptions, Button has a `focusableWhenDisabled` prop so it can be discovered where context is important.

joshwooding: Is it worth highlighting the focusableWhenDisabled also allows a tooltip to be used to explain why the action is disabled?

yunjungyeh (dismissed):

joshwooding (changes requested):
Ben and my feedback needs to be addressed/acknowledged before this is merged

site/docs/about/glossary.mdx
mark-tate: bumping  the above... we should look at before merge

mark-tate: Is the key differentiator focus?
Should the glossary be a succinct explanation, rather than suggesting alternative patterns, progressive reveal and focusWhenDisabled

Otherwise we may need another glossary item for progressive reveal or links to props.

This was LLM generated but you get the gist.

Read-Only

A read-only state in UI components allows users to view and interact with the content without modifying it. Users can select and copy text from read-only components, making it possible to use the information elsewhere. This state is typically used for displaying information that should not be altered, such as view-only fields or summaries. Visual indicators may be used to signal that the component is not editable, while still allowing for actions like text selection and copying.

Disabled

A disabled state in UI components renders them completely inactive and non-interactive. Users cannot focus on, select, or interact with disabled components, which includes the inability to copy text from them. This state is often used to indicate that an action is temporarily unavailable or not applicable, such as a button that is inactive until certain conditions are met. Disabled components are typically styled with visual cues, such as grayed-out text or reduced opacity, to signal their inactive status.

amritadesmet: Incorporating all of these versions to make a high level definition. For future consideration, do we want to include props in the glossary? How detailed / advanced should definitions in the glossary be? 

Currently I notice that the glossary provides a foundational understanding of the terms. Should users reference additional documentation if they're seeking more detailed information? If not, we may consider adding more terms here such as progressive reveal or RAG state

origami-z (approved):

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,11 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+## Disabled state
+
+A component in a disabled state isn't considered relevant to the user’s workflow at that time. As the function isn't available to the user, the component doesn't need to meet minimum contrast requirements. Consider if a disabled state is the right UI pattern, or if a progressive reveal might be more appropriate. Salt doesn't offer error, warning, or success states for disabled components, as these imply required user action; consider using a read-only state or progressive reveal instead. Some components have exceptions, such as Button, so it can be discovered where context is important.
```
joshwooding, line 42:
```suggestion
A component in a disabled state isn't considered relevant to the user’s workflow at that time. As the function isn't available to the user, the component doesn't need to meet minimum contrast requirements. Consider if a disabled state is the right UI pattern, or if a progressive reveal might be more appropriate. Salt doesn't offer validation states (error, warning, or success) for disabled components, as these imply required user action; consider using a read-only state or progressive reveal instead. Some components have exceptions, such as Button, so it can be discovered where context is important.
```
Buttons have exceptions to what?

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,11 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
...
+
+## Read-only state
+
+A component in a read-only state is considered relevant to the user’s workflow, allowing users to view and interact with the content without modifying it. It must meet accessibility requirements, including minimum contrast. The system supports status messages for read-only components, as RAG states may require visibility for recognition and further action. Consider if a read-only state is the right UI pattern, or if a disabled state might be more appropriate when modification is not needed.
```
joshwooding, line 46:
```suggestion
A component in a read-only state is considered relevant to the user’s workflow, allowing users to view and interact with the content without modifying it. It must meet accessibility requirements, including minimum contrast. The system supports validation states (error, warning or success) for read-only components, as RAG states may require visibility for recognition and further action. Consider if a read-only state is the right UI pattern, or if a disabled state might be more appropriate when modification is not needed.
```
status messages seems too narrow?
This copy introduces the term RAG states which isn't previously defined and I'm not sure how well known it is.
The last sentence seems wrong, disabled is very different non-modifiable which is what read-only is.

origami-z (dismissed):

origami-z (dismissed):

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,11 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
 Salt is under active development. At this stage, we don't recommend theming unless absolutely necessary.
 
 For an overview of theming, refer to our [theming documentation](/salt/themes/index).
+
+## Disabled state
+
+A component in a disabled state is completely non-interactive as it isn't considered relevant to the user’s workflow at that time. As the function isn't available to the user, the component doesn't need to meet minimum contrast requirements and is often displayed as visually muted. Consider if a disabled state is the right UI pattern, or if a progressive reveal might be more appropriate. Salt doesn't offer validation states (error, warning, or success) for disabled components, as these imply required user action.
```
joshwooding, line 42:
```suggestion
A component in a disabled state is completely non-interactive as it isn’t considered relevant to the user’s workflow at that time. As the function isn’t available to the user, the component doesn’t need to meet minimum contrast requirements and is often displayed as visually muted. Consider if a disabled state is the correct UI pattern or if a progressive reveal might be more appropriate. Salt doesn’t offer validation states (error, warning, or success) for disabled components, as these imply required user action.
```

site/docs/about/glossary.mdx
```diff
@@ -36,3 +36,11 @@ Salt is the J.P. Morgan design system, an open-source solution for building exce
...
+
+## Read-only state
+
+A component in a read-only state is considered relevant to the user’s workflow, allowing users to view and interact with the content without modifying it. It must meet accessibility requirements, including minimum contrast. The system supports validation states (error, warning or success) for read-only components. While disabled states are usually not detected by screen readers, read-only elements are accessible by screen reader.
```
joshwooding, line 46:
```suggestion
A component in a read-only state is considered relevant to the user’s workflow, allowing users to view and interact with the content without modifying it. It must meet accessibility requirements, including minimum contrast. The system supports validation states (error, warning or success) for read-only components. While disabled states are usually not detected by screen readers, read-only elements are accessible by screen readers.
```

mark-tate (approved):

PR #4366, opened 2024-10-31
[MenuItem] Align icons with text.

brianblakely:
MenuItems have default flex item alignment, this change centers them on the cross axis instead.

**Before:**
image

**After:**
image


joshwooding:
Unfortunately this won't work for us, our design expects the icon to align to the first baseline in the component, with this change it would align to the middle. We've solved this in different ways in the library but we're going to re-visit our icon sizing soon.

PR #4367, opened 2024-10-31
Fix Tooltip status override

### LINKED ISSUE #3868: Tooltip status getting overridden with form validation status
### Area

UI Components

### The problem

We are using Tooltip inside Form field. This tooltip is used to show the description regarding the field. We have set the status as info for the tooltip.

When there is error in the field and Form field validation is set to error then this overrides the status of tooltip as well.

Since this is just an Info regarding the field it's status should not change.

I have added an example [here](https://codesandbox.io/p/sandbox/react-hook-form-salt-input-with-info-text-5tkd9s).

Steps to replicate:

1. Hover on info icon and check the status and icon present in tooltip.

2. Click on submit without entering any text.

3. Error will be shown and when you hover again on info icon then the status of tooltip also changes to error even when we have explicitly set the status to info.

Is there any way to not override the status of tooltip?

### The solution

Tooltip takes its `status` prop as priority than form field status

### Alternatives and examples

?

liamsms:
Closes #3868

Tooltip now prioritizes it's `status` prop over the status inherited from a parent FormField.

### DIFF
--- a/.changeset/slow-bags-give.md
+++ b/.changeset/slow-bags-give.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed Tooltip to prioritize it's `status` prop over the status inherited from a parent FormField.
--- a/.changeset/slow-bags-give.md
+++ b/.changeset/slow-bags-give.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fixed Tooltip to prioritize it's `status` prop over the status inherited from a parent FormField.
+Fixed Tooltip to prioritize its `status` prop over the status inherited from a parent FormField.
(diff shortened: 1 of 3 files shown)


origami-z (dismissed):

.changeset/slow-bags-give.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed Tooltip to prioritize it's `status` prop over the status inherited from a parent FormField.
```
joshwooding, line 5:
```suggestion
Fixed Tooltip to prioritize its `status` prop over the status inherited from a parent FormField.
```

joshwooding (approved):

origami-z (approved):

PR #4111, opened 2024-09-16
Card design consolidation

origami-z:
Kept `--card-accent-color` in case anyone using it. 

Moved hover block before active, to make border/shadow styles correct.

#3875

PR #3301, opened 2024-04-24
Update stylelint to include no deprecated warning

origami-z:
Print warning of deprecated token in style lint. Keep it warning so we can gradually fix it.

Sample warning in core

```
packages/core/src/button/Button.css
 55:44  ⚠  No deprecated token (--salt-size-unit) should be used in component  salt/correct-theme-token-usage
packages/core/src/card/Card.css
 35:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 44:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 51:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 60:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
packages/core/src/checkbox/Checkbox.css
 28:55  ⚠  No deprecated token (--salt-size-selectable) should be used in component  salt/correct-theme-token-usage
packages/core/src/checkbox/CheckboxIcon.css
 2:24  ⚠  No deprecated token (--salt-size-selectable) should be used in component  salt/correct-theme-token-usage
packages/core/src/dialog/DialogHeader.css
 33:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
packages/core/src/file-drop-zone/FileDropZone.css
 2:49  ⚠  No deprecated token (--salt-text-primary-foreground) should be used in component  salt/correct-theme-token-usage
packages/core/src/interactable-card/InteractableCard.css
 38:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 47:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 54:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 63:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
packages/core/src/link-card/LinkCard.css
 38:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 47:14  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
 54:15  ⚠  No deprecated token (--salt-size-accent) should be used in component  salt/correct-theme-token-usage
[...]

PR #4417, opened 2024-11-18
Undeprecated --salt-accent-background-disabled

joshwooding:
In #4111 the token was added back to characteristics but it wasn't removed from the deprecated file.

### DIFF
--- a/.changeset/nervous-maps-fix.md
+++ b/.changeset/nervous-maps-fix.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/theme": patch
+---
+
+Undeprecate `--salt-accent-background-disabled`
--- a/packages/theme/css/deprecated/characteristics.css
+++ b/packages/theme/css/deprecated/characteristics.css
@@ -137,7 +137,6 @@
   /* Accent */
   --salt-accent-fontWeight: var(--salt-typography-fontWeight-semiBold); /* Use --salt-text-notation-fontWeight */
   --salt-accent-foreground-disabled: var(--salt-palette-accent-foreground-disabled);
-  --salt-accent-background-disabled: var(--salt-palette-accent-background-disabled);
   --salt-accent-borderColor-disabled: var(--salt-palette-accent-border-disabled); /* Use --salt-container-primary-borderColor-disabled */
 
   /* Track */


origami-z (approved):
Maybe stylelint rule needs to be enforced in core...?

joshwooding:
> Maybe stylelint rule needs to be enforced in core...?

Yeah, I think it's time :) Accordion is the only component left using deprecated tokens but I'm pretty sure they should be transparent since design doesn't have a style applied.

amritadesmet (approved):

pseys:
> > Maybe stylelint rule needs to be enforced in core...?
> 
> Yeah, I think it's time :) Accordion is the only component left using deprecated tokens but I'm pretty sure they should be transparent since design doesn't have a style applied.

Out of interest what are the stylelint rules?

origami-z:
> > > Maybe stylelint rule needs to be enforced in core...?
> > 
> > 
> > Yeah, I think it's time :) Accordion is the only component left using deprecated tokens but I'm pretty sure they should be transparent since design doesn't have a style applied.
> 
> Out of interest what are the stylelint rules?

We have custom linting rules to not use deprecated tokens, it was introduced (#3301) as warning to gradually adopt the rule and eventually should fail the build and stop any core component to use deprecated tokens
 https://github.com/jpmorganchase/salt-ds/blob/131dd9a13eda20f5729cb58460afa4ec1b225788/tooling/stylelint/correct-theme-token-usage/index.js#L62-L64

pseys:
> > > > Maybe stylelint rule needs to be enforced in core...?
> > > 
> > > 
> > > Yeah, I think it's time :) Accordion is the only component left using deprecated tokens but I'm pretty sure they should be transparent since design doesn't have a style applied.
> > 
> > 
> > Out of interest what are the stylelint rules?
> 
> We have custom linting rules to not use deprecated tokens, it was introduced (#3301) as warning to gradually adopt the rule and eventually should fail the build and stop any core component to use deprecated tokens
> 
> https://github.com/jpmorganchase/salt-ds/blob/131dd9a13eda20f5729cb58460afa4ec1b225788/tooling/stylelint/correct-theme-token-usage/index.js#L62-L64

That's awesome. Thank you!

PR #4291, opened 2024-10-14
Date picker review2

mark-tate:

PR #4425, opened 2024-11-20
Add back custom format date picker example

origami-z:
Feedback from user this capability is not clear

### DIFF
--- a/site/docs/components/date-picker/examples.mdx
+++ b/site/docs/components/date-picker/examples.mdx
@@ -28,7 +28,7 @@ An uncontrolled `DatePicker`, composed of a `DateInput` and a `Calendar` compone
 
 <LivePreview componentName="date-picker" exampleName="CustomFormat">
 
-## Custom Format
+## Custom Formatter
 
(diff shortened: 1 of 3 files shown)


joshwooding (approved):
@mark-tate does this change a lot in #4291 ?

mark-tate:
I would park this..
Format is no longer a function, it's a string.

origami-z:
> I would park this.. Format is no longer a function, it's a string.

will the new version be released very soon? This is being asked in support quite a lot

mark-tate:
format is now a string based and implemented by adapter.. Will need to revisit this example, if it is still required?

Let's get some feedback on the Adapter based version of date-picker to see whether this example is needed.

Now things are string based and re-use the date framework's API, I wonder if we need to document recommended formats.
I added a doc that lists the currently used formats to our docs. 
@ivan-calderon worth reviewing, see the LocalizationProvider which also documents adapters.

In theory, you could also extend the adapter and implement your own custom formatting, if there was an issue with the current formatting approach.

PR #4440, opened 2024-11-25
Update date picker based on community feedback

origami-z:
- https://github.com/jpmorganchase/salt-ds/pull/4291#discussion_r1808314876
- https://github.com/jpmorganchase/salt-ds/pull/4291#discussion_r1829350795


mark-tate:
This PR was absorbed into our other branch.
We changed the role to dialog, to avoid the biome error @jake-costa
Might switch that back, if the more abstract modal is preferred.

Issue #4447, opened 2024-11-26
Replace hash with link icon on external site headings


amritadesmet:
Change to "URL has been copied", Replace hash with link icon. For some reason, Salt website is not using Mosaic's copy feature
https://github.com/jpmorganchase/mosaic/blob/e2821e10afa243c0ea7d720e1a7767a00f7ea5ac/packages/components/src/Markdown/AnchorHeading/index.tsx#L69-L89

Edit scroll to prevent jumping to section, below css should help
```
scroll-margin-top: calc(4rem + 20px);
```

origami-z:
Trying to use Tag for the copy action, too big...

mark-tate:
@honey-chang we need you to add this to the editorial component suite.
This is a link that when you hover on it, displays a clickable adornment that can be used to copy the link to the clipboard.

honey-chang:
@mark-tate This issue is belong to external site. Internal site also has similar feature and I can add this editorial pattern list. Can we fix the issue on external site as well?

PR #4449, opened 2024-11-26
Add site heading link scroll top

origami-z:
This should fix the confusion that user clicking the header link, got lost... i.e. when landing on a page with hash id, headings shouldn't be covered up by the app header

Part of #4447


origami-z:
@joshwooding do you think this is the right solution?

origami-z:
Fixed in #4784

Issue #4344, opened 2024-10-28
Date Picker set `selectedDate` to `null` doesn't clear date input


origami-z:
### Latest version

- [X] I have tested the latest version

### Description

Can't set `selectedDate` to `null` to clear date input

### Steps to reproduce

Control `selectedDate` set to `null`

https://stackblitz.com/edit/salt-template-tfbgv5?file=App.tsx

Workaround: control date input value as well: https://stackblitz.com/edit/salt-template-mlhorv?file=App.tsx

### Expected behavior

My guess `if` is wrong? https://github.com/jpmorganchase/salt-ds/blob/9e6676c0a973cc131f3ce9c689f39ffa905cfddb/packages/lab/src/date-input/DateInputSingle.tsx#L242-L245

To be fixed before https://github.com/jpmorganchase/salt-ds/issues/4096

### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

lab 52

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [X] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

PR #4459, opened 2024-11-28
Fix controlled date picker set to null

origami-z:
#4344

Questions to be answered 

- Single date picker, when to callback date with `undefined` vs `null` vs `Date` (invalid vs valid)
    - e.g. clear input value to a empty string?
- `DateInputSingle`
    - `useEffect`: sync input value state to the date state
        - why: 
            1. when `date` is controlled to be `null`, we want to clear the input
            1. when `date` is controlled to be a valid date, we want the formatted input value in the input
            1. when `date` is controlled to be a invalid date, we want leave as it is
        - when: 
           - option1: only when controlled input value prop changed, sync to internal date state
           - option2: remove sync logic from `apply`
- `DatePickerSingleInput` vs `DateInputSingle`

### DIFF
--- a/packages/lab/src/date-input/DateInputRange.tsx
+++ b/packages/lab/src/date-input/DateInputRange.tsx
@@ -214,7 +214,7 @@ export const DateInputRange = forwardRef<
       endInputRef: endInputRefProp,
       locale,
       parse: parseProp,
-      placeholder = "dd mmm yyyy",
+      placeholder = format,
       readOnly: readOnlyProp,
       validationStatus: validationStatusProp,
       variant = "primary",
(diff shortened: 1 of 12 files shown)


packages/lab/src/date-input/DateInputRange.tsx
```diff
@@ -214,7 +214,7 @@ export const DateInputRange = forwardRef<
       endInputRef: endInputRefProp,
       locale,
       parse: parseProp,
-      placeholder = "dd mmm yyyy",
+      placeholder = format,
```
mark-tate, line 217: nice

origami-z: Casing actually matters, in certain library, lower case and upper case means different things.

origami-z:
replace by  #4465

PR #4465, opened 2024-11-29
- empty values now are parsed as unset in range

mark-tate:
- date format now can be passed falsy values as well as strings
- updates tests
- updates stories
- updates docs

### DIFF
--- a/packages/lab/src/__tests__/__e2e__/date-input/DateInputRange.cy.tsx
+++ b/packages/lab/src/__tests__/__e2e__/date-input/DateInputRange.cy.tsx
@@ -10,7 +10,7 @@ import {
 } from "@salt-ds/date-adapters";
 import {
   DateInputRange,
-  DateParserField,
+  type DateParserField,
   type DateRangeSelection,
 } from "@salt-ds/lab";
 import { es as dateFnsEs } from "date-fns/locale";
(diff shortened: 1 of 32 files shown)


packages/lab/src/__tests__/__e2e__/date-input/DateInputRange.cy.tsx
```diff
@@ -58,70 +67,54 @@ function assertDateChange(
...
-    expect(date.startDate).to.be.undefined;
-  } else if (expectedValidStartDate) {
-    // assert valid start date matches expected date
+  if (expectedValidStartDate) {
+    // assert valid start date
     expect(date.startDate).not.to.be.undefined;
     expect(adapter.format(date.startDate, "DD MMM YYYY")).to.equal(
       adapter.format(expectedDate.startDate, "DD MMM YYYY"),
     );
+  } else if (expectedDate?.startDate === undefined) {
```
origami-z, line 76: Feels range's `undefined` has slightly different meaning than single date's `Date | null | undefined`...

origami-z (approved):
We can release this in labs and iterate further

origami-z:
Changed target to `main` branch, so we can get a preview

packages/lab/src/date-input/DateInputSingle.tsx
```diff
@@ -160,17 +161,17 @@ export const DateInputSingle = forwardRef<
...
       onClick,
       emptyReadOnlyMarker = "—",
       endAdornment,
-      format = "DD MMM YYYY",
       inputProps = {},
       inputRef: inputRefProp = null,
-      locale,
       parse: parseProp,
-      placeholder = "dd mmm yyyy",
+      placeholder = format.toLowerCase(),
```
origami-z, line 174: lower case and upper case in date formatter are different. A few suggestions
- Remove, leave empty (no other Salt component sets this)
- Set as generic text, e.g. "Set Date", "Start Date", etc.
- Or leave as `format`

mark-tate: @ivan-calderon / @jake-costa / @honey-chang  one for you to opine on..
Feels like this falls into the formatting pattern, an overall guide on recommended formats?
Simple change to make once we agree on what it should be (compare with UITK?)

origami-z (approved):

Issue #4470, opened 2024-12-02
DatePicker ignores readOnly


APop123:
### Latest version

- [X] I have tested the latest version

### Description

Adding `readOnly` to a `DatePicker` renders the field as readonly, but the calendar button still appears, and the user can still change the date.

### Steps to reproduce

```
    <DatePicker
      selectionVariant="single"
      readOnly
    >
      <DatePickerTrigger>
        <DatePickerSingleInput />
      </DatePickerTrigger>
      <DatePickerOverlay>
        <DatePickerSinglePanel />
      </DatePickerOverlay>
    </DatePicker>
```

### Expected behavior

User should not be able to change the date

### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

1.0.0-alpha.56

### Browser

- [X] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [ ] Windows
- [X] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [X] I am an employee of JPMorgan Chase & Co.

mark-tate:
Should be in our next release

red Eye : review/release by EOS

dplsek:
Requires status updated from @mark-tate

joshwooding:
This will be released in the next version.

bhoppers2008:
Updated. Close issue.

PR #4472, opened 2024-12-04
Date picker fix readOnly overlay

origami-z:
#4470

### DIFF
--- a/packages/lab/src/date-picker/DatePickerOverlayProvider.tsx
+++ b/packages/lab/src/date-picker/DatePickerOverlayProvider.tsx
@@ -94,11 +94,15 @@ interface DatePickerOverlayProviderProps {
    * The content to be rendered inside the overlay provider.
    */
   children: ReactNode;
+  /**
+   * When true, shouldn't open the overlay.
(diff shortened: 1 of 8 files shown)


origami-z:
@mark-tate  don't know how you feel about the `setOpen` change

mark-tate (approved):

Fercas123 (approved):

packages/lab/src/date-picker/DatePickerOverlayProvider.tsx
```diff
@@ -130,6 +134,10 @@ export const DatePickerOverlayProvider: React.FC<
       reason?: OpenChangeReason | undefined,
     ) => {
       if (newOpen) {
+        if (readOnly) {
```
joshwooding, line 137: I haven't looked carefully but does this mean that providing defaultOpen and open will still open the overlay in read-only?

mark-tate: it will but can pickup in my PR, which is testing this code.

PR #4473, opened 2024-12-04
Updates to headers alignment

Fercas123:
fix alignment for titles spanning into multiple lines, wrap dialog header in H2 to avoid breaking change

### DIFF
--- a/packages/lab/src/dialog/DialogHeader.css
+++ b/packages/lab/src/dialog/DialogHeader.css
@@ -31,6 +31,7 @@
 /* Styles applied to the status indicator icon overriding its default size */
 .saltDialogHeader .saltStatusIndicator.saltIcon {
   --icon-size: var(--salt-text-h2-lineHeight);
+  padding-top: calc((var(--salt-size-base) - var(--salt-text-h2-lineHeight)) / 2);
 }
 
 /* Styles applied to DialogHeader when accent={true} */
--- a/packages/lab/src/dialog/DialogHeader.css
+++ b/packages/lab/src/dialog/DialogHeader.css
@@ -3,11 +3,13 @@
   padding-bottom: var(--salt-spacing-300);
   padding-left: var(--salt-spacing-300);
   padding-right: var(--salt-spacing-300);
-  align-items: center;
   display: flex;
   flex-direction: row;
   gap: var(--salt-spacing-100);
   box-sizing: border-box;
+}
+.saltDialogHeader-header {
+  margin: 0;
 }
 
 .saltDialogHeader-container {
--- a/packages/lab/src/dialog/DialogHeader.tsx
+++ b/packages/lab/src/dialog/DialogHeader.tsx
@@ -1,4 +1,5 @@
 import {
+  H2,
   StatusIndicator,
   Text,
   type ValidationStatus,
@@ -86,7 +87,7 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
             {preheader && (
               <Text className={withBaseName("preheader")}>{preheader}</Text>
             )}
-            {header}
+            <H2 className={withBaseName("header")}>{header}</H2>
           </div>
           {description && (
             <Text color="secondary" className={withBaseName("description")}>
--- a/packages/lab/src/dialog/DialogHeader.tsx
+++ b/packages/lab/src/dialog/DialogHeader.tsx
@@ -83,12 +83,10 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
       >
         {status && <StatusIndicator status={status} />}
         <div className={withBaseName("container")}>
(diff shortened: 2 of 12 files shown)


packages/lab/src/dialog/DialogHeader.tsx
```diff
@@ -86,7 +87,7 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
             {preheader && (
               <Text className={withBaseName("preheader")}>{preheader}</Text>
             )}
-            {header}
+            <H2 className={withBaseName("header")}>{header}</H2>
```
joshwooding, line 90: There's a change no matter what here, the preheader is no longer part of the header. @jake-costa Do you see this as a better outcome or do you think the previous was better?

jake-costa: @joshwooding good to hear the preheader is no longer part of the header. Also, I'd like to connect to go over this if you have some time this week but personally I think this approach would be better. 

Lastly, I want to add that we need to be careful about assigning h4 to the overlay, given it is also technically a modal it should probably be an h2 as well.

Fercas123: updated h4s in our stories for h2 that look like h4s

joshwooding: @Fercas123 can you set up a call with Jake please.

Fercas123: updated the h2 to wrap preheader again and swapped overlay header to do the same

packages/lab/src/dialog/DialogHeader.tsx
```diff
@@ -82,12 +83,10 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
       >
         {status && <StatusIndicator status={status} />}
         <div className={withBaseName("container")}>
-          <div className={withBaseName("header")}>
-            {preheader && (
-              <Text className={withBaseName("preheader")}>{preheader}</Text>
-            )}
+          <H2 className={withBaseName("header")}>
```
joshwooding, line 86: Do we want to hide the h2 when there's no header?

packages/lab/src/overlay/OverlayHeader.tsx
```diff
@@ -45,12 +45,10 @@ export const OverlayHeader = forwardRef<HTMLDivElement, OverlayHeaderProps>(
     return (
       <div className={clsx(withBaseName(), className)} {...rest} ref={ref}>
         <div className={withBaseName("container")}>
-          <div className={withBaseName("header")}>
-            {preheader && (
-              <Text className={withBaseName("preheader")}>{preheader}</Text>
-            )}
+          <H2 styleAs="h4" className={withBaseName("header")}>
```
joshwooding, line 48: Same here

packages/lab/stories/overlay/overlay.qa.stories.tsx
```diff
@@ -39,11 +39,19 @@ export const Default: StoryFn<QAContainerProps> = (props) => {
...
+          style={{
+            width: "30ch",
+          }}
+        >
           <OverlayHeader
             preheader="Preheader"
             description="Description"
-            header={<H4>Header block</H4>}
+            header={
+              <H2 styleAs="h4">
```
joshwooding, line 51: This should be checked as it will nest headings

packages/lab/stories/overlay/overlay.stories.tsx
```diff
@@ -98,7 +128,11 @@ export const HeaderWithCloseButton = ({ onOpenChange }: OverlayProps) => {
         <OverlayHeader
           preheader="Preheader"
           description="Description"
-          header={<H4 id={id}>Header block</H4>}
+          header={
+            <H2 styleAs="h4" id={id}>
```
joshwooding, line 132: Same here

amritadesmet (dismissed):

joshwooding (approved):

packages/lab/src/dialog/DialogHeader.css
```diff
@@ -29,6 +31,7 @@
 /* Styles applied to the status indicator icon overriding its default size */
 .saltDialogHeader .saltStatusIndicator.saltIcon {
   --icon-size: var(--salt-text-h2-lineHeight);
+  padding-top: calc((var(--salt-size-base) - var(--salt-text-h2-lineHeight)) / 2);
```
joshwooding, line 34: We should double check how they are going to achieve this in Figma.

chrispcode (approved):

chrispcode (approved):
Minor comment / Personal opinion: I think the name of a story and the title of the story should match. In the sense that If the set title is `Lab/Dialog Header` than the file should also be named something along the lines of `lab/[dialog]/dialog-header.stories.tsx`

PR #4485, opened 2024-12-10
Border layout sticky example update

Fercas123:


site/docs/components/border-layout/examples.mdx
```diff
@@ -72,6 +72,8 @@ Use the `sticky` prop of the border item wrapper to define a region as sticky. W
 
 This is particularly useful if you require a header, (in the North region) or side navigation (in the West region) to remain visible at all times.
```
origami-z, line 73: I think the problem exists here, where sticky is interpreted as needed for both header and side nav. And if that's applied, weird things happens (e.g. west region scrolls on top of north).

At the same time, most apps doesn't build like this example. Most likely it's `body` scrolls (e.g. Salt website, react.dev). What's the best option for someone new to Salt not misuse this? Is border layout fit for purpose here?

origami-z: Yes, makes sense. Let's just show how `sticky` works for now, maybe not getting into full app layout description here, which can probably be looked at as a pattern at later stage.

amritadesmet (approved):

origami-z (dismissed):

site/docs/components/border-layout/examples.mdx
```diff
@@ -70,7 +70,7 @@ Use the `verticalAlignment` and `horizontalAlignment` properties of the border i
 
 Use the `sticky` prop of the border item wrapper to define a region as sticky. When you set it to `true`, content in surrounding regions will scroll beneath a `sticky` Border Item.
```
joshwooding, line 71: We could link to https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky to help

joshwooding (approved):

amritadesmet (approved):

PR #4552, opened 2025-01-09
Remove validation status for disabled dropdown

### LINKED ISSUE #4548: Validation status shouldn't be applied when Dropdown is disabled
### Latest version

- [X] I have tested the latest version

### Description

Following our other components, validation shouldn't be applied.

### Steps to reproduce

https://storybook.saltdesignsystem.com/?path=/story/core-dropdown--disabled&args=validationStatus:error

### Expected behavior

_No response_

### Package name(s)

Core (@salt-ds/core)

### Package version(s)

_No response_

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

tayyabataimur:
Fixes #4548 

### Screenshots

**Before**
image
_________________________________________
**After**
image

### DIFF
--- a/packages/core/src/dropdown/Dropdown.tsx
+++ b/packages/core/src/dropdown/Dropdown.tsx
@@ -160,7 +160,7 @@ export const Dropdown = forwardRef(function Dropdown<Item>(
   const validationStatus = validationStatusProp ?? formFieldValidationStatus;
   const required = formFieldRequired
     ? ["required", "asterisk"].includes(formFieldRequired)
-    : (undefined ?? requiredProp);
+    : requiredProp;
   const listControl = useListControl<Item>({
     open,
     defaultOpen,
@@ -456,7 +456,7 @@ export const Dropdown = forwardRef(function Dropdown<Item>(
         >
           {!valueText ? placeholder : valueText}
         </span>
-        {validationStatus && <StatusAdornment status={validationStatus} />}
+        {!disabled && validationStatus && <StatusAdornment status={validationStatus} />}
         {!readOnly && <ExpandIcon open={openState} />}
         <div className={withBaseName("activationIndicator")} />
       </button>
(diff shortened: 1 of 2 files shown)


packages/core/src/dropdown/Dropdown.tsx
```diff
@@ -157,10 +157,12 @@ export const Dropdown = forwardRef(function Dropdown<Item>(
...
   const disabled = Boolean(disabledProp) || formFieldDisabled;
   const readOnly = Boolean(readOnlyProp) || formFieldReadOnly;
-  const validationStatus = validationStatusProp ?? formFieldValidationStatus;
+  const validationStatus = !disabled
+    ? (validationStatusProp ?? formFieldValidationStatus)
+    : undefined;
   const required = formFieldRequired
     ? ["required", "asterisk"].includes(formFieldRequired)
-    : (undefined ?? requiredProp);
+    : requiredProp;
```
joshwooding, line 165: Although this is a correct way to do it, I think we should be consistent with Input and conditionally render the StatusAdornment to fix the issue. Thoughts?

tayyabataimur: @joshwooding That was my first thought as well but then wanted to reduce duplicate conditions where we will need to add the same condition here in the [classname](https://github.com/jpmorganchase/salt-ds/blob/main/packages/core/src/dropdown/Dropdown.tsx#L426) so that it doesn't unnecessarily apply it. But happy to prioritise a consistent approach over this.

joshwooding: @tayyabataimur I think it depends on what level we expect the rule to live. If the class is kept teams could have a different visual result. I do prefer reducing duplicate conditions but we should figure out a consistent approach and adopt it. Adopting the current method is an easier task.

joshwooding (approved):

Issue #4481, opened 2024-12-09
Z-Index foundation page 


amritadesmet:
- [x] Write new documentation
- [ ] Update external site docs with new content

**Red-eye goal:** Documentation complete

mark-tate:
Red eye goal:  publish by EOS

amritadesmet:
On hold until discussed w/ @pseys

dplsek:
Still pending sync with @amritadesmet and @pseys

amritadesmet:
**18 Dec Update:** Working on content documentation

amritadesmet:
PR created #4553

dplsek:
on track, PR raised, @joshwooding on the case

amritadesmet:
**Mon 13 Jan update:** PR being updated. Will require review

amritadesmet:
PR Squashed and merged

PR #4553, opened 2025-01-09
z-index page

amritadesmet:
In relation to #4481


site/docs/foundations/elevation/z-index.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: Z-Index
+layout: DetailTechnical
+sidebar:
+  priority: 4
+---
+
+Z-index is a CSS property that controls how certain components stack on top of each other. The stack order is defined by an element with greater z-index appearing in front of an element with a lower stack order.
```
tayyabataimur, line 8:
```suggestion
Z-index is a CSS property that controls how certain components stack on top of each other. The stack order is defined by an element with greater z-index appearing on top of an element with a lower stack order. 
```

Nit: Would it make more sense to phrase it as "on top" instead of "in front" based on the rest of the documentation? 🤔

amritadesmet: Yes, good catch! Thanks for the feedback

DavieReid: I was just about to comment to say this should in front...

site/docs/foundations/elevation/z-index.mdx
```diff
@@ -0,0 +1,32 @@
...
+
+Z-index and shadow are used together to help distinguish each layer in the stack. Shadow provides the visual element and z-index provides the structural element.  
+
+## Z-Index ramp
+
+Salt provides z-index tokens that determine an item’s position in the layering order. By providing consistent z-indexes, the design system ensures a predictable hierarchy of visual elements when stacked on top of one another.
+
+| Token                  | Value |
+| ---------------------- | ----- | 
+| `-zIndex-default`      | 1     |
```
DavieReid: Out of interest why is the default 1?

site/docs/foundations/elevation/z-index.mdx
DavieReid: What does "certain components" mean?

DavieReid:
The Digram isn't a great one.  You can google and see diagrams that show the stacking context from the perspective of the viewport.

site/docs/foundations/elevation/z-index.mdx
amritadesmet: Great point, thanks for the feedback. Will adjust this accordingly.

site/docs/foundations/elevation/z-index.mdx
amritadesmet: I can point you to @joshwooding who might be able to answer that question better!

joshwooding: We're going to be updating the ramp soon, this PR is just focusing on documenting what currently exists

site/docs/foundations/elevation/z-index.mdx
```diff
@@ -0,0 +1,33 @@
+---
+title: Z-Index
+layout: DetailTechnical
+sidebar:
+  priority: 4
+---
+
+Z-index is a CSS property that controls how elements stack on top of each other. The stack order is defined by an element with a higher z-index appearing on top of an a element with a lower z-index.
```
pseys, line 8: "top of **an a** element"

site/docs/foundations/elevation/z-index.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+Z-index and shadow are used together to help distinguish each layer in the stack. Shadow provides the visual element and z-index provides the structural element.
```
pseys, line 10: Can we link to the Shadow foundation here?

site/public/img/foundations/z-index.png
pseys, line 1: @amritadesmet could we add values to the stack order? The table shows lowest value to highest reading down the table, but the diagram would be showing the highest value at the top of the stack.

joshwooding (approved):

PR #4562, opened 2025-01-10
Update AG Grid version

tayyabataimur:
Updating to a version that is present in Artifactory.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,8 @@
     "whatwg-url": "^14.0.0",
     "@joshwooding/vite-plugin-react-docgen-typescript": "0.4.2",
     "cookie": "^0.7.0",
-    "esbuild": "^0.24.0"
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "@types/tinycolor2": "^1.4.3",
     "@vitejs/plugin-react": "^4.1.0",
     "ag-grid-community": "^32.0.0",
-    "ag-grid-enterprise": "^32.0.0",
+    "ag-grid-enterprise": "^32.2.0",
     "ag-grid-react": "^32.0.0",
```
joshwooding, line 80: I think all the packages should be the same version, maybe we should keep these as they are and add a resolution to 32.2.1

joshwooding (approved):

DavieReid (approved):

PR #4573, opened 2025-01-14
Flex layout and items margin and paddings

Fercas123:
add margin and paddings to flex layout as a multiplier of salt-spacing

### DIFF
--- a/packages/core/src/flex-layout/FlexLayout.tsx
+++ b/packages/core/src/flex-layout/FlexLayout.tsx
@@ -1,6 +1,8 @@
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
 import { clsx } from "clsx";
 import { type ElementType, type ReactElement, forwardRef } from "react";
-
+import { useBreakpoint } from "../breakpoints";
 import {
   type PolymorphicComponentPropWithRef,
   type PolymorphicRef,
@@ -8,10 +10,6 @@ import {
   makePrefixer,
   resolveResponsiveValue,
 } from "../utils";
-
-import { useComponentCssInjection } from "@salt-ds/styles";
-import { useWindow } from "@salt-ds/window";
-import { useBreakpoint } from "../breakpoints";
 import flexLayoutCss from "./FlexLayout.css";
 
 const withBaseName = makePrefixer("saltFlexLayout");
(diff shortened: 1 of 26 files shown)


Fercas123:
⁠i had added examples to all layouts built from flex, then removed them because it seemed too repetitive... there are 3 options i considered but could use some feedback 

- we add examples to all components built with flex
- we add the examples to foundations/spacing and link to that section in the layouts
- we add the example in flexLayout and link to that section from the other layouts that build from flex

mark-tate:
needs a changeset

chrispcode (dismissed):

libertymayc:
this is great :) we also have to add 100% width/height to so many of our layout components... if there was a prop to do this too that would be awesome

chrispcode (approved):

site/docs/components/split-layout/examples.mdx
```diff
@@ -14,9 +14,9 @@ data:
 
 ## Default
 
-Split layouts display items horizontally at either end of the container by default.
+Split layouts display items horizontally at either end of the container by default with its items align centrally in the layout.
```
joshwooding, line 17:
```suggestion
Split layouts display items horizontally at either end of the container by default with its items aligned centrally in the layout.
```
?

site/docs/components/split-layout/examples.mdx
```diff
@@ -82,6 +82,15 @@ The flex item component's `grow` prop enables specific items in the layout to gr
...
 In contrast to the `grow` prop, the `shrink` prop shrinks specific items as small as possible within the layout. By default, `shrink` is set to 1. If an item is set to 2, that item would attempt to take half as much space.
 
+</LivePreview>
+<LivePreview componentName="split-layout" exampleName="Spacing">
+
+## Spacing
+
+Split layout displays items with a gap between items that is a multiple of the [Salt base unit](/salt/foundations/spacing) controlled by the `gap` property and is by default set to 3.
+
+The spacing properties `padding` and `margin` are set to 0 by default and can be used to control the layout's inner and outer spacing also as multiples of the Salt base unit.
```
joshwooding, line 92:
```suggestion
The spacing properties `padding` and `margin` are set to 0 by default and can be used to control the layout's inner and outer spacing (also as multiples of the Salt base unit).
```
?

packages/core/src/flex-layout/FlexLayout.tsx
```diff
@@ -71,7 +79,7 @@ function parseAlignment(style: string | undefined) {
   return style === "start" || style === "end" ? `flex-${style}` : style;
 }
 
-function parseSpacing(value: number | string | undefined) {
+export function parseSpacing(value: number | string | undefined) {
```
joshwooding, line 82: We should be careful that we don't export this publicly, I think right now we are

joshwooding (approved):

tayyabataimur (approved):

Issue #4578, opened 2025-01-16
Update Slider component implementation


tayyabataimur:
- [ ] Separate Range Slider and Single Slider components to incorporate improved readability, best practices and accessibility. 
- [ ] Use semantic elements (`<input type='range' />`) to get correct default behaviour.
- [ ] Simplify logic, ensure correct types usage.
- [ ] Update stories and tests

(Follow-up ticket to #4076)

**Discussion points**

- Should Range Slider and Single Slider be exposed as two different components in Salt similar to DatePicker or should we render one or the other based on prop values?

mark-tate:
Tecno Goal:
- Code updates merged to Labs by EOS 
- Docs/Examples will be delivered in Lenovo
- Identify a stakeholder to review the work by EOS

mark-tate:
PR draft raised
User experiences to be provided by EOS

mark-tate:
Updated approach, requires review by EOS
Design answers needed - @navkaur76 can you sync up with @tayyabataimur on your return

tayyabataimur:
Had a chat with @navkaur76 and @jake-costa and we decided that a better approach would be to not use `<input type=range>` since it is limiting in terms of how we interact with the slider thumb (e.g. hovering over thumb and showing tooltip). So going with the custom implementation approach.

tayyabataimur:
Need specs for `disabled` variant of the slider and the tooltip appearance in dark mode. To discuss with @navkaur76 once she's back.

mark-tate:
Oregano Goal:  
- Functionally complete including Range
- Docs and examples in review
- Can we do a vertical slider ? what is the cost?
This will impact our decision on when to move to Core

mark-tate:
needs design input on discussions created from demo

mark-tate:
adding tests and examples, meeting Nav to discuss design alignment by EOS
Aim to get something merged into the next release.

tayyabataimur:
PR up for review.

Need to decide if disabled state is required.

Extent of changes to add vertical slider:
1. CSS changes 
2. Update calculations of thumb movements to work in vertical orientation.

There should be no breaking changes when adding support for vertical slider.

tayyabataimur:
PR ready for review!

joshwooding:
Fixing a visual bug with disabled range sliders, working to address feedback left on PR

tayyabataimur:
[...]

Issue #4076, opened 2024-09-11
Address Slider feedback


joshwooding:
There's feedback from #3296 that needs to be addressed before it moves to core.

```[tasklist]
### Feedback
- [x] A11y review
- [x] Remove a11y cursor interaction docs since not specific to slider
- [x] With input docs
- [x] With form field docs
- [x] Custom step docs and examples
- [x] Aria-label behaviour
- [x] onChange not returning an event (needs to be tested)
- [x] realHover usage in tests
- [x] missing tests for min and max, passing in a custom step vs not. etc.
- [x] max-width
- [x] padding on the label
- [x] default values not being inline
- [x] JSDoc review
- [x] defaultValue relation to min
- [x] clamp behaviour between controlled and uncontrolled
- [x] extra type definitions
- [x] type cast usage
- [x] focussed vs focused
- [x] Label being used instead of body text (semantics and styling)
- [x] Using window directly instead of targetWindow
- [x] toFixed(1) being used
- [x] non-nullable being used
- [x] setValue logic
- [x] Inconsistent format from percentage based utils
- [x] QA stories being separate
- [x] aria-labels used in stories/examples
- [x] Slider description on the site only mentions single slider
- [x] Check Keyboard behaviour docs covers all behaviour
```

mark-tate:
Can pull back into Q1 if there is a stakeholder who needs it, contact the team

bhoppers2008:
To be addressed in Q2.

tayyabataimur:
Took a look at the Slider component and its feedback and can split the work as following for easier reviews:
1. **Code refactoring, improvements and optimization**
2. **Update site documentation and QA stories**
3. **A11Y review with Jake and Josh and perform any follow-up tasks**

PR #4614, opened 2025-01-23
Slider update

### LINKED ISSUE #4579: Update Slider's site documentation and QA stories
The current documentation about the Slider on the site does not correctly describe the component in some places, a few examples have been listed below. Also, the QA stories are separate and need to be consistent with other components.

- [ ] Remove a11y cursor interaction docs since not specific to slider
- [ ] With input docs
- [ ] With form field docs
- [ ] Custom step docs and examples
- [ ] JSDoc review
- [ ] QA stories being separate
- [ ] aria-labels used in stories/examples
- [ ] Slider description on the site only mentions single slider
- [ ] Check Keyboard behaviour docs covers all behaviour

If we decide to separate and expose Range Slider and Single Slider into 2 components, the documentation for that will need to be updated accordingly as well. 

Follow-up to #4076

Is blocked by #4578

tayyabataimur:
Updated the `Slider` component with the following:

1. Separated `Slider` and `RangeSlider` components.
2. Added `showTooltip` prop to allow users to show or hide the tooltip when the thumb is hovered.
3. Added `minLabel` and `maxLabel` props to allow users to add inline labels. 
4. Added `marks` prop to allow users to add marks with labels on specific values along the track.
5. Added `onChangeEnd` callback prop which will be triggered when the user stops dragging the pointer and when values are changed during keyboard navigation.
6. Added `stepMultiplier` prop which will decide the size of the step when `PageUp` and `PageDown` keys are pressed.
7. Added `format` callback prop which allows users to format the minimum and maximum labels, the tooltip text as well as the `aria-valuetext` for non-numeric values.

8. Added more tests and improved accessibility. 

Closes #4579, #4578 and #4076


.yarnrc.yml
```diff
@@ -1,7 +1,15 @@
...
 
+httpProxy: "http://approxy.jpmchase.net:10443"
+
+httpsProxy: "http://approxy.jpmchase.net:10443"
+
 nodeLinker: node-modules
 
 yarnPath: .yarn/releases/yarn-4.6.0.cjs
+
+npmRegistryServer: "https://artifacts-read.gkp.jpmchase.net/artifactory/api/npm/npm"
```
joshwooding, line 15:
```suggestion
compressionLevel: mixed

enableGlobalCache: false

nodeLinker: node-modules

yarnPath: .yarn/releases/yarn-4.6.0.cjs
```

packages/lab/src/custom-slider/CustomSlider.tsx
```diff
@@ -0,0 +1,208 @@
...
+          className={clsx(withBaseName("track"))}
+          onMouseDown={handleMouseDown}
+          ref={sliderRef}
+          style={
+            {
+              "--salt-slider-progressPercentage": `${progressPercentage}%`,
+            } as React.CSSProperties
+          }
+        >
+          <div
```
joshwooding, line 182: Can we make this an `<input>` ?

packages/lab/src/custom-slider/CustomSlider.tsx
```diff
@@ -0,0 +1,208 @@
...
+      <div
+        className={clsx(
+          withBaseName(),
+          withBaseName(`${labelPosition}Labels`),
+          { dragging: isDragging },
+        )}
+        ref={ref}
+        {...rest}
+      >
+        <Text as="label" className={clsx(withBaseName("minLabel"))}>
```
joshwooding, line 169:
```suggestion
        <Text as="label" className={withBaseName("minLabel")}>
```
More of a nit, we don't need to use `clsx` here.

packages/lab/src/custom-slider/CustomSlider.tsx
```diff
@@ -0,0 +1,208 @@
...
+    };
+
+    const progressPercentage = ((value - min) / (max - min)) * 100;
+
+    return (
+      <div
+        className={clsx(
+          withBaseName(),
+          withBaseName(`${labelPosition}Labels`),
+          { dragging: isDragging },
```
joshwooding, line 164:
```suggestion
          { [withBaseName("dragging")]: isDragging },
```

packages/lab/src/custom-slider/CustomSlider.tsx
```diff
@@ -0,0 +1,208 @@
...
+      if (!isDragging) setTooltipVisible(false);
+    };
+
+    const handleFocus = () => setTooltipVisible(true);
+
+    const handleBlur = () => setTooltipVisible(false);
+
+    const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
+      let newValue = value;
+      switch (event.key) {
```
joshwooding, line 132: We should align with StepperInput's keyboard handling i.e. PageUp or Shift + ArrowUp

packages/lab/src/custom-slider/CustomSlider.tsx
```diff
@@ -0,0 +1,208 @@
...
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-slider",
+      css: sliderCss,
+      window: targetWindow,
+    });
+
+    const [value, setValue] = useControlled({
+      controlled:
+        valueProp === undefined ? undefined : clamp(valueProp, min, max),
```
joshwooding, line 78: Does this clamp only for controlled or default too?

packages/lab/src/slider/internal/SliderTrack.tsx
```diff
@@ -1,111 +1,112 @@
...
+              "--slider-progressPercentageStart": `${progressPercentageRange?.[0]}%`,
+              "--slider-progressPercentageEnd": `${progressPercentageRange?.[1]}%`,
+            } as React.CSSProperties
+          }
+        >
+          {children}
+        </div>
+        {/* Max label */}
+        {!markers && (
+          <Text as="label" className={withBaseName("maxLabel")}>
```
joshwooding, line 84:
```suggestion
          <Text styleAs="label" className={withBaseName("maxLabel")}>
```

Worth checking with @jake-costa but right now markers are `<label>` elements.

packages/lab/src/slider/internal/SliderTrack.tsx
```diff
@@ -1,111 +1,112 @@
...
+              (marker: { label: string; value: number | string }) => (
+                <Text
+                  className={withBaseName("marker-label")}
+                  style={
+                    {
+                      "--slider-marker-percentage": `${calculateMarkerPosition(marker.value, min, max)}%`,
+                    } as React.CSSProperties
+                  }
+                  key={marker.value}
+                  as="label"
```
joshwooding, line 101:
```suggestion
                  styleAs="label"
```

packages/lab/src/slider/internal/SliderTrack.tsx
jake-costa: @tayyabataimur Yeah, we shouldn't have those as labels and we should also provide aria-hidden to these as well.

packages/lab/src/slider/Slider.css
joshwooding, line 1: There's a conversation here around if we are happy using 1 CSS file for all of Slider. Typically our classnames align to the exported component's name e.g. RangeSlider and Slider but that would result in duplicate CSS

tayyabataimur: We could look at unifying both Slider and Range Slider into one component?

tayyabataimur: Can break the CSS down as well so that both can have their own

origami-z (commented):
Leave a trace - the thumb is too narrow :)

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
```diff
@@ -0,0 +1,169 @@
...
+      <RangeSlider
+        style={{ width: "400px" }}
+        min={5}
+        max={125}
+        step={5}
+        defaultValue={[50, 100]}
+      />,
+    );
+
+    // TODO Finish tests
```
joshwooding, line 15: What's left to finish?

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
```diff
@@ -0,0 +1,169 @@
...
+    });
+    cy.get(".saltSliderTrack-rail").trigger("pointerup");
+
+    // First thumb
+    cy.findAllByRole("slider").eq(0).should("have.attr", "aria-valuenow", "2");
+    // First thumb
+    cy.findAllByRole("slider").eq(1).should("have.attr", "aria-valuenow", "7");
+  });
+
+  it("should change thumb positions based on keyboard navigation", () => {
```
joshwooding, line 55: Page Up and Page down seem to be missing?

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
```diff
@@ -18,26 +18,26 @@ describe("Given a Slider", () => {
...
+      cy.get(".saltSliderTrack-rail").trigger("pointerdown", {
         button: 0,
         clientX: 50,
         clientY: 50,
       });
       cy.get("@changeSpy").should("have.callCount", 1);
     });
 
-    it("THEN keyboard navigation can be used to change the slider position", () => {
+    it("should change the thumb position on slider based on keyboard navigation", () => {
```
joshwooding, line 32: Page Up and Page Down are missing here too

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
joshwooding, line 1: Do we have enough test coverage for dragging?

joshwooding: Is format tested?

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
```diff
@@ -18,26 +18,26 @@ describe("Given a Slider", () => {
...
+    it("should fire onChange on pointer down on slider track", () => {
       const changeSpy = cy.stub().as("changeSpy");
       cy.mount(<Slider style={{ width: "400px" }} onChange={changeSpy} />);
-      cy.get(".saltSliderTrack").trigger("pointerdown", {
+      cy.get(".saltSliderTrack-rail").trigger("pointerdown", {
         button: 0,
         clientX: 50,
         clientY: 50,
       });
       cy.get("@changeSpy").should("have.callCount", 1);
```
joshwooding, line 29: We should test the correct values are returned when we test callbacks

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
joshwooding: Are both uncontrolled and controlled tested

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,174 @@
...
-            {max}
-          </Label>
-        )}
-        {marks === "all" && <SliderMarks max={max} min={min} step={step} />}
-      </div>
-    </SliderContext.Provider>
+      <SliderThumb
+        aria-describedby={formFieldDescribedBy}
+        aria-label={ariaLabel}
+        aria-labelledby={formFieldLabelledBy}
```
joshwooding, line 155: aria-labelledby and aria-describedby being passed in as props should also be handled (and tested)

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,174 @@
...
+        aria-valuemax={max}
+        aria-valuemin={min}
+        aria-valuetext={ariaValueText}
```
joshwooding, line 158: Should we be providing guidance over this? @jake-costa

packages/lab/src/slider/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./Slider";
-export * from "./types";
+export * from "./RangeSlider";
+export * from "./internal/useSliderThumb";
```
joshwooding, line 3: We should be careful that this isn't exposing this from the package unnecessarily, import {} from "@salt-ds/core"

packages/lab/src/slider/internal/SliderThumb.css
```diff
@@ -0,0 +1,56 @@
+.saltSliderThumb {
+  position: absolute;
+  width: var(--salt-size-border-strong);
+  height: var(--salt-size-selectable);
+  background: var(--salt-accent-borderColor);
+  transform: translateX(-50%);
+  cursor: pointer;
```
joshwooding, line 7: This should probably be using a cursor token

packages/lab/src/slider/internal/SliderThumb.css
```diff
@@ -0,0 +1,56 @@
...
+.saltSliderThumb-focused {
+  outline-style: var(--salt-focused-outlineStyle);
+  outline-width: var(--salt-focused-outlineWidth);
+  outline-offset: var(--salt-focused-outlineOffset);
+  outline-color: var(--salt-focused-outlineColor);
+}
+
+.saltSliderThumb:active,
+.saltSliderThumb-dragging {
+  cursor: grabbing;
```
joshwooding, line 26: This should be using a token

packages/lab/src/slider/internal/SliderThumb.css
```diff
@@ -0,0 +1,56 @@
...
+
+.saltSliderThumb-secondThumb {
+  z-index: 10;
+}
+
+/* Visually hide accessible text. Can not
+  use visibility: hidden as we need the screen
+  reader to announce it */
+.saltSliderThumb-accessibleText {
+  position: absolute;
```
joshwooding, line 54: I think we should use the standard approach here? There should be an example elsewhere in the code

packages/lab/src/slider/internal/SliderThumb.tsx
```diff
@@ -1,82 +1,163 @@
...
 
 const withBaseName = makePrefixer("saltSliderThumb");
 
-export interface SliderThumbProps extends ComponentPropsWithoutRef<"div"> {
-  index: ThumbIndex;
-  activeThumb: ActiveThumbIndex;
-  setActiveThumb: (index: ActiveThumbIndex) => void;
+interface SliderThumbProps
+  extends Omit<
+    HTMLAttributes<HTMLInputElement>,
```
joshwooding, line 19:
```suggestion
    ComponentPropsWithoutRef<"input"> ,
```

Will cover more types

packages/lab/src/slider/internal/SliderThumb.tsx
```diff
@@ -1,82 +1,163 @@
...
+      setTimeout(() => {
+        setIsTooltipVisible(false);
+      }, 300);
+    };
 
-  const thumbPosition = getPercentage(min, max, thumbValue as number);
+    const handleFocus = () => {
+      // We add focus to the input to get the keyboard
+      // interactions for the slider out of the box.
+      if (inputRef.current) inputRef.current.focus();
```
joshwooding, line 90: Is this actually necessary? It seems to call focus on the input when the input is focused?

packages/lab/src/slider/internal/SliderThumb.tsx
```diff
@@ -1,82 +1,163 @@
...
-      className={withBaseName("container")}
-      onPointerOver={handlePointerOver}
-      onFocus={handleFocus}
-      onBlur={handleBlur}
-    >
+    return (
       <div
-        className={clsx(withBaseName("tooltip"), {
-          [withBaseName("tooltip-visible")]: showTooltip,
+        aria-describedby={`sliderTooltip-${id}-${index}`}
```
joshwooding, line 112: Is this correct? @jake-costa

packages/lab/src/slider/internal/SliderTooltip.css
```diff
@@ -0,0 +1,31 @@
...
+  position: absolute;
+  background: var(--salt-container-primary-background);
+  border-radius: var(--salt-palette-corner-weak);
+  bottom: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border: 1px solid var(--salt-accent-borderColor);
+  padding: var(--salt-spacing-50) var(--salt-spacing-100);
+  user-select: none;
+  z-index: 1;
```
joshwooding, line 11: Just curious is there a reason we didn't use the z-index used in Tooltip here?

packages/lab/src/slider/internal/SliderTooltip.css
```diff
@@ -0,0 +1,31 @@
+.saltSliderTooltip {
+  position: absolute;
+  background: var(--salt-container-primary-background);
+  border-radius: var(--salt-palette-corner-weak);
+  bottom: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border: 1px solid var(--salt-accent-borderColor);
```
joshwooding, line 8: Missing tokens?

packages/lab/src/slider/internal/SliderTooltip.css
```diff
@@ -0,0 +1,31 @@
...
+  padding: var(--salt-spacing-50) var(--salt-spacing-100);
+  user-select: none;
```
joshwooding, line 10: We should check this behaviour, although I assume this is included to prevent selecting text when dragging. When not dragging should selecting text work? @jake-costa  @navkaur76

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -0,0 +1,123 @@
...
+  width: 100%;
+  background: linear-gradient(
+    to right,
+    var(--slider-track-fill) calc(var(--slider-progressPercentage) - 1px),
+    transparent calc(var(--slider-progressPercentage) - 1px),
+    transparent calc(var(--slider-progressPercentage) + 1px),
+    var(--slider-track-background) calc(var(--slider-progressPercentage) + 1px)
+  );
+  height: var(--salt-size-bar);
+  cursor: pointer;
```
joshwooding, line 54: Same here

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -0,0 +1,123 @@
...
+
+.saltSliderTrack-rail {
+  border-radius: var(--salt-palette-corner-strongest);
+  flex: 100%;
+  order: 0;
+  position: relative;
+  width: 100%;
+  background: linear-gradient(
+    to right,
+    var(--slider-track-fill) calc(var(--slider-progressPercentage) - 1px),
```
joshwooding, line 48: Are all these `1px` related? Should we save this in a variable to make it easier to understand

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -0,0 +1,123 @@
...
+    transparent calc(var(--slider-progressPercentage) + 1px),
+    var(--slider-track-background) calc(var(--slider-progressPercentage) + 1px)
+  );
+  height: var(--salt-size-bar);
+  cursor: pointer;
+}
+
+.saltSliderTrack-minLabel,
+.saltSliderTrack-maxLabel {
+  user-select: none;
```
joshwooding, line 59: We should probably just check if this is necessary everywhere

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -0,0 +1,123 @@
...
+}
+
+.saltSliderTrack-minLabel {
+  order: 1;
```
joshwooding, line 63: I always get a bit nervous when I see `order` being used, due to accessibility worried @jake-costa  maybe something to have a quick look at

packages/lab/stories/slider/slider.qa.stories.tsx
```diff
@@ -3,94 +3,65 @@ import type { Meta, StoryFn } from "@storybook/react";
 import { QAContainer, type QAContainerProps } from "docs/components";
 
 export default {
-  title: "Lab/Slider/Slider QA",
```
joshwooding, line 6: This should be changed back, it makes it easier to tell what story you're dealing with on Chromatic

site/docs/components/slider/examples.mdx
```diff
@@ -9,24 +9,21 @@ data:
...
 <LivePreviewControls>
-
 <LivePreview componentName="slider" exampleName="Single">
 
 ## Single
 
-By default the minimum and maximum values are shown, and can be displayed either inline or below the track (using the `marks` prop to `"bottom"`).
-
-Alternately, incremental marks can be shown along the track by setting the `marks` prop to `"all"`. The incremental marks will match the `step` prop value (i.e. for a `step` value of `5`, the marks will display as "0", "5", "10", "15", etc.) Note that when all marks are shown, they are always displayed below the track.
+Selection of a single value from a horizontally oriented slider track. Values can be displayed either inline or below the track.
```
joshwooding, line 16: values or labels?

joshwooding, line 16: I can understand why but since we only support one orientation for now we don't need to add "horizontally oriented" everywhere

site/docs/components/slider/examples.mdx
```diff
@@ -73,8 +94,30 @@ You can display inputs with either a single or range slider. Inputs are used to
...
+- For single slider, display one input to the left of the track.
+- Only use the inline variant of slider alongside an input. This ensures the size-base alignment of the overall component layout is maintained.
+
+</LivePreview>
+
+<LivePreview componentName="slider" exampleName="WithMarkers">
+
+## With Markers
+
+You can display markers under the values of your choice with custom labels.
```
joshwooding, line 106: This wording confused me a bit, "markers" are the labels so we might need to clear up the wording a bit

joshwooding (commented):
Looking good 🎉, left a few comments/thoughts. Not all of these need to be resolved before going to labs but it's better to raise them now.

packages/lab/src/slider/internal/SliderTooltip.css
tayyabataimur: @joshwooding Yep, this was intentional to prevent selecting the tooltip's text when dragging. Good question about being able to select the text when not dragging, keen to hear Jake and Nav's thoughts on this.

packages/lab/src/slider/internal/SliderThumb.tsx
tayyabataimur: Have removed this based on Jake's feedback

packages/lab/src/slider/internal/SliderTooltip.css
tayyabataimur: Good point, will add in missing tokens

site/docs/components/slider/examples.mdx
tayyabataimur: So the way I think of this is, we always have min/max `labels` and if we want to add more labels, we pass them through `markers` or `marks` to add those labels at certain `marks`.

packages/lab/src/slider/internal/SliderThumb.css
```diff
@@ -0,0 +1,59 @@
...
+
+.saltSliderThumb-secondThumb {
+  /* The second thumb needs to sit on top the first thumb */
+  z-index: 2;
+}
+
+/* Visually hide accessible text. Can not
+  use visibility: hidden as we need the screen
+  reader to announce it */
+.saltSliderThumb-accessibleText {
```
Fercas123, line 56: NIT.  we recently added the following class to button for this purpose, i am copying it in for carousel accessible's text, might be worth using the same here?

```.saltButton-sr-only {
  position: fixed;
  top: 0;
  left: 0;
  transform: translate(-100%, -100%);
}```

Comment (origami-z) [reply packages/lab/src/slider/internal/SliderThumb.css:56]: Need more testing on this, we received bug report on above code in `Step.SROnly.css`

**Review by Fercas123 — DISMISSED** [review]

File: packages/lab/src/slider/internal/SliderThumb.css (hunk 32 above)
Comment (tayyabataimur) [reply packages/lab/src/slider/internal/SliderThumb.css:56]: Since the slider has a `position: relative` set, adding these attributes results in the text remaining visible on the slider 😞

**Review by Fercas123 — APPROVED** [review]

File: packages/lab/src/slider/internal/SliderThumb.css (hunk 32 above)
Comment (joshwooding) [reply packages/lab/src/slider/internal/SliderThumb.css:56]: Could you try:

```css
position: absolute;
  height: 1px;
  width: 1px;
  padding: 0;
  margin: -1px;
  overflow: hidden;
  clip: rect(0, 0, 0, 0);
  white-space: nowrap;
  border-width: 0;
```

packages/lab/stories/slider/slider.stories.tsx
```diff
@@ -1,44 +1,29 @@
...
   FlexLayout,
   FormField,
-  FormFieldHelperText,
   FormFieldLabel,
   Input,
   StackLayout,
 } from "@salt-ds/core";
-import { Slider, type SliderProps, type SliderValue } from "@salt-ds/lab";
 import type { StoryFn } from "@storybook/react";
+import { Slider, type SliderProps } from "packages/lab/src/slider/Slider";
```
joshwooding, line 9:
```suggestion
import { Slider, type SliderProps } from "@salt-ds/core";
```

tayyabataimur: `@salt-ds/lab`*

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
tayyabataimur: Removed comment

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
tayyabataimur: Added

packages/lab/src/slider/internal/SliderThumb.css
```diff
@@ -0,0 +1,56 @@
+.saltSliderThumb {
+  position: absolute;
+  width: var(--salt-size-border-strong);
+  height: var(--salt-size-selectable);
+  background: var(--salt-accent-borderColor);
+  transform: translateX(-50%);
+  cursor: pointer;
+  outline: none;
+  top: -5px;
```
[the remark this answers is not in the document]
tayyabataimur: Replaced with token calculation

packages/lab/src/slider/internal/SliderThumb.css
tayyabataimur: Done 👍

origami-z:
Purely on storybook, i found it very hard to click accurately on the super narrow track.... is this something UX should help?

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,178 @@
...
+    clamp,
+    isDragging,
+    sliderRef,
+  } = useSliderThumb({
+    min,
+    max,
+    step,
+    valueState,
+    // @ts-ignore onChange can accept both number and [number, number] value types
+    onChange,
```
origami-z, line 119: [Not blocking PR] Should `onChange` fire when at min/max (use keyboard up/down)? `<input type="range">` doesn't fire change event when at limit.

.changeset/smart-buttons-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Update Slider component implementation to have separate Slider and Range Slider components, with improved accessibility and test coverage.
```
origami-z, line 5: If possible,  add a diff of before / after to help migrate

packages/lab/src/slider/Slider.tsx
origami-z: Similarly, should it fire when one thumb move towards another thumb in range picker?

origami-z: Another note, [storybook page](https://jpmorganchase-slider-update.saltdesignsystem-storybook.pages.dev/?path=/story/lab-rangeslider--with-marks) becomes unresponsive when i drag the handle quickly. Is there any throttling or unnecessary render happening?

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
```diff
@@ -0,0 +1,286 @@
...
+    cy.get("@changeSpy").should("have.callCount", 1);
+    cy.get("@changeSpy").should(
+      "have.been.calledWith",
+      Cypress.sinon.match.any,
+      Cypress.sinon.match.array.deepEquals([0, 7]),
+    );
+    cy.get(".saltSliderTrack-rail").trigger("pointerup");
+  });
+
+  it("should set thumb positions correctly when clicked on the track", () => {
```
origami-z, line 34: [non-blocking] is there anything to test to determine which thumb should be moved

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
```diff
@@ -0,0 +1,286 @@
...
+    // Focus and move first thumb
+    cy.findAllByRole("slider").eq(0).focus().realPress("PageUp");
+    cy.findAllByRole("slider").eq(0).should("have.value", "10");
+
+    // Focus and move second thumb
+    cy.findAllByRole("slider").eq(1).focus().realPress("PageDown");
+    cy.findAllByRole("slider").eq(1).should("have.value", "20");
+  });
+
+  it("should not allow thumbs to overlap", () => {
```
origami-z, line 157: both are 8, not overlap?

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+      "aria-label": ariaLabel,
+      "aria-labelledby": clsx(formFieldLabelledBy, ariaLabelledBy),
+      "aria-valuemax": max,
+      "aria-valuemin": min,
+      "aria-valuetext": ariaValueText,
+      disabled: disabled,
+      format: format,
+      max: max,
+      min: min,
+      name: name,
```
origami-z, line 148: Worth to find out whether same `name` passed down to 2 thumbs with `input` would cause any trouble

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
```diff
@@ -1,169 +1,200 @@
...
+    );
+
+    cy.findByText("Custom Min Label").should("exist");
+    cy.findByText("Custom Max Label").should("exist");
+  });
+
+  it("should be disabled when set", () => {
+    cy.mount(<Default disabled defaultValue={2} />);
+
+    cy.findByRole("slider").should("be.disabled");
```
origami-z, line 188: worth to add functional pieces like click / focus won't work

packages/lab/src/slider/internal/SliderThumb.tsx
```diff
@@ -1,82 +1,194 @@
...
+          onBlur={handleBlur}
+          className={withBaseName("input")}
+          value={value}
+          onChange={handleInputChange}
+          onKeyDown={handleKeydownOnThumb}
+          aria-labelledby={ariaLabelledBy}
+          aria-valuenow={value}
+          aria-valuetext={ariaValueText}
+          aria-label={ariaLabel}
+          aria-describedby={`saltSlider-${id}-${index}`}
```
origami-z, line 173: very minor, if we compute an `id`, do it once and pass down the same. or it can be broken later easily

packages/lab/src/slider/internal/SliderTooltip.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import clsx from "clsx";
+import sliderTooltipCss from "./SliderTooltip.css";
+
+const withBaseName = makePrefixer("saltSliderTooltip");
+
+interface SliderTooltipProps {
+  value: number | string;
+  isVisible?: boolean;
```
origami-z, line 11: doesn't matter too much, `Tooltip` has `open` prop

origami-z (dismissed):
good change for lab
preliminary 👍

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
tayyabataimur: I think the correct term should be "being able to cross each other"? Since technically they can overlap, they just can't pass through the other.

origami-z: yeah sure

packages/lab/src/slider/RangeSlider.tsx
tayyabataimur: Checked a number of different component libraries and they all seem to have it this way so hopefully it shouldn't be a problem

.changeset/smart-buttons-applaud.md
mark-tate: "Range Slider" to `RangeSlider` so we create familiarity with the component names and wrap `Slider` in backticks, so both stand out as component names.

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Change handler to be used when in a controlled state
+   */
+  onChange?: (event: SyntheticEvent, value: [number, number]) => void;
+  /**
+   * Minimum interval the slider thumb can move
+   */
+  step?: number;
+  /**
+   * The multiplier for the step when using PageUp and PageDown keys via keyboard interaction
```
mark-tate, line 66: `using PageUp and PageDown keys via keyboard interaction`

these repeat themselves perhaps?

could be

`The multiplier for the step when using PageUp and PageDown keys`

mark-tate: equally, if you compare this to step description, we could say 
`Maximum interval the slider thumb can move when using PageUp and PageDown keys`

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+      value: valueProp,
+      defaultValue = [min, min + step],
+      ...rest
+    },
+    ref,
+  ) {
+    const [valueState, setValue] = useControlled({
+      controlled: valueProp,
+      default: defaultValue,
+      name: "Slider",
```
mark-tate, line 103: should this be RangeSlider

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+      isDragging,
+      sliderRef,
+      thumbIndexState,
+      preventThumbOverlap,
+    } = useSliderThumb({
+      min,
+      max,
+      step,
+      valueState,
+      // @ts-ignore onChange can accept both number and [number, number] value types
```
mark-tate, line 127: should useSliderThumb be passed a variant, "single" or "range" and then use discriminating union, so that we do not need to ts-ignore ?

have a look at how datepicker and calendar can switch between single and range dates based on the selectionVariant

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Label for the minimum value
+   */
+  minLabel?: string;
+  /**
+   * Name of the slider input
+   */
+  name?: string;
+  /**
+   * Change handler to be used when in a controlled state
```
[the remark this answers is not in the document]
tayyabataimur: Incorrect prop description, it can be used for both controlled and uncontrolled. I'll update it 👍

packages/lab/src/slider/internal/useSliderThumb.ts
```diff
@@ -0,0 +1,240 @@
...
+  const sliderRef = useRef<HTMLDivElement>(null);
+  const targetWindow = useWindow();
+
+  useEffect(() => {
+    if (isDragging) {
+      targetWindow?.addEventListener("pointermove", handlePointerMove);
+      targetWindow?.addEventListener("pointerup", handlePointerUp);
+    } else {
+      targetWindow?.removeEventListener("pointermove", handlePointerMove);
+      targetWindow?.removeEventListener("Pointerup", handlePointerUp);
```
mark-tate, line 38: typo, will mean the listener is never removed as addEventListener is case sensitive

packages/lab/src/slider/internal/utils.ts
```diff
@@ -1,131 +1,15 @@
...
-  min: number,
-  max: number,
-) => {
-  if (typeof value === "undefined" || value.length < 1) return;
-  const a = clampValue(value[0], [min, max]);
-  if (value.length === 1) return [a] as SliderValue;
-  const b = clampValue(value[1], [min, max]);
-  if (a > b) return [a, a] as SliderValue;
-  return [a, b] as SliderValue;
+  const markerPosition = ((clampedValue - min) / (max - min)) * 100;
```
mark-tate, line 13: if max and min are equal you would get divide by zero ?

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -0,0 +1,144 @@
+.saltSliderTrack {
+  --slider-progressPercentage: "0%";
```
mark-tate, line 2: should these be quoted ?

packages/lab/src/slider/index.ts
mark-tate: if it's in internal would say that it should not be shared ?
if it needs to be shared move it out of internal?

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+import clsx from "clsx";
+import { SliderThumb } from "./internal/SliderThumb";
+import { SliderTrack } from "./internal/SliderTrack";
+import { useSliderThumb } from "./internal/useSliderThumb";
+import { toFloat } from "./internal/utils";
+
+export interface RangeSliderProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, "onChange" | "defaultValue"> {
+  /**
+   * Initial value of the slider
```
mark-tate, line 18: I felt Initial was not how we had described other uncontrolled defaults.
I then copied this from Tabs.  As a reader of docs, good to have consistent descriptions.

The default value. Use when the component is not controlled.

site/src/examples/slider/WithFormFieldRange.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+export const WithFormFieldRange = (): ReactElement => {
+  return (
+    <StackLayout style={{ width: "400px" }}>
+      <FormField labelPlacement="left">
+        <FormFieldLabel>Field Label</FormFieldLabel>
+        <RangeSlider min={0} max={50} step={10} />
+      </FormField>
+
+      <FormField style={{ width: "400px" }}>
```
mark-tate, line 13: Maybe ask Design if this is a good representation of formfield labels?
As we put two sliders together with different label placements, it makes the "left" alignment look weird due to the size of the slider (400px). 
ie. the gap is rather large between label and component

packages/lab/src/slider/Slider.tsx
mark-tate: Open in it's own Canvas it' ok, although does it suggest a performance issue?
Worth checking.

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,178 @@
...
-   * Value of the slider, to be used when in a controlled state
+   * Name of the slider input
    */
-  value?: SliderValue;
+  name?: string;
   /**
    * Change handler to be used when in a controlled state
    */
-  onChange?: SliderChangeHandler;
+  onChange?: (event: SyntheticEvent, value: number) => void;
```
mark-tate, line 60: I think we need an additional API handler, equivalent of  onChangeCommitted (Mui), onChangeEnd (Mantine) or onChangeEnd (Chakra).
All these APIs enable you to just listen for the final value rather than everything in-between.

packages/lab/src/slider/internal/useSliderThumb.ts
```diff
@@ -0,0 +1,240 @@
...
+
+      // Range slider
+      if (Array.isArray(sliderValue)) {
+        const values = preventThumbOverlap(
+          newValue,
+          sliderValue as [number, number],
+          thumbIndexState,
+        );
+        setValue(values);
+        onChange?.(event, values);
```
mark-tate, line 131: Because we have steps, we have to check here if the value has changed and only call onChange when there is a new value. Otherwise, there will be a high number of change events called and we would create performance issues in the consumers code.

mark-tate: This might be related to the perf issues in Storybook ?

site/src/examples/slider/CustomStep.tsx
```diff
@@ -6,19 +6,25 @@ export const CustomStep = (): ReactElement => (
   <StackLayout gap={10} style={{ width: "400px" }}>
     <FormField>
       <FormFieldLabel>Step: 1 (default)</FormFieldLabel>
-      <Slider min={-1} max={1} marks="all" />
+      <Slider min={-1} max={1} />
     </FormField>
     <FormField>
       <FormFieldLabel>Step: 0.2</FormFieldLabel>
-      <Slider min={-1} max={1} step={0.2} marks="all" />
```
mark-tate, line 13: I get a strange rounding issue on 0.6 an -0.6 on this example, see the tooltip, it will have several decimal places

site/src/examples/slider/WithMarkers.tsx
```diff
@@ -0,0 +1,60 @@
+import { StackLayout } from "@salt-ds/core";
+import { RangeSlider, Slider } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+export const WithMarkers = (): ReactElement => (
```
mark-tate, line 5: maybe not for the mvp release but is there a use case where markers are set but steps are un-equal?
if markers and steps are equal then you can use the two props to only change between the visible steps.
But what if the markers have gaps, so the step is not uniform.
See this example in mantine
https://mantine.dev/core/slider/#restrict-selection-to-marks

site/docs/components/slider/examples.mdx
```diff
@@ -36,9 +33,9 @@ If two values are provided (either to the `value` or `defaultValue` props), then
...
+Custom step values can be set to any number, including decimal values up to two decimal places.
 
-If a custom step value is provided that is not a multiple of the range, the slider will round the maximum selectable value to the largest multiple of the step that is less than the maximum value. **Note:** This will render the final portion of the slider track unusable (the portion between the rounded maximum value and the actual maximum value). Step calculations are made from `0`, so the same rounding logic applies to negative values, but in the opposite direction.
+If a custom step value is provided that is not a multiple of the range, the slider will round the maximum selectable value to the largest multiple of the step that is less than the maximum value. **Note:** This will render the final portion of the slider track inaccessible by keyboard navigation but still draggable by pointer (the portion between the rounded maximum value and the actual maximum value). Step calculations are made from `0`, so the same rounding logic applies to negative values, but in the opposite direction.
 
 ### Best practices
```
mark-tate, line 40: best practices uses a subjective term, "tedious" selection

Could we rewrite 

Use a step value that aligns with the level of detail required on the scale. This ensures users can make selections efficiently, minimizing unnecessary adjustments and improving the overall usability of the slider.

site/src/examples/slider/WithInput.tsx
```diff
@@ -83,7 +75,7 @@ export const SingleWithInput = () => {
 const RangeWithInput = () => {
   const bounds: [number, number] = [0, 100];
 
-  const [value, setValue] = useState<SliderValue>([20, 60]);
+  const [value, setValue] = useState<[number, number]>([20, 60]);
```
mark-tate, line 78: Set the value to an invalid value with 3 digits and the input truncates the value

site/src/examples/slider/WithInput.tsx
```diff
@@ -44,22 +41,17 @@ export const SingleWithInput = () => {
...
+    setValue(Number.parseFloat(inputValue));
   };
 
-  const handleChange: SliderChangeHandler = (value: SliderValue) => {
+  const handleChange = (_e: SyntheticEvent, value: number) => {
     setValue(value);
   };
 
-  useEffect(() => {
-    const valid = validateSingle(inputValue, bounds);
```
mark-tate, line 54: Looks like we removed validation from the single slider ?
Range still has it but not single

site/src/examples/slider/WithMarkers.tsx
mark-tate, line 5: What are markers ?
Design / a11y question @jake-costa @navkaur76 
If we depend on labels to denote markers/steps and the labels are not clickable, what part of the track should be clicked to get the marker/step.. It's kind of quicker, easier to have some form of visual cue that says, click me, to get this value, Otherwise, I'm having to decide where on the track to click. Also, as steps and labels are optional, there is not obvious visual differentation between a slider with steps and a decimal slider.
Mantine has these little dots on the track which help 
https://mantine.dev/core/slider/#decimal-values

site/src/examples/slider/Single.tsx
```diff
@@ -4,14 +4,20 @@ import type { ReactElement } from "react";
 
 export const Single = (): ReactElement => (
   <StackLayout gap={3}>
-    <Slider aria-label="single" min={0} max={100} style={{ width: "400px" }} />
     <Slider
```
mark-tate, line 7: I start dragging a slider and move the cursor out of the example, the slider still drags as the mouse is held down but the cursor reverts back to the pointer.

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+   */
+  maxLabel?: string;
+  /**
+   * Label for the minimum value
+   */
+  minLabel?: string;
+  /**
+   * Name of the slider input
+   */
+  name?: string;
```
mark-tate, line 56: what do we use name for ?
(is it format related, can't see this added on other components)

jake-costa:
@tayyabataimur "With Non Numeric Values" example now conveys the label of the day, however the range is saying "minimum 1, maximum 7."

packages/lab/src/slider/internal/useSliderThumb.ts
tayyabataimur: Ah, good catch!

packages/lab/src/slider/internal/utils.ts
tayyabataimur: Can a slider be valid if max and min are equal? I could add a safeguard for that but just wondering if this use-case is possible 🤔

packages/lab/src/slider/internal/useSliderThumb.ts
```diff
@@ -0,0 +1,240 @@
...
+      sliderValue,
+      thumbIndexState,
+      setValue,
+      onChange,
+      getClickedPosition,
+      preventThumbOverlap,
+    ],
+  );
+
+  // ******* Event Handlers ***********
```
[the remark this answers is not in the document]
tayyabataimur: Will remove 👍

packages/lab/src/slider/RangeSlider.tsx
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Change handler to be used when in a controlled state
+   */
+  onChange?: (event: SyntheticEvent, value: [number, number]) => void;
```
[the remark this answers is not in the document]
tayyabataimur: It can be ChangeEvent

packages/lab/src/slider/internal/useSliderThumb.ts
tayyabataimur: You're right! I'll update this so onChange is only fired when the value changes.

packages/lab/src/slider/RangeSlider.tsx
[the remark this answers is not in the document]
tayyabataimur: Actually, the reason that it is more generic is that the ChangeEvent can either be a PointerEvent, ChangeEvent or a React.PointerEvent since we are firing it via keyboard and mouse interactions.

site/docs/components/slider/examples.mdx
tayyabataimur: I'll update this description since we now allow the thumb to be dragged all the way to the maximum.

packages/lab/src/slider/RangeSlider.tsx
tayyabataimur: I'll remove it 👍  Initially added it thinking might be needed for forms but I don't think we actually do.

site/src/examples/slider/Single.tsx
tayyabataimur: I would consider this standard since we can only control the cursor within the bounds of the component. E.g. [mantine](https://mantine.dev/core/slider/) and [Ant](https://ant.design/components/slider) have the cursor pointer only within the component.

origami-z: Interesting to see both examples having bigger / taller hitbox on dragging. UX change to consider

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,178 @@
...
 export const Slider = forwardRef<HTMLDivElement, SliderProps>(function Slider(
   {
-    min = defaultMin,
-    max = defaultMax,
-    step = defaultStep,
-    value: valueProp,
-    defaultValue = [0],
-    onChange,
-    className,
     "aria-label": ariaLabel,
```
jake-costa, line 77: @tayyabataimur FYI I think it only effects the examples but we shouldn't use use a value for aria-label and aria-labelledby on the same element because aria-labelledby will take precedence over aria-label if both are applied

packages/lab/src/slider/RangeSlider.tsx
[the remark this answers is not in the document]
tayyabataimur: Actually, since it is being triggered by both input and div pointer events, it needs to be generic. I'll update the prop description to reflect this.

packages/lab/src/slider/internal/utils.ts
mark-tate: validation is a userland responsibility but at the same time worth checking this can't throw through normal use.

resolve conversation if your satisfied this is not an issue.

site/src/examples/slider/WithMarkers.tsx
mark-tate: What's the design view of this? @navkaur76 

If we say that the track can have markers, does that provide a better user experience?

When do the markers appear, on hover  or all the time, or is it configurable ?

FYI: Link Pokemon/PR is adding a prop for underline that says when the adornment appears, so there is a precedent  where we  can configure, when something appears?

Perhaps a key question from design and maybe @jake-costa is what do the markers represent?
How would we describe their use ?

Are they a visual indicator on the track so we do not need labels on every step?
ie. 10 markers, with labels for 0, 50 and 100
here the markers would tell for 10, 20, 30 ,40 without having to add more labels.
However, they also help the user know where to click the track without having to search for it.

packages/lab/src/slider/Slider.tsx
tayyabataimur: I have updated site example docs 👍

packages/lab/src/slider/internal/utils.ts
tayyabataimur: I have added a safeguard 👍

packages/lab/src/slider/Slider.tsx
tayyabataimur: Added 👍

packages/lab/src/slider/RangeSlider.tsx
tayyabataimur: After trying out union types, I realised that it would be far more simpler to separate the range and single slider hooks simply because of the complexity the types were introducing. So I have now separated the logic for the single and range slider hooks.

origami-z (approved):
A good change for lab in the current form

site/docs/components/range-slider/examples.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+`RangeSlider`'s labels indicate the minimum and maximum values of the range, and can be located either inline (default), or at the bottom of the `RangeSlider`.
+
+</LivePreview>
+
+<LivePreview componentName="range-slider" exampleName="CustomStep">
+
+## Custom Step
+
+By default, the `RangeSlider` step increment is set to 1. This can be customized by using the ‘step’ prop. Custom step values can be set to any number, including decimal values up to two decimal places.
```
mark-tate, line 34: should be back ticks on step

packages/lab/src/slider/Slider.tsx
```diff
@@ -1,127 +1,178 @@
...
+  /**
+   * A callback to format the display value in the tooltip
+   */
+  format?: (value: number) => string | number;
+  /**
+   * Position of the labels
+   */
+  labelPosition?: "bottom" | "inline";
+  /**
+   * The markers to show under the slider to label some values
```
mark-tate, line 34: this needs rewording - markers and labels are different things

tayyabataimur: Updated 👍

site/docs/components/range-slider/examples.mdx
```diff
@@ -0,0 +1,122 @@
...
+
+## With Non Numeric values
+
+You can implement a `RangeSlider` that allows users to select non-numeric values, with a focus on accessibility through the use of `aria-valuetext`.
+
+The `RangeSlider` maps numeric values to meaningful non-numeric options, enabling intuitive selection beyond simple numbers. The `aria-valuetext` attribute is dynamically updated to announce the selected option, providing clear feedback for users with assistive technologies. Each option is marked on the slider, offering visual guidance on available selections.
+
+### Best practices:
+
+- State Management: Utilizes `useState` to track the selected option.
```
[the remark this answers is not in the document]
tayyabataimur: Updated 👍

packages/lab/src/slider/Slider.tsx
tayyabataimur: Fixed 👍

site/src/examples/slider/WithMarkers.tsx
tayyabataimur: Will add in follow up PR 👍

tayyabataimur: Design meeting scheduled, will add potential implementation in follow up PR 👍

origami-z (approved):

Fercas123 (approved):

.changeset/smart-buttons-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Update Slider component implementation to have separate `Slider` and `RangeSlider` components, with improved accessibility and test coverage.
```
mark-tate, line 5: can we add a basic example in code, so when reading the changelog I get  an understanding in code of what the API looks like, see other changelogs for when we added new components.

mark-tate (approved):

origami-z (approved):

PR #4635, opened 2025-01-31
Carousel

Fercas123:
Updates to carousel


packages/lab/src/carousel/Carousel.tsx
```diff
@@ -1,170 +1,47 @@
...
-   * The animation when the slides are shown.
-   * Optional. Defaults to `slide`
-   **/
-  animation?: "slide" | "fade";
+  activeSlideIndex?: number;
   /**
    * If this props is passed it will set the aria-label with value to the carousel container.
    * Optional. Defaults to undefined
    */
   carouselDescription?: string;
```
joshwooding, line 14: We should probably just use aria-label in this case

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,150 @@
...
+        if (containerRef.current) {
+          setSliderW(containerRef.current.offsetWidth);
+        }
+      };
+      targetWindow?.addEventListener("resize", handleResize);
+      return () => targetWindow?.removeEventListener("resize", handleResize);
+    }, []);
+
+    useEffect(() => {
+      if (process.env.NODE_ENV !== "production") {
```
joshwooding, line 106: Not sure if we need to warn about this

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,150 @@
...
+    useEffect(() => {
+      if (containerRef.current) {
+        setSliderW(containerRef.current.offsetWidth);
+      }
+      const handleResize = () => {
+        if (containerRef.current) {
+          setSliderW(containerRef.current.offsetWidth);
+        }
+      };
+      targetWindow?.addEventListener("resize", handleResize);
```
joshwooding, line 101: Is it possible to use a resize observer instead?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,75 +1,62 @@
...
-              [withBaseName("textContainerLeft")]: contentAlignment === "left",
+      <div role="group" ref={ref} className={withBaseName()}>
+        {media}
+        <div
+          className={clsx(withBaseName("content"), {
+            [withBaseName("bordered")]: bordered,
+          })}
+        >
+          {children}
+          <SplitLayout
```
joshwooding, line 54: It's fine to use layout components in explorations but I think we should rely on CSS in our components

packages/lab/src/carousel/Carousel.tsx
```diff
@@ -1,170 +1,47 @@
...
-   * Optional. Defaults to false
-   **/
-  compact?: boolean;
-  /**
-   * It sets the id for the Carousel Container.
-   * String. Optional
+   * A boolean. When `true`, the slider will receive a full border.
    */
-  id?: string;
+  bordered?: boolean;
```
joshwooding, line 18: Styles might need to be checked, currently this shifts the alignment

packages/lab/src/carousel/CarouselControls.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const slidesCount = slides.length;
+
+  const isOnFirstSlide = activeSlide === 0;
+  const isOnLastSlide = activeSlide === slidesCount - 1;
+
+  const ControlsLabel = () => (
+    <Text as="span">
+      <strong>
+        {/* TODO: check w dev, this will need changing or a formatter*/}
```
joshwooding, line 41: Probably fine for now, we should align with Pagination

packages/lab/src/carousel/CarouselControls.tsx
```diff
@@ -0,0 +1,84 @@
...
+        {activeSlide + 1} of {slidesCount}
+      </strong>
+    </Text>
+  );
+  const Controls = () => (
+    <div role="group" className={withBaseName("controls")}>
+      {/*  TODO: check w dev, currently some props are in the context and some out, should title go in context? */}
```
joshwooding, line 48: Personally I don't think title should go in context

packages/lab/src/carousel/CarouselControls.tsx
```diff
@@ -0,0 +1,84 @@
...
+      {!title && <ControlsLabel />}
+    </div>
+  );
+
+  return (
+    <SplitLayout
+      align="center"
+      className={withBaseName()}
+      ref={ref}
+      startItem={title ? <H2>{title}</H2> : <Controls />}
```
joshwooding, line 79: I don't think we can assume H2 semantically?

packages/lab/src/carousel/Carousel.tsx
joshwooding: Just thought a bit more and I think the title in the controls will be programatically linked to the carousel so this won't be needed most of the time anyway

packages/lab/src/carousel/CarouselControls.tsx
Fercas123: does `CarouselHeader` make more sense than controls given the title is included in this component too? I think that is what making me noise

packages/lab/src/carousel/CarouselControls.tsx
Fercas123: you mean secondary & no bold? pagination seem to be forcing the of in, no formatter... should there be one to allow users to format the info?

packages/lab/src/carousel/Carousel.tsx
Fercas123: that's intentional, but the prop name might not give it very clearly... I will check this with @ivan-calderon but there is a Hero (transparent) version and a...norther(bordered) one hahaha will check the naming

jake-costa (commented):
A few additional items:

- When the previous or next button is pressed the position should also be conveyed to screen reader users (e.g. "slide label" 2 of 5)

- When previous button is pressed moving the carousel back to slide 1 (if we are not allowing wrapping) and the button becomes disabled, focus should be placed on the next button. Similarly when pressing next button for the last slide, when the next button is disabled, focus should be placed on the previous button.  

- Just checking, for mobile view/small viewport did we want to leave the previous/next buttons at the top?

packages/core/src/utils/index.ts
```diff
@@ -16,6 +16,7 @@ export * from "./useForkRef";
 export * from "./useId";
 export * from "./useIsFocusVisible";
 export * from "./useIsomorphicLayoutEffect";
+export * from "./useIntersectionObserver";
```
joshwooding, line 19: Could we put this in labs first please?

site/docs/components/carousel/index.mdx
```diff
@@ -0,0 +1,11 @@
+---
+title: Carousel
+data:
+  description: |
+    `Carousel` groups small amounts of content into discrete slides, that are viewed in a linear sequence.
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/core/src/carousel"
```
joshwooding, line 6:
```suggestion
  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/lab/src/carousel"
```

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,103 @@
...
+          setIsScrolling(false);
+        }, 300);
+      }
+    },
+    [isScrolling, nextSlide, prevSlide],
+  );
+};
+
+export const CarouselSlider = forwardRef<HTMLDivElement, CarouselSliderProps>(
+  function CarouselSlider({ children, onKeyDown: onKeyDownProp }, ref) {
```
joshwooding, line 60: Should there be extra props here?

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,103 @@
...
+  children: Array<ReactElement<CarouselSlideProps>>;
+}
+
+const withBaseName = makePrefixer("saltCarouselSlider");
+
+const useKeyNavigation = ({
+  nextSlide,
+  prevSlide,
+}: Pick<CarouselContextValue, "nextSlide" | "prevSlide">) => {
+  const [isScrolling, setIsScrolling] = useState(false);
```
joshwooding, line 33: I thought we had made changes that removed this?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,106 @@
...
+  type ReactNode,
   forwardRef,
+  useEffect,
   useRef,
 } from "react";
+import { useCarousel } from "./CarouselContext";
+import carouselSlideCss from "./CarouselSlide.css";
 
-import { clsx } from "clsx";
+export const CarouselSlideAppearanceValues = [
```
joshwooding, line 15: I think this is to match button but can we simplify this and just a union please "bordered" | "transparent"

site/docs/components/carousel/accessibility.mdx
```diff
@@ -0,0 +1,33 @@
...
+</KeyboardControl>
+
+<KeyboardControl keyOrCombos="Enter">
+
+</KeyboardControl>
+
+<KeyboardControl keyOrCombos="Up arrow / Down arrow / Left arrow / Right arrow">
+
+</KeyboardControl>
+</KeyboardControls>
```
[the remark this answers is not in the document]
Fercas123: might be a next sprint task to finish site docs, will remove them for now

packages/lab/src/__tests__/__e2e__/carousel/Carousel.cy.tsx
```diff
@@ -0,0 +1,100 @@
...
+import { composeStories } from "@storybook/react";
+import { checkAccessibility } from "../../../../../../cypress/tests/checkAccessibility";
+
+const composedStories = composeStories(carouselStories);
+const { Default, WithActions } = composedStories;
+describe("GIVEN a 100% width slides carousel", () => {
+  checkAccessibility(composedStories);
+  describe("WHEN the default is rendered with slides", () => {
+    it("SHOULD render carousel", () => {
+      cy.mount(<Default />);
```
tayyabataimur, line 11: Should we also add the assertion here that the carousel component exists? 🤔

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+  useState,
+} from "react";
+
+export interface CarouselContextValue {
+  firstVisibleSlide: number;
+  visibleSlides: number;
+  slidesCount: number;
+  nextSlide: (event: MouseEvent | KeyboardEvent) => void;
+  prevSlide: (event: MouseEvent | KeyboardEvent) => void;
+  updateActiveFromScroll: (scrollLeft: number) => void;
```
mark-tate, line 20: how do you see Active and Visible?
Are they the same thing. On first read I have to read through code to try and understand the difference and I think they could be the same thing.

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+      }
+      focusSlide(previousSlide);
+    }
+    scrollToSlide(previousSlide);
+  };
+
+  return (
+    <CarouselContext.Provider
+      value={{
+        visibleSlides,
```
mark-tate, line 138: in date picker we have state and helpers.
is this a pattern that would make the API easier to understand?

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+  activeSlideIndex = 0,
+  visibleSlides = 1,
+  id,
+}: {
+  children: ReactNode;
+  activeSlideIndex?: number;
+  visibleSlides?: number;
+  id?: string;
+}) {
+  const [firstVisibleSlide, setFirstVisibleSlide] = useState(activeSlideIndex);
```
mark-tate, line 52: this suggests the prop name is not correctly named?
Should this be defaultVisibleSlideIndex (or similar) to tell the consumer, this is an initial value of an uncontrolled value ?

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const [visibleFocus, setVisibleFocus] = useState(0);
+  const [sliderW, setSliderW] = useState(0);
```
mark-tate, line 54: I guess it's width, could we rename this sliderWidth if so, and remove the ambiguity ?

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+  const slides = useRef<Map<number, HTMLDivElement>>(new Map());
+  const [slidesCount, setSlidesCount] = useState(slides.current.size);
```
mark-tate, line 56: Do we need to maintain this state or can we derive it from `slides.current.size` ?

Fercas123: slides.current.size would not update immediately, so the updates get to the consumer components one render late if i depend on it.

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+      left: targetScrollLeft,
+      behavior: "smooth",
+    });
+  };
+
+  const focusSlide = (index: number) => {
+    slides.current.get(index)?.focus();
+  };
+
+  const nextSlide = (event: MouseEvent | KeyboardEvent) => {
```
mark-tate, line 109: Next Slide and focus slide get combined here, we pass in an event, only to check it's type.
We then use that type to either give the card focus, so we can enable next/prev keyboard control or advance the card. 

I then went to useKeyNavigation and could see that we needed this code to support an internal hook. We pushed complexity into context which was created by a hook only used by CarouselNavigation.

Was not expecting this code in here, perhaps needs refactoring, so nextSlide, always advances the slide and focus is provided elsewhere?

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+export const CarouselContext = createContext<CarouselContextValue | null>(
+  "CarouselContext",
+  null,
+);
+
+export function useCarousel() {
+  const context = useContext(CarouselContext);
+  if (!context) {
+    throw new Error("useCarousel must be used within carousel provider");
```
mark-tate, line 36: should this be 
"useCarousel must be used within CarouselProvider" or similar

packages/lab/src/carousel/CarouselControls.tsx
```diff
@@ -0,0 +1,122 @@
...
+import carouselControlsCss from "./CarouselControls.css";
+
+const withBaseName = makePrefixer("saltCarouselControls");
+
+export interface CarouselControlsProps
+  extends Omit<HTMLAttributes<HTMLDivElement>, "title"> {
+  /**
+   * Callback when back button is clicked.
+   */
+  onMoveBack?: (event: SyntheticEvent<HTMLButtonElement>) => void;
```
mark-tate, line 22: we changed the language on this api, should it be onNext and onPrevious, or similar ?
this will align with the rest of the code

packages/lab/src/carousel/CarouselNavigationBar.tsx
```diff
@@ -0,0 +1,36 @@
...
+  title?: ReactNode;
+}
+
+export const CarouselNavigationBar = forwardRef<
+  HTMLDivElement,
+  CarouselNavigationBarProps
+>(function CarouselNavigationBar({ title, className, ...rest }, ref) {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-carousel-controls",
```
mark-tate, line 25: this re-uses the id from controls

packages/lab/src/carousel/CarouselNavigationBar.tsx
```diff
@@ -0,0 +1,36 @@
+import { makePrefixer } from "@salt-ds/core";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { type HTMLAttributes, type ReactNode, forwardRef } from "react";
+
+import { CarouselControls } from "./CarouselControls";
+import carouselControlsCss from "./CarouselNavigationBar.css";
```
mark-tate, line 7: should this be carouselNavigationCss ?

packages/lab/src/carousel/index.ts
```diff
@@ -1,2 +1,5 @@
 export * from "./Carousel";
 export * from "./CarouselSlide";
+export * from "./CarouselSlider";
+export * from "./CarouselNavigationBar";
+export * from "./CarouselControls";
```
mark-tate, line 5: Does this component need to exist or be exported?
CarouselNavigationBar and CarouselControls seem like the same component?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
+      if (slideRef.current) {
+        const assignedIndex = registerSlide(slideRef.current);
+        setIndex(assignedIndex);
+      }
+      return () => {
+        if (index !== null) {
+          unregisterSlide(index);
+        }
+      };
+    }, [registerSlide, unregisterSlide]);
```
mark-tate, line 78: this will not cleanup as index is not in the dependencies ?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
+
+    const isVisible =
```
mark-tate, line 80: feels like it could be owned by context helpers perhaps, to simplify this code and have a single source of truth?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
-  Media: ReactElement;
-  description?: string;
-  title?: string;
-  contentAlignment?: "center" | "left" | "right";
+  /**
+   * Actions to be displayed in the content footer.
+   */
+  actions?: ReactNode;
+  /**
+   * The media to be displayed inside the slide.
```
mark-tate, line 22: description could be improved, it's not clear without reading the code, what media is and how it differs from children

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
+   */
+  media?: ReactNode;
+  /**
+   * The appearance of the slide. Options are 'bordered', and 'transparent'.
+   * 'transparent' is the default value.
+   * */
+  appearance?: "bordered" | "transparent";
+  /**
+   * Header text
```
mark-tate, line 31: is it text, could be something else as it's a node ?

packages/lab/src/carousel/CarouselSlider.css
```diff
@@ -0,0 +1,9 @@
+.saltCarouselSlider {
+  display: flex;
+  scroll-snap-type: x mandatory;
+  overflow-x: auto;
+  scroll-behavior: smooth;
+  scrollbar-width: none;
```
mark-tate, line 6: support exceeds our min browser recommendations, what's the impact ?

packages/lab/src/utils/useIntersectionObserver.ts
```diff
@@ -0,0 +1,55 @@
...
+import { type RefObject, useEffect, useRef } from "react";
+
+export interface UseIntersectionObserverProps {
+  ref: RefObject<HTMLElement>;
+  onIntersect: (index: number) => void;
+  threshold?: number;
+  rootMargin?: string;
+}
+
+export function useIntersectionObserver({
```
mark-tate, line 11: should this live in Carousel, does it need to be in utils ?

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,99 @@
...
+  prevSlide,
+}: Pick<CarouselContextValue, "nextSlide" | "prevSlide">) => {
+  return useCallback(
+    (event: KeyboardEvent<HTMLDivElement>) => {
+      if (event.key === "ArrowRight" || event.key === "ArrowLeft") {
+        if (event.repeat) return;
+        event.stopPropagation();
+        if (event.key === "ArrowRight") {
+          nextSlide?.(event);
+        } else if (event.key === "ArrowLeft") {
```
mark-tate, line 36: should could just be an else

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+export const CarouselSlider = forwardRef<HTMLDivElement, CarouselSliderProps>(
+  function CarouselSlider(
+    { children, onKeyDown: onKeyDownProp, ...rest },
+    propRef,
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-carousel-slide",
+      css: carouselSliderCss,
```
mark-tate, line 53: carouselSliderCss -> carouselSlideCss

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,99 @@
...
+        } else if (event.key === "ArrowLeft") {
+          prevSlide?.(event);
+        }
+      }
+    },
+    [nextSlide, prevSlide],
+  );
+};
+
+export const CarouselSlider = forwardRef<HTMLDivElement, CarouselSliderProps>(
```
mark-tate, line 45: Calling it Slider made me think of the Slider component, is there a better name for this ?

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,99 @@
...
+  function CarouselSlider(
+    { children, onKeyDown: onKeyDownProp, ...rest },
+    propRef,
+  ) {
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-carousel-slide",
```
mark-tate, line 52: "salt-carousel-slide -> "salt-carousel-slider

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,99 @@
...
+      prevSlide,
+    });
+
+    // Handlers
+    const handleScroll = useCallback(() => {
+      if (containerRef.current) {
+        const scrollLeft = containerRef.current.scrollLeft;
+        updateActiveFromScroll(scrollLeft);
+      }
+    }, [containerRef, updateActiveFromScroll]);
```
mark-tate: We offer an uncontrolled component but what if we offered a controlled component then folks could do alot of things themselves.

For instance, they could create an autoPlay  and we could document how to create it ?

Most other Carousels provide an API that provide events for slides becoming visible ?

jake-costa (changes requested):

packages/lab/src/carousel/CarouselContext.tsx
mark-tate: Any need to have an infinite carousel, so when it gets to the end, it goes back to the beginning?

packages/lab/src/carousel/Carousel.tsx
```diff
@@ -1,170 +1,59 @@
...
-   * The animation when the slides are shown.
-   * Optional. Defaults to `slide`
-   **/
-  animation?: "slide" | "fade";
-  /**
-   * If this props is passed it will set the aria-label with value to the carousel container.
-   * Optional. Defaults to undefined
-   */
-  carouselDescription?: string;
+  activeSlideIndex?: number;
```
mark-tate: If I set activeSlideIndex, to 2 then when the page loads, you get scroll movement, as it scrolls to the activeSlideIndex.

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
 export interface CarouselSlideProps extends HTMLAttributes<HTMLDivElement> {
-  ButtonBar?: ElementType;
-  Media: ReactElement;
-  description?: string;
-  title?: string;
-  contentAlignment?: "center" | "left" | "right";
+  /**
+   * Actions to be displayed in the content footer.
+   */
+  actions?: ReactNode;
```
mark-tate, line 20: A Carousel's responsibility is to just navigate content, it does not define what the content is.

Actions on the API , this forces a single pattern on every slide.
Ideally you can just compose Salt Cards in the Carousel and then this is a Cards pattern ?

Maybe this extends to media?

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
+    };
+    const ref = useForkRef(refProp, slideRef);
     return (
-      <div ref={ref}>
-        {Media && <div className={withBaseName("mediaContainer")}>{Media}</div>}
-        <div className={withBaseName("fixedContainer")} ref={buttonBarRef}>
+      <div
+        role="group"
+        aria-roledescription="slide"
+        aria-labelledby={clsx(ariaLabelledBy)}
```
joshwooding, line 96:
```suggestion
        aria-labelledby={ariaLabelledBy}
```

packages/lab/src/carousel/CarouselSlide.tsx
Fercas123: if the content has navigable items and we dont ensure those are hidden then tab would move slides

packages/lab/src/carousel/CarouselSlide.tsx
```diff
@@ -1,76 +1,134 @@
...
+    } = useCarousel();
 
+    useEffect(() => {
+      if (slideRef.current) {
+        const assignedIndex = registerSlide(slideRef.current);
+        setIndex(assignedIndex);
+      }
+      return () => {
+        if (index !== null) {
+          unregisterSlide(index);
```
joshwooding, line 75: Index isn't a unique identifier, is this going to affect this? Should we just use an id from useId instead?

packages/lab/src/carousel/CarouselSlide.tsx
Fercas123: index would only change when a slide is registered

origami-z (commented):
Some naming / props questions

jake-costa (commented):
Just need to ensure previous/next buttons receive focus first even if placed below the carousel. Then we should be all set to move to labs and start getting feedback.

jake-costa:
@Fercas123 Just adding some thoughts (not violations), maybe fore future release:

- Do we want to add a Peek behavior so a small part of the previous/next slides are visible?
- Do we want to add a tabbed carousel option as well (with dots)?
- Do we want to add an option to allow the text content to be displayed on the right/left side of the image instead of under only? (but remain dynamically responsive based on the viewport size)
- Do we need to specify aspect ratio for images or is that handled automatically?
- Are we doing to look into drag/touch slide control as well to better support use on mobile?
- Can we change the heading levels default to h2 instead of h3's? and allow the devs to change the heading level based on their use case? (I think this would be really helpful)

ivan-calderon:
Very good feedback @jake-costa. Few quick answers here:

**- "Do we want to add a Peek behavior so a small part of the previous/next slides are visible?"** We chatted with Fer about this when scoping. This is highly recommended specially for mobile devices. Castillo, Fernanda (CIB Tech, GBR) please correct me if I am wrong but the parent container width can be configure to allow this behavior, correct? If so, should we add an example of it in the docs with a best practice section?
**- "Do we want to add a tabbed carousel option as well (with dots)?"** This was not in scope for this release because of a few reasons: 1) Labels work more efficiently than dots considering that dots require users to remember the slide content location while labels provide a clearer indication of the content. 2) We had doubts about the target size of the dots and we will need further exploration to put this in place with the right recommendations. 3) We recommend not to add more than 5 slides. If this is respected we can still be in a safe place and add a few extra clicks wihtout the dots. But this surely needs to be considered in the future + plus a I convinced we will receive feedback about adding the dots. 
**- "Do we want to add an option to allow the text content to be displayed on the right/left side of the image instead of under only? (but remain dynamically responsive based on the viewport size)"** We should. This is also something we discussed and can be a great addition to this component.
**- Do we need to specify aspect ratio for images or is that handled automatically?** Very good question. This one is very tricky. With Fer we chatted that the parent container width can be configurable to be more flexible when showing the number of slides based on the width of the parent container. For instance, if we want to use a responsive full-width hero carousel with a single slide displayed, the spec ratio might need a different treatment than using the same carousel displaying 2 slides at a time in a widget. Stakeholders feedback might help on this considering that they are building the experiences.
- "Are we going to look into drag/touch slide control as well to better support use on mobile?" @Fercas123 , please correct me if am wrong but I remember this has been implemented already for mobile.
**- "Do we have a limit to the amount of text a slide can contain?"** Very good feedback. We should. I was not able to find studies recommending a specific about of text. But we might recommend that the text be short and concise and not to surpass the 2 sentences. I think that is the ideal scenario.
**- "Can we change the heading levels default to h2 instead of h3's? and allow the devs to change the heading level based on their use case? (I think this would be really helpful)"** Big like here + very good feedback. We were using H3 because for the Carousel title (optional) we use H2. However, our initial thought, with Fer, was to use H2 for the slides heading and move the carousel title to H1. Do you think this is a change we might consider now Costa, Jacob (CIB, USA)?

packages/lab/src/__tests__/__e2e__/carousel/Carousel.cy.tsx
```diff
@@ -0,0 +1,120 @@
...
+        "true",
+      );
+      cy.findAllByRole("button", { name: "Next slide" }).click();
+      cy.findAllByText("2 of 4").should("exist");
+    });
+
+    it("SHOULD move to the previous slide with button", () => {
+      cy.mount(<Default />);
+      // move one to the left
+      cy.findAllByRole("button", { name: "Next slide" }).click();
```
mark-tate, line 31: note.. these buttons could do nothing and you would never know the test was failing.
See lines 17 and 18... it tests the intial state and on line 34 the same state,
We don't check for any in-between state.

packages/lab/src/__tests__/__e2e__/carousel/Carousel.cy.tsx
```diff
@@ -0,0 +1,120 @@
...
+      cy.findAllByRole("group").get('[tabindex="0"]').focus();
+      cy.findAllByRole("group").get('[tabindex="0"]').realPress("ArrowRight");
+      cy.findAllByText("2 of 4").should("exist");
+      cy.wait(100);
+      cy.findAllByRole("group").get('[tabindex="0"]').realPress("ArrowLeft");
+      cy.findAllByText("1 of 4").should("exist");
+    });
+    describe("WHEN navigating with keyboard keys", () => {
+      it("SHOULD NOT move slides when tabbing out of actions within", () => {
+        cy.mount(<WithActions />);
```
mark-tate, line 78: initial state and end state match but no checks in-between that the state changes, so could provide a false positive?

packages/lab/src/carousel/CarouselContext.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [visibleSlides]);
+
+  useEffect(() => {
+    dispatch({
+      type: "moveToIndex",
+      payload: activeSlideIndex,
+    });
+  }, [activeSlideIndex]);
+
+  useEffect(() => {
```
mark-tate, line 67: is this code needed ?
if activeSlideIndexProp prop changes then `activeSlideIndex` will change and that would be picked up by the other useEffect ?

packages/lab/src/carousel/CarouselReducer.ts
```diff
@@ -0,0 +1,94 @@
...
+  activeSlideIndex: number;
+  visibleSlides: number;
+  focusedSlideIndex: number;
+  carouselId?: string;
+}
+export type CarouselReducerAction =
+  | { type: "register"; payload: [CarouselSlideId, CarouselSlideMeta] }
+  | { type: "unregister"; payload: CarouselSlideId }
+  | { type: "updateSlideCount"; payload: number }
+  | { type: "move"; payload: CarouselSlideId }
```
mark-tate, line 16: One for the radar, might need to update Stepper reducer which may still have goto. APIs should re-use naming and be consistent.

packages/lab/src/carousel/CarouselReducer.ts
```diff
@@ -0,0 +1,94 @@
...
+        ...state,
+        activeSlideIndex: index,
+      };
+    }
+    case "moveToIndex": {
+      const { slides } = state;
+
+      const index = action.payload;
+
+      if (index === -1 || index > slides.size) return state;
```
mark-tate, line 63: NIT do we enforce/prefer braces on single line conditions to create better readability ?

.changeset/purple-beers-laugh.md
```diff
@@ -0,0 +1,47 @@
...
+    {items.map((slide) => (
+        <CarouselSlide
+          key={slide.title}
+          header={<H3>{slide.title}</H3>}
+          actions={<Link href="#">{slide.link}</Link>}
+        >
+          <Text>{slide.content}</Text>
+          </CarouselSlide>
+      ))}
+</CarouselSlider>
```
mark-tate, line 45: formatting broken here

packages/lab/src/carousel/CarouselSlider.tsx
```diff
@@ -0,0 +1,136 @@
...
+      const container = containerRef.current;
+      if (!container) return;
+
+      const scrollBehavior = hasRun.current ? "smooth" : "instant";
+      const slideWidth = container.offsetWidth / visibleSlides;
+
+      requestAnimationFrame(() => {
+        container.scrollTo({
+          left: focusedSlideIndex * slideWidth,
+          // @ts-ignore
```
mark-tate, line 114: needs an explanation why this line is ignored

packages/lab/stories/carousel/carousel.stories.tsx
```diff
@@ -1,41 +1,196 @@
...
+  CarouselSlider,
+} from "@salt-ds/lab";
 import "./carousel.stories.css";
-import { Button, FlexLayout } from "@salt-ds/core";
+import { Button, H2, H3, Link, StackLayout, Text } from "@salt-ds/core";
+import { useState } from "react";
+import carouselSlide1 from "../assets/carouselSlide1.png";
+import carouselSlide3 from "../assets/carouselSlide3.png";
+import carouselSlide4 from "../assets/carouselSlide4.png";
+import carouselSlide5 from "../assets/carouselSlide5.png";
```
mark-tate, line 15: @Fercas123 this file is missing so storybook cannot build, can you push and then we should be able to get this merged to labs,

ivan-calderon (approved):

jake-costa (approved):

mark-tate (approved):

PR #4650, opened 2025-02-03
Splitter

chrispcode:


packages/lab/src/splitter/SplitHandle.tsx
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Change which sides get a border displayed
+   *
+   * Default is based on the orientation and appearance
+   * set on the parent Stepper components, ex.
+   * bordered + horizontal = left-right
+   * bordered + vertical = top-bottom
+   * transparent = none
+   */
+  accent?: SplitHandleAccent;
```
joshwooding, line 47: Is accent the right terminology? cc @ivan-calderon

jake-costa:
@chrispcode Wanted to let you know that it looks like each separator lacks an accessible name (either via aria-label or aria-labelledby) and aria-controls is pointing to a data-panel-id attribute rather than a genuine id.

packages/lab/src/splitter/Splitter.tsx
```diff
@@ -0,0 +1,44 @@
...
+  appearance: appearanceProp,
+  className,
+  ...props
+}: SplitterProps) {
+  const appearanceContext = useContext(AppearanceContext);
+  const appearance = appearanceProp ?? appearanceContext;
+
+  return (
+    <OrientationContext.Provider value={orientation}>
+      <AppearanceContext.Provider value={appearance}>
```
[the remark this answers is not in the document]
chrispcode: I consider separating contexts as much as possible is a good practice. Passing objects in context without proper memoization can introduce unnecessary re-renders.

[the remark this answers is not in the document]
joshwooding: Separating contexts that either update frequently (to minimise re-renders) makes sense but these are read in the same place and provide no real benefit. Also you could just memoise in this case with two values...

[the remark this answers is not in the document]
chrispcode: They are used in the same places now, but in the future that might not be the case. The two values are not logically bound, so makes little sense to me why should be combined into a singular context.

site/docs/components/splitter/examples.mdx
```diff
@@ -0,0 +1,105 @@
...
+  sidebar:
+    exclude: true
+  data:
+    $ref: ./#/data
+---
+
+<LivePreviewControls>
+
+The Salt Splitter leverages the popular open-source library [react-resizable-panels](https://react-resizable-panels.vercel.app/), whilst adding styling, introducing extra functionality and making adjustments to component and prop names to better align with the rest of the components present in our design system.
+
```
Fercas123, line 14:
```suggestion

```

Fercas123: I would say we get rid of this line, as we tend to go straight into examples in all other docs so it feels a bit on it's own there

packages/lab/src/splitter/Splitter.tsx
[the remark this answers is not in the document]
joshwooding: I don't think it really matters but this imo is premature optimisation

.changeset/seven-penguins-fly.md
```diff
@@ -0,0 +1,33 @@
...
+
+`SplitPanelProps["variant"]` - Change the background color of the SplitPanel component. The variant prop takes one of three options - `primary` (default), `secondary`, `tertiary`. Follows the same behavior as our Panel component.
+
+`SplitHandle["variant"]` - Change the background color of the SplitHandle component. It behaves similarly to the `SplitPanelProps["variant"]` prop and it is meant to be used in combination with it. The variant prop can take one of three options - `primary` (default), `secondary`, `tertiary`. Follows the same behavior as our Panel component.
+
+`SplitHandle["border"]` - Override the border placement of the SplitHandle component. The border prop takes one of seven options - `top`, `bottom`, `left`, `right`, `top-bottom`, `left-right`, `none`. The default value is computed based on the orientation and appearance of the parent Splitter component. If the parent splitter's orientation is `horizontal` and the appearance is `bordered`, the default value will be `top-bottom`. If the orientation is `vertical` and the appearance is `bordered`, the default value will be `left-right`. If the appearance is `transparent`, the default value will be `none`.
```
joshwooding, line 33: I don't think this needs to be in the changelog, realistically to consumers all they want to know is that Splitter is added

packages/lab/src/__tests__/__e2e__/splitter/Splitter.cy.tsx
```diff
@@ -0,0 +1,63 @@
+// import { Splitter, SplitPanel, SplitHandle } from "@salt-ds/lab";
```
joshwooding, line 1:
```suggestion
```

packages/lab/src/splitter/SplitHandle.css
joshwooding, line 1: The fallbacks need to be cleared up, and unnecessary ones should be removed. `--salt-content-primary-foreground` is a valid value and shouldn't need a fallback

packages/lab/src/splitter/SplitHandle.tsx
```diff
@@ -0,0 +1,93 @@
...
+  type PanelResizeHandleProps,
+} from "react-resizable-panels";
+
+import splitHandleCSS from "./SplitHandle.css";
+import { AppearanceContext, OrientationContext } from "./Splitter";
+import { computeAccent, computeVariant } from "./utils";
+
+const withBaseName = makePrefixer("saltSplitHandle");
+
+export type SplitHandleVariant =
```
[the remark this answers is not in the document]
chrispcode: Yes, if somebody wants to use the type for some reason they would be able to.

site/docs/components/splitter/accessibility.mdx
```diff
@@ -0,0 +1,58 @@
...
+data:
+  $ref: ./#/data
+---
+
+## Best Practices
+
+- Set an Appropriate `aria-label` or `aria-labelledby` for `SplitHandle`:  
+   Make sure to assign a meaningful and descriptive `aria-label` to the `SplitHandle`. This will help screen readers and other assistive technologies accurately convey the purpose of the handle to users.
+
+- Use `id` with `SplitPanel` for `aria-controls`:
```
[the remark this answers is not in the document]
chrispcode: Yes, if not passed each handle points to the PanelGroup (Splitter) component instead

packages/lab/src/splitter/SplitHandle.tsx
chrispcode: Changed to `border` as agreed

.changeset/seven-penguins-fly.md
chrispcode: Moved this files contents in a dedicated docs page in storybook

packages/lab/src/splitter/SplitHandle.css
chrispcode: Done!

site/docs/components/splitter/examples.mdx
chrispcode: Done!

packages/lab/src/splitter/SplitHandle.tsx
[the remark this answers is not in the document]
joshwooding: They can use it via the normal method SplitHandleProps["variant"] adding extra exports seems redundant.

packages/lab/src/splitter/SplitHandle.css
joshwooding: > Done!

I'm still seeing fallback values in the CSS file? `--splitHandle-background: var(--salt-container-primary-background, rgba(255, 255, 255, 1));`

chrispcode: Now should be good!

site/docs/components/splitter/accessibility.mdx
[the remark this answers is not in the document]
chrispcode: Added a solution that uses `useId()` and automatically assigns an id to every panel, thus always referencing a valid `aria-controls`

packages/theme/css/characteristics/separable.css
```diff
@@ -4,4 +4,11 @@
   --salt-separable-primary-borderColor: var(--salt-palette-alpha-strong);
   --salt-separable-secondary-borderColor: var(--salt-palette-alpha);
   --salt-separable-tertiary-borderColor: var(--salt-palette-alpha-weak);
+
+  --salt-separable-foreground: var(--salt-palette-neutral-primary-foreground);
+  --salt-separable-foreground-hover: var(--salt-palette-neutral-primary-foreground);
+  --salt-separable-foreground-active: var(--salt-palette-interact-cta-foreground);
+  --salt-separable-background: var(--salt-palette-alpha-none);
+  --salt-separable-background-hover: var(--salt-palette-alpha-weak);
```
Fercas123, line 12:
```suggestion
  --salt-separable-background-hover: var(--salt-palette-alpha-none);
```

Fercas123: looking at figma i would have thought it was this one, but not super sure so feel free to ignore me

Fercas123 (approved):

packages/theme/css/characteristics/separable.css
joshwooding: @Fercas123 There's a comment in Figma to figure out which is the right value

.changeset/seven-penguins-fly.md
```diff
@@ -0,0 +1,30 @@
+---
+"@salt-ds/lab": minor
+"@salt-ds/theme": minor
```
joshwooding, line 3: Minor but can we split the theme and labs changelog please, that way each change shows against the specific library.

packages/lab/src/splitter/SplitHandle.css
```diff
@@ -0,0 +1,198 @@
...
+
+.saltSplitHandle[data-panel-group-direction="vertical"] {
+  flex-direction: row;
+  height: 4px;
+}
+
+.saltSplitHandle[data-panel-group-direction="horizontal"]::after {
+  content: "";
+  position: absolute;
+  left: -10px;
```
joshwooding, line 67: Is this related to size base? if so can we just use a calcuation

packages/lab/src/splitter/SplitHandle.css
```diff
@@ -0,0 +1,198 @@
...
+  background: var(--splitHandle-background-hover);
+}
+
+.saltSplitHandle[data-resize-handle-state="drag"] {
+  --splitHandle-background: var(--splitHandle-background-active);
+}
+
+.saltSplitHandle[data-panel-group-direction="horizontal"] {
+  flex-direction: column;
+  width: 4px;
```
joshwooding, line 56: Should this use the split handle size token?

packages/lab/src/splitter/SplitHandle.css
chrispcode: Not related, the splitter stays 24px no matter the density

packages/lab/src/splitter/SplitHandle.css
```diff
@@ -0,0 +1,198 @@
...
+  outline-style: var(--salt-focused-outlineStyle);
+  outline-width: var(--salt-focused-outlineWidth);
+  outline-color: var(--salt-focused-outlineColor);
+  outline-offset: var(--salt-focused-outlineOffset);
+}
+
+.saltSplitHandle-border-top {
+  border-top-width: var(--splitHandle-borderWidth);
+  border-top-color: var(--splitHandle-borderColor);
+  border-top-style: solid;
```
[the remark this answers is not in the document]
chrispcode: The token exists in the legacy theme, but not in next.

.changeset/seven-penguins-fly.md
chrispcode: Done!

packages/lab/src/splitter/SplitHandle.css
chrispcode: Done!

packages/lab/src/splitter/SplitHandle.css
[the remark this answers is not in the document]
joshwooding: @chrispcode the way the theme works means the token will be available in next too

packages/lab/src/splitter/SplitHandle.css
ivan-calderon: @joshwooding we don't have a token currently for it. 24px is a fixed value based on the A11y recommendation about target size, independently of density.

ivan-calderon (approved):

packages/theme/css/characteristics/separable.css
chrispcode: Just talked with @ivan-calderon, and we concluded that the correct value is `alpha-weak`

packages/lab/src/splitter/SplitHandle.css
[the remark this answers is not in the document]
chrispcode: Done!

joshwooding (approved):

PR #4653, opened 2025-02-04
initial POC to demo style contracts

mark-tate:
- style contracts enable you to create a schema of tokens that can be overriden
- default props provider enables you to specify default props based on component name
- ComponentBase is an abstract base component that is used to add common props to base elements (size/spacing)

All ideas, naming and implementation, represent a POC


origami-z:
First quick thoughts 


- I found it quite tricky to comment on the POC, given this feels to solve multiple different problems at once. 
- It's good to see cut down of component level variables. We need to make sure our own use cases are all covered, e.g., i remember some use case on icon size (cross Salt component override vs public consumer api). And not too sure whether the change is actually fully backwards compatible...
- I can see the example contract is first split into system vs component levels.
    - system are `@salt-ds/theme` tokens? We need to evaluate the maintenance cost for both Salt and BLM teams, especially comparing runtime methods vs pure CSS solutions.
    - component contains a mixture of Salt component class name, token name as well as custom class names. First two exponentially increases Salt's control area, so would negatively impact maintenance i guess. Not too sure how custom class names (from BLM) would impact Salt code base
- I only see default props provider useful in very limited use cases.
    - how many default props decision can actually be "universal" decision? Given those are component specific decisions, we need to consider benefit for both first time build and well as reusability. IMO, there is greater risk when reusing components across different system than the benefit of time saving at first build.
- i don't think `ComponentBase` would be feasible in real life, given each of our components is designed differently, which can't be easily modeled using the same set of size/spacing. If this can be done, then modifying theme tokens would be enough to achieve similar effect i'd assume.
- `MockTicket` mixes local style override on top of contract, so hard to tell the purpose of it

mark-tate:
**_What parts did this POC explore ?_**

The POC explored 3 parts

- What does adding a Base component like Chakra (Box) do for us?
This does not seem that useful at present but applying classname or data-attributes like small/medium/large might be revisited when we get to sizing concerns of some components
- Added a DefaultPropsProvider
Users don't want to have to constantly set props, when all they want is bordered controls. The POC demonstrates the ability but not entirely sure if this is the solution, it does mean components do not need to be wrapped if this was required.
- Contracts, ability to apply extended styling to Salt components, outside of the system.
This was the most interesting part of this POC as it provides an escape hatch that mean products don't need to be blocked by Salt and can manage additional styles and a layer of business semantics, without wrapping components.

**_What advantages do Contracts have ?_**

- reduces complexity as only the contract owner needs to understand Salt and everyone else can use components out the box
- no wrapping of components required, this is one thing stakeholders have asked for. It does not mean wrapped components do not exist but means there maybe less of them.
- typed values, dependent on the complexity of the contract, means you don't need to read the docs or understand the system to configure a contract, someone else has decided what the choices are.
- contracts turn unknown CSS overrides into known overrides, previously overrides may of been owned by components, with contracts they are federated from a single source of truth that is easy to update/maintain
- apply a contract to all or part of your app, targeting either component or system overrides
- no need to use inline styles and depend on untyped tokens
- no build tools required, some internal teams are often challenged by tooling based solutions as there is no standard or common version
- can be used outside of Salt with your own contracts, selectors or tokens, meaning one solution  which can be used for Salt and more business level semantics
- more of a known quantity compared to styling options which overload the SaltProvider with too much responsibility and won't scale, how many props can we add to SaltProvider ? Also, those props can be mixed and matched to create unknown quantities
- you can turn them on/off, they are more specific than the out-the-box default styles, so you can always see what the recommended standard should be, without your overrides applied.
- Salt's "out the box" systemic theming approach provides the recommended defaults and a golden path, but the delivery may move slower than required. The system is a holistic solution to theming whilst a contract provides support of corner cases that do not fit into the system. Ideally, in time, everything ends up being systemic (aligned to a known standard and Salt is universally, used out of the box) but that will take time, so contracts enable the migration journey to start.
- Values in the contract can be removed and fallback to the system, so aligning to the recommended standard is as simple a s removing your overrides. Compare this to editing CSS in a file, where tokens have been spread across multiple components.
- By offering a solution where components do not need to be wrapped, there is a higher chance of cross LOB use.
- Change is easier to implement, rather than having to update many individual/wrapped components, you can update one central file.
- Rather than relying on visual regression to tell you something is broken, you have types as well.

**_Why not use static CSS instead?_**

- over time it can lead to a monolithic solution where the lines between Salt overrides and product concerns become muddled. With contracts Salt could provide a path for teams to re-align back to the system/out-the-box solution. 
- adding more CSS files to override corner cases was something that teams have said they do not want, as they see it as blackbox and something they don't have an understanding of
- with a contract we have defined rules and types that help you configure contracts, the next step could (?) be to generate static CSS from them, so that you can opt for either runtime or static solution.

**_How risky is the code ?_**

Well this is a POC, in this form, we won't be merging anything, the branch is for demos.
That said...
Contracts implementation is simple and uses pre-existing Core APIs, emulating somewhat, the approach of inline/additional styles, that can set Salt tokens. Except the tokens are not exposed to everyone and only need to be defined once, in the contract. I would not see contracts becoming part of the Core codebase, more an opt-into solution provided by it's own package.

**_Does this not mean that anyone can override the defaults and do what they want?_**

All our solutions use different levels of CSS, if someone wants to ignore standards and override something they will. 
Something like contracts makes what has been overridden clearer /visible and easier to remove.

**_Does this mean that Salt has to support contracts created by other products, they could be fragile and subject to breaking changes ?_**

Yes, they could, if that concerns you, then using Salt out the box is the best option.
When you elect to deviate from a known standard, there is a responsibility to own those changes and maintain them yourselves. 

As part of the POC, we cleaned up some CSS styles, which expose a finite set of CSS variables for each component that map to CSS attributes. Use these tokens and they will be much more stable as CSS attributes don't change. Obviously if those variables could use values which are system tokens, these could change, but that is manageable as tokens are documented. Selectors are probably the most fragile part but when writing a contract you can weigh that up in your decision. If deemed to fragile then use Salt out the box until such time as the system supports it.

**_What are the typical use-cases?_**

- short-lived tactical overrides to enable a migration journey to the brand/out-the-box solution
- corner cases that will never be part of the system but want to still use Salt, without wrapping
- business semantics can be added on top of Salt without re-wrapping or maintaining additional CSS
- could be a more scalable replacement for the current styling options on SaltProvider

**_How big should contracts be ?_**
- Small and applied to a whole app preferably, if you were creating a contract per view then something must be wrong with your design. This is not a replacement for CSS or CSS in JS solution, it's specifically intended to configure pre-existing styles via tokens/CSS variables.


**_What is the MockTicket?_**

MockTicket is representative of a ticket built in vanilla, out the box, Salt. 
You apply the contract and toggle between overrides which apply the overrides. 
Equally you can remove the overrides and see the out the box state. 
Whilst you can use Salt out the box to build a ticket, there is always "glue" CSS that manages some additional layout.
Salt layouts only get you so far.

What's next for this POC?
- At present getting feedback and adding in more examples till we understand the value/approach better
- If contracts solve one problem but  creates another, is that problem bigger or smaller?
- We have decided from this POC to clean up our CSS files so that if we go ahead with this approach there are a stable set of component tokens we can wire contracts to
- We may publish the contract provider package as a snapshot, so folks can try things for themselves.

If you are a Salt user and have thoughts, please add them.

packages/style-contract/stories/RaichuContract.ts
```diff
@@ -0,0 +1,47 @@
+import type { ResponsiveProp } from "@salt-ds/core";
+
+type SystemProperties = {
+  "salt-text-action-textTransform"?: "capitalize";
+};
+
+type SaltAlphaCardProperties = {
```
mark-tate, line 7: This is the Salt card contract

origami-z:
> * no wrapping of components required, this is one thing stakeholders have asked for.

What feature / requirement does "no wrapping" include? With different levels of potential customization (see below), this feels too wild a claim. We need to help stakeholders categorize their customization requirement, before taking "no wrapping" as the most important requirement to fulfill. 

> Contracts, ability to apply extended styling to Salt components, outside of the system.

> * apply a contract to all or part of your app, targeting either component or system overrides

I think a breakdown of supported POC contract features is needed, and discussed individually.
- Split system vs component from top level of contract: are we promoting component level change to the same level of system token change? Or should only universal changes (both system and component) be promoted, i.e., no situational change?
- Component level contract consists multiple features (e.g., universal component property override, create custom component by extending Salt component, situational component property override). 

## Salt usages

Can use below as different levels of potential Salt usages (at this moment of time) 

1. pick a theme, e.g., Legacy vs Brand
1. pick styling option, e.g., Light vs dark modes, accent color, corner radius
1. _universal_ theme token override, e.g., change action text transform to 'none'
1. pattern component created with specific configuration, e.g., app header
1. _universal_ component customization
    1. change default prop value, e.g., input `bordered`
    1. change certain styling, e.g., color/opacity of status dialog
1. extend component
    1. additional variant, e.g. Button's buy/sell variant, dimensional button
1. _situational_ component override
    1. override with extended theme token set, e.g. update card corner to `--salt-palette-corner-stronger`
1. one time override
    1. override certain CSS property to fit the layout, e.g. change card padding to 0 for full width image example

Is it possible for us to settle on above list, then let stakeholders comment on expectations?

origami-z:
> 6. extend component
>     - additional variant, e.g. Button's buy/sell variant, dimensional button

Brainstormed with Pedro on variant extension use case, using example of adding "non-default" to Input's validation status, mainly focus on the "least" amount of maintenance from BLM tech team perspective.

Given most of variant extension is styling change, as long as component prop can accept "more" status options, a new class name will automatically make it work, without the need of _wrapping_ component. Effectively, it means in React, `<Input validationStatus="nonDefault" />` would give `.saltInput-nonDefault`, with the exception of TypeScript error. TS can potentially be managed via generics.

`.saltInput-nonDefault` may raise some concern about confusion of mixture of owner of Salt vs other teams. This maybe mitigated by either additional prefix, e.g. `.saltInput-mtkNonDefault`, or potentially make `makePrefixer` offer more flexibility at run-time and/or build-time. 


Another discussion around component CSS API, overlap with #190 . With the current set up, for any new status extension needed, BLM team needs to write corresponding CSS for all components, e.g., 

```
.saltInput-nonDefault {
      --input-background: var(--extended-editable-nonDefault-background);
      --input-background-active: var(--extended-editable-nonDefault-background-active);
...
}

.saltDropdown-nonDefault {
      --dropdown-background: var(--extended-editable-nonDefault-background);
      --dropdown-background-active: var(--extended-editable-nonDefault-background-active);
...
}
``` 

Alternative implementation may save the need to understand individual component availability

```
/* shared .css */
.saltEditable-secondary {
  --salt-editable-background:  var(--salt-palette-background-secondary);
  ....
}
```

Then add `.saltEditable` to components
```
`.saltInput {
    background: var(--salt-editable-background);
}

.saltInput.saltEditable-secondary {
    /** automatic */
}
```

With above, extending variant could mean as simple as authorizing a new `.saltEditable-nonDefault`, without the need to author component specific CSS files.

joshwooding:
Closing this for now, we can use it as a reference for future PRs

PR #4669, opened 2025-02-10
Improve responsive density story

### LINKED ISSUE #3303: Better way to support density switch on breakpoints
### Area

UI Components

### The problem

Some site needs to switch density depending on breakpoints, Salt website does this as well. There isn't a straightforward way to implement this at the moment, where `density` needs to be provided as a prop to `SaltProvider` but breakpoint can only be accessed inside the provider (e.g., via nested component). It creates unnecessary nesting to implement this feature. 

### The solution

Add callback to `SaltProvider` / its `useDensity` hook to update density, similar to POC for mode in #3237.

### Alternatives and examples

Not so ideal implementation with nested component just for a callback: https://stackblitz.com/edit/salt-template-56hpdn?file=App.tsx

joshwooding:
Closes #3303

### DIFF
--- a/.changeset/many-crabs-end.md
+++ b/.changeset/many-crabs-end.md
@@ -1,0 +1,11 @@
+---
+"@salt-ds/core": minor
+---
+
+Make `density` a responsive prop.
+
+```tsx
+<SaltProvider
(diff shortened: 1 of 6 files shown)


joshwooding:
Closing for now as we're still unsure about the guidance with density and breakpoints

PR #4698, opened 2025-02-11
refactor button styles implementation to more maintainable approach w…

mark-tate:
…ith simpler naming

### DIFF
--- a/.changeset/tough-hats-draw.md
+++ b/.changeset/tough-hats-draw.md
@@ -1,0 +1,54 @@
+---
+"@salt-ds/core": patch
+---
+
+Refactor internal implementation of button styles
+
+- this is a non-breaking refactor of how we organise the styles within the CSS file.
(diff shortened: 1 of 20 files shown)


origami-z:
before i forget, one of the reasons this was needed, was to enable cross component override (e.g. Icon). More testing is needed to confirm whether 1 type of variable is enough, and/or alternative approach

.changeset/tough-hats-draw.md
```diff
@@ -0,0 +1,54 @@
+---
+"@salt-ds/core": patch
+---
+
+Refactor internal implementation of button styles
+
+- this is a non-breaking refactor of how we organise the styles within the CSS file.
```
libertymayc, line 7: checking, is this definitely considered non breaking/..only a patch? I personally have used a lot of --saltComponent tokens myself which would have to be changed by this

PR #4720, opened 2025-02-17
Stepped Tracker - Reset

chrispcode:
- add reset action
- fix immutability issue


.changeset/happy-chairs-ring.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/lab": minor
+---
+
+When using the `stepped-tracker` with the `useStepReducer` hook, there was an edge case where the state indicated that all steps were completed, but the `ended` boolean remained `false`. This issue stemmed from an immutability problem in the state handling logic.This bug is resolved now, whilst ensuring proper immutability and fixing the `ended` boolean behavior. Additionally, the `reset` action has been introduced, which, as the name suggests, resets the state so that the first step becomes active and all others are marked as pending.
```
origami-z, line 5: Changset will become changelog / release note for our consumers, thus our standard starting with "Fixes", "Added", etc. 

Half of this paragraph fits more nicely within PR description and/or site documentation...?

origami-z (commented):
Fix with test is nice 👍

tayyabataimur (dismissed):
🚀

packages/lab/src/stepped-tracker/stepReducer.types.ts
```diff
@@ -16,7 +16,8 @@ export type StepReducerAction =
   | { type: "previous" }
   | { type: "error" }
   | { type: "warning" }
-  | { type: "clear" };
+  | { type: "clear" }
+  | { type: "reset" };
```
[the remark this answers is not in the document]
chrispcode: Fair enough, I also saw some developers dispatching clear and expecting the reset functionality. Any ideas on what should we rename clear to?

How about?

```tsx
export type StepReducerAction =
  | { type: "next" }
  | { type: "previous" }
  | { type: "reset" }
  | { type: "status/error" }
  | { type: "status/warning" }
  | { type: "status/clear" };
```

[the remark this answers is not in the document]
chrispcode: Done!

.changeset/happy-chairs-ring.md
chrispcode: Done!

packages/lab/src/stepped-tracker/stepReducer.ts
```diff
@@ -4,121 +4,141 @@ import { assignSteps, autoStageSteps, flattenSteps, resetSteps } from "./utils";
...
-  }
 
-  if (action.type === "warning") {
-    if (state.activeStep) {
+    case "status/warning": {
+      if (!state.activeStep) {
+        return state;
+      }
+
       state.activeStep.status = "warning";
```
origami-z, line 103: Is this assignment actually mutating `state`?

origami-z (approved):

packages/lab/src/stepped-tracker/stepReducer.ts
origami-z: if someone is depending on `state.activeStep`, the same reference will be held?

chrispcode: In theory the component that depends on activeStep will have a parent that depends on state itself, and whenever this action is dispatched, react will go down and update the parent and the child components.

mark-tate: in theory.. but this is a reducer and a golder rule of reducers is to avoid mutating state

Threw this into LLM to see what it thought

```
import type { StepReducerAction, StepReducerState } from "./stepReducer.types";
import { assignSteps, autoStageSteps, flattenSteps, resetSteps } from "./utils";

export default function stepReducer(
  state: StepReducerState,
  action: StepReducerAction,
): StepReducerState {
  const { activeStep, activeStepIndex, flatSteps, steps } = state;

  const updateSteps = (newSteps: typeof steps, newIndex: number) => {
    const newFlatSteps = flattenSteps(newSteps);
    return {
      steps: newSteps,
      flatSteps: newFlatSteps,
      activeStepIndex: newIndex,
      activeStep: newFlatSteps[newIndex] ?? null,
      previousStep: newFlatSteps[newIndex - 1] ?? null,
      nextStep: newFlatSteps[newIndex + 1] ?? null,
      started: newIndex >= 0,
      ended: newIndex >= newFlatSteps.length,
    };
  };

  switch (action.type) {
    case "next": {
      if (activeStep?.status === "error") return state;

      const newIndex = activeStepIndex + 1;
      if (newIndex > flatSteps.length - 1) {
        return updateSteps(assignSteps(steps, "completed"), newIndex);
      }

      const newSteps = autoStageSteps(resetSteps(steps), { activeStepId: flatSteps[newIndex].id });
      return updateSteps(newSteps, newIndex);
    }

    case "previous": {
      if (activeStep?.status === "error") return state;

      const newIndex = activeStepIndex - 1;
      if (newIndex < 0) {
        return updateSteps(assignSteps(steps, "pending"), newIndex);
      }

      const newSteps = autoStageSteps(resetSteps(steps), { activeStepId: flatSteps[newIndex].id });
      return updateSteps(newSteps, newIndex);
    }

    case "status/error":
    case "status/warning":
    case "status/clear": {
      if (!activeStep) return state;

      const newStatus = action.type === "status/error" ? "error" :
                        action.type === "status/warning" ? "warning" : undefined;

      const updatedActiveStep = { ...activeStep, status: newStatus };
      const updatedFlatSteps = flatSteps.map((step, index) =>
        index === activeStepIndex ? updatedActiveStep : step
      );

      return { ...state, flatSteps: updatedFlatSteps, activeStep: updatedActiveStep };
    }

    case "reset": {
      const firstStepId = flatSteps[0].id;
      const newSteps = autoStageSteps(resetSteps(steps, { resetStatus: true }), { activeStepId: firstStepId });
      return updateSteps(newSteps, 0);
    }

    default: {
      const exhaustiveCheck: never = action;
      throw new Error(`Unhandled action: ${exhaustiveCheck}`);
    }
  }
}
```

half the number of lines, no contentious mutation.

Needs sanity checking but might help us align to a purer reducer?

packages/lab/src/stepped-tracker/utils.ts
```diff
@@ -6,17 +6,39 @@ export function assignSteps(
...
+      ...step,
+      stage,
+      substeps: assignSteps(step.substeps, stage),
+    };
   });
 }
 
-export function resetSteps(steps: StepRecord[]): StepRecord[] {
-  return assignSteps(steps, undefined);
+export function resetSteps(
```
mark-tate, line 21: Would this work ?

```
export function resetSteps(
  steps: StepRecord[],
  { resetStatus = false } = {}
): StepRecord[] {
  return steps.map((step) => ({
    ...step,
    stage: undefined,
    status: resetStatus ? undefined : step.status,
    substeps: step.substeps ? resetSteps(step.substeps, { resetStatus }) : undefined,
  }));
}
```

mark-tate (commented):
Side comments:

what is the difference between the Basic and Horizonal doc examples, do we need both ?
what would happen if I have overly long status text, no design guidance on limits of the text. I can imagine someone trying to add an error message here. Because just suppose a step does fail, it would need explanation. From a design POV, how do we do that, if there are limits on description ?
- The useStepReducer hook is smart! If you add stage: 'active' to any step, it will automatically determine the stages of all other steps (autoStage), both before and after the active one. 
If this is value we want to callout then I would say add an example, so folks can visualize it.

packages/lab/src/stepped-tracker/stepReducer.ts
chrispcode: Mutation replaced with an immutable solution

packages/lab/src/stepped-tracker/utils.ts
```diff
@@ -1,22 +1,65 @@
-import type { StepRecord, StepStage } from "./Step.types";
+import type { StepRecord, StepStage, StepStatus } from "./Step.types";
 import type { StepReducerOptions, StepReducerState } from "./stepReducer.types";
 
-export function assignSteps(
+export function assignStepsStage(
```
origami-z, line 4: Mention rename? This is public and users can use .?

chrispcode: Nope, this is internal  use only. Users have to go through `useStepReducer` to utilize that functionality.

packages/lab/src/stepped-tracker/utils.ts
chrispcode: Both achieve the exact same thing, I just like having a clearly defined exit condition at the top of a recursive function, makes it a touch more readable.

chrispcode:
> Side comments:
> 
> what is the difference between the Basic and Horizonal doc examples, do we need both ? what would happen if I have overly long status text, no design guidance on limits of the text. I can imagine someone trying to add an error message here. Because just suppose a step does fail, it would need explanation. From a design POV, how do we do that, if there are limits on description ?
> 
> * The useStepReducer hook is smart! If you add stage: 'active' to any step, it will automatically determine the stages of all other steps (autoStage), both before and after the active one.
>   If this is value we want to callout then I would say add an example, so folks can visualize it.

- Removed the basic example, kept the orientation horizontal one
- I do have an example [here](https://saltdesignsystem-nxue8z4vd-fed-team.vercel.app/salt/components/stepped-tracker/examples#hook). Check steps with id `step-1-3-1`

packages/lab/src/stepped-tracker/stepReducer.spec.tsx
```diff
@@ -0,0 +1,160 @@
+import { describe, expect, it } from "vitest";
+
```
mark-tate, line 2: Is there a convention in place already to put tests be in `packages/lab/__tests__` ?

chrispcode: Done!

P.S. Personal opinion units tests should be placed close to source code

origami-z (approved):

mark-tate (approved):
approving and then will review in entirety from Labs as a pre-cursor to understanding what we need to do , to move this to Core.

Issue #4724, opened 2025-02-18
TabsNext doesn't adhere to max width and overflows


origami-z:
### Latest version

- [x] I have tested the latest version

### Description

`TabsNext` exceeds available space, especially after selecting an item from the overflow menu. It also appears that at least 2 tabs will be visible when selecting from the overflow menu, works fine when selecting between visible tabs.

Split from #4617 

Community report from [SO#97197](https://go/stack/questions/97197) [SO#97131](https://go/stack/questions/97131)

### Steps to reproduce

https://stackblitz.com/edit/vitejs-vite-nmkvothf?file=src%2FApp.tsx

```
      <div style={{ maxWidth: 200, background: 'yellow' }}>
        <TabsNext defaultValue={tabs[0]}>
          <TabBar inset divider>
            <TabListNext style={{ margin: 'auto' }}>
              {tabs.map((label) => (
                <TabNext value={label} key={label}>
                  <TabNextTrigger>{label}</TabNextTrigger>
                </TabNext>
              ))}
            </TabListNext>
          </TabBar>
        </TabsNext>
      </div>
```

### Expected behavior

Tabs shouldn't exceed given space, when selecting from overflow menu

### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

`@salt-ds/lab@1.0.0-alpha.60`

### Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #4738, opened 2025-02-21
Fix `TabsNext` overflow when only having enough space for one tab

origami-z:
#4724

### DIFF
--- a/packages/lab/stories/tabs-next/tabs-next.stories.tsx
+++ b/packages/lab/stories/tabs-next/tabs-next.stories.tsx
@@ -203,18 +203,31 @@ WithBadge.args = {
 };
 
 export const Overflow: StoryFn<typeof TabsNext> = (args) => {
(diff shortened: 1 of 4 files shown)


joshwooding (approved):

packages/lab/stories/tabs-next/tabs-next.stories.tsx
```diff
@@ -203,18 +203,31 @@ WithBadge.args = {
 };
 
 export const Overflow: StoryFn<typeof TabsNext> = (args) => {
+  const [width, setWidth] = useState(400);
```
joshwooding, line 206: Is this actually needed, it should work with the resize in the toolbar?

origami-z: Easiest for a test i guess

dplsek (approved):

Issue #4719, opened 2025-02-17
Add weight options to Action Font styles


pseys:
Action Font style currently only provides one weight option whereas all other styles (with the exception of `code`) provides three options; default, strong, and small.

Add strong and small options to the Action Font style in both code and Figma enabling consumers to change the weight of text within individual buttons.

mark-tate:
discuss on tue

pseys:
I've created a design spec for the two additional styles. We need to add the following:

- `--salt-text-action-fontWeight-small` = regular
- `--salt-text-action-fontWeight-strong` = bold

mark-tate:
Paprika: Closed by end of wk1

PR #4758, opened 2025-02-26
Add action fontWeight support

origami-z:
#4719

### DIFF
--- a/.changeset/action-fontweight.md
+++ b/.changeset/action-fontweight.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added strong and small font weight support to `Text`.
--- a/.changeset/action-fontweight.md
+++ b/.changeset/action-fontweight.md
@@ -2,5 +2,5 @@
 "@salt-ds/core": minor
 ---
 
-Added strong and small font weight support to `Text` when `styleAs="action"`. 
+Added strong and small font weight support to `Text` when `styleAs="action"`.
 `<strong>` and `<small>` work in `Button` as well.
--- a/.changeset/action-fontweight.md
+++ b/.changeset/action-fontweight.md
@@ -2,4 +2,5 @@
 "@salt-ds/core": minor
 ---
 
-Added strong and small font weight support to `Text` when `styleAs="action"`.
+Added strong and small font weight support to `Text` when `styleAs="action"`. 
+`<strong>` and `<small>` work in `Button` as well.
--- a/.changeset/action-fontweight.md
+++ b/.changeset/action-fontweight.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": minor
 ---
 
-Added strong and small font weight support to `Text`.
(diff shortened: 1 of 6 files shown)


joshwooding (commented):
Do we want to also add support for <Button><strong>Text</strong></Button> ?

origami-z:
> Do we want to also add support for `<Button>\<strong\>Text\</strong\></Button>` ?

I need to sync with @pseys on this. Current implementation feels more a "fix", Text action story was broken before

joshwooding:
> > Do we want to also add support for `<Button>\<strong\>Text\</strong\></Button>` ?
> 
> I need to sync with @pseys on this. Current implementation feels more a "fix", Text action story was broken before

Are you happy for this to go in before then? We can add greater support after

pseys (approved):

pseys:
> > > > Do we want to also add support for `<Button>\<strong\>Text\</strong\></Button>` ?
> > > 
> > > 
> > > I need to sync with @pseys on this. Current implementation feels more a "fix", Text action story was broken before
> > 
> > 
> > Are you happy for this to go in before then? We can add greater support after
> 
> Will make this work for action text first. Button needs more discussion, one Pokemon use case is making all Button stronger, which `<Button>\<strong\>` won't be the solution.

Is it possible, based on what we've added here, for the text within a single Button to be changed to either `--small` or `--strong`? For now that's the main issue we need to resolve. Making all buttons strong or small across a UI is a separate issue IMO.

pseys (dismissed):

.changeset/action-fontweight.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added strong and small font weight support to `Text`.
```
joshwooding, line 5:
```suggestion
Added strong and small font weight support to `Text` when `styleAs="action"`.
```

joshwooding: ?

pseys (approved):

packages/core/src/button/Button.css
```diff
@@ -391,6 +391,15 @@
   cursor: progress;
 }
 
+.saltButton strong {
+  font-weight: var(--salt-text-action-fontWeight-strong);
+}
+
+.saltButton small {
```
mark-tate, line 398: needs a QA test ? (can follow up)

.changeset/action-fontweight.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/core": minor
+---
+
+Added strong and small font weight support to `Text` when `styleAs="action"`. 
+`<strong>` and `<small>` work in `Button` as well.
```
mark-tate, line 6: Does this change need any new docs or existing docs updated? @dplsek @pseys

mark-tate (approved):

mark-tate (approved):

Fercas123 (approved):

Issue #4718, opened 2025-02-17
Add ability to switch font case in Button


pseys:
Add the ability for consumers to set all Buttons (and Button derived components such as Toggle Button) to sentence case instead of the UPPERCASE default.

~~Proposal to achieve this by adding a new system level 'style option' called `action-case` with two options: `actionCase:uppercase`, `actionCase:sentencecase`~~ Not styling option but theme extension via token

Font case cannot be controlled through Variables in Figma, therefore the style option will need to be achieved by providing an additional set of Text Styles with case set on the style.

mark-tate:
Tue meeting.

origami-z:
> * sentencecase

I don't think there is a way to actually achieve this options from [`text-transform`](https://developer.mozilla.org/en-US/docs/Web/CSS/text-transform)

origami-z:
"asTyped" | "uppercase" (default)

origami-z:
[Figma api](https://www.figma.com/plugin-docs/api/TextCase/) uses "ORIGINAL" | "UPPER"

pseys:
> [Figma api](https://www.figma.com/plugin-docs/api/TextCase/) uses "ORIGINAL" | "UPPER"

IMO `asTyped` is a more obvious option, with words like 'legacy' and 'next' etc already floating around, I'm not sure how well `original` would be received.

Can we keep it as `asTyped` and `uppercase`?

origami-z:
> > [Figma api](https://www.figma.com/plugin-docs/api/TextCase/) uses "ORIGINAL" | "UPPER"
> 
> IMO `asTyped` is a more obvious option, with words like 'legacy' and 'next' etc already floating around, I'm not sure how well `original` would be received.
> 
> Can we keep it as `asTyped` and `uppercase`?

ok

pseys:
Speaking to Ben this morning, and following yesterday's discussion on Style Options, I don't believe we should add this as a style option at this point. We have added styles to Figma to allow designers to switch case within their design files. In code this change can be achieved by a dev targeting `--salt-text-action-textTransform` and overriding the value of 'uppercase' with 'none'. 

@origami-z @joshwooding - are you both happy with that?

In design, to better reflect the options in Figma, that styles will be labelled as 'As typed' and 'Uppercase', the latter remaining the default.

mark-tate:
Paprika: closed by end of wk1

origami-z:
[...]

PR #4765, opened 2025-02-27
Add action case styling option

origami-z:
#4718


origami-z:
No need to be a styling option. Change `--salt-text-action-textTransform` in a custom theming solution.

PR #4781, opened 2025-03-04
Splitter - Promote to Core

chrispcode:


.changeset/hip-paws-run.md
```diff
@@ -0,0 +1,6 @@
+---
+"@salt-ds/core": minor
+"@salt-ds/lab": minor
+---
+
+Promote splitter to core
```
mark-tate, line 6: A Core user would see this and ask what is Splitter ?
Have a look through the changelog and see how other components have been announced in Core's changelog.
Something like

`Add `Dialog`, `DialogHeader`, `DialogContent`, `DialogActions`, and `DialogCloseButton` to core`

or Dropdown's changelog entry

we then give an example of the code

packages/core/src/splitter/Splitter.tsx
mark-tate, line 1: The context's don't use a prefix that makes it clear the ownership.
Whilst these contexts belong to the component, any context error might be confusing for the end user.
    <OrientationContext.Provider value={orientation}>
      <AppearanceContext.Provider value={appearance}>
In other components we have given contexts a prefix of the component name.
ie. SplitterAppearanceContext

As appearance is something which is part of the foundation semantics, it could make the difference between someone fixing the issue themselves or raising a bug.

Also.. why do we need 2 contexts which just have single values in each and they are both used by the same component.

So we could create a single context with a name like SplitterContext and in doing so
- make the naming more relatable
- remove one context and type

packages/core/src/splitter/SplitHandle.tsx
mark-tate, line 1: Convo to be had here, we need to align the language on variant with the pokemon work.
Split Handle has a variant primary, secondary, tertiary and transparent
whilst the Button updates introduces a ghost variant.
If transparent remains, it may need to move to appearance, rather than variant
@pseys @dplsek @origami-z

site/src/examples/splitter/Collapsible.tsx
mark-tate, line 1: @ivan-calderon  when collapsing something, you are making existing content smaller, so there is a pattern here that might be worth highlighting

https://react-resizable-panels.vercel.app/examples/collapsible

See how the text based options switch to icons when the available space is smaller.

Is this something which we should have an example of? 
The current examples explain the functionality but not the design opportunities.

packages/core/src/splitter/SplitHandle.css
mark-tate, line 1: @ivan-calderon  is it intended that it does not re-size between density, how well does a small control, work on touch based screen sizes ?

site/docs/components/splitter/index.mdx
mark-tate, line 1: @ivan-calderon @chrispcode @joshwooding 
Thoughts?
We have not called out that our panels are react-resizable-panels. 
The persistance and external persistence features are extremely useful features for our stakeholders.
You can get them just by adding props, we don't need to do anything but we could make folks aware of the value we are creating with the decision to use this third-party library.

site/docs/components/splitter/index.mdx
```diff
@@ -4,7 +4,7 @@ data:
   description: "`Splitter` divides window content into separate regions called split panels that can be dragged and resized, allowing users to customize the layout of their workspace."
   sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/lab/src/splitter"
```
mark-tate, line 5: sourceCodeUrl contains a lab reference to source code

site/docs/components/splitter/usage.mdx
mark-tate, line 1: @ivan-calderon  do we need to update the dashboard pattern and highlight the splitter ?

site/src/examples/splitter/MultiOrientational.tsx
```diff
@@ -1,5 +1,10 @@
...
-import { SplitHandle, SplitPanel, Splitter } from "@salt-ds/lab";
+import {
+  FlexLayout,
+  SplitHandle,
+  SplitPanel,
+  Splitter,
+  Text,
+} from "@salt-ds/core";
 
 import styles from "./splitter.module.css";
```
mark-tate, line 9: in splitter.module.css we have 
/* stylelint-disable salt/correct-theme-token-usage */

this is the only example in core that has this ?

site/src/examples/splitter/MultiOrientational.tsx
mark-tate, line 1: nit: MultiOrientational is a made up word, we can just say Multiple Orientations

site/src/examples/splitter/Transparent.tsx
mark-tate, line 1: NIT: created a style called boxGrey but it's wired up to secondary color which might not always be grey, so the naming is fragile

site/src/examples/splitter/MultiOrientational.tsx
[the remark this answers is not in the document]
chrispcode: There is an example for that it's called [here](https://saltdesignsystem-fut3gc823-fed-team.vercel.app/salt/components/splitter/examples#multi-orientational)

packages/core/src/splitter/Splitter.tsx
chrispcode: - Renaming the contexts makes sense
- Combining the contexts, on the other hand, I am agains. The way they are setup now, they both contain simple strings, which in JS are compared by values, if I wrap them in an object, I am introducing a unfounded relation between them, as well as forced to memoize the joint object.

site/src/examples/splitter/Collapsible.tsx
mark-tate: nit: Consistency
in some examples we use Center and in some we say Middle
I thought it might be because we use middle for vertical references and center for horizontal but Size examples is a case where we use middle in the horizontal plane.

packages/core/src/splitter/Splitter.tsx
mark-tate, line 1: We have consciously/un-consciously decided to not match react-resizeable-panels API?
Horizontal in react-resizable-panels is our vertical
and vertical in react-resizable-panels is our horizontal ?
https://react-resizable-panels.vercel.app/examples/horizontal

packages/core/stories/splitter/splitter.mdx
mark-tate, line 1: @ivan-calderon could the visual hierarchy ever change, after resizing the panels?
It feels like by making one panel more visible than the others, that it might be an better (in some use-cases) for the user of a panel to want one panel to now become the "primary" panel ?

is this a design callout, is it something that a pattern exists for already in design?

I can see this in collapse as well, you collapse a panel and because it's smaller you may want it to change it's background, because although something is smaller, you want to make it standout

packages/core/src/splitter/Splitter.tsx
ivan-calderon: I think this is related to the position/orientation prop added to the component. If so, @chrispcode spotted what you have mentioned above. However, in UITK the position/orientation is treated as is in the component at the moment. Would you recommend to follow @chrispcode and your direction instead?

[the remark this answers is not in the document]
jake-costa: @mark-tate We can add it. As you noted they are optional, initial focus was to handle the required items and target size issues. 

@chrispcode thoughts on adding F6?

packages/core/src/splitter/SplitPanel.css
mark-tate, line 1: this uses `has` which is not supported by our minimum browser ?
That's fine, if there is just cosmetic degradation but does it mean that the design specs won't match for some stakeholders ?
Is there someone where we need to call this out from a design POV @bhoppers2008  to avoid a defect or support issue. Are we creating a precedent here (a good one?) that enables us to start using  
@supports selector(:has(*)) {}
(or similar)
How does this impact the minimum browser docs ?
Perhaps we agree that we can start to use "@supports" more,  but we should share that conversation with the whole team to all be in alignment.
Moving the needle is great but we need to do things in a way that is clear and can be consistently applied.

packages/core/package.json
```diff
@@ -17,7 +17,8 @@
     "@salt-ds/icons": "workspace:^",
     "@salt-ds/styles": "workspace:*",
     "@salt-ds/window": "workspace:*",
-    "clsx": "^2.0.0"
+    "clsx": "^2.0.0",
+    "react-resizable-panels": "^2.1.7"
```
mark-tate, line 21: @joshwooding  not everyone needs splitter but we are forcing everyone to install react-resizable-panels.
That feels different to @floating-ui which has multiple use-cases across components and everyone needs.
Should this be a separate package, or if not, what is the rationale, when to have a separate  package (date adapters) and when we can extend the core dependencies (react-resizable-panels) ?

packages/core/stories/splitter/splitter.mdx
ivan-calderon: @chrispcode can correct me if I am wrong. He made of the component flexible enough to match backgrounds (primary and secondary background) between the splitHangle and the spliPanel. This is specially useful when only one border is used in the splitHandle. He documents that in the [Variants](https://saltdesignsystem-fut3gc823-fed-team.vercel.app/salt/components/splitter/examples#variant) section. 

However, I am not sure how the programmatic part of it can work (Media queries?). That will require to adjust color on re-sizing interactions (From what I understand). If so, that will require to revisit background color combinations when the splitter is fully bordered, partially borders or transparent; how should those work on primary, secondary or tertiary background; and how to provide guidance which splitPanel is hierarchical on top of the others.  

About the visual hierarchy. That can be achieved in different ways: size and scale, color and contrast, typography (in this case content in each panel can convey hierarchy as well). 

I think this is a very specific use case. But this is worthy to consider if we consider it a requirement.

@chrispcode any thoughts?

packages/core/src/splitter/SplitHandle.tsx
chrispcode: Additional explanation for each prop available here: https://jpmorganchase-splitter.saltdesignsystem-storybook.pages.dev/?path=/docs/documentation-core-splitter--docs

site/src/examples/splitter/Collapsible.tsx
chrispcode: We do have a collapsible example that does what the base library does [here](https://saltdesignsystem-fut3gc823-fed-team.vercel.app/salt/components/splitter/examples#collapsible). Teams can assign a @container query based on the size and implement that functionality themselves.

chrispcode: > nit: Consistency in some examples we use Center and in some we say Middle I thought it might be because we use middle for vertical references and center for horizontal but Size examples is a case where we use middle in the horizontal plane.

Yes, the middle and center are used based on the orientation of the `Splitter`.

packages/core/src/splitter/Splitter.tsx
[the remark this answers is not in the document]
chrispcode: How would it work if the focused element is something inside one of the panels and the user clicks F6?

packages/core/src/splitter/utils.ts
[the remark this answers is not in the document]
chrispcode: I am 50/50 about this one. It doesn't hurt having it in a different file

packages/core/src/splitter/SplitPanel.css
chrispcode: Quick explanation of `@supports`: allows us to check if the browser that the current user uses support a particular feature of CSS. The CSS `:has` selector is supported on `93.62%` of the browsers, but not supported on some of the lower versions of chrome and firefox we are bound to support. 

What features wouldn't work if the user is on the 6% of browsers that are not supported:
- Splitter inside Splitter, that also have appearance set to transparent

packages/core/src/splitter/Splitter.tsx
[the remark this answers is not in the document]
jake-costa: @chrispcode Pressing F6 when in a split window environment would move focus from one panel to another, placing focus on the panel itself instead of the elements inside it.

chrispcode: Done!

site/src/examples/splitter/MultiOrientational.tsx
chrispcode: Done!

site/src/examples/splitter/Transparent.tsx
chrispcode: Done!

site/src/examples/splitter/MultiOrientational.tsx
chrispcode: Done!

site/docs/components/splitter/index.mdx
chrispcode: Done!

site/src/examples/splitter/Collapsible.tsx
mark-tate: Check Size and Variant examples, both have vertical splitters but they use different descriptions, Center and Middle.

site/src/examples/splitter/MultiOrientational.tsx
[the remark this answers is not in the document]
mark-tate: What is PanelGroup for, if not used ?

packages/core/src/splitter/Splitter.tsx
mark-tate: If we are using an API from react-resizable-panels would say we align to that rather than invert it just to match UITK. There can/should be a migration guide for UITK components.  Any +1s ?

packages/core/stories/splitter/splitter.mdx
mark-tate: If we ignore the hows for a minute... Why do folks move resize panels? they typically resize panels to see more content. That is sometimes at the expense of some of the other UI. However that UI might be crucial to operation and you could find yourself in a situation where UI controls are hidden or not usable, unless re-expanded. What we see in this example is 
https://react-resizable-panels.vercel.app/examples/collapsible
is that when the user resizes the panel they swap one content for another.
ie.  we give the user more space but adapt the content so the UI is still usable.

Our examples are rather like wireframes but there is some design patterns that might be used with a Splitter, like swapping to an icon bar for page controls. 

Equally when using panels, what does the visual hierachy mean? 
Are there patterns around  the one that has focus, being the primary and other secondary and the collapsed ones tertiary. Trying to make sense of what the semantics mean so we can curate some great examples of things you can do with this? (our examples right now, are empty skeletons, they show what the Splitter does but not how we should use it, in practice).

packages/core/src/splitter/utils.ts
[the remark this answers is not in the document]
mark-tate: if we apply that rule everywhere, we would have thousands of files and the rationale would be hard to understand as each file would of been created because it was someone's preference.

My rationale on this..
If the code is shared across concerns or exported then perhaps a separate file shows that this file can be used outside of the component but here that's not the case, it's just clutter and means a developer has to change file and back again to follow the code.

packages/core/src/splitter/SplitPanel.css
mark-tate: When changing the standard, we need to consider the bigger picture, what do we need to change to support it and how to go forward from here? As a team we also need to be careful how we use this, so we manage it's usage.

On this PR can you add this to our Supported Platform docs

```
## Progressive Enhancement with Newer Features

While Salt supports the above browser versions, we also embrace progressive enhancement by utilizing newer browser features, such as CSS `:has()`, when available. These features are wrapped in `@supports` rules to ensure they are only applied if the user's browser supports them. This approach allows us to deliver an enhanced experience for users with modern browsers while maintaining compatibility with older versions.

Important Note: Users on older browsers that do not support these newer features may experience a sub-optimal user experience. We recommend keeping your browser up-to-date to take full advantage of the latest features and improvements.
```

@joshwooding @bhoppers2008 @origami-z for visibility
We are introducing a feature not supported by our current minimum browser.
I believe we can manage a small amount of usage on a case-by-case basis, initially on component with a low-usage such as Splitter. 

Adding this disclaimer to our docs, will help communicate but equally provide an incentive for folks to update. 

Thoughts?

site/src/examples/splitter/MultiOrientational.tsx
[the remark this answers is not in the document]
chrispcode: Yes, `SplitPanel` is `Splitter` in the case of `salt-ds`. More information about remapping can be found [here].(https://jpmorganchase-splitter.saltdesignsystem-storybook.pages.dev/?path=/docs/documentation-core-splitter--docs)

packages/core/src/splitter/Splitter.tsx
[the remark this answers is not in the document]
chrispcode: That will take some time, I would suggest we implement it after we promote this component to core. It would be a nice feature to implement at a later point.

packages/core/stories/splitter/splitter.mdx
chrispcode: Here is the react-resizable-panels example done in salt: https://jpmorganchase-splitter.saltdesignsystem-storybook.pages.dev/?path=/story/core-splitter--collapsible-container-query

site/src/examples/splitter/Collapsible.tsx
chrispcode: Done! Changed all `Middle` to `Center`.

packages/core/src/splitter/SplitHandle.css
chrispcode: Yes, it remains 24px despite density

site/src/examples/splitter/Collapsible.tsx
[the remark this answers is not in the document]
ivan-calderon: @mark-tate I was researching a little about add/remove splitPanels pattern. I have not been able to find any or a consistent use cases of it. Is there any specific LOB requesting that? I can reach them out. TY!

packages/core/src/splitter/Splitter.tsx
ivan-calderon: Understood. That make sense! @chrispcode made the same recommendation.

chrispcode: It has been changed now -  our `orientation prop` matches the base library's `direction` prop

site/src/examples/splitter/Collapsible.tsx
chrispcode: Add above example to side, the one that contains the container query

packages/core/src/splitter/Splitter.tsx
chrispcode: Decided to keep it the same as the previous splitter, the same as the divider component.

packages/core/stories/splitter/splitter.mdx
chrispcode: We've had a discussion and noted that we should consider the value of variant based on the side of a panel

packages/core/src/splitter/Splitter.tsx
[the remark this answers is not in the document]
chrispcode: If you need this feature, raise a feature request and we'll do it!

site/src/examples/splitter/Collapsible.tsx
chrispcode: Done!

packages/core/src/splitter/SplitPanel.css
chrispcode: Done!

site/src/examples/splitter/Collapsible.tsx
chrispcode: https://saltdesignsystem-ic3err9kx-fed-team.vercel.app/salt/components/splitter/examples#collapsible-with-container-query

packages/core/src/splitter/SplitPanel.css
chrispcode: https://saltdesignsystem-ic3err9kx-fed-team.vercel.app/salt/about/supported-platforms

site/docs/components/splitter/usage.mdx
ivan-calderon: Sure. I'll create an issue for that NP. Please, be aware that this will be seated for a time in the Figma backlog due to priorities.

site/docs/components/splitter/index.mdx
mark-tate: in the review updates - I added a storybook example and acknowledged that Splitter uses react-sizeable-panel, should be enough to show the value/opportunity

Issue #4814, opened 2025-03-10
Update Accordion A11y Docs


jake-costa:
### Related page(s)

https://www.saltdesignsystem.com/salt/components/accordion/accessibility

### Kind of issue

Missing information

### Description

Adding Known Issue with Safari/VoiceOver - does not report state changes dynamically. (e.g. expanded/collapsed) Not a Salt Bug

Is a Safari/Mac/VoiceOver bug.

Add A11y best practices

### Browser

- [ ] Chrome
- [x] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #4819, opened 2025-03-10
Update accordion accessibility docs

jake-costa:
Adding Known Issue with Safari/VoiceOver - does not report state changes dynamically. (e.g. expanded/collapsed) Not a Salt Bug

Is a Safari/Mac/VoiceOver bug.

Add A11y best practices

Closes issue: https://github.com/jpmorganchase/salt-ds/issues/4814

### DIFF
--- a/site/docs/components/accordion/accessibility.mdx
+++ b/site/docs/components/accordion/accessibility.mdx
@@ -38,6 +38,6 @@ data:
 </KeyboardControls>
 
 ## Known Issue
-- VoiceOver/Safari on Mac does not report state changes dynamically for Accordion buttons.
+- VoiceOver/Safari on Mac bug: VoiceOver does not report state changes dynamically for Accordion buttons. __(This is not a Salt bug)__
   - Only workaround is to tab away and shift+tab back to this control then it will announce the state.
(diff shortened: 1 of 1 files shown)


site/docs/components/accordion/accessibility.mdx
```diff
@@ -8,6 +8,15 @@ data:
   $ref: ./#/data
 ---
 
+## Best practices
+- Hide the chevron icon from screen readers.
```
joshwooding, line 12: @jake-costa IMO, this isn't a best practice as we should do it for them. If that's not the case it's a bug :)

jake-costa: @joshwooding Thanks for the feedback. I have received a lot of questions about how the chevron should be labeled (though mainly from designers but some devs) so I figured it would be good to clarify in the docs. But you raise a valid point since we already do it for them. 

what are your thoughts on something like:
The chevron icon is hidden from screen readers by default. The state is conveyed though aria-expanded.
 
 Or do you still think just to remove it completely?

origami-z (approved):

joshwooding (approved):

origami-z:
Failed test not related to doc change in this PR

PR #4820, opened 2025-03-10
Add manual GH action lint fix

origami-z:
Trying to help non-devs updating content, and ran into Linting issue.

### DIFF
--- a/.github/workflows/lint-auto-fix.yml
+++ b/.github/workflows/lint-auto-fix.yml
@@ -1,0 +1,31 @@
+name: "Link auto fix"
+on:
+  workflow_dispatch:
+
+permissions:
+    contents: write
+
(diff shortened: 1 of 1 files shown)


origami-z:
@joshwooding do you think this will work, for those PRs edited via online editor and failing prettier (e.g. by design / content)

joshwooding:
> @joshwooding do you think this will work, for those PRs edited via online editor and failing prettier (e.g. by design / content)

It might be better to trigger with label? What do you think?

origami-z:
> > @joshwooding do you think this will work, for those PRs edited via online editor and failing prettier (e.g. by design / content)
> 
> It might be better to trigger with label? What do you think?

Oh yeah, i forgot about that. Let me change.

PR #4829, opened 2025-03-12
Update stepper input

tayyabataimur:


packages/core/src/status-adornment/StatusAdornment.tsx
```diff
@@ -27,9 +27,9 @@ export interface StatusAdornmentProps extends IconProps {
 }
 
 const statusToAriaLabelMap: Record<AdornmentValidationStatus, string> = {
-  error: "error",
-  warning: "warning",
-  success: "success",
+  error: "error.",
```
joshwooding: This change will affect other components, has this been fully tested?

site/docs/components/number-input/examples.mdx
```diff
@@ -10,118 +10,122 @@ data:
 
 ## Default
 
-Number input has an increment step value of 1 and a step block value of 10, and you can only use it to accept numeric characters.
+The `NumberInput`, by default, features an increment `step` of 1, enabling precise value adjustments. For larger value adjustments, users can utilize the `Shift` key in combination with the `Arrow` keys to apply a `stepMultiplier` which is 2 by default.
```
joshwooding: I don't think we typically use backticks this way e.g. `Shift`, `Arrow`. Typically they are reserved for component names, prop names and literal values.

site/docs/components/number-input/examples.mdx
```diff
@@ -10,118 +10,122 @@ data:
...
+
+<LivePreview componentName="number-input" exampleName="Limits" />
+
+## Clamp behaviour
 
-To style the number input with a full border, set `bordered={true}`. This styling is ideal when the input field uses the same fill color as the background, such as a primary fill color on a primary background.
+By default, when `min` and `max` values are set and user types a value that is out of range, the value is not clamped. You can set the `clampBehaviour` to `default` to clamp the value to `max` or `min`, whichever is closest when the input loses focus (on blur). Setting the `clampBehavior` to `strict` will not allow the user to even type an out of range value.
 
-<LivePreview componentName="number-input" exampleName="Bordered" />
+<LivePreview componentName="number-input" exampleName="ClampBehaviour" />
```
joshwooding, line 48: We're technically not following our best practices with the max value here, I had to check the code to find what value is the maximum allowed

site/docs/components/number-input/examples.mdx
```diff
@@ -10,118 +10,122 @@ data:
...
-- Use the form field error state to inform users who manually type in a value outside the accepted range that the value is invalid.
+With Intl Number Format, start and end adornments and custom formatting.
 
-<LivePreview componentName="number-input" exampleName="Limits" />
+<LivePreview componentName="number-input" exampleName="Formatting" />
 
 ## Custom steps
 
-Customize the default, increment and block values to suit your use case.
+You can customize the `default` value of the `NumberInput` along with `step` and `stepMultiplier` to tailor the values according to your usecase.
```
joshwooding: `default` would also not be typically done, since `defaultValue` is the prop name

site/docs/components/number-input/examples.mdx
```diff
@@ -10,118 +10,122 @@ data:
...
+## Formatting
 
-- Clearly state the accepted value range in the form field’s helper text.
-- Use the form field error state to inform users who manually type in a value outside the accepted range that the value is invalid.
+With Intl Number Format, start and end adornments and custom formatting.
 
-<LivePreview componentName="number-input" exampleName="Limits" />
+<LivePreview componentName="number-input" exampleName="Formatting" />
 
 ## Custom steps
```
joshwooding, line 86: Is there a use-case for a strict step clamp? Or is that already possible. It might be a controlled example or handled by validation

joshwooding:
Also NumberInput currently doesn't follow the rest of our form controls with it's bordered style, this should probably be aligned @navkaur76

packages/core/src/status-adornment/StatusAdornment.tsx
tayyabataimur: Have to revert this, unintentional

packages/lab/src/number-input/internal/useFormatControlled.ts
```diff
@@ -0,0 +1,69 @@
...
+type UseFormatControlledProps<T> = {
+  controlled?: T;
+  default: T;
+  name: string;
+  state: string;
+  format?: (value: T) => T;
+};
+
+export const useFormatControlled = <S>({
+  controlled,
```
mark-tate: this appear like a copy of useControlled and internal to NumberInput, if we go down this route, then suspect we move this to utils?

Could this hook not just call the other hook, just return formatted results ?

packages/lab/src/number-input/internal/useFormatControlled.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (process.env.NODE_ENV !== "production") {
+      if (!isControlled && defaultValue !== defaultProp) {
+        console.error(
+          [
+            `Salt: A component is changing the default ${state} state of an uncontrolled ${name} after being initialized. ` +
+              `To suppress this warning opt to use a controlled ${name}.`,
+          ].join("\n"),
+        );
+      }
+    }
```
mark-tate: is this missing support for uncontrolled formatting, which is currently only applied via useState ?

Something like
```
  const setValueIfUncontrolled: Dispatch<SetStateAction<S>> = useCallback(
    (newValue) => {
      if (!isControlled) {
        setValue(format(newValue));
      }
    },
    [],
  );

  return [value, setValueIfUncontrolled, isControlled];
  ```

File: packages/lab/src/number-input/NumberInput.tsx
Comment (mark-tate) [inline packages/lab/src/number-input/NumberInput.tsx:1]: Quite a hard one to review, as lots of responsibilities going on in one file.

things I picked out so far
- inputOnBlur is not called from handleInputBlur, so userland onBlur events are not supported
- these 2 lines could be problematic
```parse?.(value) || value```
```const rawValue = parse?.(inputValue) || inputValue;```
I would switch to nullish coalescing

File: packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
Comment (mark-tate) [inline packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx:1]: I asked LLM what was missing, worth reviewing to see if any of these should be added

 
 Edge Case Tests:

Test behavior with extremely large or small numbers, including edge cases like Infinity or NaN.
Verify behavior when rapidly changing values, such as holding down increment/decrement buttons or keys.
Error Handling Tests:

Test scenarios where the parse or format functions might throw errors, ensuring that the component handles these gracefully.
Focus and Blur Behavior:

Test focus and blur behavior, especially when switching between inputs or interacting with other elements on the page.
Read-Only Mode:

Verify that the component behaves correctly when readOnly is set, ensuring that no changes can be made and that the component is announced as read-only.
Custom Styling and Class Names:

Test custom styling and class names to ensure that they are applied correctly and do not interfere with the component's functionality.
Dynamic Prop Changes:

Test behavior when props like min, max, step, or value are changed dynamically after the component has mounted.
Button Visibility:

Verify that the increment and decrement buttons are correctly hidden or shown based on the hideButtons prop.
Validation Status:

Test different validation statuses (error, warning, success) to ensure that they are displayed correctly and affect the component's appearance and behavior.
Internationalization and Localization:

Test behavior with different locales, especially those with different number formats (e.g., using commas as decimal separators).

File: .changeset/silent-carrots-greet.md
Comment (mark-tate) [inline .changeset/silent-carrots-greet.md:1]: made some notes/comments, maybe not all are actionable, just highlighting for awareness
- typing a non-numeric character moves the caret to the end, it that expected?
- I can only find the component in search on the site, if I type "number i",  it makes this component hard to find at the mo (maybe something missing)
- when using the arrow key, the caret moves to the end, is that expected ?
- do we need an exponential example
- parsing is somewhat buried in the docs, within formatting
- testing is based on 5 examples but there are 12 examples, so maybe not all functionality is tested (check LLM comment)
- once you have an error state you can type whatever you like -  I cannot type -2 into min/max example but if I type 222 and then I can change it to -2 by typing ( is this an issue or  a rationale for it)
- if I type 0.9 (valid) into min/max example and then decrement I get -0.9 and an error but I could not get that before when I tried to step to it from 0
- the emptyReadOnlyProp is mentioned in the docs but not in the code/props table
- do we need better explanations on some of the examples, compact notation is not explained but it supports multiple notations, you only know if you read the code (but not all users of our site are developers)
- the start and end adornments are somewhat far from the value, wonder if this contributes to visual inconsistency as input widths are dictated by layout, so multiple number inputs on a page may look different.
- do we need something like `renderText`
- does it support leading zeros
- how much duplication is there with input? The input component is hardcoded, wonder if being able to provide a custom input would make this behaviour be able to plugin to any input

Comment (mark-tate) [reply .changeset/silent-carrots-greet.md:1]: changelog could be better?
What's the impact to someone who already is using the current version?

Does not explain what changes are, 
ie. what does clampBehaviour do.

Looks like we may of made more changes ?
decimalPlaces -> decimalScale

File: packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx [hunk 8]
```diff
@@ -1,18 +1,32 @@
...
+  Default,
+  ControlledFormatting,
+  MinAndMaxValue,
+  ResetAdornment,
+  ReadOnly,
+  UncontrolledFormatting,
+} = composedStories;
 
 describe("Number Input", () => {
+  it("should have no a11y violations on load", () => {
```
joshwooding: checkAccesibility() should replace this

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -117,24 +131,25 @@ describe("Number Input", () => {
...
     );
 
     cy.findByRole("spinbutton").focus().realPress("ArrowDown");
-    cy.findByRole("spinbutton").should("have.value", "0").realPress("PageDown");
-    cy.findByRole("spinbutton")
-      .should("have.value", "-100")
-      .realPress(["Shift", "ArrowDown"]);
-    cy.findByRole("spinbutton").should("have.value", "-200").realPress("Home");
+    cy.findByRole("spinbutton").should("have.value", "0");
+    cy.findByRole("spinbutton").focus().realPress("PageDown");
```
joshwooding:
```suggestion
    cy.realPress("PageDown");
```
Do we need to focus the element each time?

packages/lab/src/number-input/NumberInput.css
```diff
@@ -1,24 +1,34 @@
 /* Styles applied to number container */
 .saltNumberInput {
-  --numberInput-border: none;
+  --numberInput-border: var(--salt-size-border) var(--salt-container-borderStyle) var(--numberInput-borderColor);
```
joshwooding:
```suggestion
  --numberInput-border: var(--salt-size-fixed-100) var(--salt-container-borderStyle) var(--numberInput-borderColor);
```

packages/lab/src/number-input/NumberInput.css
```diff
@@ -140,7 +134,7 @@
 }
 
 /* Styling when focused */
-.saltNumberInput-focused {
+.saltNumberInput:focus-within {
```
joshwooding: This doesn't seem right, even though it's tempting to use :focus-within this means that when the focus is on the adornments two focus-rings appear cc @navkaur76

packages/lab/src/number-input/NumberInput.css
```diff
@@ -275,17 +245,47 @@
...
+
+.saltNumberInput-hiddenButtons .saltNumberInput-buttonContainer {
+  display: none;
 }
 
 /* Styles applied to number buttons */
 .saltNumberInput-numberButton {
-  --saltButton-height: calc((var(--salt-size-base) - var(--numberInput-buttonGap)) * 0.5);
-  --saltButton-width: var(--salt-size-base);
+  --saltButton-height: calc((var(--salt-size-base) - (var(--salt-spacing-50) * 2) - var(--salt-size-border-strong)) / 2);
```
joshwooding:
```suggestion
  --saltButton-height: calc((var(--salt-size-base) - (var(--salt-spacing-50) * 2) - var(--salt-size-fixed-200)) / 2);
```

packages/lab/src/number-input/NumberInput.css
```diff
@@ -275,17 +245,47 @@
...
-  --saltButton-borderRadius: 0 0 var(--salt-palette-corner-weak, 0) var(--salt-palette-corner-weak, 0);
+  --saltButton-borderRadius: 0 0 var(--salt-palette-corner-weaker, 0) var(--salt-palette-corner-weaker, 0);
+}
+
+.saltNumberInput-activationIndicator {
+  left: 0;
+  bottom: 0;
+  width: 100%;
+  position: absolute;
+  border-bottom: var(--salt-size-border) var(--salt-editable-borderStyle) var(--salt-editable-borderColor);
```
joshwooding:
```suggestion
  border-bottom: var(--salt-size-fixed-100) var(--salt-editable-borderStyle) var(--salt-editable-borderColor);
```

packages/lab/src/number-input/NumberInput.css
```diff
@@ -275,17 +245,47 @@
...
+}
+
+.saltNumberInput:hover .saltNumberInput-activationIndicator {
+  border-bottom-style: var(--salt-editable-borderStyle-hover);
+  border-bottom-color: var(--salt-editable-borderColor-hover);
+}
+
+.saltNumberInput:focus-within .saltNumberInput-activationIndicator {
+  border-bottom: var(--salt-size-border-strong) var(--salt-editable-borderStyle-active) var(--salt-editable-borderColor-active);
```
joshwooding:
```suggestion
  border-bottom: var(--salt-size-fixed-200) var(--salt-editable-borderStyle-active) var(--salt-editable-borderColor-active);
```

packages/lab/src/number-input/NumberInput.css
tayyabataimur: Yeah you're right, nice catch! I was trying avoid using state for focus mainly because there's already so much state management going on in the component. I'll find a way to get rid of the two focus rings.

packages/lab/src/number-input/NumberInput.css
```diff
@@ -275,17 +245,47 @@
...
+}
+
+.saltNumberInput-bordered .saltNumberInput-activationIndicator {
+  border-bottom-width: 0;
+}
+
+.saltNumberInput-bordered.saltNumberInput:focus-within .saltNumberInput-activationIndicator {
+  border-bottom-width: calc(var(--salt-size-border-strong) - var(--salt-size-border));
```
joshwooding:
```suggestion
  /* Activation indicator width minus the border from the number input. */
  border-bottom-width: var(--salt-size-fixed-100);
```

packages/lab/src/number-input/internal/useCaret.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const recordCaret = useCallback(() => {
+    if (inputRef.current) {
+      const { selectionStart: start, selectionEnd: end } = inputRef.current;
+      caretPositionRef.current = {
+        start,
+        end,
+      };
+    }
+  }, [inputRef.current]);
```
joshwooding: ref's shouldn't be used in dependency arrays, is this needed?

site/docs/components/number-input/examples.mdx
```diff
@@ -10,118 +10,135 @@ data:
 
 ## Default
 
-Number input has an increment step value of 1 and a step block value of 10, and you can only use it to accept numeric characters.
+The `NumberInput` is designed with a default increment step of 1, which is adjustable via the `step` prop for precise value changes. For larger adjustments, users can hold the Shift key while using the Arrow keys to apply a `stepMultiplier`, a prop that defaults to 2.
```
joshwooding: The wording here is a little odd: "is designed with" makes it sound like we don't recommend changing it

site/src/examples/number-input/ButtonAdornments.tsx
```diff
@@ -4,29 +4,62 @@ import {
...
+
+const ResetAdornment = () => {
   const defaultValue = 1000;
   const [value, setValue] = useState<number | string>(defaultValue);
+  const [accessibleText, setAccessibleText] = useState("");
+
+  const formFieldLabel = "Number Input with Reset adornment";
+  const accessibleTextId = useId();
+
+  const clearAccessibleText = () =>
```
joshwooding, line 28: I think this should use `useAriaAnnouncer()`

.changeset/silent-carrots-greet.md
```diff
@@ -0,0 +1,12 @@
+---
+"@salt-ds/lab": minor
+---
+
+Updated the `NumberInput`:
+
+1. Added support for `format` and `parse` callbacks for formatting capabilities.
+2. Added `clampValue` prop to allow enforcement of values within the min and max range.
+3. Refactored `stepBlock` prop to be `stepMultiplier` to be consistent with `Slider` and `RangeSlider` components and to ensure that valid values remain reachable.
+4. Refactored `decimalPlaces` to be `decimalScale`.
```
mark-tate, line 10: see the changelog for carousel
you can diff with a code example
https://github.com/jpmorganchase/salt-ds/blob/baebcea6808662d45c65641cf2fb8b5db2b69f7d/.changeset/dirty-months-sip.md
Might make the changes easier to consume

Whilst labs are un-stable components we can mitigate the friction by providing great changelogs

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
mark-tate, line 1: the controlled test does not have any check that onChange is called only on user changes 
...and all other spys are for tests related to uncontrolled, so this looks like a gap in the test coverage

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -209,21 +253,21 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
       ? ["required", "asterisk"].includes(formFieldRequired)
       : inputRequired;
 
+    const [isEditing, setIsEditing] = useState(false);
```
joshwooding, line 251: I haven't checked but we should check if state is correct for isEditing and isAdjusting. Refs might be able to be used instead.

packages/lab/src/number-input/useNumberInput.ts
mark-tate, line 1: If I click and hold the spin button and mouse out, should the increment stop ?

site/docs/components/number-input/examples.mdx
mark-tate, line 1: Whilst usage says you can't type exponential numbers, you can pass them in. So it might be worth adding that.
Curious, is It technically feasible to make it work with, if the component is controlled and has custom formatting and parsing.

site/src/examples/number-input/index.ts
mark-tate, line 1: the custom button example does not increment on hold, we might need to explain how to reproduce this behaviour if using a custom button.

packages/lab/src/number-input/NumberInput.tsx
mark-tate, line 1: onChange is called for every step, when you press and hold. Suspect this needs to be configurable.
Some cases you want a ramp and in some case you just want to know the final value.

packages/lab/src/number-input/internal/useCaret.ts
tayyabataimur: Not really now that I think about it. Added it as a quick fix to silence the biome warnings. I'll remove it.

packages/lab/src/number-input/NumberInput.tsx
tayyabataimur: For `isEditing`, I need it to be a state because it triggers a re-render that formats the value. For `isAdjusting`, can make it a ref, I'll look into it

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -192,6 +221,22 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
     const isDisabled = disabled || formFieldDisabled;
     const isReadOnly = readOnlyProp || formFieldReadOnly;
     const validationStatus = formFieldValidationStatus ?? validationStatusProp;
+    const isEmptyReadOnly = isReadOnly && !defaultValueProp && !valueProp;
+    const defaultValue = isEmptyReadOnly
```
mark-tate, line 225: this matches input but would show the emptyReadOnlyMarker, when 0.
not sure that is the intended use, as it's designed to avoid empty fields ?

joshwooding (approved):

PR #4841, opened 2025-03-15
Stepper review additions

mark-tate:


packages/core/src/__tests__/unit/stepper/StepReducer.spec.tsx
mark-tate, line 1: filename still has StepReducer

mark-tate (commented):
can fix up the test file name  on the other branch

packages/core/src/__tests__/unit/stepper/StepReducer.spec.tsx
chrispcode: Renamed!

site/src/examples/stepped-tracker/VerticalWithLinks.tsx
```diff
@@ -0,0 +1,115 @@
...
+      <SteppedTracker orientation={"vertical"}>
+        {state.steps.map((step) => (
+          <Step
+            key={step.id}
+            {...step}
+            render={(props) => {
+              if (props.stage === "completed") {
+                return (
+                  <Tooltip content={"reset to first step"} placement={"right"}>
+                    <Link
```
mark-tate: this renders a step as a link

Issue #4737, opened 2025-02-20
aXe: Checkbox Nested in Option is being flagged for nested interactive


jake-costa:
### Latest version

- [x] I have tested the latest version

### Description

Options that include nested checkbox inputs (e.g. multi-select) are being flagged for [nested-interactive](https://dequeuniversity.com/rules/axe/4.10/nested-interactive?application=AxeChrome)

Components effected:
Multi-select combo box
Multi-select dropdown
Multi-select listbox

In addition, Nvda/Firefox is failing to convey which options are selected/not-selected.

### Steps to reproduce

Open chrome
Navigate to one of the components listed above
ensure the listbox with options containing the checkboxes is displayed
run aXe

### Package name(s)

Core (@salt-ds/core)

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #4882, opened 2025-03-20
Fix nested checkbox in option

origami-z:
Replace checkbox (with input) with checkbox icon in option.

~~Also added `aria-checked` to fix NVDA selected/unselected announcement. Need to keep both on multi-select, as VoiceOver / Safari only works with `aria-selected`~~ (to create new issue)

#4737

### DIFF
--- a/.changeset/clean-turkeys-fold.md
+++ b/.changeset/clean-turkeys-fold.md
@@ -1,5 +1,0 @@
----
-"@salt-ds/core": patch
----
-
-Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox.
--- a/.changeset/clean-turkeys-fold.md
+++ b/.changeset/clean-turkeys-fold.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox.
--- a/.changeset/clean-turkeys-fold.md
+++ b/.changeset/clean-turkeys-fold.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox using NVDA.
+Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox.
--- a/.changeset/clean-turkeys-fold.md
+++ b/.changeset/clean-turkeys-fold.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox.
+Fixed Option not conveying selected state when used in multiselect ComboBox, Dropdown and ListBox using NVDA.
--- a/.changeset/option-nested-interactive.md
+++ b/.changeset/option-nested-interactive.md
@@ -1,0 +1,7 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed multiselect Option having nested interactive elements, impacted ComboBox, Dropdown and ListBox.
+
+Fixes #4737.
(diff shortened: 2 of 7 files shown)


jake-costa (dismissed):
Hi @origami-z, This does remove the aXe flag but I have a few comments. 

1. It might be a storybook issue but I am noticing with NVDA/Firefox, the selected option state is not being conveyed all I get is "option label x of x" but missing "selected" or "not selected". we would need at least 1 or the other. We had this on the previous example but I was hoping removing the nested checkbox would resolve it.  

This is only with NVDA/Firefox, seems to be working correctly with the other AT/Browser combinations though. 

2. After an option is selected and is added to the list in the combobox, the close icon "X" svg contains a label and is not hidden. We should add aria-hidden"true" to the close icon svg and add aria-label to the button (e.g. "option label" click to remove)

That said, we can merge and close this PR since it does address the original issue that was called out and create a new one for the comments above if needed.

origami-z:
> This is only with NVDA/Firefox, seems to be working correctly with the other AT/Browser combinations though.

NVDA seems to only recognize `aria-checked` in multiselect, where others understand `aria-selected` (e.g. VoiceOver only sees selected not checked). I updated the option code in this PR to use both when multiselect, although not recommended by [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected#option), but seem to work in my testing. Can you have a try as well?



> 2. the close icon "X" svg contains a label and is not hidden

Is this a real issue? The label there seem to be helping, when focused on the pill, it reads out "<option> click to close, button"

jake-costa:
> NVDA seems to only recognize `aria-checked` in multiselect, where others understand `aria-selected` (e.g. VoiceOver only sees selected not checked). I updated the option code in this PR to use both when multiselect, although not recommended by [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-selected#option), but seem to work in my testing. Can you have a try as well?

To clarify, while you are correct that NVDA does not announce "selected" it should announce "not selected" 
Here is NVDA's take on that [#12893](https://github.com/nvaccess/nvda/issues/12893):
"Only announcing one of the states (selected vs not selected) reduces verbosity, the other state is implied."

From the [ARIA spec regarding aria-selected](http://www.w3.org/TR/wai-aria/states_and_properties#aria-selected):

If no DOM element in the widget is explicitly marked as selected, assistive technologies MAY convey implicit selection which follows the keyboard focus of the managed focus widget.

So, because the item is focused, it's already implicitly selected. Therefore, setting aria-selected="true" doesn't change anything and thus NVDA doesn't announce any change. If you want the item to be unselected when focused, you need to explicitly set aria-selected="false".

That said, I think we should investigate why "not selected" is not being announced instead of adding both (aria-checked and aria-selected)

> > 2. the close icon "X" svg contains a label and is not hidden
> 
> Is this a real issue? The label there seem to be helping, when focused on the pill, it reads out " click to close, button"

Yes, this is an issue. While tabbing to the first selected option NVDA will announce the button correctly, when using arrow keys these are treated separately (arrow up = "click to close, button", arrow up again "option label, button"). In this case the svg is decorative and should not have an aria-label and should be hidden using aria-hidden. The button, should contain a single aria-label that should contain ("visible text" click to close). This prevents confusion and ensures consistent and clear announcements across navigation methods.

To help explain this, you can also try selecting multiple options then using NVDA/Firefox, close the listbox using esc, this will place focus on the input, then navigate to those options pressing shift+tab which will take you to the list of option buttons, then try to navigate the list of options using tab/shift+tab, then try using arrow keys to navigate the selected options.

origami-z:
I don't have time to work on additional fixes. Revert to only fix checkbox

jake-costa:
@chrispcode Would you have time to look at this?

jake-costa (approved):
This fixed Nested interactive aXe flag.

Will create new issue for the 2 other items

Fercas123 (approved):

Issue #3852, opened 2024-07-19
Progression and urgency pattern


navkaur76:
- Revisit the use of progress and status icons across the system
- Document and publish a pattern covering use cases and how statuses are interpreted + communicated with use of colour
- Consider the existing status indication pattern from UITK (figma only)

bhoppers2008:
@amritadesmet @navkaur76 this would be a good one to revisit. Potentially alongside the requirement of a semantic token set that would fit a required asterisk. @dplsek @pseys

mark-tate:
Cayenne Goal: update pattern guidance in Figma and supporting tokens, pattern published to site

pseys:
- Indication Pattern documentation created in Figma and reviewed by @dplsek 
- Tag 'Status' override in Figma documented on Stackoverflow

origami-z:
Indication pattern published on site

pseys:
Once the pattern is published this issue can be closed.

mark-tate:
merged

PR #4883, opened 2025-03-20
Indication pattern

origami-z:
Made some changes to [Figma](https://www.figma.com/design/p6aJv3hER4wRSQrU2poDtj/Indication-Pattern-(DS)?node-id=1-1591&m=dev). Preview [link to page](https://saltdesignsystem-git-3852-indication-pattern-fed-team.vercel.app/salt/patterns/indication).

- Table with very long text doesn't work out nicely on site layout. Breakout usage guidance on top of table
- Does not support 2 colored dots in the table
- Added inline code example as well as storybook (can be removed when general layout be updated)
- Added same theme switch panel from component pages, so icon examples can be toggled between 2 themes

#3852

### DIFF
--- a/packages/core/stories/patterns/indication/indication.stories.tsx
+++ b/packages/core/stories/patterns/indication/indication.stories.tsx
@@ -1,0 +1,117 @@
+import { Button, FlowLayout } from "@salt-ds/core";
+import {
+  ErrorIcon,
+  InfoIcon,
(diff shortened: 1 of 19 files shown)


bpiven (approved):
Edits made, review completed.

pseys (approved):

PR #4886, opened 2025-03-21
Fix component layout on mobile

joshwooding:


site/src/layouts/DetailComponent/DetailComponent.module.css
```diff
@@ -28,3 +28,7 @@
 .overlay {
   max-width: 100vw;
 }
+
+.content {
```
mark-tate, line 32: is it just an issue with this layout ?

mark-tate (approved):
is it just an issue with this layout ?

site/src/layouts/DetailComponent/DetailComponent.module.css
origami-z: is this fixing tablet layout? but it makes desktop content alignment to the left (not aligned to heading section)

origami-z (approved):

mark-tate (approved):

PR #4899, opened 2025-03-25
Slider core design feedback

tayyabataimur:
Actioned the following feedback:

- the white gap should match the thumb width (2px)
- Use spacing-100 here with inline labels
- use alpha-50 (white or black) for track
- the first tick looks funky... the handle seems to be splitting pixels causing a blur and appearing out of alignment. 
- Tick height should not be a hard-coded value
- Improved form field examples


origami-z (dismissed):

.changeset/rotten-chefs-approve.md
```diff
@@ -0,0 +1,10 @@
+---
+"@salt-ds/lab": patch
+---
+
+Improved QA stories and added design updates, including
```
origami-z, line 5: Changeset turns into release not, QA stories isn't something our user would need to know.

origami-z (approved):

.changeset/rotten-chefs-approve.md
tayyabataimur: @origami-z Updated 👍

origami-z (dismissed):

.changeset/rotten-chefs-approve.md
```diff
@@ -0,0 +1,10 @@
+---
+"@salt-ds/lab": patch
+---
+
+Added design updates, including
+
+- white gap on both sides of the `Slider` and `RangeSlider` thumbs.
```
mark-tate, line 7: we already have a changeset for Slider in the Core PR, does this one make sense?

packages/lab/src/__tests__/__e2e__/slider/RangeSlider.cy.tsx
```diff
@@ -389,10 +389,13 @@ describe("Given a Range Slider", () => {
     cy.findByText("Custom Max Label").should("exist");
   });
 
-  it("should be disabled when set", () => {
+  it("should be disabled when set and not receive focus", () => {
```
mark-tate, line 392: NIT... grammar issues
should not receive focus when disabled

packages/lab/src/__tests__/__e2e__/slider/Slider.cy.tsx
```diff
@@ -310,10 +310,12 @@ describe("Given a Slider", () => {
     cy.findByText("Custom Max Label").should("exist");
   });
 
-  it("should be disabled when set", () => {
+  it("should be disabled when set and not receive focus", () => {
```
mark-tate, line 313: NIT (as previously) ... grammar issues
should not receive focus when disabled

.changeset/rotten-chefs-approve.md
tayyabataimur: We made some design changes (e.g. white gap on both sides of the thumb), so if we want to keep track of that? Otherwise I can remove it.

origami-z: You can edit existing changeset file to mention some of these

packages/lab/src/slider/internal/SliderTrack.css
```diff
@@ -104,65 +117,76 @@
...
 
-.saltSliderTrack-withMarkTicks .saltSliderTrack-markTickHidden {
-  visibility: hidden;
+.saltSliderTrack-tick:last-of-type {
+  transform: translateX(-100%);
+}
+
+.saltSliderTrack-marks {
+  position: absolute;
+  user-select: none;
```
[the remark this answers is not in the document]
tayyabataimur: Yes, otherwise the marks get highlighted during dragging.

packages/lab/src/slider/internal/utils.ts
```diff
@@ -50,9 +50,11 @@ export const clamp = (
   }
   // Round to the nearest multiple of the step
   let roundedValue = Math.round(clampedValue / step) * step;
-  // Ensure the rounded value does not exceed max
+  // Ensure the rounded value does not exceed max or min
   if (roundedValue > max) {
```
mark-tate, line 54: NIT could of used Math.min/Math.max

```
const clampedValue = Math.min(Math.max(roundedValue, min), max);
const formattedValue = clampedValue.toFixed(decimalPlaces);
return Number.parseFloat(formattedValue);
```

site/src/examples/range-slider/Disabled.tsx
```diff
@@ -7,6 +7,6 @@ export const Disabled = (): ReactElement => (
     minLabel="0"
     maxLabel="10"
     disabled
-    style={{ width: "400px" }}
+    style={{ width: "600px" }}
```
mark-tate, line 10: what's the impact of this on mobile examples ?

chrispcode (approved):

Fercas123 (approved):

PR #4931, opened 2025-04-02
Carousel docs

Fercas123:


site/docs/components/carousel/accessibility.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+</KeyboardControl>
+<KeyboardControl keyOrCombos="Enter">
+- if focus is on next button, it moves to the next slide and keeps focus within the button.
+- if focus is on previous button, it moves to the previous slide and keeps focus within the button.
+</KeyboardControl>
+
+<KeyboardControl keyOrCombos="Shift + Tab">
+  
+- If focus is outside carousel and carousel is the previous element in the tabbing order, moves focus into the Carousel, setting focus on the first active button.
```
jake-costa, line 28: If focus is on the element after the carousel, Shift-tab should set focus on the last interactive element on the last visible slide or the last visible slide if it doesn't contain any interactive elements

jake-costa (commented):
Added comments as well as a known a11y issue (Not Salt)

jake-costa:
@Fercas123 I think we may want to add these but not sure if it should be on the main tab or the a11y tab, thoughts?

Best practices:
- Use carousel to showcase 2 to 6 slides.
  - Limit the number of slides used in a carousel. Many users engage only with the first couple of slides.
- Avoid using infographics, maps, charts as images or [images of text](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-text-presentation.html).
- Order slides by importance, with the most critical info first. 
  - The first slides receive more exposure than later ones.

mark-tate:
i looked on mobile and could not see what the multi slide example did different to the default example. Perhaps the title should change between slides on the Title example?

mark-tate (approved):

Fercas123:
> i looked on mobile and could not see what the multi slide example did different to the default example. Perhaps the title should change between slides on the Title example?

the example is responsive, so in smaller devices/viewports it would swap to 1 slide, might be worth adding that to the description

PR #4993, opened 2025-04-11
Add search pattern

### LINKED ISSUE #4827: Search pattern
Jess and Len have completed major work, focusing on refining components and restructuring the documentation. 
Aiming to publish the search pattern sticker and documentation on the site by the Cayenne.

honey-chang:
Closes #4827


site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+### 1. Initial interaction
+
+Search field displays with "Search" text, inviting the user to begin their search.
+<Image
+  src="/img/patterns/search/search-pattern-4.png"
+  srcDark="/img/patterns/search/search-pattern-4-dark.png"
+  alt="BSearch pattern layout"
+/>
+
+### 2. Activation the search
```
prashantswami, line 69: Activating the search instead?

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+<Diagram>
+
+### Recent searches
+
+If there are recent searches, a list of past queries appears, providing quick access to previous results.
+
+<Image
+  src="/img/patterns/search/search-pattern-6.png"
+  srcDark="/img/patterns/search/search-pattern-6-dark.png"
+  alt="Example of search with recnet searches in dropdown"
```
prashantswami, line 99: recent?

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+### Layout
+- Use [`Combo box`](/salt/components/combo-box) to display search field.
+- Use Option Group to group the search results.
+- Use Option to display a search result.
+
+<Image
+  src="/img/patterns/search/search-pattern-3.png"
+  srcDark="/img/patterns/search/search-pattern-3-dark.png"
+  alt="BSearch pattern layout"
```
prashantswami, line 54: BSearch is intended?

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+## Basic menu-driven search
+
+### 1. Initial interaction
+
+Search field displays with "Search" text, inviting the user to begin their search.
+<Image
+  src="/img/patterns/search/search-pattern-4.png"
+  srcDark="/img/patterns/search/search-pattern-4-dark.png"
+  alt="BSearch pattern layout"
```
prashantswami, line 66: same here

mark-tate:
This pattern limits the results, it looks like it’s aimed at sites, but  some account searches can return 1000s of results. 

Typically a user types A and you get a large set of results, suppose we limit the results to the recent history or top alphabetical matches. How do we get to the others results starting with A?

Giving a user a 100 results is not an optimal ux so we ask them to type more letters to narrow the results? How do we make the user aware they need to type more, to refine the results?

But maybe the data is the problem, there is no order to the data so a user wants to browse the entire result set?

This creates challenges for performance, which often leads to folks asking for virtualisation which is not a11y, so would suggest we page the results to reduce the need for virtualisation.

How long is it recommended for search to take before the user disconnects and goes somewhere else?

How important is the dataset to a good search experience?

Should you provide a way to clear previous search results ?

How important is for the user to know the size of the possible matches ? 

What happens when a previous result no longer exists or is it relevant? 

What happens if search is not available or an error is returned by the search ?

There is another pattern, think amazon and ebay, where you filter, sort results or group results in categories.

How does keyboard navigation and focus work in this pattern ?

How does screen reader work with this pattern ?

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+
+### Anatomy
+
+A search pattern can be present in an application, as a global or focused search, depending of the scope of the search functionality. Independently of how a search is presented, the following elements are present in a search pattern:
+
+1. **Search input field:** A single input field where users type their queries. 
+2. **Search icon:** The magnifying glass icon indicates search.
+3. **Placeholder text:** Hints to users what information can be searched for. Placeholder text should be clear and tell the user exactly what can be searched for if results are limited. For example, if only page titles can be searched, placeholder text should read “Search page titles.”
+4. **Close button:** Close button that appears within the search input field, allowing users to quickly clear their current query and start a new search.
+5. **Expand button:** An chevron icon button typically used to indicate that a dropdown menu can be expanded or collapsed.
```
amritadesmet, line 36:
```suggestion
5. **Expand button:** A chevron icon button typically used to indicate that a dropdown menu can be expanded or collapsed.
```

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+The following are examples of the types of results items you can create.
+
+<Image
+  src="/img/patterns/search/search-pattern-14.png"
+  srcDark="/img/patterns/search/search-pattern-14-dark.png"
+  alt="Image examples of search result item types"
+/>
+
+
+1. **Basic Search Result:** Displays only the primary search result value, ideal for straightforward results where no additional context is needed.
```
amritadesmet, line 225:
```suggestion
1. **Basic search result:** Displays only the primary search result value, ideal for straightforward results where no additional context is needed.
```

site/docs/patterns/search.mdx
```diff
@@ -0,0 +1,302 @@
...
+  srcDark="/img/patterns/search/search-pattern-17-dark.png"
+  alt="Example of search is located in header"
+/>
+**Center:** Prominent on homepages, used when search is a primary function.
+<Image
+  src="/img/patterns/search/search-pattern-18.png"
+  srcDark="/img/patterns/search/search-pattern-18-dark.png"
+  alt="Example of search is located in center"
+/>
+**Expandable Search:** Starts as an icon button, expanding to a search field or dialog, saving space in minimalist designs.
```
amritadesmet, line 284:
```suggestion
**Expandable search:** Starts as an icon button, expanding to a search field or dialog, saving space in minimalist designs.
```

bhoppers2008:
@mark-tate and I had a quick chat...

> This pattern limits the results, it looks like it’s aimed at sites, but some account searches can return 1000s of results. **We have 1+ use cases in GB that already want to use the pattern, which aren't site related, so no need to call out 'site' as a specific intent. This is just the first version, capturing what we know is technically possible... we'll add to it with other docs including Filtering, Search Results, Handling large results, etc. Separately, we need to document the limitations with our own Combo Box that we can resolve in a Combo Box update.**

> But maybe the data is the problem, there is no order to the data so a user wants to browse the entire result set? **If we know there are technical limitations with our Combo Box regarding number of results... let's capture that on this pattern page. Do we know what max sized dataset we would recommend with this technical solution?**
 
> What happens if search is not available or an error is returned by the search ? **This is a good one to follow with... we have a loading state but no edge or extreme cases yet. We should follow with this**

> How does keyboard navigation and focus work in this pattern ?
> How does screen reader work with this pattern ?
**Presumably both of these would be the same as the current Combo Box implementation? We're not documenting anything here that isn't already possible or documented in the Combo Box component docs.**

honey-chang:
@mark-tate
This version of the search pattern includes only the features that Salt can utilize out of the box, which are currently being used on the Salt site. As a result, there are several limitations. More features, such as a search results page, will be added in future versions.

Typically, a user types A and you get a large set of results, suppose we limit the results to the recent history or top alphabetical matches. How do we get to the others results starting with A? -> **The search results will display recent history first, if applicable, followed by other results. Displaying 100 results in a single dropdown list is not ideal. Ideally, a "View all results" link would be added at the bottom to lead to a results page. However, this is not currently possible as we don't have a pattern for a results page. This is briefly mentioned in the Free Text Search Capability section.**

Giving a user a 100 results is not an optimal ux so we ask them to type more letters to narrow the results? How do we make the user aware they need to type more, to refine the results? -> **There is no indication for users to type more letters, as it is assumed that users will naturally continue typing until they see the desired results. Currently, the Salt site search activates when a user types three letters or more, but this can be adjusted to activate from first letter or two letters.**

But maybe the data is the problem, there is no order to the data so a user wants to browse the entire result set? This creates challenges for performance, which often leads to folks asking for virtualisation which is not a11y, so would suggest we page the results to reduce the need for virtualisation. -> **This can be addressed through the search results page in the next pattern. Paging the results is a more accessible solution than virtualization.**

How long is it recommended for search to take before the user disconnects and goes somewhere else? -> **Not sure about it. @origami-z Any ideas?**

How important is the dataset to a good search experience? -> **Not sure about it.**

Should you provide a way to clear previous search results? -> If it's possible to show a close button when hovering over a recent search result in the current combobox, this is something we can provide. Is it possible?@origami-z  What do you think? @bhoppers2008

What happens if search is not available or an error is returned by the search? -> I'll add this one.

There is another pattern, think amazon and ebay, where you filter, sort results or group results in categories. -> This filtering feature can be considered in the next version.

amritadesmet (approved):

yunjungyeh (approved):

amritadesmet (dismissed):

origami-z (dismissed):

amritadesmet (approved):

origami-z (approved):

PR #5002, opened 2025-04-15
add center alignment to menu items

libertymayc:
If salt-size-base is changed at theme level, menu item text is misaligned

### DIFF
--- a/.changeset/spicy-peas-beam.md
+++ b/.changeset/spicy-peas-beam.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Adds center alignment styling to `MenuItem`
+Added center alignment styling to `MenuItem`
--- a/.changeset/spicy-peas-beam.md
+++ b/.changeset/spicy-peas-beam.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/spicy-peas-beam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Adds center alignment styling to `MenuItem`
```
libertymayc, line 5:
```suggestion
Added center alignment styling to `MenuItem`
```

origami-z (commented):
Is it possible to design a QA story to show this? I felt this needs to be fixed holistically given `size-base` would touch a lot more components? @pseys

packages/core/src/menu/MenuItem.css
```diff
@@ -1,4 +1,5 @@
 .saltMenuItem {
+  align-items: center;
```
joshwooding, line 2: This would not be correct as it doesn't follow the current alignment rules of list items, where iconography and other elements should be aligned to the first line of text.

joshwooding (changes requested):
This change would break the alignment of the component. Also I'm not sure that we support size variables being changed or altered cc @pseys

PR #4965, opened 2025-04-07
Color + Typography foundation fixes

amritadesmet:
Closes #4924

PR #5007, opened 2025-04-16
Show Salt Next characteristics

libertymayc:


origami-z:
This approach feels similar with #4965. Personally, i'd prefer to use a theme switch, something I first tried in #4883.

@amritadesmet @pseys to opine on this, @honey-chang on theme switch toggle on pages (component page has 3 switches, but foundation / theme / pattern pages only needs one)

amritadesmet:
In an ideal world, it'd be great to have all swatches in one page under their respective sections to prevent page jumping. With regards to implementation, I'm not sure. We've previously discussed using Tabs however there are cons to this. Thoughts @honey-chang ? For now, this looks great!

origami-z:
Merging this to the other branch to do further update

PR #5008, opened 2025-04-16
Add to radio button how to use doc about deselection

libertymayc:
UITK mentions that this isn't possible, but I can't see anywhere Salt doc mentions it - has been questioned a few times by my team


origami-z (approved):
This is default HTML `input type="radio"` behavior, but unfortunately i couldn't find specific reference in MDN talk about it. [HTML spec](https://html.spec.whatwg.org/multipage/input.html#radio-button-state-(type=radio)) indirectly states only switch checkness to false when other becomes true

origami-z (approved):

tayyabataimur (approved):

site/docs/components/radio-button/usage.mdx
```diff
@@ -20,6 +20,7 @@ data:
 - When you have fewer than five options and multiple selection is necessary. Instead, use [`CheckboxGroup`](../checkbox).
 - When you have more than five and fewer than 10 options. Instead, use [`Dropdown`](../dropdown).
 - When you have more than 10 options. Instead, use [`ComboBox`](../combo-box).
+- When it should be possible to deselect after a single option has been selected. Instead, consider [`Switch`](../switch), [`Checkbox`](../checkbox) or [`Dropdown`](../dropdown).
```
joshwooding, line 23: Has this been checked by design? Have we thought about encouraging the usage of a reset button instead?

joshwooding (changes requested):

PR #5009, opened 2025-04-17
Site code block color adjustment

origami-z:
- Add background color to both inline code and code block to make it more obvious
- Adjust inline code alignment


libertymayc (approved):

honey-chang (approved):
Looks good

PR #5044, opened 2025-04-30
carousel improvements

Fercas123:
Added passed prop classname to carousel components, 
added interactive example and alter controlled example with side button

### DIFF
--- a/packages/lab/src/carousel/Carousel.tsx
+++ b/packages/lab/src/carousel/Carousel.tsx
@@ -42,6 +42,7 @@ export const Carousel = forwardRef<HTMLDivElement, CarouselProps>(
       activeSlideIndex,
       visibleSlides: visibleSlidesProp = 1,
       children,
+      className,
       controlsPlacement = "top",
       id: idProp,
       ...rest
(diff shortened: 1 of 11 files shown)


packages/lab/src/carousel/Carousel.tsx
```diff
@@ -70,9 +71,13 @@ export const Carousel = forwardRef<HTMLDivElement, CarouselProps>(
       >
         <section
           role="region"
-          className={clsx(withBaseName(), {
-            [withBaseName(controlsPlacement)]: controlsPlacement === "bottom",
-          })}
+          className={clsx(
+            withBaseName(),
+            {
+              [withBaseName(controlsPlacement)]: controlsPlacement === "bottom",
```
libertymayc, line 77: wondering if it's better to add -top classname to the CSS and have

  grid-template-areas: "controls" "slider";
  grid-template-rows: auto 1fr;
  
belong to that class, rather than that being in the root class name and only having the -bottom class to override

mark-tate: makes sense to me as we have to create increased specifity on bottom to help it apply it.

mark-tate:
see related pr 
https://github.com/jpmorganchase/salt-ds/pull/5185

Issue #4806, opened 2025-03-07
Investigate replacing lineHeight values with universal rem value


pseys:
All lineHeight values in Salt are calculated as 1.3x the respective font size. In order to provide greater font size options it would be beneficial to introduce a universal rem value for lineHeight across the design system. 

This will:
- Significantly reduce the number of tokens we need to manage, therefore making a lighter-weight design system and reducing complexity
- Provide greater styling flexibility to consumers e.g. data dense applications could reduce the lineHeight value from 1.3 to 1.2, more experiential UIs may wish to increase it to 1.4 or 1.5. 

Changing lineHeight to use rem is a step closer to providing a more accessible experience across the design system, if the same approach is taken with font size it would allow more control to end users.

mark-tate:
@pseys split into 2 tickets

mark-tate:
Likely we can only do as part of a breaking change

pseys:
Related to #4200

PR #5063, opened 2025-05-07
[DNM] Test pure TD for site

origami-z:
Test whether how far away is pure TD from editorial theme

Most obvious areas need further design

- App header being taller, shall we reduce top/bottom spacing? App header pattern needs to be updated as well
- Home page stacking layout spacing is larger
- We can potentially apply specific override to markdown related line height (e.g. to 1.6-1.8), without associate with a overall theme, before #4806 is addressed in the next breaking change 
- Look at secondary (right) side bar, currently MD override is applied to make it smaller than left primary sidebar


joshwooding:
Closing since this isn't a priority at the moment, we can refer back to it if necessary.

PR #5088, opened 2025-05-16
Trigger onPillRemove callback when a pill is removed in ComboBox

### LINKED ISSUE #5087: ComboBox onPillRemove listener not being called
### Latest version

### Description

In the multi-select combobox that uses Pills, when a Pill is removed, the `onPillRemove` callback is not triggered. 

### Steps to reproduce

In the following example of the ComboBox, when a pill is removed, the `onPillRemove` callback is not triggered.
```
<ComboBox
      {...{
        multiselect: true,
        value: inputValue,
        selected: values,
        valueToString: value => {
          const isTopSources = valuesHaveAllTopSources(value);

          if (isTopSources) {
            return TOP_SOURCE.label;
          } else {
            let vendor;

            if (mode === VENDOR_SPECIFIC) {
              vendor = value.slice(0, 3);
            }

            const index = values.findIndex(_value => _value === value);
            const label = labels[index];

            return [label, vendor].filter(v => v).join(' ');
          }
        },
        onPillRemove: (e, value) => {
          console.log('onPillRemove');
        },
        onChange(e) {
          onChangeInputValue(e.target.value);
        },
        onSelectionChange(e, values) {
          onChangeSelection(values);
        },
        onKeyDown(e) {
          onKeyDown(e.code);
        }
      }}
    >
```

### Expected behavior

_No response_

### Package name(s)

Core (@salt-ds/core)

### Package version(s)

_No response_

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

tayyabataimur:
Closes #5087

### DIFF
--- a/.changeset/smart-brooms-juggle.md
+++ b/.changeset/smart-brooms-juggle.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Added the trigger to `onPillRemove` callback when a pill is clicked and removed in a `ComboBox`.
--- a/.changeset/smart-brooms-juggle.md
+++ b/.changeset/smart-brooms-juggle.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Added the trigger to `onPillRemove` callback when a pill is clicked and removed in a `ComboBox`.
+Fixed `onPillRemove` not being called when a pill is clicked and removed in a `ComboBox`.
--- a/packages/core/src/combo-box/ComboBox.tsx
+++ b/packages/core/src/combo-box/ComboBox.tsx
@@ -88,6 +88,7 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
     valueToString = defaultValueToString,
     truncate,
     bordered = false,
+    onPillRemove,
     ...rest
   } = props;
 
@@ -342,6 +343,7 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
     event.stopPropagation();
     const removed = selectedState[index];
     removePill(event, removed);
+    onPillRemove?.(event, index);
   };
 
   const handleListMouseOver = () => {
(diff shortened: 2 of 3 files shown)


.changeset/smart-brooms-juggle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Added the trigger to `onPillRemove` callback when a pill is clicked and removed in a `ComboBox`.
```
origami-z, line 5: I think this is a "fixed"

origami-z (approved):

.changeset/smart-brooms-juggle.md
origami-z:
```suggestion
Fixed `onPillRemove` not being called when a pill is clicked and removed in a `ComboBox`.
```

tayyabataimur: Done ✅

amritadesmet (approved):

packages/core/src/combo-box/ComboBox.tsx
```diff
@@ -342,6 +343,7 @@ export const ComboBox = forwardRef(function ComboBox<Item>(
     event.stopPropagation();
     const removed = selectedState[index];
     removePill(event, removed);
+    onPillRemove?.(event, index);
```
mark-tate, line 346: @tayyabataimur can we explain the original user's issue with this?
We now have 2 apis which do the same thing onSelectionChange and onPillRemove, one is value driven and one is index driven ?
Also, it's only called when the pill is removed and not if I deselect in the list, so there is a danger of 2 apis returning different values ?

onPillRemove is part of Pill's API, it's bled into the ComboBox API, but probably should of been removed from the ComboBox API to avoid this confusion?

Any thoughts?

mark-tate: ie... onPillRemove is only called when you interact with the pill but another API onSelectionChange is called when the list changes.

PR #5100, opened 2025-05-19
Add tabIndex to step with tooltip story

origami-z:
Make keybaord access to the tooltip. It is just an example we don't document on site...

@jake-costa


jake-costa:
I have an open doc update PR (see [Update A11y Docs For Tooltip #5217](https://github.com/jpmorganchase/salt-ds/pull/5217/files#top))

My pending tooltip-docs PR clarifies that tooltips should sit on interactive elements. Adding `tabIndex` onto a static step just to display a Tooltip goes against that guidance. While you could do this, there are several accessibility reasons why we should not encourage this that is captured in the doc update.

FYI If the steps are links the could add a tooltip to those for that use case.

PR #5107, opened 2025-05-19
[wip] add brand prop to salt provider

libertymayc:


libertymayc:
@origami-z question (as asked also to: @bhoppers2008) - with SaltProviderNext we change e.g. `--salt-palette-negative` to red-600 from red-500 in SaltProvider. is this an update that should apply to both legacy and jpmc brand themes? or is it specific to jpmc brand? this is just an example; question is very broad, i.e. more examples... `--salt-color-background-jet` only provided in SaltProviderNext - is it specific to jpmc?  `--salt-palette-background-primary`, `background` isn't in the palette as defined by original SaltProvider - again is it specific to jpmc or a wider theme update?

origami-z:
> @origami-z question (as asked also to: @bhoppers2008) - with SaltProviderNext we change e.g. `--salt-palette-negative` to red-600 from red-500 in SaltProvider. is this an update that should apply to both legacy and jpmc brand themes? or is it specific to jpmc brand? this is just an example; question is very broad, i.e. more examples... `--salt-color-background-jet` only provided in SaltProviderNext - is it specific to jpmc? `--salt-palette-background-primary`, `background` isn't in the palette as defined by original SaltProvider - again is it specific to jpmc or a wider theme update?

Yes. Color foundations are different, so as the mapping between the 2 themes. All the examples above are specific to brand theme. Another type of example, `color.css` contains colors only for the legacy theme, e.g., `red-50`, which shouldn't be defined for brand theme.

cc @pseys

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -190,6 +198,11 @@ interface SaltProviderBaseProps {
    * A string. Specifies custom theme name(s) you want to apply, similar to `className`.
    */
   theme?: ThemeName;
+  /**
+   * Either "legacy" or "jpmc".
+   * Specifies branding to be used for the theme.
+   */
+  brand?: BrandName;
```
origami-z, line 205: I don't think `jpmc` should appear in Salt code? This is technically jpm not jpmc brand.
Is "legacy" a brand...?

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -435,11 +457,13 @@ interface SaltProviderNextAdditionalProps {
...
 }
 
+/** @deprecated use `SaltProviderProps` with `brand="jpmc"` prop */
 export type SaltProviderNextProps = SaltProviderProps &
   SaltProviderNextAdditionalProps;
-/** @deprecated use `SaltProviderNextProps` */
+/** @deprecated use `SaltProvider` with `brand="jpmc"` prop */
 export type UNSTABLE_SaltProviderNextProps = SaltProviderNextProps;
 
+/** @deprecated use `SaltProvider` with `brand="jpmc"` prop */
```
origami-z, line 466: Let's not deprecate this yet, until everything can be proven working. Don't want to cause panic for teams like Ring / Pepper

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -262,6 +277,7 @@ function InternalSaltProvider({
   const density = densityProp ?? inheritedDensity ?? DEFAULT_DENSITY;
   const themeName =
     themeProp ?? (inheritedTheme === "" ? DEFAULT_THEME_NAME : inheritedTheme);
+  const brandName = brandProp ?? inheritedBrand ?? DEFAULT_BRAND;
   const mode = modeProp ?? inheritedMode;
   const breakpoints = breakpointsProp ?? DEFAULT_BREAKPOINTS;
   const corner = cornerProp ?? inheritedCorner ?? DEFAULT_CORNER;
```
origami-z, line 283: (I haven't put too much thought into this.) Will styling option inheritance cause any issue, when nest provider between brand and legacy?

packages/core/stories/salt-provider/salt-provider.stories.tsx
```diff
@@ -30,7 +33,67 @@ export const Default = () => {
...
+              <ToggleButton aria-label="dark theme" value="dark">
+                Dark
+              </ToggleButton>
+            </ToggleButtonGroup>
+          </FlexLayout>
+          <p>{`This Card is wrapped with a SaltProvider, brand is ${brand}, mode is ${mode}`}</p>
+
+          <Checkbox label="Example Choice 1" />
+          <Checkbox defaultChecked label="Example Choice 2" />
+          <Checkbox defaultChecked indeterminate label="Example Choice 3" />
```
origami-z, line 75: When clicking this on jpmc brand, it shows different color. Not sure whether there is a problem

packages/core/src/salt-provider/SaltProvider.tsx
libertymayc: np! I just did this as temporary, without worrying too much about prop name

packages/core/src/salt-provider/SaltProvider.tsx
libertymayc: thanks, will investigate this

packages/core/stories/salt-provider/salt-provider.stories.tsx
libertymayc: looking at the Figma, I think it's wrong - https://github.com/jpmorganchase/salt-ds/pull/5117

packages/theme/css/palette/text.css
libertymayc, line 1: Thought: `.salt-theme` class e.g. here without the data-brand attribute would mean these tokens are default (for example   `--salt-palette-text-fontFamily: var(--salt-typography-fontFamily-openSans);`) - is that true and would it make sense for future iterations? 

technically speaking, they could be defined under both `.salt-theme[data-brand="uitk"]` and `.salt-theme[data-brand="jpm"]`, but I can see that, in terms of backwards compatibility, this may cause confusion and issues

origami-z: Will there be a situation where an old version of theme package be used with newer version of core package? If not, CSS rules would be reflective of SaltProvider code.

mark-tate:
@pseys we need a meeting, so this can progress

packages/theme/scripts/build.mjs
```diff
@@ -24,9 +24,11 @@ const context = await esbuild.context({
   absWorkingDir: cwd,
   entryPoints: [
     "index.css",
-    "css/theme.css",
     "css/global.css",
-    "css/theme-next.css",
+    "css/deprecated/index.css",
```
joshwooding: I think this needs to be reverted, with the new ones added? Not sure if deprecated needs to be an entry

packages/theme/index.css
```diff
@@ -1,2 +1,4 @@
 @import url(css/global.css);
-@import url(css/theme.css);
+
+@import url(css/foundations/index.css);
+@import url(css/deprecated/index.css);
```
joshwooding, line 3: Shouldn't uitk be imported here?

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -274,6 +300,9 @@ function InternalSaltProvider({
   const applyClassesTo =
     applyClassesToProp ?? (isRootProvider ? "root" : "scope");
 
+  /* If brand prop is provided as something other than legacy (default), don't apply the deprecated themeNextProps */
+  const applyThemeNextProps = themeNext && brand === "uitk";
```
libertymayc: @joshwooding @pseys this should solve breaking change issues for styling props since they're only applied for the next provider. anyone with an existing `SaltProviderNext` it won't break for since "uitk" is the default brand and so the salt-theme-next class and styling attributes are applied, if they decide to explicitly add "uitk" as the brand and they're confused that it doesn't look how it should we can document (if needed) that they need to get rid of the deprecated data-attributes for it to work, and for anyone using the new brand prop with other values, they should anyway be comfortable to remove at the same time their styling props

PR #5117, opened 2025-05-20
Fix checked indeterminate Checkbox styling 

libertymayc:
When a checkbox has checked/defaultChecked set to true but is indeterminate, blue color styles are overriding the what should be gray (according to Figma spec) styles for an indeterminate checkbox as noticed here by @origami-z  https://storybook.saltdesignsystem.com/?path=/story/core-salt-provider--toggle-theme&globals=density:low 

Before

image

After

image

### DIFF
--- a/.changeset/rich-walls-chew.md
+++ b/.changeset/rich-walls-chew.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix `CheckboxIcon` checked color styles applying when checkbox is `indeterminate`
--- a/.changeset/rich-walls-chew.md
+++ b/.changeset/rich-walls-chew.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fix `CheckboxIcon` checked color styles applying when checkbox is `indeterminate`
+Fixed `Checkbox`'s checked color styles applying when `Checkbox` is indeterminate.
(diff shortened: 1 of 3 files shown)


packages/core/src/checkbox/CheckboxIcon.tsx
```diff
@@ -68,7 +68,7 @@ export const CheckboxIcon = ({
       className={clsx(
         withBaseName(),
         {
-          [withBaseName("checked")]: checked,
+          [withBaseName("checked")]: checked && !indeterminate,
```
joshwooding, line 71: I don't think this fix is correct, this would make it harder to apply a different style if you wanted to. Can we fix this by adjusting the CSS specificity please

joshwooding (changes requested):

packages/core/src/checkbox/CheckboxIcon.tsx
libertymayc: yes, it was my first thought to use CSS but i wasn't sure on the standard for this type of specificity, let me know if there's something preferred than :not selector

joshwooding (changes requested):
See comment

.changeset/rich-walls-chew.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix `CheckboxIcon` checked color styles applying when checkbox is `indeterminate`
```
joshwooding, line 5:
```suggestion
Fixed `Checkbox`'s checked color styles applying when `Checkbox` is indeterminate.
```

PR #5124, opened 2025-05-21
Add SID input in issue template

origami-z:
This should help us reach out. Currently this is a checkbox, but there is no way to track back who submitted the issue if the checkbox is ticked.


joshwooding:
I'm not sure this is allowed, SID have previously been treated as internal only

joshwooding (changes requested):
This needs to be checked for compliance

PR #5128, opened 2025-05-22
Upgrade to Typescript 5

joshwooding:


origami-z:
Is this a bumping requirement for our users? What if they can't upgrade yet...

joshwooding:
> Is this a bumping requirement for our users? What if they can't upgrade yet...

It doesn't change anything for users, it only affects our repo and unblocks tooling support where TS v5 is required

origami-z:
> > Is this a bumping requirement for our users? What if they can't upgrade yet...
> 
> It doesn't change anything for users, it only affects our repo and unblocks tooling support where TS v5 is required

Interesting, I thought we'd have to do [`typesVersions`](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions), in case anyone is still using TS v4

origami-z (approved):

PR #4978, opened 2025-04-09
Group slider docs and add change events example

tayyabataimur:
- Grouped slider docs on the site.
- Added examples for change events.

PR #5147, opened 2025-05-28
Date picker site pages split to group

origami-z:
- Split date picker to 2 pages
- Move localization provider to a top callout


tayyabataimur (commented):
Maybe not for this PR but the `RC` label appears on the `DatePicker` group title and the `Range Date Picker` title in the sidebar. I experienced the same with the Slider when it was in labs. A nicer UI would be to just have the `RC` label on the `DatePicker` group title in the side bar and not for its child items since it would already be implied that they are RCs. The example pages would still have the RC label so that would be fine.

Here's the Slider PR where I implemented this: https://github.com/jpmorganchase/salt-ds/pull/4978/files

origami-z:
> Maybe not for this PR but the `RC` label appears on the `DatePicker` group title and the `Range Date Picker` title in the sidebar. I experienced the same with the Slider when it was in labs. A nicer UI would be to just have the `RC` label on the `DatePicker` group title in the side bar and not for its child items since it would already be implied that they are RCs. The example pages would still have the RC label so that would be fine.
> 
> Here's the Slider PR where I implemented this: https://github.com/jpmorganchase/salt-ds/pull/4978/files
> 
> Link to deployment to see it: https://saltdesignsystem-git-slider-site-fed-team.vercel.app/salt/components/slider/examples

I removed the lab package from range pick doc. It's mixed problem, now the range picker page doesn't have "Release Candidate" on top

tayyabataimur:
Just saw the latest deployment and it seems to be perfect now!

tayyabataimur (approved):

PR #5151, opened 2025-05-29
Rename splitter page

origami-z:
rename `/salt/components/react-resizable-panels-theme` to `/salt/components/splitter`


libertymayc:
does it need a redirect added for existing bookmarks etc?

mark-tate:
> > does it need a redirect added for existing bookmarks etc?
> 
> It's unclear what should be done with splitter page. Some suggests rename route, some suggests redirects. Not up to me to decide at this point.

redirects are hard to maintain.. you never remove them and eventually get redirects of redirects.

@honey-chang an IA change here could enable us to support the different types of components.
ie. Once we make this change, what should happen to the AG Grid theme?

Would any changes to the page anatomy, help differentiate between a themed component and implementation?
Perhaps we use the tag component and have some form of categorisation of components ?

bpiven:
Here's what I would propose:

1) Kindly approve this PR that @origami-z initiated to restore the original Splitter URL, thus fixing broken link and nav

2) Per point made by @libertymayc, yes we should have a redirect from RRP for anyone who bookmarked

3) Excellent suggestion @mark-tate @honey-chang to do longer-term IA work around subset(s) of components that are grids / layouts / themes / styles

joshwooding (approved):

PR #5179, opened 2025-06-06
Initial Collapsible work

joshwooding:


packages/lab/src/collapsible/Collapsible.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { makePrefixer, useControlled } from "@salt-ds/core";
+import { CollapsibleContext } from "./CollapsibleContext";
+import { clsx } from "clsx";
+
+export interface CollapsibleProps
+  extends Omit<ComponentPropsWithoutRef<"div">, "onToggle"> {
+  /**
+   * Whether the accordion is expanded.
+   */
+  expanded?: boolean;
```
mark-tate, line 17: NIT: it's called a Collapsible and it has expanded props ?

wondered if `open` aligned more with other components ?

Issue #5216, opened 2025-06-11
Update Tooltip Docs (to add info on non-interactive elements)


jake-costa:
### Related page(s)

https://www.saltdesignsystem.com

### Kind of issue

Missing information

### Description

We need to update our tooltip documentation to provide guidance on the interactivity of the Tooltip's anchor to address questions and use cases where devs are trying to add a tooltip to an non-interactive element.

Information that needs to be added:
Clarification on a Tooltip's anchor and when to use
Alternatives 

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #5217, opened 2025-06-11
Update A11y Docs For Tooltip

jake-costa:
We need to update our tooltip documentation to provide guidance on the interactivity of the Tooltip's anchor to address questions and use cases where devs are trying to add a tooltip to an non-interactive element.

Information that needs to be added:
Clarification on a Tooltip's anchor and when to use
Alternatives

Closes: https://github.com/jpmorganchase/salt-ds/issues/5216

### DIFF
--- a/site/docs/components/index.mdx
+++ b/site/docs/components/index.mdx
@@ -18,6 +18,10 @@ in Figma when logged in to the JPMC network.
 They include accessibility and design specifications for each component,
 with designs for light and dark modes in all four densities.
 
+Please note: Digital accessibility is a shared responsibility and while our components include standard accessibility
+behavior and attributes, product teams are responsible for implementing components
+correctly and ensuring their digital products meet all current accessibility requirements.
+
 ## Lab components
 
 Stable components do not feature an accompanying symbol in their navigation link titles. You can find them in the `@salt-ds/core` package.
--- a/site/docs/components/tooltip/accessibility.mdx
+++ b/site/docs/components/tooltip/accessibility.mdx
@@ -10,4 +10,45 @@ data:
 
 ## Best practices
 
-The tooltip doesn't receive focus, so we recommend that you don't include interactive elements as the user can't access them.
+- Keep the use of tooltips to a minimum and use only when providing extra (non-essential) information for the user.
+- Focus does not move to the tooltip. Since a tooltip does not receive focus, do not include interactive elements in the tooltip as the user cannot access them.
+- Only one tooltip should be displayed at a time.
+- Keep the text for a tooltip short and concise.
+- Avoid rich content such as bold text, italics, headings, icons, etc. as these will not be conveyed through `aria-describedby`.
+- Do not use a timeout to hide the tooltip.
+
+## Anti-Pattern
+
+Avoid applying a tooltip to non-interactive elements. As a rule, non-operable elements should not receive keyboard focus to avoid confusion and unnecessary navigation steps.
+A common anti-pattern is to use an (i) icon (or another non-interative element) as the tooltip anchor however this is not recommended for the following reasons:
+
+- Icons (and other non-interative elements) do not receive focus by default.
+- Adding `tabindex` to make the icon focusable introduces an additional tab stop.
+- An (i) icon or non-interative element will not provide an appropriate programmatic association to what the tooltip is visually associated with. These tooltips become an orphaned note that screen readers announce out of context.
+- Screen readers are designed to announce tooltips when a user focuses on an interactive element. When a tooltip is attached to a static element (even using `tabindex`) Screen readers may not reliably announce its content as it is not part of the expected interaction model.
+
+Note: When the content of a tooltip is not about an interactive UI element, consider adding the message directly on the page or use an [`Overlay`](../overlay), [`Dialog`](../dialog), or [`Banner`](../banner) instead.
+
+## Mobile Consideration
+
+Standard tooltips do not display on touch-only devices when attached to [`Links`](../link), [`Buttons`](../button) or non-interative elements and there is currently no workaround. That said, tooltips should display when attached to a [`Form Field`](../form-field).
+
+## Keyboard interactions
+
+<KeyboardControls>
+  <KeyboardControl keyOrCombos="Tab">
+
+    - Tooltips do not receive focus, so Tab does not move focus to the tooltip. Instead, Tab moves focus to the tooltip trigger element to display the tooltip.
+
+  </KeyboardControl>
+  <KeyboardControl keyOrCombos="Escape">
+
+    - If the tooltip is displayed, pressing Escape dismisses the tooltip.
+
+  </KeyboardControl>
+</KeyboardControls>
+
+## Resources
+
+- [ARIA Tooltip Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip/)
+- [WCAG 1.4.13 Content on Hover or Focus](https://www.w3.org/TR/WCAG21/#content-on-hover-or-focus)
--- a/site/docs/components/tooltip/examples.mdx
+++ b/site/docs/components/tooltip/examples.mdx
@@ -10,9 +10,13 @@ data:
 
 ## Default
 
-By default, `Tooltip` displays a message and an arrow pointing to the relevant UI element. Its default placement is "right," and appears after 300 milliseconds.
+By default, `Tooltip` displays a message and an arrow pointing to the relevant UI element. Its default placement is "right," and appears after 300 milliseconds. Use the `content` prop to pass a simple string to display in the tooltip.
 
-Use the `content` prop to pass a simple string to display in the tooltip.
+Tooltips are triggered by both focus and hover events and remain visible and persistent until:
+
+- Hover or focus is removed from the trigger element
+- The user manually dismisses it (e.g.`Esc` key)
+- The content is no longer valid
 
 <LivePreview componentName="tooltip" exampleName="Default" />
 
--- a/site/docs/components/tooltip/index.mdx
+++ b/site/docs/components/tooltip/index.mdx
@@ -1,7 +1,7 @@
 ---
 title: Tooltip
 data:
-  description: "`Tooltip` displays a brief message to the user that provides additional information about a UI element. The tooltip appears after the user’s mouse pointer hovers over the target element for a certain time.  It can communicate new or supporting information, errors, warnings or successful completion of a process or task."
+  description: "`Tooltip` is a non-modal overlay, initially hidden, that appears on hover or focus. It provides concise, text-only content related to the associated interactive element, typically offering supplementary (non-essential) information without disturbing the users workflow."
   sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/core/src/tooltip"
   package:
     name: "@salt-ds/core"
--- a/site/docs/components/tooltip/usage.mdx
+++ b/site/docs/components/tooltip/usage.mdx
@@ -12,7 +12,7 @@ data:
 
 ### When to use
 
-- When you want to display a brief message to the user that provides additional information about a UI element.
+- When you want to display a brief message to the user that provides additional information about an interactive UI element.
 
 ### When not to use
 


site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,22 @@ data:
...
 ## Best practices
+Tooltips must be triggered by both focus and hover events and remains on the screen as long as the trigger has the focus. The focus does not move to the tooltip.
+Since a tooltip doesn't receive focus, don't include interactive elements as the user can't access them.
+The expand/collapsed state is intentionally not exposed
+
+
+## Anti-Pattern
+Never apply a tooltip to non-interactive elements. As a rule, non-operable elements should not receive keyboard focus to avoid unnecessary navigation steps.
+A common design pattern is to use an (i) icon as the tooltip anchor however this is not recommended for the following reasons:
+* Icons do not recieve focus by default
```
tycup: Tiny spelling error: receive*

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,22 @@ data:
   $ref: ./#/data
 ---
 
+## What is a tooltip
+A tooltip should only be used to provide extra (non-essential) information about a form field, a link, a button, or other focusable element.
```
mark-tate: should  they be links to the component ?
``` about a form field, a link, a button, ```

File: site/docs/components/tooltip/accessibility.mdx [hunk 3]
```diff
@@ -8,6 +8,22 @@ data:
...
+Tooltips must be triggered by both focus and hover events and remains on the screen as long as the trigger has the focus. The focus does not move to the tooltip.
+Since a tooltip doesn't receive focus, don't include interactive elements as the user can't access them.
+The expand/collapsed state is intentionally not exposed
+
+
+## Anti-Pattern
+Never apply a tooltip to non-interactive elements. As a rule, non-operable elements should not receive keyboard focus to avoid unnecessary navigation steps.
+A common design pattern is to use an (i) icon as the tooltip anchor however this is not recommended for the following reasons:
+* Icons do not recieve focus by default
+* Adding tabindex to make the icon focusable introduces an addtional tab stop
```
mark-tate: tabindex is an attribute should it be in backticks `tabindex`

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,22 @@ data:
...
+
 ## Best practices
+Tooltips must be triggered by both focus and hover events and remains on the screen as long as the trigger has the focus. The focus does not move to the tooltip.
+Since a tooltip doesn't receive focus, don't include interactive elements as the user can't access them.
+The expand/collapsed state is intentionally not exposed
+
+
+## Anti-Pattern
+Never apply a tooltip to non-interactive elements. As a rule, non-operable elements should not receive keyboard focus to avoid unnecessary navigation steps.
+A common design pattern is to use an (i) icon as the tooltip anchor however this is not recommended for the following reasons:
```
mark-tate: Should this read: A common anti pattern

site/docs/components/tooltip/accessibility.mdx
mark-tate, line 1: any value in you reviewing the APIs
```
export interface UseTooltipProps
  extends Partial<
    Pick<UseFloatingUIProps, "onOpenChange" | "open" | "placement">
  > {
  /**
   * When `true`, the tooltip will be disabled.
   */
  disabled?: boolean;
  /**
   * Do not respond to focus events.
   */
  disableFocusListener?: boolean;
  /**
   * Do not respond to hover events.
   */
  disableHoverListener?: boolean;
  /**
   * The number of milliseconds to wait before showing the tooltip.
   * This prop won't impact the enter touch delay (`enterTouchDelay`).
   */
  enterDelay?: number;
  /**
   * The number of milliseconds to wait before hiding the tooltip.
   * This prop won't impact the leave touch delay (`leaveTouchDelay`).
   */
  leaveDelay?: number;
}
```
any comments needed on the enter or leave delay ?

Defaults are currently (300) enter and leave is 0

jake-costa: @mark-tate Thanks for bringing this up.

There is no guidance in the wcag regarding the delay on showing a tool tip however I would highly recommend that it be extremely short. I'll do some research to see if the w3c has any recommendations on this.

As for the auto dismiss, Success criteria 1.4.13 states for content that becomes visible on focus or hover (e.g. tooltip), the additional content must remain visible until hover or focus is removed, the user has dismissed it, or its information is no longer valid. Given this we should not allow the tooltip to automatically dismiss (leave) after a period of time. Having it auto dismiss after a period of time would be a violation.

joshwooding: This is based off our duration foundation btw, so if we change anything it could have a ripple effect. Also there isn't an auto dismiss functionality for Tooltip so we should be fine. The props control the delay before showing and the delay on hiding once you remove the trigger action e.g. mouse off

tycup (approved):

pseys (dismissed):

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
   $ref: ./#/data
 ---
 
+Digital accessibility is a shared responsibility and while the standard behavior and attributes for a tooltip are already handled for you as part of our component, product teams are responsible for implementing the component correctly and ensuring their digital products meet all current accessibility requirements.
```
joshwooding: It's not necessarily wrong but we typically don't introduce our accessibility docs like this

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+
+## What is a tooltip
+A tooltip is a non-modal overlay that is hidden by default and displays text-only content on hover or focus. It should only be used to provide extra (non-essential) information about a [`Form Field`](../form-field), [`Link`](../link), [`Button`](../button), or other interactive element.
```
joshwooding: This is arguably the description shown on  the component page, we could update that instead of providing a description here

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+
 ## Best practices
+- Tooltips are triggered by both focus and hover events and remain visible and persistent until:
```
joshwooding: Not sure this is a best practice? It's describing the behaviour Tooltip already has and doesn't need configuration to achieve

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+  - Hover or focus is removed from the trigger element
+  - The user manually dismisses it (e.g.`Esc` key)
+  - The content is no longer valid 
+- Focus does not move to the tooltip. Since a tooltip does not receive focus, do not include interactive elements in the tooltip as the user cannot access them.
+- Only one tooltip should be displayed at a time.
```
joshwooding, line 15: IMO "best practices" start here

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+- Keep the text for a tooltip short and concise.
+- Avoid rich content such as bold text, italics, headings, icons, etc. as these will not be conveyed through `aria-describedby`.
+- Do not put essential information in tooltips.
+- Do not use a timeout to hide the tooltip.
+- The expand/collapsed state is intentionally not exposed.
```
joshwooding: Not sure what this means?

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+- Keep the use of tooltips to a minimum and use only when providing extra (non-essential) information for the user.
```
joshwooding: This is similar to an above point:

"Do not put essential information in tooltips."

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+- When the content of a tooltip is not about an interactive UI element, consider adding the message directly on the page or using [`Overlay`](../overlay), [`Dialog`](../dialog), or [`Banner`](../banner) instead.
```
joshwooding: Is this the same as the anti-pattern below?

site/docs/components/tooltip/accessibility.mdx
```diff
@@ -8,6 +8,56 @@ data:
...
+A common anti-pattern is to use an (i) icon (non-interative element) as the tooltip anchor however this is not recommended for the following reasons:
+
+- Icons (and other non-interative elements) do not receive focus by default.
+- Adding `tabindex` to make the icon focusable introduces an additional tab stop.
+- An (i) icon or non-interative element will not provide an appropriate programmatic association to what the tooltip is visually associated with. These tooltips become an orphaned note that screen readers announce out of context.
+- Screen readers are designed to announce tooltips when a user focuses on an interactive element. When a tooltip is attached to a static element (even using `tabindex`) Screen readers may not reliably announce its content as it is not part of the expected interaction model.
+
+
+## Mobile Consideration
+Standard tooltips do not display on touch-only devices when attached to [`Links`](../link), [`Buttons`](../button) or non-interative elements and there is currently no workaround. That said, tooltips should display when attached to a [`Form Field`](../form-field).
```
joshwooding, line 34: Tested this and our tooltips show on mobile, although I don't think the interaction is consistent and would work in a realistic example e.g. performing an action

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding I have been on multiple discussions in which people think that using salt components alone means the application will be accessible and figured that adding this to the beginning of the accessibility documentation might help better convey that especially considering we do not have our accessibility statement on our public site. If you have any suggestions on changes or would rather we remove it, let me know. Though I do think that it would be good to add.

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding that's fine with me. Are you thinking in the "how to use" section or the text below the Tooltip heading?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding  while it is the behavior the tool tip already has, I do think it is good for this documentation to be informative for those that might not be familiar with why. In addition, this documentation might be important for those that are thinking about overriding it as well. Are you thinking of moving it or removing it?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding the first two points are related to the item above. Thoughts?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding can you clarify what you are unsure on? Is it all of these points?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding  yeah, good call

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding yes, they are similar however, for the anti pattern, my goal was to specifically highlight the common issue where a tool tip is applied to a non-interactive element. That said, I do think the last point could go in either section. Would you prefer it be moved to the anti-pattern section below?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding As you mentioned, the examples that we are using on the site are just buttons that don't have an action so on a mobile device, they are being displayed, however if a tool tip is being added to a button or a link that does have an action, the user would trigger the action instead of display the tool tip. I have tried using talk back and focusing on the buttons, and the tool tip is not being displayed either. I do think it would be better if we updated our examples so that we can accurately test this and maybe even include some form field examples to show other use cases. Thoughts?

site/docs/components/tooltip/accessibility.mdx
jake-costa: Maybe we can include it on the components overview page instead? That way it wouldn't have to be added it to all of the components, which might be a little cleaner and more appropriate. Thoughts?

joshwooding: I'm happy to add it to a more centralised place , we should revisit if we need an accessibility statement page on the site

site/docs/components/tooltip/accessibility.mdx
joshwooding: If we want it for people who override the behaviour we should include it in the example where we do so.

site/docs/components/tooltip/accessibility.mdx
joshwooding: line 27, is this referring to aria-expanded not being used? Because developers can tell if the tooltip is open or hidden so this might cause confusion

site/docs/components/tooltip/accessibility.mdx
joshwooding: I meant line 22. GitHub shows previous lines for context when you comment

site/docs/components/tooltip/accessibility.mdx
joshwooding: We typically only show examples directly related to the component on the same page, it might be worth revisiting the idea but not sure if we can do anything about that now. FormField and Tooltip is on the FormField page. We can add Storybook examples to test the functionality though

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding correct it was to clarify that we are not using aria-expanded as some developers are adding it. I did remove it from the last round of changes, but do you think it might be better to clearly state that I'm referring to aria-expanded?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding Ok, do you have an idea where that would be? Maybe the first example "default"?

site/docs/components/tooltip/accessibility.mdx
jake-costa: @joshwooding I did move it to the component overview page. Let me know what you think? And I definitely think we should include an accessibility statement page on the site and believe that we plan on it. We were just having trouble figuring out exactly where it should go.

site/docs/components/tooltip/accessibility.mdx
joshwooding: I think it's fine as it is, I'm surprised some developers are adding it. Are they being told to?

site/docs/components/tooltip/accessibility.mdx
joshwooding: I think we could add one, we have the follow functionality already:

disableHoverListener  | Option to remove the hover listener. 
disableFocusListener  | Option to remove the focus listener.

But we don't expose it on the site. Are you happy for us to add it?

jake-costa: Is there a reason why we allow the ability to disable the hover or focus listener for a tooltip? I believe, adding either of these would result in an accessibility violation for WCAG and wouldn't align with the guidance in the APG. Maybe there's a use case that I'm not considering?

joshwooding: I'm not sure of any usecases apart from making it only show on click. It was added because it was previously a feature

joshwooding: > In addition, this documentation might be important for those that are thinking about overriding it as well

What were you thinking of when you said this?

jake-costa: I was thinking of users that might try to disableHoverListener or disableFocusListener. But, wasn't aware we offered these either. I do not think we should offer the ability to disable either of these. If they want to use a tooltip both need to be enabled.

jake-costa:
Pending review/feedback

joshwooding (dismissed):

site/docs/components/index.mdx
```diff
@@ -18,6 +18,10 @@ in Figma when logged in to the JPMC network.
 They include accessibility and design specifications for each component,
 with designs for light and dark modes in all four densities.
 
+Please note: Digital accessibility is a shared responsibility and while our components include standard accessibility
```
joshwooding, line 21: We can refine this later

joshwooding (approved):

site/docs/components/tooltip/accessibility.mdx
mark-tate, line 1: can we review the examples and how it correlates to this page, there feels like a disconnect ?

Our page says:
Tooltip displays a brief message to the user that provides additional information about a UI element. The tooltip appears after the user’s mouse pointer hovers over the target element for a certain time.

- our docs talk about tooltips appearing for a finite time  and we give 3 examples of different times?
What is the a11y view of this and timing ?

Any guidance needed on:
- on positioning and strategies for flipping ?
- should care be taken when not using arrows ?
- Keep the text for a tooltip short and concise.  is there a max length of tooltip text or language to use within tooltips ?

- this says avoid rich content such as bold text, italics, headings, icons, etc 

but our examples use bulleted lists and icons ?

joshwooding: Our example is not correct (content). I think we can address this feedback after

joshwooding: https://github.com/jpmorganchase/salt-ds/pull/5301

mark-tate (approved):

joshwooding (approved):

PR #5226, opened 2025-06-16
Upgrade Tooltip's max-width

### LINKED ISSUE #3694: Update Tooltip max width to follow content guide
Tooltip uses a fixed 230px value for max-width, which isn't density aware

https://github.com/jpmorganchase/salt-ds/blob/baa5aaa6e8443cae49ce08d1d62409bb184635b9/packages/core/src/tooltip/Tooltip.css#L22

Update tooltip to use 60 characters max width per line, to follow content guideline.

Figma - https://www.figma.com/design/0qHqLDEobXhRGb1PzQ0I5q/Tooltip?node-id=2048-4137&t=O1JAUiQpX4DUu9EJ-1

Reference - https://stackoverflow.com/a/48858102

joshwooding:
Closes #3694

### DIFF
--- a/packages/core/src/dialog/DialogHeader.css
+++ b/packages/core/src/dialog/DialogHeader.css
@@ -10,7 +10,6 @@
 }
 .saltDialogHeader-header {
   margin: 0;
-  overflow-wrap: anywhere;
 }
 
 .saltDialogHeader-container {


packages/core/src/dialog/DialogHeader.css
```diff
@@ -10,6 +10,7 @@
 }
 .saltDialogHeader-header {
   margin: 0;
+  overflow-wrap: anywhere;
```
libertymayc: should this be in this pr/does it need cs?

joshwooding: > should this be in this pr/does it need cs?

Ah, shouldn't be included

Issue #3791, opened 2024-07-10
Hide Tooltip on click


joshwooding:
### Area

UI Components

### The problem

If a user clicks they already know the intent of the button and probably don't need the tooltip.

### The solution

This is possible to do with the controlled API but, we should consider if this should be default behaviour.

### Alternatives and examples

.

### Are you a JPMorgan Chase & Co. employee?

- [ ] I am an employee of JPMorgan Chase & Co.

PR #5252, opened 2025-06-20
Fix Tooltip not working with MenuTrigger and OverlayTrigger

joshwooding:


libertymayc (approved):

packages/core/src/overlay/OverlayTrigger.tsx
```diff
@@ -1,28 +1,40 @@
...
 
-  const { reference, getReferenceProps } = useOverlayContext();
+    const { reference, getReferenceProps } = useOverlayContext();
 
-  const triggerRef = useForkRef(getRefFromChildren(children), reference);
+    const handleFloatingRef = useForkRef(
+      getRefFromChildren(children),
+      reference,
+    );
+    const handleRef = useForkRef(handleFloatingRef, ref);
```
tayyabataimur, line 25: For my understanding, why does this reassignment of ref fix the issue? 🤔

tayyabataimur (approved):

packages/core/src/overlay/OverlayTrigger.tsx
joshwooding: Floating UI uses refs to attach events, OverlayTrigger and MenuTrigger did not support ref forwarding but now they do so floating ui works

packages/core/stories/overlay/overlay.stories.tsx
```diff
@@ -337,3 +337,28 @@ export const WithActions = ({ onOpenChange }: OverlayProps) => {
     </Overlay>
   );
 };
+
+export const WithTooltip: StoryFn<OverlayProps> = ({ ...args }) => {
```
mark-tate, line 341: @pseys is it an issue for the user that overlay and tooltip look somewhat identical?
have a look at this example in Storybook, to verify

joshwooding: @dplsek

mark-tate: Perhaps, once an overlay is displayed, then you do not need the tooltip anymore ?
Seems strange (perhaps) to have both ?

joshwooding: > Perhaps, once an overlay is displayed, then you do not need the tooltip anymore ? Seems strange (perhaps) to have both ?

@mark-tate I've got an issue to solve that: https://github.com/jpmorganchase/salt-ds/issues/3791

packages/core/src/menu/MenuTrigger.tsx
```diff
@@ -11,39 +17,42 @@ export interface MenuTriggerProps {
...
+      getRefFromChildren(children),
+      refs?.setReference,
+    );
+    const handleRef = useForkRef(handleFloatingRef, ref);
 
-  if (!children || !isValidElement<{ ref?: Ref<unknown> }>(children)) {
-    // Should we log or throw error?
-    return <>{children}</>;
-  }
+    if (!children || !isValidElement<{ ref?: Ref<unknown> }>(children)) {
```
mark-tate, line 34: NIT 
 * The trigger element for the menu. This should be a single child element and accept a ref.
 
 We cover in the props that the child element should receive a ref... Wonder if that is enough and needs to be in usage  docs ?

mark-tate (approved):

packages/core/stories/overlay/overlay.stories.tsx
dplsek: i'm not concerned that the components look the same since they have different and distinct uses cases... no different than system status and banner imo. however, i agree that showing both simultaneously doesn't make sense. once the button is clicked/toggled into the active state, i would assume the tool-tip wouldn't appear until the button reverts back to default state. also, this example is weak. we should use something that clearly shows the difference between the two components since it's not the appearance. perhaps change the label in the tool-tip to "Click to show content"... this helps to imply that tool-tips appear on hover and help aid a user about an action. and i'd use the same overlay (or another realistic example) like we did for 'With Actions'. thoughts?

PR #5277, opened 2025-06-27
Move withDataMock to stories that require it

tycup:
Highcharts internally uses Date to manage animations. Having this decorator globally prevents it from working as expected.


joshwooding (approved):

joshwooding (changes requested):
Hmm, something isn't working. The QA stories aren't being mocked

.storybook/preview.tsx
```diff
@@ -232,7 +231,6 @@ const preview: Preview = {
     withTheme,
     withLocalization,
```
joshwooding, line 232: I think because this is higher up in the tree than withDateMock now, things aren't being mocked

joshwooding (approved):

PR #5284, opened 2025-07-01
Deprecate target background and border color

libertymayc:


pseys:
@libertymayc following review of this @joshwooding and @dplsek have both raised concerns about removing the Characteristic. The consensus of opinion is that there are several requirements on the horizon to include drag/drop functionality to several components. There is evidence of these in some BPLs e.g. Tree, List. The tokens we have may not be appropriate but the feeling was that we should leave them as is for now so that they can be extended if necessary. Can we cancel this PR please?

PR #5287, opened 2025-07-01
Disable wrapping of code components on the site

joshwooding:


pseys (approved):

honey-chang:
Most of the pages don't have a problem, but the How to Use page's props table for components has an issue because the code strings are too long for the narrow table cells. (https://saltdesignsystem-git-site-code-no-wrap-fed-team.vercel.app/salt/components/accordion/usage)I suggest redesigning the structure of the props table first and then applying this change.

PR #5297, opened 2025-07-02
Adjust alpha ramp

joshwooding:


packages/theme/css/characteristics/separable-next.css
```diff
@@ -1,9 +1,9 @@
 .salt-theme.salt-theme-next {
   --salt-separable-borderStyle: solid;
 
-  --salt-separable-primary-borderColor: var(--salt-palette-alpha-contrast-mediumLow);
-  --salt-separable-secondary-borderColor: var(--salt-palette-alpha-contrast-low);
-  --salt-separable-tertiary-borderColor: var(--salt-palette-alpha-contrast-lower);
+  --salt-separable-primary-borderColor: var(--salt-palette-alpha-contrast-mediumHigh);
+  --salt-separable-secondary-borderColor: var(--salt-palette-alpha-contrast-medium);
+  --salt-separable-tertiary-borderColor: var(--salt-palette-alpha-contrast-mediumLow);
```
pseys, line 6: `--salt-separable-background-hover` is referencing a deprecated token. It should be pointing to `--contrast-lower` and therefore changing to `--contrast-mediumLow`

pseys (dismissed):

.changeset/green-parrots-brush.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/theme": minor
+---
+
+Deprecated 6 alpha foundation tokens.
+
+| Token                  |
```
pseys, line 7: Do we need to suggest the nearest neighbour to choose in their absence?

joshwooding: Not sure if it's necessary but if you think we should I'm happy to add it. The other alternative is for consumers to define it themselves

amritadesmet (approved):

PR #5317, opened 2025-07-07
Update foundational colors

joshwooding:


.changeset/sweet-rules-remember.md
```diff
@@ -0,0 +1,17 @@
+---
+"@salt-ds/theme": minor
+---
+
+Added an experimental brown ramp.
```
pseys, line 5: Is there away we can flag this as 'beta' or is that implied in the note that this is experimental?

joshwooding: It's implied by mentioning experimental, this was mostly waiting for something I could copy from @dplsek's post. If we add it to the site we can put more guidance there

joshwooding: Also there's not really a way to flag it as experimental unless we change the token name to reflect that?

pseys: OK great, thanks for the clarification. I thought that was the case. I don't want to alter the token names. Hopefully the release announcement will provide clarity.

joshwooding: @pseys We could also move it to a new file called experimental.css ?

pseys: No I think it's fine. We're not tying it to any components or characteristics yet and even if the values change the color token name won't. You highlight it in the changelog and we have it marked as beta so I think it's fine as it is.

pseys (approved):

PR #5339, opened 2025-07-11
Update navigable tokens

joshwooding:


pseys (approved):

packages/theme/css/characteristics/navigable-next.css
```diff
@@ -1,4 +1,8 @@
 .salt-theme.salt-theme-next {
-  --salt-navigable-indicator-hover: var(--salt-palette-neutral);
-  --salt-navigable-indicator-active: var(--salt-palette-accent);
+  --salt-navigable-accent-indicator-hover: var(--salt-palette-neutral);
+  --salt-navigable-accent-indicator-active: var(--salt-palette-accent);
```
libertymayc, line 3: wondering what the reason was for adding accent to these - what if the these tokens in another theme aren't desired to be styled the same as the accent color?  i assume we are adding accent because there will be other variants, but it ties this quite strongly here to accent, which is itself a different semantic and if we are saying components using this token should always be the accent color maybe they should be within that characteristic itself as a navigable variant?

pseys: Hey @libertymayc we’ve tied the tokens to accent because we know there is a requirement coming, once the brown ramp is added, for the navigation ‘accent’ to be brown. We’re adding brown through the sentiment structure to sit along ‘accent’ as ‘brand’. So in the near future nav will have a color prop with two variants; accent and brand.

PR #5343, opened 2025-07-13
remaining a11y recommendations for first release

mark-tate:


jake-costa:
# A11y Feedback:
## Structure and roles
* ❌ Remove aria-roledescription="slide" from all slide containers with role=tabpanel
* ❌ Add role="tabpanel" to each slide container (Does not look to be added to Multi slide example)
* ❌ For the multi-slide tab pattern - one tab cannot control multiple tabpanels according to ARIA specification for tabs. Need to look into potential solutions for this (one idea below) or use the previous pattern for this example. 
* ✅ Tabs have proper role="tab", aria-selected, and aria-controls
* ✅ Tablist has role="tablist" and aria-label
* ✅ Tablist correctly has tabindex="-1" to prevent focus

## Focus management
* ✅ Prev/Next buttons correctly have tabindex="-1"
* ✅ Tab focus management is correct (selected tab has tabindex="0")
* ✅ Arrow keys navigate between tabs when a tab in the tablist has focus, Tab/Shift+Tab moves focus in/out of tablist
* ❌ When slides contain interactive elements (links/buttons), those elements receive focus before the selected tab in the tablist - this breaks the expected focus order. 

## Carousel heading/title
* ❌ The visible "Title" is a `<div>` with H2 styling, not a semantic `<h2>` element
* ❌ The carousel uses aria-label="carousel example" instead of aria-labelledby pointing to the visible title (if available)

## Additional Considerations:
* The tabpanel containing individual slides should receive focus to allow arrow keys to move between slides instead of allowing slide navigation using arrow keys when focus is on the interactive element(s) in the tabpanel.  
* Screen readers are announcing the label for the tabs 3 times (Looks to be a result of the updated visual position label "slide 1 of 4") 
* An idea for the multi-slide example is to consolidate multiple slides into a single tab panel. Each slide could then have role="group", an aria-roledescription=slide and appropriate label attribute, allowing the slides, rather than the tab panel, to receive focus. This would enable users to use arrow keys to navigate between slides when one is focused. The tab aria-label (remove aria-labelledby) could then indicate the visible range, such as "slides 1-2 of 7" for example, based on how many slides are displayed.

packages/embla-carousel/src/CarouselCard.tsx
```diff
@@ -58,7 +58,6 @@ export const CarouselCard = forwardRef<HTMLDivElement, CarouselCardProps>(
     return (
       <div
         role="tabpanel"
-        aria-roledescription="slide"
```
mark-tate, line 61: @jake-costa I removed this but  APG example has it still

packages/embla-carousel/src/CarouselProgressLabel.tsx
```diff
@@ -29,32 +30,25 @@ export function CarouselProgressLabel({
...
-        numberOfSlides,
-      );
-
-      if (startSlideNumber === endSlideNumber) {
-        setCurrentSlide(startSlideNumber.toString(10));
-      } else {
-        setCurrentSlide(`${startSlideNumber}-${endSlideNumber}`);
-      }
-      setTotalSlides(numberOfSlides);
+      const startSlideNumber = visibleSlides[0];
```
libertymayc: what if visibleSliders is empty?  I think this is possible

packages/embla-carousel/src/CarouselTabList.tsx
```diff
@@ -81,42 +86,47 @@ export const CarouselTabList = forwardRef<HTMLDivElement, CarouselTabListProps>(
...
+        {scrollSnaps.map((_, scrollSnapIndex) => {
+          const visibleSlides = getVisibleSlideIndexes(
+            emblaApi,
+            scrollSnapIndex,
           );
-          const label =
-            startSlideNumber === endSlideNumber
-              ? `Slide ${startSlideNumber}`
-              : `Slides ${startSlideNumber}-${endSlideNumber} of ${numberOfSlides}`;
+          const startSlideNumber = visibleSlides[0];
```
libertymayc: same here

packages/embla-carousel/src/CarouselProgressLabel.tsx
mark-tate: good shout

packages/embla-carousel/src/CarouselTabList.tsx
mark-tate: good shout

jake-costa:
# Clarification on the use of `aria-roledescription` property
* [APG Carousel](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) for the tabbed pattern says:
  * Each slide container has role [tabpanel](https://w3c.github.io/aria/#tabpanel) in lieu of group, and it does not have the `aria-roledescription` property.

* However on the [Auto-Rotating Image Carousel with Tabs for Slide Control Example](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/examples/carousel-2-tablist/) the `aria-roledescription` property is used and the "Role, Property, State, and Tabindex Attributes" section says
  * The `aria-roledescription` property informs assistive technologies to identify the element as a "slide" rather than as a "tabpanel".

* After some research and testing we have decided to utilize the `aria-roledescription` property on our slide containers that contain role [tabpanel](https://w3c.github.io/aria/#tabpanel) moving forward

jake-costa:
# A11y Feedback: Carousel Component

## URL's Reviewed:
* https://saltdesignsystem-fu9za1ej7-fed-team.vercel.app/
* https://jpmorganchase-carousel-a11y.saltdesignsystem-storybook.pages.dev/

## Carousel container 
* ✅ Contains both `role="region"` and `aria-roledescription="carousel"`
* ✅ Can use both `aria-labelledby` and `aria-label` to provide an accessible name
  * Refer to additional considerations (example specific) below for more information

## Single Slide Carousel (Tabbed Pattern)
* ✅ The container that holds the `tab` elements contains `role="tablist"` with `aria-label`
* ✅ Individual tabs have `role="tab"` with appropriate `aria-selected` states
* ✅ Selected tab has `tabindex="0"`, unselected tabs have `tabindex="-1"`
* ❌ Tab elements should have `aria-label` that matches the visual label of the slide they control (if available) 
	* Default example on the site has the `aria-label` as "Slide 1" but the visible slide label is "Your Accounts, on the move"
* ❌ Tabs have empty `aria-controls` 
  * `aria-controls` needs to reference the associated tabpanel ID
  * Example: `<button class="saltCarouselTab saltCarouselTab-selected" role="tab" aria-label="Slide 1" aria-selected="true" tabindex="0" aria-controls=""></button>`
* ❌ Tabpanels with `role="tabpanel"` are missing `aria-roledescription="slide"` (does not apply to multi-slide or autoplay)
  * Was there originally but was removed do to previous conflicting APG guidance
  * Refer to the clarification on the use of `aria-roledescription` property above
* ❌ Tabpanels are missing IDs for the `aria-controls` relationship
  * Example: `<div role="tabpanel" class="saltCarouselCard carousel_carouselSlide__gc0XJ" aria-labelledby=":r1v:-0">`
* ✅ Tabpanels contain either `aria-label` or `aria-labelledby` referencing their associated tab elements

### Focus Management
* ❌ Containers holding tabpanel slides have `tabindex="0"`, this should be removed
	* Individual tabpanels should have `tabindex="-1"` and enable programmatic focus management
	* When user presses Tab when focus is on the currently selected tab, focus should move to the associated tabpanel, Not the container that holds all tabpanel slides
  * VoiceOver issue: After selecting the `tab` for slide 2 and pressing Tab, focus looks to be going to visually hidden slide 1 and the Announcement contains the label for slide one as well. 
* ✅ Carousel controls receive focus first

### Live Region
* ❌ Separate SR-only div with `aria-live="polite"` looks like it might be causing duplicate announcements when combined with built-in screen reader announcements for the tab pattern.
  * Ideally we should use 1 `aria-live="polite"` region and should only convey information that the screen reader doesn't already announce for the component
  * The main slides container explicitly declares `aria-live="off"`. This attribute does not look to be used and can be removed

## Multi-Slide Carousel (Basic Pattern)

### Slides
* ✅ Individual slides have `role="group"` with `aria-roledescription="slide"`
* ✅ Individual slides in the examples have `aria-labelledby` pointing to their heading elements.
* ❌ Individual **visible** slides do not receive focus.
	* The container holding all slides should not receive focus. 

### Live Region
* ❌ Screen readers are announcing every visible slide's label when navigating between slide groups.
	* "Slides 1-2 of 4" announcement via `aria-live="polite"` is all that is needed 

## Autoplay Carousel (Basic Pattern)

### Autoplay Control
* ✅ Autoplay control button exists and is first in the control order
* ✅ Autoplay control button is always visible
* ✅ When focus or mouse hover is on the carousel slide content auto-rotation is paused.
* ❌ When focus or mouse hover is on the autoplay and navigation controls the auto-rotation is not paused. 
	* When keyboard focus or mouse hover is on any part of the carousel. Autoplay needs to be paused.
	* For keyboard focus, autoplay should not restart unless the user explicitly requests it to do so.
* ❌ Button label "Resume/Pause play" should be changed to "Start/Stop automatic slide rotation" for better clarity

### Live Region - Autoplay
* ❌ SR-only div has `aria-live="polite"` - should be "off" when autoplay is on to prevent constant interruptions

## Density/Reflow information
* ✅ Carousel controls meet target size requirements 2.5.8 (24x24) in all densities
  * **Note:** In high density the tabs in the tab list do not meet target size requirements but the previous/next buttons do, fulfilling the equivalent exception
* ✅ When using 320x256 viewport the examples do not require scrolling in 2 dimensions

## Light/Dark
* ✅ Examples meet for both text and non-text contrast requirements. 
	* **Note**: While examples pass, this is conditional and we might want to let developers know that this will need to be evaluated when implemented in their specific application 

## Additional Considerations
* ❓ Tab lists have `tabindex="-1"`.
  * Tab lists by default shouldn't receive focus, we need to verify if this is needed. Otherwise this should be removed.
* ❓ When using 320x256 viewport the examples change from tabbed pattern to basic pattern. Even though the viewport is smaller it would be good to keep the tabbed pattern (excluding multi-slide which should likely switch to single slide view, especially if displaying more than two slides at a given time)


## Additional Considerations (**Example specific**)
* When an example shows a visible heading `<h2>Title</h2>`, the carousel container that has `aria-roledescription="carousel"` should be using `aria-labelledby` pointing to the heading's ID,  examples should reflect this.
	* `aria-label` is appropriate only when no visible heading is available
* For elements that contain `aria-roledescription="carousel"` the accessible label should not contain the word "carousel"
* For elements that contain `aria-roledescription="slide"` the accessible label should not contain the word "slide"

site/docs/components/carousel/accessibility.mdx
```diff
@@ -12,9 +12,23 @@ data:
...
 - Avoid using infographics, maps, charts as images or [images of text](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-text-presentation.html).
 - Order slides by importance, with the most critical info first.
+- If composing with navigation controls (including tablist), ensure they are the first element to receive focus, enabling them to control the carousel, upon focus.
+
+### With tablist
+
+If composing with a tablist:
+
+- If composing with a tablist, ensure they are the first element to receive focus, enabling them to control the carousel, upon focus.
 - Tablist components provide keyboard navigation, making them less suitable for mobile devices where space is limited and users often prefer gesture-based slide navigation.
```
jake-costa: We should remove/modify this. While sighted, non-AT users might prefer to swipe through the slides with gestures, the tab list itself is a primary and essential navigation method for all users on mobile, including those using screen readers.

* When a mobile screen reader is active, it repurposes swipe gestures to move focus between elements, not to scroll content. This makes the tab list the essential mechanism for these users to navigate the carousel slide by slide.
* The tab list provides clear context on the user's position (e.g., slide 3 of 5) while also serving as an additional navigation method. This allows any user to directly select a specific slide, regardless of how they are accessing the page.

site/docs/components/carousel/accessibility.mdx
```diff
@@ -12,9 +12,23 @@ data:
...
-- If slides contain focusable elements, then add carousel controls to the top, so they receive focus first.
-- If slides do not contain focusable elements, then carousel controls can be placed at the top or bottom.
+- Remove the next and prev buttons from the focus order, as the tablist provides sufficient navigation controls.
+- If multiple cards are visible, within one slide, do not include a tablist, as a tablist should only be used with single cards.
```
jake-costa: Suggestion: "If multiple slides are visible at one time, do not include a tablist, as the tabs in a tablist should be associated with individual slides."

site/docs/components/carousel/accessibility.mdx
```diff
@@ -12,9 +12,23 @@ data:
...
+- If using a tablist, ensure each card has`role="tabpanel"`.
+
+### Without tablist
+
+If composing without a tablist:
+
+- If not using a tablist, ensure each card has`role="group" and `aria-roledescription="slide".
```
jake-costa: Missing backticks: `role="group"` and `aria-roledescription="slide"`

jake-costa (commented):
We need to change this: "ensure screen readers announce changes in slide content during autoplay, providing context for visually impaired users."

To "ensure screen readers do **not** announce changes in slide content during autoplay" as the constant interruptions make the page unusable for a screen reader user.

[Carousel APG](https://www.w3.org/WAI/ARIA/apg/patterns/carousel/) aria-live is set to off if the carousel is automatically rotating.

site/docs/components/carousel/accessibility.mdx
mark-tate: fixed but you feel free to edit the doc in the PR rather than just comment.

site/docs/components/carousel/accessibility.mdx
mark-tate: Will remove for now, feel free to  add any specific mobile guidance

mark-tate:
@jake-costa pushed a set of a11y updates based on your feedback, thanks++
Can we collate any remaining issues into a new list, comprising just of remaining failures.
If there are doc changes feel free to push a commit yourself or edit inline within github as a suggestion.

jake-costa:
# Updated A11y Feedback: Carousel Component

## URLs Reviewed

* https://saltdesignsystem-btstb5w6h-fed-team.vercel.app/salt/components/carousel/examples
* https://jpmorganchase-carousel-a11y.saltdesignsystem-storybook.pages.dev/?path=/story/patterns-carousel

## Tabbed Pattern Slides

* ❌ `tabpanel` missing `aria-roledescription="slide"` and `aria-labelledby` (applied on a parent wrapper)
  * **Affected:** Storybook: Card · Vercel: Default, Bordered
  * **Expected:** Apply `aria-labelledby` and `aria-roledescription="slide"` directly on the `tabpanel` 

### Focus Management

* ❌ Parent wrapper holding all slides has `tabindex="0"` (Applies to Multi-slide as well)
  * **Expected:** Tabbing from the selected `tab` moves into its associated `tabpanel`, not the outer parent wrapper
  * **Note:** This also impacts the announcement made by screen readers

### ARIA Controls (recommendation)

* ⚠️ `aria-controls` points to the outer slide wrapper `id` that contains the `tabpanel` (valid but indirect).
  * **Affected:** Storybook: Card · Vercel: Default, Bordered
  * **Expected:** Point `aria-controls` to the `id` of the `tabpanel`

### Visually Hidden Slides

* ❌ Visually hidden slides are still exposed to AT (Applies to Multi-slide as well)
  * **Expected:** Conditionally apply `aria-hidden="true"` to any non-visible slide (`tabpanel` or `role="group"`)

## Basic Pattern Slides

### Slides

* ❌ `role="group"` slides missing `aria-roledescription="slide"`
  * **Affected:** Vercel example (Storybook is correct)
  * **Expected:** Add `aria-roledescription="slide"` to the element with `role="group"`

### Live Region

* ❌ Screen readers announce every visible slide label when navigating between multi-slide groups
  * **Expected:** Announce a single concise message (e.g., “Slides 3-4 of 4”) via one `aria-live="polite"` region
  * **Note:** Announcements when navigating single-slides are correct
 
## Autoplay

* ❌ Autoplay doesn't pause when keyboard focus is on the autoplay or navigation controls

* ❌ Control label is inverted in relation to the actual state
  * **Expected:** When paused → “Start automatic slide rotation”; when playing → “Stop automatic slide rotation”

* ❌ Control uses `aria-pressed`
  * **Expected:** Remove `aria-pressed` as the button uses dynamically updated action-based label
 
* ❌ Parent `div` of autoplay progress indicator SVG contains unnecessary `aria-label` attribute
  * **Expected:** Remove the aria-label from this div as the toggle button already provides play/pause state

## (Storybook examples only)

* ⚠️ Some examples with visible heading use `aria-label`
  * **Expected:** Use `aria-labelledby` pointing to the visible H2

* ⚠️ Reflow at 320 × 256 viewport does not appear to work as expected in Storybook (looks to be working on Vercel site)

site/src/examples/carousel/BorderedCard.tsx
```diff
@@ -20,48 +13,57 @@ import { sliderData } from "./exampleData";
...
+            const id = `${carouselId}-card${index}`;
+            return (
+              <CarouselCard
+                className={styles.carouselSlide}
+                key={`slide-${id}`}
+                aria-labelledby={`title-${id}`}
+                appearance="bordered"
+                media={
+                  <img
+                    alt={`stock content to show in carousel slide ${index}`}
```
joshwooding: @jake-costa Would you expect images to have an accessible name here?

jake-costa: @joshwooding For this example we can apply aria-hidden and clear the alt since they are decorative but if the images were meaningful this would be correct.

mark-tate:
@jake-costa can you re-review.

The remaining concerns are:

Your direction to get the slide title from the title for Cards, means the Slide uses ariaLabelledBy to link to the H2 etc, but then you also ask for the tabs to be the source of truth for the screenreader and for the slide to link back to the tab.

I've  achieved want you described but suspect would be better to use ariaLabelledBy for tab, which points to slide and the slide to use ariaLabelledBy to point to Title, or I could wire both directly to the title.

Any preference?

Also, your direction to use aria-roledescription="slide", means we can get a different role announcement for next/previous (slide) as to when you use tab (tab). Suspect, when navigating a tab list, this should be consistent between next/prev and tab.

packages/embla-carousel/src/CarouselCard.tsx
```diff
@@ -55,13 +56,15 @@ export const CarouselCard = forwardRef<HTMLDivElement, CarouselCardProps>(
...
 
+    const { ariaVariant } = useCarouselContext();
+
     return (
       <div
-        role="tabpanel"
-        aria-roledescription="slide"
-        className={clsx(withBaseName(), className)}
-        {...rest}
+        aria-roledescription={ariaVariant === "tabpanel" ? "tab" : "slide"}
```
jake-costa: Should be aria-roledescription="slide" for both tabpanel and group

jake-costa:
>>Also, your direction to use aria-roledescription="slide", means we can get a different role announcement for next/previous (slide) as to when you use tab (tab). Suspect, when navigating a tab list, this should be consistent between next/prev and tab.

> Regarding aria-roledescription="slide", The different announcements (tab vs slide) between these 2 patterns is expected and appropriate.

Wanted to clarify in case of confusion, my response was to ensure both patterns use `aria-roledescription="slide"` even if the screen reader announcements differ when navigating using the controls.

site/docs/components/carousel/accessibility.mdx
```diff
@@ -25,7 +38,18 @@ If using Autoplay, then additional considerations are required
...
+- unset `aria-roledescription` to align with JPM accessibility recommendations.
+- if autoplay of content starts upon init, the slides should pause upon hover.
+- if the user has to press play to see more content, the slides do not need to pause upon hover.
+
+### Multiple slides per tab
+
+If using with multiple slides per tab, this will create increased complexity for the users relying on assistive technologies.
+
+- consider whether a single slide view is sufficient for your use case, as it simplifies navigation and interaction.
+  - unset `aria-roledescription` to align with JPM accessibility recommendations.
```
jake-costa: Do we want to remove this if handled by the component or change it to set instead of unset?
applies to the "Autoplay" section above as well.

site/docs/components/carousel/usage.mdx
```diff
@@ -15,22 +15,31 @@ Carousel is a composable pattern of Salt components that uses the [Embla](https:
...
 
 A `tablist` provides a visual indicator of the current slide and includes clickable elements for direct navigation to specific slides, along with keyboard navigation controls. Ensure the tablist is positioned at the top or bottom of the carousel, alongside other controls like `CarouselNextButton` and `CarouselPreviousButton`.
 
-<LivePreview componentName="carousel" exampleName="TabList" />
+`CarouselTabList` is an aria `tablist` providing support for screen readers to announce the slides as the associated`tab`.
 
+<LivePreview componentName="carousel" exampleName="TabListTop" />
 <br />
+
+If you position a `tablist` or navigation controls at the bottom of the carousel, the slides will initially receive focus. To ensure accessibility, navigation controls should receive focus before the slides. You can achieve this by using an additional `div` to reverse the flex direction, thereby ensuring the navigation controls are focused first.
```
jake-costa: Is this something the dev will need to implement or are the navigation controls receiving focus initially handled by the component itself?

jake-costa:
Hi @mark-tate, made some changes to the docs, let me know what you think. Also, do we want Ben to review?

Lastly, I added (aria-label="slides") as recommended by the APG but the screen reader announcements were not ideal. Given this, I removed it so it is back to how you had it before which has a much better screen reader announcement.

site/docs/components/carousel/accessibility.mdx
mark-tate: removed, as replaced by recent improvements/recommendations

site/docs/components/carousel/accessibility.mdx
```diff
@@ -12,22 +12,28 @@ data:
 
 - Avoid using infographics, maps, charts as images or [images of text](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-text-presentation.html).
 - Order slides by importance, with the most critical info first.
-- If composing with navigation controls (previous/next slide or tablist), ensure they are the first elements to receive focus, enabling the user to control the carousel.
+- If composing with navigation controls (previous/next buttons or tablist), ensure they are the first elements to receive focus, enabling the user to control the carousel, on initial focus.
+
+### Accessible name
+- If the carousel has a visible label, its accessible label is provided by aria-labelledby on the carousel container set to the ID of the the visible label. Otherwise, an accessible label is provided using aria-label set on the carousel container.
```
mark-tate: should use back-ticks, will update

jake-costa (dismissed):

tayyabataimur (dismissed):

tayyabataimur (approved):

joshwooding:
I think there are some weird issues with focus timing, if you have recently focused a slider and shift+tab away from it quick enough your focus moves back to the slide.

This is sometimes causing focus to be trapped in a loop between slides in the multi-slide example

jake-costa:
FYI looks like a number of doc changes I added have been reverted/removed. For example "Refrain from using progress dots on mobile devices, as space is limited and users favor gesture-based navigation." not sure we should have this as both dots and previous/next buttons are valid. In addition, sighted non-AT users may favor gesture-based navigation but not AT users as the gestures when AT is enabled does not function the same.

packages/embla-carousel/src/__tests__/__e2e__/Carousel.cy.tsx
```diff
@@ -69,19 +69,26 @@ describe("Given a Carousel", () => {
...
   };
 
-  it("should render the carousel with four slides", () => {
+  it("should render the carousel with four slides as a tabbed list", () => {
     cy.mount(<Default />);
     cy.findByRole("region").should("exist");
-    cy.get('[aria-label="carousel example"]').should("exist");
-    cy.get('[aria-roledescription="slide"]').should("have.length", 4);
+    cy.get('[aria-label="Numbered tab example"]').should("exist");
+    cy.get('[role="group"]').should("not.exist");
```
joshwooding: Is there a reason findByLabel or findByRole isn't being used in the new tests?

packages/embla-carousel/src/CarouselTabList.tsx
```diff
@@ -67,56 +70,71 @@ export const CarouselTabList = forwardRef<HTMLDivElement, CarouselTabListProps>(
...
 
+          const selected = selectedIndex === scrollSnapIndex;
           const ariaControls = slideNodes?.length
             ? slideNodes[startSlideNumber - 1].id
             : undefined;
+
           return (
             <CarouselTabRenderer
-              key={`carouselTab-${tabIndex}}`}
+              key={`carouselTab-${scrollSnapIndex}}`}
```
joshwooding, line 125: Why is this key needed?

packages/core/src/aria-announcer/AriaAnnouncerContext.ts
```diff
@@ -1,9 +1,23 @@
 import { createContext } from "react";
 
-export type AnnounceFn = (announcement: string, delay?: number) => void;
+export type AnnounceFnOptions = {
+  duration?: number;
```
joshwooding, line 4: We should add jsdocs to explain what these options do

.changeset/nasty-planes-sell.md
```diff
@@ -0,0 +1,19 @@
+---
+"@salt-ds/core": minor
+---
+
+- `AriaAnnouncer` improvements, previously `delay` was used to queue announcements, but it could create overlapping announcements. We have deprecated `delay`, although it is still supported. Instead, we have added a new options API.
+- `duration`: provides a sequential delay between announcements, ensuring they do not overlap.
```
joshwooding, line 6: Is duration teminology?

site/docs/components/carousel/usage.mdx
```diff
@@ -8,47 +8,54 @@ data:
...
+<LivePreview componentName="carousel" exampleName="Card" />
 
-Refer to the [Embla documentation](https://www.embla-carousel.com/) for the full API.
+### With focusable slide content
 
-<LivePreview componentName="carousel" exampleName="ControlledCarousel" />
+Incorporate actions, such as a `Link`, into a slide. `CarouselCard` supports actions, through the `actions` prop.
+
+<GuidanceCallout type="positive" customPillText="Best practices">
+  - actions such as[`Link`](../link), can be incorporated into a carousel slide,
```
joshwooding: This list isn't displaying correctly

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -1,66 +1,127 @@
...
+  type AnnounceFnOptions,
+  AriaAnnouncerContext,
+} from "./AriaAnnouncerContext";
+
+export const ARIA_ANNOUNCE_DELAY = 150; // time between DOM updates
+export const DEFAULT_ANNOUNCEMENT_DURATION = 500; // time between announcements
 
-export const ARIA_ANNOUNCE_DELAY = 150;
+/**
+ * TODO add forwardRef at a breaking change
```
joshwooding: adding forward ref isn't a breaking change

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -1,66 +1,127 @@
...
+  const makeAnnouncement = useCallback(
+    (announcement: string, ariaLive: AnnounceFnOptions["ariaLive"]) => {
+      setAriaLive(ariaLive);
+      setCurrentAnnouncement(announcement);
+    },
+    [],
+  );
+
+  /** Each announcement has a minimum duration */
+  // biome-ignore lint/correctness/useExhaustiveDependencies: clearAnnouncement will not change
```
joshwooding, line 53: It should still be included

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -71,12 +132,15 @@ export function AriaAnnouncerProvider({
...
   const value = useMemo(() => ({ announce }), [announce]);
+
   return (
     <AriaAnnouncerContext.Provider value={value}>
       {children}
       <div
         aria-atomic="true"
-        aria-live="assertive"
+        aria-live={
+          (ariaLive as React.AriaAttributes["aria-live"]) ?? "assertive"
```
joshwooding: This cast is only needed because the ariaLive option is incorrectly typed

packages/core/src/aria-announcer/AriaAnnouncerContext.ts
```diff
@@ -1,9 +1,23 @@
...
+  ariaLive?: Omit<React.AriaAttributes["aria-live"], "off">;
```
joshwooding: `Omit` doesn't work in this context, `Exclude` should be used instead.

packages/core/src/aria-announcer/AriaAnnounce.tsx
```diff
@@ -1,24 +1,36 @@
...
   useEffect(() => {
     if (announcement) {
-      announce(announcement);
+      if (delay !== undefined) {
+        announce(announcement, delay);
+      }
+      announce(announcement, rest);
     }
-  }, [announce, announcement]);
+  }, [announce, announcement, delay]);
```
joshwooding, line 33: We need a better solution for this. This would cause an announcement to be made if "delay" changes

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -1,66 +1,127 @@
...
+          setTimeout(announceAllLegacy, ARIA_ANNOUNCE_DELAY);
         } else {
           isAnnouncingRef.current = false;
         }
       });
     }
   }, []);
 
+  /**
+   *  TODO default to `assertive` until a breaking change, when we should switch to `polite`
```
joshwooding: Should this be on another line?

packages/embla-carousel/src/Carousel.tsx
```diff
@@ -61,16 +78,36 @@ export const Carousel = forwardRef<HTMLElement, CarouselProps>(
       ...emblaPlugins,
     ]);
 
+    const carouselId = useId();
```
joshwooding:
```suggestion
    const carouselId = useId(id);
```

packages/embla-carousel/src/Carousel.tsx
```diff
@@ -61,16 +78,36 @@ export const Carousel = forwardRef<HTMLElement, CarouselProps>(
...
+      <CarouselContext.Provider
+        value={{
+          ariaVariant,
+          disableSlideAnnouncements,
+          emblaApi,
+          emblaRef,
+          setAriaVariant,
+          announcementState,
+          setAnnouncementState,
+          carouselId: id ?? carouselId,
```
joshwooding:
```suggestion
          carouselId,
```

packages/embla-carousel/src/CarouselContext.tsx
```diff
@@ -2,6 +2,18 @@ import { createContext } from "@salt-ds/core";
...
+
+export const CarouselAnnouncementTriggerValues = [
+  "tab",
+  "navigation",
+  "drag",
+  "focus",
+] as const;
+export type CarouselAnnouncementTrigger =
+  (typeof CarouselAnnouncementTriggerValues)[number];
+
```
joshwooding: I think we can just use standard unions

site/docs/components/carousel/accessibility.mdx
```diff
@@ -8,74 +8,112 @@ data:
...
+  components offer keyboard navigation, which may not be ideal for mobile
+  devices where space is limited and users prefer gesture-based navigation.
+</GuidanceCallout>
 
-- Avoid using infographics, maps, charts as images or [images of text](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-text-presentation.html).
-- Order slides by importance, with the most critical info first.
-- Tablist components provide keyboard navigation, making them less suitable for mobile devices where space is limited and users often prefer gesture-based slide navigation.
-- If slides contain focusable elements, then add carousel controls to the top, so they receive focus first.
-- If slides do not contain focusable elements, then carousel controls can be placed at the top or bottom.
+### Accessible name
```
joshwooding, line 16: I think since this is a H3 we are skipping headings and that would be an a11y violation. @jake-costa  can you check this please

packages/embla-carousel/stories/carousel.stories.tsx
```diff
@@ -9,89 +9,89 @@ import {
...
-      emblaPlugins={[CarouselAnnouncement()]}
-      {...args}
-    >
-      <FlexLayout justify={"space-between"} align={"center"} direction={"row"}>
-        <Text styleAs={"h2"} className="carouselHeading">
-          Title
-        </Text>
+    <Carousel aria-label="Numbered tab example" className="carousel" {...args}>
+      <FlexLayout gap={1} wrap={true} align={"center"}>
+        <CarouselPreviousButton tabIndex={-1} appearance="transparent" />
```
joshwooding: I think appearance should be updated in CarouselPreviousButton and CarouselNextButton instead.

site/docs/components/carousel/examples.mdx
```diff
@@ -10,33 +10,67 @@ data:
...
+- Pause the animation upon hover. 
+- Stop the slideshow upon any user interaction with content and do not resume, unless the user restarts it with the play button.
+
+</GuidanceCallout>
+
+The example below demonstrates accessible autoplay for illustrative purposes.
 
 <LivePreview componentName="carousel" exampleName="Autoplay" />
+
+### Embla
```
joshwooding: I'm guessing this is a controlled example? Should the h3 be changed to a h2 and the title be updated?

site/docs/components/carousel/usage.mdx
```diff
@@ -63,12 +70,16 @@ Refer to the [Embla documentation](https://www.embla-carousel.com/) for the full
...
 - To display less than 2 or more than 5 pieces of content.
 
-### Best practices
-
-- Always provide navigation controls for the carousel, alongside a visual indicator of the current slide.
-- Include 5 or fewer slides. The most important slide always comes first.
-- Not all content is visible within a carousel until selected, ensure that important content is also placed on another page of your site.
-- To provide arrow key navigation add a tablist.
+<GuidanceCallout type="positive" customPillText="Best practices">
+  - Always provide navigation controls for the carousel, alongside a visual
```
joshwooding: This has the same formatting issue

packages/embla-carousel/src/__tests__/__e2e__/Carousel.cy.tsx
mark-tate: no,  updated

packages/embla-carousel/src/CarouselTabList.tsx
mark-tate: it's rendering a map so needed

site/docs/components/carousel/usage.mdx
mark-tate: ah... prettier/formatting breaks it.. may need to pull it from the Callout

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
mark-tate: added it

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
mark-tate: it's because we do not want to support "off" in the API
re-worked the type so the cast is not needed

packages/core/src/aria-announcer/AriaAnnouncerContext.ts
mark-tate: reworked

packages/embla-carousel/src/Carousel.tsx
mark-tate: good one

site/docs/components/carousel/accessibility.mdx
mark-tate: my bad, added back the h2 for best practices that I had replaced with a callout.

packages/embla-carousel/stories/carousel.stories.tsx
mark-tate: good shout

site/docs/components/carousel/examples.mdx
mark-tate: Controlled is nicer

site/docs/components/carousel/usage.mdx
mark-tate: prettier, a new line before/after seems to fix it

packages/core/src/aria-announcer/AriaAnnounce.tsx
jake-costa: FYI a delay should not be applied to the aria live region and it should be polite not assertive.

jake-costa:
@mark-tate I am open to accepting this one defect (assertive live announcements) on the Carousel if we are going to open a PR to handle the announcer changes since it would be a bug fix for a11y. If so, do you want me to raise an issue about the aria announcer so we can add polite and switch the default to polite?

Let me know if you want to connect or what you think next steps should be. 

>Typically the DOM needs around 500 msecs to see any change, it then decides whether or not to announce it.
However, when it maybe announcing things in quick succession it can drop things.
A pre-timer delay can all fire at once (<500 msecs) so the DOM changes overwrite one another and updates are lost.
By using a duration/post-delay the DOM changes can be seen.

Also, I just want to clarify that while this current example does not have any violations and there are use cases where controlling the announcements would be beneficial, in general controlling the polite announcement timing manually could be problematic, and relying on built-in screen reader behavior is often better even if a polite announcement is cut off by another polite announcement. 

So as you mentioned, for the carousel specifically, I would expect if a user using a screen reader navigates to slide 2 and the announcement says "slide 2" the live region is cleared after a short period of time but if the user is pressing the next button in quick succession and lands on slide 4 they would expect the screen reader to only announce "slide 4" not 2 and 3 depending on the AT/browser combination. 

Lastly, also just wanted to note that the tab selection seems to be on mouse down-event. While not an ADA violation either, I would expect selection to be made on the up-event.

jake-costa (approved):

PR #5346, opened 2025-07-15
Add NumberInput enhancements

tayyabataimur:


packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -373,14 +372,18 @@ describe("Number Input", () => {
...
-    cy.findByRole("spinbutton").should("have.value", "100.00");
+    cy.findByRole("spinbutton").should("have.value", "100");
 
     cy.findByRole("spinbutton").focus().clear();
     cy.realType("12.1234");
     cy.realPress("Tab");
     cy.findByRole("spinbutton").should("have.value", "12.12");
   });
 
+  it("allows trailing zeros when decimal scale is set and fixedDecimalScale is set to true", () => {
```
mark-tate: should there be an assertion here ?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -106,16 +109,17 @@ export interface NumberInputProps
...
    */
-  onChange?: (
-    event: SyntheticEvent | undefined,
-    value: number | string,
-  ) => void;
+  onChange?: (event: SyntheticEvent | undefined, value: number) => void;
+  /**
+   * Callback for final value
+   */
+  onChangeEnd?: (event: SyntheticEvent | undefined, value: number) => void;
```
mark-tate, line 116: is there a test for this being called ?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -160,10 +164,14 @@ export interface NumberInputProps
    * @default "primary"
    */
   variant?: "primary" | "secondary";
+  /**
+   * Callback to validate input
+   */
+  isAllowed?: (value: string) => boolean;
```
mark-tate: is there a test for this?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -345,34 +337,43 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
-      // Update the value if it has changed
-      if (clampedValue.toString() !== value.toString()) {
-        setValue(clampedValue);
-        onChangeProp?.(event, clampedValue);
+      const parsed = parse ? parse(inputValue) : sanitizeInput(inputValue);
+      const clampAndFixed = !Number.isNaN(toFloat(parsed))
+        ? clampAndFix(toFloat(parsed))
+        : toFloat(parsed);
+      if (value !== clampAndFixed) {
+        setValue(clampAndFixed || "");
```
mark-tate: does this work with 0 ?

site/docs/components/number-input/examples.mdx
```diff
@@ -71,15 +74,15 @@ When configuring the number of decimal places, ensure it aligns with the specifi
 
 ## Formatting
 
-You can format the value of the `NumberInput` using the `format` callback prop. It typically works alongside a `parse` callback for complex formatting.
+The `NumberInput` component offers the ability to format its value using the `format` callback prop. This feature is typically used in conjunction with a `parse` callback, which determines the numerical value, and an `isAllowed` callback to regulate permissible input values.
```
tycup: I think @mark-tate also brought this up yesterday in the demo; was isAllowed settled as the right prop naming? The language around the prop, and that it's "doing" something ("regulates permissible input values") feels like it belongs in the same naming category as "format" or "parse" (so maybe "validate"). Also since above, "fixedDecimalScale" is a boolean and doesn't include the auxiliary verb "is".

feel free to resolve if alreayd discussed and decided upon

tayyabataimur: Thanks for bringing this up! There are ongoing discussions on naming `isAllowed` and `fixedDecimalScale`, will update it once finalised. Feel free to drop in your suggestions as well.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -194,6 +194,34 @@ describe("Number Input", () => {
...
+    cy.mount(<Default onChangeEnd={changeEndSpy} />);
+    cy.findByLabelText("increment value")
+      .realMouseDown()
+      .wait(1000)
+      .realMouseUp();
+
+    cy.get("@changeEndSpy").should(
+      "have.been.calledOnceWith",
+      Cypress.sinon.match.any,
+      7,
```
mark-tate, line 208: curious - how do we know the value will be 7 after  1 second..

probably missed something.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -194,6 +194,34 @@ describe("Number Input", () => {
...
 
+  it("calls the `onChangeEnd` callback with the final value after continuous increments/decrements", () => {
+    const changeEndSpy = cy.stub().as("changeEndSpy");
+    cy.mount(<Default onChangeEnd={changeEndSpy} />);
+    cy.findByLabelText("increment value")
+      .realMouseDown()
+      .wait(1000)
+      .realMouseUp();
+
+    cy.get("@changeEndSpy").should(
```
mark-tate, line 205: can see in the next test, we type 7 but how does this test pass, before we type 7 ?
How do we know after 1 sec, it will be called with 7.

Maybe missed something.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -367,18 +419,56 @@ describe("Number Input", () => {
...
     cy.realPress("Tab");
-    cy.findByRole("spinbutton").should("have.value", "100.00");
+    cy.findByRole("spinbutton").should("have.value", "100");
+    cy.get("@changeSpy").should(
+      "have.been.calledWith",
+      Cypress.sinon.match.any,
+      100,
+    );
+    cy.get("@changeEndSpy").should(
+      "have.been.calledWith",
```
mark-tate, line 451: have.been.calledOnceWith

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -367,18 +419,56 @@ describe("Number Input", () => {
...
     cy.realType("12.1234");
     cy.realPress("Tab");
     cy.findByRole("spinbutton").should("have.value", "12.12");
+    cy.get("@changeSpy").should(
+      "have.been.calledWith",
+      Cypress.sinon.match.any,
+      12.12,
+    );
+    cy.get("@changeEndSpy").should(
+      "have.been.calledWith",
```
mark-tate: have.been.calledOnceWith

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -57,7 +56,7 @@ export interface NumberInputProps
   /**
    * The default value. Use when the component is uncontrolled.
    */
-  defaultValue?: number | string;
+  defaultValue?: number;
```
mark-tate, line 59: we still assign a "" as a default 

We assign a defaultValue which can be emptyReadOnlyMarker.

Both these are strings, given the change of type, if we did not assign these values as defaults then the controlled value would be inferred/typed as number. Wonder if that helps with code type safety?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -106,16 +109,16 @@ export interface NumberInputProps
...
+   */
+  onChangeEnd?: (event: SyntheticEvent | undefined, value: number) => void;
   /**
    *
-   * A callback to parse the value of the `NumberInput`. To be used alongside
-   * the `format` callback.
+   * A callback to parse the value of the `NumberInput`. To be used alongside the `format` callback.
    */
-  parse?: (value: number | string) => string | number;
+  parse?: (value: number | string) => number | undefined;
```
mark-tate: does parse need to support number and string, we call it once and it seems to just been with the blur value, which would be a string ?
It's also a bit clearer, there a few cases of number | string dotted around which creates some complexity and it could be that we can simplify this

I also see format as something which gets called which parsed values (number) and returns a string.

Worth checking the typings to see if we can simplify to 

  parse?: (value: string) => number | undefined;
  and 
    format?: (value: number) => string;
    
    Input does not have a format prop, but it could do the work of emptyReadOnlyMarker?
    I guess we are adding to make it consistent but it feels low value?
    I was curious why parse can return undefined but format cannot receive it.
    
    @joshwooding  what are your thoughts on emptyReadOnlyMarker and format, does it make sense to have both or to allow  format to format undefined values ?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -267,9 +266,20 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
 
     const [displayValue, setDisplayValue] = useState<string | number>(value);
 
-    const clampAndFix = (value: number) => {
+    const clampAndFix = (value: number): number | string => {
```
mark-tate, line 269: might be cleaner if only returned one type, number rather than both number | string ?

I think then you can change format to just accept numbers ?

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
tayyabataimur: So maybe this test is not the most accurate in testing this behaviour but I tried it manually like holding the mouse down for 1 second and the value increased from 0 to 7 in 1 second so that's why. 

In the next test, it's simply typing 7 and asserting that the value will be 7 because onChangeEnd will be called on keyboard interaction as well.

packages/lab/src/number-input/NumberInput.tsx
tayyabataimur: Yes the intention was to have parse just accept a string that would return a number. I can safely update that in the type definition 👍 

Will try to reduce the `number | string` complexity. It mainly stems from being able to support empty / blank inputs since that could either be considered null/undefined/empty string or 0.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
tayyabataimur: `onChangeEnd` is called when the user is typing values so in this example, since we are simulating typing 1000000, `onChangeEnd` will be called more than once. This would be the equivalent behaviour that we have in slider like when we are using the keyboard, `onChangeEnd` is called on every step

packages/lab/src/number-input/NumberInput.tsx
tayyabataimur: I have set defaultValue to be a number and have added an implementation for emptyReadOnlyMarker which is assigned to the display value.

packages/lab/src/number-input/NumberInput.tsx
tayyabataimur: @mark-tate Needs to be able to return a string as well in case zeros are padded at the end since `toFloat` gets rid of padded zeros.

joshwooding:
Converted to draft for now, until #5441 is resolved

PR #5347, opened 2025-07-15
Numberinput fractions demo (will be closed soon)

tayyabataimur:
POC - not intended to be merged - please don't review. Will be closed soon.

### DIFF
--- a/packages/lab/stories/number-input/number-input.stories.tsx
+++ b/packages/lab/stories/number-input/number-input.stories.tsx
@@ -510,16 +510,11 @@ export const Fractions: StoryFn<NumberInputProps> = (args) => {
             );
           }}
           parse={(val) => {
-            console.log("parsing: ", val);
+            const match = String(val).match(
+              /^\d+(-\d{0,2}(\d|\+)?|\d{0,2}(\d|\+)?)?$/,
+            );
+            if (!match) return;
             return Fraction32ndParser.from32nd(val.toString());
-          }}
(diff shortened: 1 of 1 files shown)


packages/lab/stories/number-input/number-input.stories.tsx
```diff
@@ -5,13 +5,15 @@ import {
...
   StackLayout,
+  Text,
   useId,
 } from "@salt-ds/core";
 import { AddIcon, RefreshIcon, RemoveIcon } from "@salt-ds/icons";
 import { NumberInput, type NumberInputProps } from "@salt-ds/lab";
 import type { Meta, StoryFn } from "@storybook/react-vite";
 import { toFloat } from "packages/core/src/slider/internal/utils";
 import { useState } from "react";
+import { Fraction32ndParser, RoundingEnum } from "./formatter";
```
mark-tate, line 15: This code I guess has come from friends.
My worry is that, this code gets copied and pasted around and has no tests.
It's too large/important/complex to be code that is copy and pasted.
So from a PR POV, this is a good POC but not sure we want this code to be part of our story code.

Lets have a discussion offline, could be an example we push to an internal storybook of our stakeholders, who own/maintain/test this code.

tayyabataimur: @mark-tate It's not meant to be in the story or our codebase. This PR will get closed once stakeholders have tested it.

PR #5356, opened 2025-07-16
Refactor multiline input

joshwooding:


libertymayc:
does it need a cs?

packages/core/src/multiline-input/MultilineInput.css
```diff
@@ -1,23 +1,32 @@
...
-  padding-left: var(--salt-spacing-100);
-  padding-right: var(--salt-spacing-100);
   position: relative;
   width: 100%;
   box-sizing: border-box;
   overflow: hidden;
+  padding: 0 var(--salt-spacing-100);
+}
+
+.saltMultilineInput {
```
tycup: Separate .saltMultilineInput for readability..?

packages/core/src/multiline-input/MultilineInput.css
```diff
@@ -183,8 +192,9 @@
 }
 
 .saltMultilineInput.saltMultilineInput-withAdornmentRow {
-  display: flex;
-  flex-wrap: wrap;
+  grid-template-areas: "start-adornment textarea textarea" "end-adornment end-adornment end-adornment";
+  grid-template-columns: min-content 1fr;
```
tycup, line 194: small thing; do we get an implicit column here silently added since you have three named areas? To satisfy number of columns implied by grid-template-areas?

packages/core/src/multiline-input/MultilineInput.css
joshwooding: It was but I've merged them

packages/core/src/multiline-input/MultilineInput.css
joshwooding: Because the area spans multiple columns you only need one value, but we can simplify the area definition

tycup (dismissed):

tayyabataimur (approved):

mark-tate (approved):

PR #5365, opened 2025-07-22
Fix weak/weakest tokens not aligned with figma

libertymayc:
These tokens should be -weakest, as they are in design

### DIFF
--- a/.changeset/eight-news-give.md
+++ b/.changeset/eight-news-give.md
@@ -1,0 +1,14 @@
+---
+"@salt-ds/theme": minor
+---
+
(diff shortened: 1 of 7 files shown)


.changeset/eight-news-give.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/theme": minor
```
libertymayc, line 2: is this patch or minor? @joshwooding

pseys (approved):

joshwooding (approved):

.changeset/eight-news-give.md
joshwooding: I'd class it as a minor due to the deprecation

PR #5410, opened 2025-08-05
Site a11y tweaks

joshwooding:


site/src/layouts/DetailComponent/ComponentPageHeading.tsx
```diff
@@ -61,7 +61,11 @@ export default function ComponentPageHeading({ title, id }: PageHeadingProps) {
             <Overlay>
               <Tooltip content="Theme controls">
```
jake-costa: tooltip doesn't seem to be displaying on focus for this. Other tooltips are working on focus.

jake-costa (approved):
Multiple A11y fixes added. Will create another PR for tooltip issue

PR #5418, opened 2025-08-05
Hide Tooltip if Content is null

chase-manning:
The tooltip currently doesn't show if you pass `undefined` or an empty string. However if you pass `null` it shows a weird error state on the UI, where it looks like a tiny empty tooltip with a gap between the arrow and the container. I suspect this behaviour is unexpected, at least it was for me.

This PR changes the Tooltip so that it also doesn't show if `null` is passed. I was going to add `&& !== null`, but I think that's functionally the same as just doing `!!content` so went with that as was a bit cleaner.

### DIFF
--- a/packages/core/src/tooltip/Tooltip.tsx
+++ b/packages/core/src/tooltip/Tooltip.tsx
@@ -124,7 +124,7 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
     const triggerRef = useForkRef(getRefFromChildren(children), reference);
 
     const floatingRef = useForkRef<HTMLDivElement>(floating, ref);
-    const hasContent = content !== undefined && content !== "";
+    const hasContent = !!content;
 
     return (
       <>


packages/core/src/tooltip/Tooltip.tsx
```diff
@@ -124,7 +124,7 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
     const triggerRef = useForkRef(getRefFromChildren(children), reference);
 
     const floatingRef = useForkRef<HTMLDivElement>(floating, ref);
-    const hasContent = content !== undefined && content !== "";
+    const hasContent = !!content;
```
joshwooding, line 127: This would make Tooltip with content `0` not render when previously it would (which is breaking the test). I think we can just do:


```suggestion
    const hasContent = content != null && content !== "";
```

joshwooding (changes requested):
See comment

packages/core/src/tooltip/Tooltip.tsx
mark-tate: what about returning early with just the children if no content?

We are adding elements to the dom that do nothing.  

Makes sense to render if content is defined and then you are just changing state for open/disabled but if there is no content there is an opportunity to not render a tooltip.

joshwooding, line 127:
```suggestion
    const hasContent = content != null && content !== "";
```

joshwooding: I think this makes sense but we'd need to check desktop support but also FloatingComponent returns null if it's not open so no extra dom elements should be present.

PR #5420, opened 2025-08-06
Add `textAlign` prop to TH and TD

### LINKED ISSUE #5389: Add numeric cell support to Table
### Description

The AG Grid theme supports numeric cells but the Table component doesn't

### Expected behavior

The text should be right aligned in a numeric cell.

### Alternatives and examples

.

### Context

.

joshwooding:
Closes #5389


site/docs/components/table/examples.mdx
```diff
@@ -1,3 +1,6 @@
+import {
+  LivePreview
+} from "../../../src/components/index";
 ---
```
joshwooding:
```suggestion
---
```

site/docs/components/table/examples.mdx
```diff
@@ -48,3 +48,13 @@ The `variant` prop can also be applied to the table footer. This allows you to c
...
 
 <LivePreview componentName="table" exampleName="TFootVariant" />
+
+## Text alignment
+
+You can control the text alignment of table cells and headers using the `textAlign` prop. It accepts: `left` or `right`.
+
+## Best practices
+
+- Use `textAlign="right"` for numeric data to improve readability.
```
joshwooding:
```suggestion
-  Numeric data should be right-aligned to improve readability.
```

packages/lab/src/table/Table.css
```diff
@@ -199,4 +203,9 @@ table.saltTable-divider-none tbody tr {
 table.saltTable td {
   height: 100%;
   padding: var(--salt-spacing-100);
+  text-align: start;
```
libertymayc, line 220: is it worth putting this/th equiv in `table.saltTable td.saltTable-td-align-left` otherwise that class is added but redundant?

joshwooding: I don't feel strongly about it, moving it to a new class increases the amount of CSS for the component (only by a small amount). 

There are also other places in the code where we haven't scoped our styling to another class.

I don't think it's redundant though, the class can be used later or used if we support component overrides.

libertymayc (dismissed):

packages/lab/src/table/TD.tsx
```diff
@@ -6,10 +6,17 @@ import { withTableBaseName } from "./Table";
...
 import tableCss from "./Table.css";
 
-export interface TDProps extends ComponentPropsWithoutRef<"td"> {}
+export interface TDProps extends ComponentPropsWithoutRef<"td"> {
+  /**
+   * Specifies the alignment of the text within the `TD`.
+   *
+   * @default "left"
+   */
+  textAlign?: "left" | "right";
```
mark-tate, line 15: we have props elsewhere called “align” with “left”, “right” and here we have “text-align”
which maps to CSS, but that uses”start”, “end”, so feels
likes the wires are crossed on our naming ?


Good one to discuss on our bi-weekly, how we approach naming.

joshwooding: This aligns with the Input textAlign prop https://github.com/jpmorganchase/salt-ds/blob/main/packages/core/src/input/Input.tsx#L57

libertymayc (approved):

PR #5468, opened 2025-08-19
Update Biome.js

joshwooding:


biome.jsonc
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.1.2/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.2.0/schema.json",
```
tycup: Out of curiosity, why don't we get this from node_modules?

cypress/support/commands.tsx
```diff
@@ -57,10 +57,9 @@ declare global {
 
       /**
        * Set the date locale used by the date adapter
-       * @param any
+       * @param locale any
```
tycup:
```suggestion
       * @param locale unknown
```

biome.jsonc
joshwooding: No particular reason, it was the default option when we first started using biome

cypress/support/commands.tsx
joshwooding: I was thinking about this but was too lazy to check if unknown was supported. Update this as the jsdoc format is wrong anyway

tycup (approved):

packages/lab/src/__tests__/__e2e__/tree/Tree.cy.tsx
```diff
@@ -1,13 +1,13 @@
...
-  }
-  return nodes;
-}
+// function createNodes(count: number, factory: (index: number) => any) {
+//   const nodes = [];
+//   for (let i = 0; i < count; i++) {
+//     nodes.push(factory(i));
+//   }
+//   return nodes;
+// }
```
tayyabataimur, line 10: Out of curiosity, why is this and the `noop` callback below commented out instead of removed?

joshwooding: It's in the code below that's also commented out so didn't want to remove it. Once we work on Tree and re-write the tests this will all change

PR #5469, opened 2025-08-19
Highcharts | Add PieChart

tycup:
- DonutChart is a variant of PieChart, so the `plotOptions.pie.innerSize` setting is moved out of the global options; otherwise, a default PieChart would automatically become a DonutChart. Consumers are instructed to supply the `innerSize` setting themselves in documentation.

- Documentation reordered to show PieChart first due to the above.

### DIFF
--- a/.changeset/quiet-hairs-try.md
+++ b/.changeset/quiet-hairs-try.md
@@ -1,0 +1,8 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+- Added PieChart
+- Since DonutChart is a variant of PieChart,the plotOptions.pie.innerSize setting is moved out of the global options; otherwise, a default PieChart would automatically become a DonutChart
+- Consumers are instructed to supply the innerSize setting themselves in documentation.
+- Documentation reordered to show PieChart first due to the above.
--- a/packages/highcharts-theme/src/highcharts-options-salt.ts
+++ b/packages/highcharts-theme/src/highcharts-options-salt.ts
@@ -21,6 +21,7 @@ export const saltThemeDefaults: HighchartsOptionsCompat = {
     headerFormat: "<span>{point.key}</span><br/>",
     pointFormat:
       '<span>{series.name}: </span><span class="value">{point.y}</span>',
+    stickOnContact: true,
   },
   legend: {
     layout: "vertical",
@@ -55,7 +56,6 @@ export const saltThemeDefaults: HighchartsOptionsCompat = {
           '{point.name} <span class="value">{point.percentage:.1f}%</span>',
         enabled: true,
       },
-      innerSize: "80%",
     },
   },
   title: {
--- a/site/docs/components/highcharts-theme/examples.mdx
+++ b/site/docs/components/highcharts-theme/examples.mdx
@@ -21,9 +21,21 @@ Use a line chart to show changes in data over time. To ensure accessible present
 
 <LivePreview componentName="highcharts-theme" exampleName="LineChart" />
 
+## Pie Chart
+
+Use a pie chart to show proportions of a whole. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
+<LivePreview componentName="highcharts-theme" exampleName="PieChart" />
+
 ## Donut Chart
 
-Use a donut chart to show proportions of a whole. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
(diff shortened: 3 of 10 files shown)


site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -21,9 +21,17 @@ Use a line chart to show changes in data over time. To ensure accessible present
...
+Use a pie chart to show proportions of a whole. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
+<LivePreview componentName="highcharts-theme" exampleName="PieChart" />
+
 ## Donut Chart
 
-Use a donut chart to show proportions of a whole. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+Use a donut chart to show proportions of a whole, with space in the center to highlight key values or additional information. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
+A donut chart is a variation of the pie chart, with the `plotOptions.pie.innerSize` set to `80%` within the `Options` object.
```
joshwooding: Do we care about it being specifically 80%?

tycup: Yep, this is the value necessary to align with the stickersheet (but also i could see designers have an opinion on this value changing depending on content inside the donut hole, so it represents a default that is likely to be adjusted based on use-case). CC: @amritadesmet

amritadesmet: Yes, in Figma there's a sticker guide I created on how to adjust the segment thickness in case consumers want to change that.

joshwooding: In that case I would change this sentence to something like:

```suggestion
A donut chart is a variation of the pie chart, with the `plotOptions.pie.innerSize` set to less than `100%` within the `Options` object.

## Best practices

- 80% `innerSize` provides a good default visual balance.
```

tycup: Meet in the middle? 

```
A donut chart is a variation of the pie chart, with the `plotOptions.pie.innerSize` set to less than `100%` within the `Options` object.

## Best practices
- Use innerSize: "80%" to align with the Salt Figma library.
```

tycup: (Good default sounds a bit too generic for me; its a precise, intentional default without which you'll be out of sync with the default sticker a designer is most likely to use)

joshwooding: How is it precise? It's a value we chose that we feel is good :) We shouldn't focus too much on figma vs code but look at it broadly.

Happy to adjust the language and talk about a recommendation instead.

tycup: I see your point - by precise, i meant matching figma. Then again, the whole point of using a donut vs a pie is to provide space for content in the middle, so i could see designers adjusting to accommodate depending on what they decide to place there. I wouldn't even be opposed to getting rid of any mention of the 80% value, since its a variable that consumers can change and its already visible in the code snippet.

but tldr; recommendation is better, thanks:) - ill update.

tycup: Stuck with "Best practices" heading to stay consistent with other docs (grid-layout also has a "best practices" header followed by a recommendation) and slightly reworded (i think starting a sentence with a numerical value breaks up the flow a little bit)  - take a look @joshwooding and hit resolve if fine

.changeset/quiet-hairs-try.md
```diff
@@ -0,0 +1,8 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+- Add PieChart
```
joshwooding:
```suggestion
- Added PieChart
```

packages/highcharts-theme/src/highcharts-options-salt.ts
```diff
@@ -21,6 +21,7 @@ export const saltThemeDefaults: HighchartsOptionsCompat = {
     headerFormat: "<span>{point.key}</span><br/>",
     pointFormat:
       '<span>{series.name}: </span><span class="value">{point.y}</span>',
+    stickOnContact: true,
```
joshwooding, line 24: We should add this on our list to document :)

joshwooding: I would create a list of stuff like this so we don't forget @jake-costa might need your help when we get to the a11y side of the docs

joshwooding (approved):

joshwooding (dismissed):

joshwooding (approved):

Issue #4736, opened 2025-02-20
Input Validation Status not conveyed via Screen reader


jake-costa:
### Latest version

- [x] I have tested the latest version

### Description

Input Validation Status not conveyed via Screen reader. [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)

Error status should have aria-invalid
For warning, and success. We could use aria-describedby to target the aria--label of the icon or use another method to add hidden accessible text to convey the status.

Also, to support [WCAG 4.1.3 Status Messages](https://www.w3.org/WAI/WCAG22/Understanding/status-messages.html), After navigating away from an empty required form field for collecting the user's first name, an error icon is added to the form field. The screen reader should announce "error: first name required." for example.

### Steps to reproduce

Using a screen reader, navigate to an input field that contains an error, warning, or success. notice that the status is not conveyed. 

### Package name(s)

Core (@salt-ds/core)

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

bhoppers2008:
Migrated to Jira.

PR #5497, opened 2025-08-28
Number Input revisions

mark-tate:


site/src/examples/number-input/MinMax.tsx
```diff
@@ -0,0 +1,76 @@
...
+      validationStatus={getValidationStatus()}
+      style={{ width: "256px" }}
+    >
+      <FormFieldLabel>Number input with limited range</FormFieldLabel>
+      <NumberInput
+        value={value}
+        onChange={(event: ChangeEvent<HTMLInputElement>) => {
+          setValue(event.target.value);
+        }}
+        onNumberChange={(newValue) => {
```
tayyabataimur: In this example, when I type a number greater than the max (e.g. 100), the value is clamped on blur to 5 and the input shows an error state. Then, when I focus in the input and press down arrow key, I get 5, 5, 4, 3, 2, 1 etc. 

Seems like clamping is always happening on blur and the `clamp` prop isn't working correctly..? Ideally, the behaviour I would expect without clamping would be that when I type a value greater than max, it retains that invalid value on blur and the input goes into an error state.

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -100,32 +100,30 @@ export interface NumberInputProps
    * @default Number.MIN_SAFE_INTEGER
    */
   min?: number;
+  /**
+   * Callback function that is triggered when the value of the input changes.
+   *
+   * @param event - The event that triggers the value change, providing access to the unformatted value.
```
tayyabataimur:
```suggestion
   * @param event - The event that triggers the value change, providing access to the raw input value.
```

Slight rephrase because "unformatted" gives the impression that it might be parsed?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -100,32 +100,30 @@ export interface NumberInputProps
    * @default Number.MIN_SAFE_INTEGER
    */
   min?: number;
+  /**
+   * Callback function that is triggered when the value of the input changes.
```
tayyabataimur:
```suggestion
   * Callback function that is triggered when the value of the `NumberInput` changes.
```

To be consistent with the prop description of `onNumberChange`. Or both can be changed to `input`. Just want to ensure users know that both are triggered when the value changes.

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +98,61 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
+
+<LivePreview componentName="number-input" exampleName="FixedDecimalScale" />
+
 ## Formatting
 
-You can format the value of the `NumberInput` using the `format` callback prop. It typically works alongside a `parse` callback for complex formatting.
+The `NumberInput` component offers the ability to format its value using the `format` callback prop. This feature is typically used in conjunction with a `parse` callback, which can be used to determine the valid input format and extract the numerical value of the formatted input.
 
-Users will be able to see the formatted value in the input, e.g. "5K". When the input gains focus, the value will be parsed into its numerical form, so "5K" becomes "5000". This allows users to edit the values with precision. When the input loses focus, the value will be formatted again.
+### Best Practice
```
tayyabataimur: Should we mention here that you can access the parsed numerical value via `onNumberChange` callback?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -500,22 +567,26 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
         {!isReadOnly && (
           <div className={clsx(withBaseName("buttonContainer"))}>
             <Button
-              className={clsx(
-                withBaseName("numberButton"),
-                withBaseName("numberButtonIncrement"),
-              )}
+              aria-hidden={true}
+              aria-label={"increment value"}
```
tayyabataimur: Increment/decrement buttons are `aria-hidden` so they don't need an `aria-label` since we don't need the screen reader to announce them.
```suggestion
```

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -500,22 +567,26 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
             >
               <IncreaseIcon aria-hidden />
             </Button>
             <Button
-              className={clsx(
-                withBaseName("numberButton"),
-                withBaseName("numberButtonDecrement"),
-              )}
+              aria-hidden={true}
+              aria-label={"decrement value"}
```
tayyabataimur:
```suggestion
```

packages/lab/stories/number-input/number-input.stories.tsx
```diff
@@ -378,62 +450,84 @@ export const CustomButtons: StoryFn<NumberInputProps> = (args) => {
...
+    <StackLayout gap={2}>
       <FormField>
-        <FormFieldLabel>With compact notation, controlled</FormFieldLabel>
+        <FormFieldLabel>Compact notation</FormFieldLabel>
         <NumberInput
           {...args}
           value={value}
-          onChange={(e, value) => {
-            setValue(value);
+          isAllowed={(inputValue) => validValue.test(inputValue)}
```
tayyabataimur: Currently, `isAllowed` is only being called in `handleInputChange`. This means that if you pass an invalid value as default to the input, it will be rendered. Incrementing or decrementing from it will result in NaN. You can test this by passing `abc` as default value and then try to increment or decrement. 

We should have a mechanism in place to clean up invalid, default values too.

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -246,139 +271,150 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
     const [value, setValue] = useControlled({
-      controlled: valueProp,
-      default: defaultValue,
+      controlled: valueProp !== undefined ? String(valueProp) : undefined,
+      default: String(defaultValueProp ?? ""),
       name: "NumberInput",
       state: "value",
     });
+    const lastNumberChange = useRef<{
+      value: number | null;
```
tayyabataimur:
```suggestion
      parsedValue: number | null;
```
Nit (not mandatory): Should `value` be renamed to something more descriptive?

packages/lab/stories/number-input/number-input.stories.tsx
```diff
@@ -186,53 +230,76 @@ export const MinAndMaxValue: StoryFn<NumberInputProps> = (args) => {
...
-        <FormFieldLabel>With clamping</FormFieldLabel>
-        <NumberInput
-          {...args}
-          defaultValue={2}
           clamp
           max={max}
           min={min}
           style={{ width: "250px" }}
+          inputProps={{
+            onBlur: () => {
```
tayyabataimur: Don't want to go into a11y too much but the screen reader announcements aren't working as expected. e.g. when I type 500, it does not announce "500 is greater than the maximum value". When I click outside the input (on blur) and it clamps the value back to 5, it announces "Maximum value reached", whereas it should have announced "500 is greater than the maximum value, value was set to 5".

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -167,40 +176,55 @@ describe("Number Input", () => {
...
+    );
 
     cy.findByLabelText("decrement value").realClick();
-    cy.get("@changeSpy").should(
-      "have.been.calledWith",
-      Cypress.sinon.match.any,
-      15,
-    );
+    cy.get("@changeSpy").should("not.have.been.called");
+    cy.get("@numberChangeSpy").should("have.been.calledWith", 15);
```
tayyabataimur: I have mixed feelings about the inconsistent event handling. `onChange` is only called when the user edits the input (i.e. by typing?) whereas `onNumberChange` is called on blur and on increment/decrement. In both their prop descriptions, it says they are called when the value of the input changes. 

Has this pattern been used elsewhere? Should the documentation reflect more accurately how these event handlers work?

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -167,40 +176,55 @@ describe("Number Input", () => {
   });
 
   it("calls the `onChange` callback when the value is decremented", () => {
```
tayyabataimur: Test description should be updated since we are asserting that `onChange` should not have been called and `onNumberChange` should have been called.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -223,77 +247,118 @@ describe("Number Input", () => {
...
+        defaultValue={Number.MAX_SAFE_INTEGER + 1}
+        onNumberChange={numberChangeSpy}
+      />,
+    );
+
+    cy.findByRole("spinbutton").should(
+      "have.value",
+      Number.MAX_SAFE_INTEGER.toString(),
+    );
+    cy.get("@numberChangeSpy").should("not.have.been.called");
```
tayyabataimur:
```suggestion
```

We haven't performed any action yet so this assertion seems redundant.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -223,77 +247,118 @@ describe("Number Input", () => {
...
+
+    cy.findByRole("spinbutton")
+      .clear()
+      .focus()
+      .realType("-83491232183712983713791237");
+    cy.realPress("Tab");
+    cy.findByRole("spinbutton").should(
+      "have.value",
+      Number.MIN_SAFE_INTEGER.toString(),
+    );
```
tayyabataimur, line 288: Should `@numberChangeSpy` call assertion be added here?

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -323,36 +388,30 @@ describe("Number Input", () => {
   });
 
   it("sanitizes input to only allow numbers, decimal points, and plus/minus symbols", () => {
-    cy.mount(<Default />);
-
-    cy.findByRole("spinbutton").focus();
-    cy.findByRole("spinbutton").clear();
-    cy.realType("abc-12.3.+-def");
+    cy.mount(<Default defaultValue={undefined} />);
```
tayyabataimur:
```suggestion
    cy.mount(<Default />);
```
Is this needed?

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -323,36 +388,30 @@ describe("Number Input", () => {
...
-  it("sanitizes default value in controlled to only allow numbers, decimal points, and plus/minus symbols", () => {
+  it("as a controlled component, emulates native input and cannot type non-numeric but can set an invalid value through default value", () => {
     cy.mount(<Default value={"abc-12.3.+-def"} />);
-    cy.findByRole("spinbutton").should("have.value", "-12.3");
-  });
-
-  it("sanitizes default value in controlled to only allow numbers, decimal points, and plus/minus symbols", () => {
-    cy.mount(<Default value={"abc-12.3.+-def"} />);
-    cy.findByRole("spinbutton").should("have.value", "-12.3");
+    cy.findByRole("spinbutton").should("have.value", "abc-12.3.+-def");
```
tayyabataimur, line 476: This is not the case in other libraries (e.g. mantine). They clean up default values in controlled state too. However, if we do wish to go this route, perhaps add graceful error handling when incrementing/decrementing since it currently results in NaN state.

tayyabataimur: Here's a sandbox with Mantine's number input: https://codesandbox.io/p/sandbox/romantic-blackburn-fhqfzg

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -541,41 +641,73 @@ describe("Number Input", () => {
...
+      cy.get("@numberChangeSpy").should("have.been.calledWith", 14);
+      cy.realPress("Tab");
+      cy.findByRole("spinbutton").should("have.value", "14%");
+    });
 
-      cy.get("@changeSpy").should(
-        "have.been.calledWith",
-        Cypress.sinon.match.any,
-        14,
+    it("should allow typing formatted values and call onChange with correct numerical values", () => {
```
tayyabataimur, line 832: Test description can be updated since `onChange` doesn't provide numerical values and we are testing for both `onChange` and `onNumberChange`.

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -597,16 +729,18 @@ describe("Number Input", () => {
     });
 
     it("displays value with correct number of decimal places when decimal scale is set", () => {
-      cy.mount(<Default decimalScale={2} defaultValue={""} />);
+      cy.mount(<Default defaultValue={undefined} decimalScale={2} />);
```
tayyabataimur:
```suggestion
      cy.mount(<Default decimalScale={2} />);
```

packages/lab/stories/number-input/number-input.stories.tsx
```diff
@@ -22,7 +21,13 @@ const Template: StoryFn = ({ ...args }) => {
   return (
     <FormField>
       <FormFieldLabel>Number input</FormFieldLabel>
-      <NumberInput {...args} />
+      <NumberInput
+        defaultValue={0}
```
tayyabataimur:
```suggestion
```
Should we leave default as is?

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -597,16 +729,18 @@ describe("Number Input", () => {
...
-      cy.realType("-12");
+      cy.realType("-12.1234");
       cy.realPress("Tab");
-      cy.findByRole("spinbutton").should("have.value", "-12.00");
+      cy.findByRole("spinbutton").should("have.value", "-12.12");
     });
 
-    it("pads with zeros to correct number of decimal places when decimal scale is set", () => {
-      cy.mount(<Default decimalScale={3} defaultValue={-5.8} />);
+    it("pads with zeros to correct number of decimal places when fixedDecimalScale is set", () => {
```
tayyabataimur: This test can be removed/updated since the API doesn't support `fixedDecimalScale` anymore.

site/docs/components/number-input/examples.mdx
```diff
@@ -28,6 +28,44 @@ By default, the `NumberInput` is transparent. You can add borders using the `bor
...
+To maintain a cohesive visual flow, numeric values are left-aligned by default, positioned directly beneath the label and other form field elements. If your specific use case or layout requires a different alignment, you can adjust the `NumberInput` to be right or center aligned using the `textAlign` prop.
+
+### Best practices
+
+- Right alignment is often used for numeric values to facilitate easier comparison. However, if number fields are mixed with text fields or do not relate to each other, left alignment may enhance the overall visual flow.
+- Center alignment can be used to emphasize a particular value, but it should be applied sparingly and only in exceptional cases to avoid disrupting the visual consistency.
+
+<LivePreview componentName="number-input" exampleName="TextAlignment" />
+
+## Customise input value
```
tayyabataimur: This should come under one of the example cases of `Formatting` where there is already an example of suffix/prefix.

packages/lab/stories/number-input/number-input.stories.tsx
tayyabataimur: @mark-tate  took a look at it and it's a funny one because `onNumberChange` is triggered on blur and it makes the screen reader announce "Max/min limit reached", so it is basically overriding the announcement that we try to trigger via the `onBlur` prop.....We could move the "reached" announcement to on blur but then the screen reader won't announce it when incrementing/decrementing.

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
 
 <LivePreview componentName="number-input" exampleName="Validation" />
 
-## Decimal scale
+## Decimal Scale
```
tayyabataimur:
```suggestion
## Decimal scale
```

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
+
+<LivePreview componentName="number-input" exampleName="FixedDecimalScale" />
+
 ## Formatting
 
-You can format the value of the `NumberInput` using the `format` callback prop. It typically works alongside a `parse` callback for complex formatting.
+The `NumberInput` component offers the ability to format its value using the `format` callback prop. This feature is typically used in conjunction with a `parse` callback, which can be used to determine the valid input format and extract the numerical value of the formatted input.
 
-Users will be able to see the formatted value in the input, e.g. "5K". When the input gains focus, the value will be parsed into its numerical form, so "5K" becomes "5000". This allows users to edit the values with precision. When the input loses focus, the value will be formatted again.
+### Best Practice
```
tayyabataimur:
```suggestion
### Best practice
```

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
-However, you can set it to any value of your choice using the `decimalScale` prop and that will have priority.
+The `decimalScale` prop allows you to define the number of decimal places displayed in the `NumberInput`. By default, the decimal scale is determined by the greater number of decimal places between the `step` value and the `value` or `defaultValue` provided.
 
 ### Best practices
 
-When configuring the number of decimal places, ensure it aligns with the specific use case. For example, when dealing with currencies such as pounds sterling or U.S. dollars in a retail context, limit the decimal places to two, as these currencies typically require only two decimal places for accuracy. Avoid using three decimal places in such scenarios to maintain consistency and relevance.
+- when dealing with currencies such as pounds sterling or U.S. dollars in a retail context, limit the decimal places to two, as these currencies typically require only two decimal places for accuracy. Avoid using three decimal places in such scenarios to maintain consistency and relevance.
```
tycup: Capitalise 'when'

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
 
 <LivePreview componentName="number-input" exampleName="Decimals" />
 
+To provide a fixed scale provide a custom format function.
```
tycup: Either a comma should be added here "To provide a fixed scale, provide a custom format function" or not sure about double "provide" - maybe "To use a fixed scale, provide a custom format function"?

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
 ## Formatting
 
-You can format the value of the `NumberInput` using the `format` callback prop. It typically works alongside a `parse` callback for complex formatting.
+The `NumberInput` component offers the ability to format its value using the `format` callback prop. This feature is typically used in conjunction with a `parse` callback, which can be used to determine the valid input format and extract the numerical value of the formatted input.
 
-Users will be able to see the formatted value in the input, e.g. "5K". When the input gains focus, the value will be parsed into its numerical form, so "5K" becomes "5000". This allows users to edit the values with precision. When the input loses focus, the value will be formatted again.
+### Best Practice
 
-### Best practices
+The `format` callback is designed to work seamlessly with most formatting libraries, such as Intl.NumberFormat. It is strongly recommended to provide a `parse` callback alongside `format` to ensure that valid input is entered and its correct numerical value is derived, maintaining precision and consistency.
```
tycup, line 115:
```suggestion
The `format` callback is designed to work seamlessly with most formatting libraries, such as `Intl.NumberFormat`. It is strongly recommended to provide a `parse` callback alongside `format` to ensure that valid input is entered and its correct numerical value is derived, maintaining precision and consistency.
```

site/docs/components/number-input/examples.mdx
```diff
@@ -58,30 +91,67 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
+<LivePreview
+  componentName="number-input"
+  exampleName="FormattingThousandsSeparator"
+/>
 
-The `format` callback has been designed to work with most formatting libraries, e.g. Intl.NumberFormat. However, it is highly recommended to provide a `parse` callback alongside `format` to determine the correct numerical value so the precision of the value can be maintained.
+### Compact notation
 
-**Note**: If a `format` callback is provided, it will override the `decimalScale`, enabling you to handle decimal place rules directly within your formatting function.
+Large numbers can be transformed into their compact forms to be used in, for example, charts or other data visualizations. You can use `format` to convert the numerical value into its shortened form and `parse` to convert the shortened form back to the numerical value. The example below supports the following conversion:
```
tycup: Would reorder this sentence a little bit for clarity and tie it directly with the headline (compact notation for shortened form)

```suggestion
You can use `format` to convert the numerical value into its compact notation and `parse` to convert compact notation back to its numerical value. This is useful when used with charts or other forms of data visualization. The example below supports the following compact notations:

```

site/docs/components/number-input/usage.mdx
```diff
@@ -37,7 +39,41 @@ To avoid misleading users, set the width to correlate to the length of the value
 
 Just like native input elements in React, number inputs can be controlled or uncontrolled. By default a number input is uncontrolled. Use the `defaultValue` prop to control the default value of the number input.
 
-If you want to control the value of the number input, use the `value` prop. The `onChange` prop is required when using the controlled approach.
+If you want to control the value of the number input:
+
+- use the `value` prop to provide the current value.
```
tycup: I think ive seen this in other docs too so can ignore but - is it an inconsistency or a rule to go between capitalising the first word in a bullet point list and not..?

site/docs/components/number-input/examples.mdx
tayyabataimur: @tycup  We reserve backticks for component and prop names

packages/lab/stories/number-input/number-input.stories.tsx
```diff
@@ -291,6 +291,9 @@ export const Clamping: StoryFn<NumberInputProps> = (args) => {
           onNumberChange={(newValue) => {
             console.log(`Number changed to ${newValue}`);
             setValue(newValue ?? "");
+            if (newValue === null) {
```
mark-tate: this does not do anything ?

tayyabataimur: Yeah, it was in the example previously. Have removed it

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -487,33 +487,36 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
+      disableDecrement: boolean,
+    ) => {
       event.preventDefault();
-      setIsEditing(false);
-      activateDecrement(event);
+      if (!disableDecrement) {
+        setIsEditing(false);
+        activateDecrement(event);
+      } else if (inputRef.current) {
+        inputRef.current.select();
```
tayyabataimur, line 507: Can we remove this from `handleDecrementMouseDown` and `handleIncrementMouseDown` because when the buttons get disabled and a user clicks on them, the text is being selected because of this.

tayyabataimur: @mark-tate is the above needed?

mark-tate: Lets gets some feedback, if you double click on the input, it should select the text, once the buttons are disabled, then double-clicking should be equivalent to double clicking the input, which would result in the value being selected.

.storybook/decorators/withResponsiveWrapper.tsx
```diff
@@ -37,10 +37,14 @@ const ResponsiveContainer = ({ children }: { children?: ReactNode }) => {
           </Tooltip>
         </ToggleButtonGroup>
         <NumberInput
-          value={containerWidth[0]}
+          value={String(containerWidth[0])}
           max={maxUnits}
           min={10}
-          onChange={(_event, nextValue) => setWidth([nextValue as number])}
+          onNumberChange={(_e, value) => {
```
wemmyo, line 43: Might be out of scope for this ticket but generally I've found it better to have the `value` as the first parameter since the `event` is usually ignored anyway.

joshwooding: It's something we've discussed as something in the future but event being first is the standard across all* our APIs.

* where possible...

packages/lab/src/number-input/NumberInput.css
```diff
@@ -250,25 +250,13 @@
   flex-direction: column;
   gap: var(--numberInput-buttonGap);
   justify-content: center;
+  cursor: var(--salt-cursor-disabled);
```
joshwooding: Is this right?

packages/lab/src/number-input/NumberInput.css
```diff
@@ -290,3 +278,32 @@
...
   border-bottom-width: 0;
 }
+
+/* Styles applied to number buttons */
+.saltNumberInput-decrement {
+  --saltButton-height: calc((var(--salt-size-base) - (var(--salt-spacing-50) * 2) - var(--salt-size-fixed-200)) / 2);
+  --saltButton-padding: calc(var(--salt-spacing-50) - var(--button-borderWidth));
+  --saltButton-borderRadius: 0 0 var(--salt-palette-corner-weaker, 0) var(--salt-palette-corner-weaker, 0);
+}
+.saltButton.saltNumberInput-decrement-disabled:hover,
```
joshwooding: Is there a reason we don't just use a disabled button? Is it due to the focus behaviour?

packages/lab/src/number-input/NumberInput.css
```diff
@@ -290,3 +278,32 @@
...
+.saltButton.saltNumberInput-decrement-disabled:active,
+.saltButton.saltNumberInput-decrement-disabled {
+  background: var(--saltButton-background-disabled, var(--button-background-disabled));
+  color: var(--saltButton-text-color-disabled, var(--button-text-color-disabled));
+  cursor: var(--saltButton-cursor-disabled, var(--salt-cursor-disabled));
+  border-color: var(--saltButton-borderColor-disabled, var(--button-borderColor-disabled));
+}
+
+.saltNumberInput-increment {
```
joshwooding: NIT: this could be merged with .saltNumberInput-decrement e.g.

```css
.saltNumberInput-decrement,
.saltNumberInput-increment {
```

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -163,29 +165,64 @@ export interface NumberInputProps
...
+      clamp,
+      decimalScale: decimalScaleProp,
       disabled,
       emptyReadOnlyMarker = "—",
       endAdornment,
-      format,
+      format: formatProp,
       hideButtons,
       id: idProp,
+      pattern = (inputValue) =>
```
joshwooding: We should probably move the default outside the component

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -246,139 +276,189 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
     });
+    const lastCommitValue = useRef<string>(value);
 
     const decimalScale =
       decimalScaleProp ||
       Math.max(getNumberPrecision(value), getNumberPrecision(step));
 
-    const [displayValue, setDisplayValue] = useState<string | number>(value);
+    const defaultFormat = (value: string): string => {
+      if (!value.length) {
```
joshwooding: Is there a chance a string with spaces gets passed here e.g. "    "

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -246,139 +276,189 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
 
     const [value, setValue] = useControlled({
-      controlled: valueProp,
-      default: defaultValue,
+      controlled: valueProp !== undefined ? String(valueProp) : undefined,
+      default: String(defaultValueProp ?? ""),
       name: "NumberInput",
       state: "value",
     });
+    const lastCommitValue = useRef<string>(value);
```
joshwooding: NIT: I would prefer this being closer to the actual `commit` function.

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -418,10 +500,49 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
+      if (!disableDecrement) {
+        setIsEditing(false);
+        activateDecrement(event);
+      } else if (inputRef.current) {
+        inputRef.current.select();
+      }
+    };
+
+    const handleContainerMouseUp = () => {
+      setIsFocused(true);
```
joshwooding, line 514: We need to destructure and call onMouseUp here

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -446,26 +569,17 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
         <input
-          aria-describedby={clsx(
+          aria-describedby={clsx(formFieldDescribedBy, inputDescribedBy)}
+          aria-labelledby={clsx(
+            formFieldLabelledBy,
+            inputLabelledBy,
             validationStatusId,
-            formFieldDescribedBy,
-            inputDescribedBy,
           )}
```
joshwooding:
```suggestion
          aria-describedby={clsx(formFieldDescribedBy, inputDescribedBy) || undefined}
          aria-labelledby={clsx(
            formFieldLabelledBy,
            inputLabelledBy,
validationStatusId,
) || undefined}
```

NIT: This means we don't get empty attributes

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -446,26 +569,17 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
           </div>
         )}
         <input
-          aria-describedby={clsx(
+          aria-describedby={clsx(formFieldDescribedBy, inputDescribedBy)}
+          aria-labelledby={clsx(
+            formFieldLabelledBy,
+            inputLabelledBy,
             validationStatusId,
```
joshwooding: We don't do this for any other form control, we should probably be consistent and look at this in https://github.com/jpmorganchase/salt-ds/issues/4736

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -446,26 +569,17 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
-            formFieldDescribedBy,
-            inputDescribedBy,
           )}
-          aria-labelledby={clsx(formFieldLabelledBy, inputLabelledBy)}
           aria-invalid={
-            !isReadOnly
+            !isReadOnly && renderedValue.length
               ? isOutOfRange(value, min, max) || validationStatus === "error"
```
joshwooding: This is different to other form controls true (Setting aria-invalid when validationStatus is set)

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -446,26 +569,17 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
+            formFieldLabelledBy,
+            inputLabelledBy,
             validationStatusId,
-            formFieldDescribedBy,
-            inputDescribedBy,
           )}
-          aria-labelledby={clsx(formFieldLabelledBy, inputLabelledBy)}
           aria-invalid={
-            !isReadOnly
+            !isReadOnly && renderedValue.length
```
joshwooding, line 564: Should this be value rather than renderedValue?

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -475,17 +589,32 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
           onFocus={handleInputFocus}
-          onKeyDown={handleInputKeyDown}
-          onBeforeInput={handleBeforeInput}
+          onKeyDown={isReadOnly ? undefined : handleInputKeyDown}
           placeholder={placeholder}
           readOnly={isReadOnly}
           aria-readonly={isReadOnly ? "true" : undefined}
           ref={handleInputRef}
           required={isRequired}
+          {...(!isReadOnly &&
```
joshwooding: shorthand spreads are generally okay, but we probably should refactor this to make it easier for people not used to the syntax

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -500,22 +629,30 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
-                withBaseName("numberButton"),
-                withBaseName("numberButtonIncrement"),
-              )}
+              aria-hidden={true}
+              aria-label={"increment value"}
               appearance="transparent"
-              {...incrementButtonProps}
+              tabIndex={-1}
+              className={clsx(withBaseName("increment"), {
+                [withBaseName("increment-disabled")]: disableIncrement,
```
joshwooding: Rather than this we might want to use aria-disabled if disabled doesn't work

packages/lab/stories/date-picker/CustomDatePickerPanel.tsx
```diff
@@ -74,6 +74,7 @@ export const CustomDatePickerPanel = forwardRef<
               List header
             </FormFieldLabel>
             <ListBox
+              tabIndex={0}
```
joshwooding: ?

packages/lab/stories/number-input/number-input.qa.stories.tsx
```diff
@@ -7,52 +7,64 @@ export default {
   component: NumberInput,
 } as Meta<typeof NumberInput>;
 
-export const ExamplesGrid: StoryFn<QAContainerProps> = (props) => {
+const formatWithDecimalScale =
+  (decimalScale: number) => (value: string | null) => {
+    if (value === null) {
+      return "";
```
joshwooding: I was going to ask about this in the tests, should we always return "" rather than null?

site/docs/components/number-input/examples.mdx
```diff
@@ -60,34 +93,71 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
+You can support custom input patterns, such as "10x", using a combination of `pattern`, `format` and `parse` callbacks.
+
+<LivePreview componentName="number-input" exampleName="CustomInputPattern" />
 
 ## Custom steps
 
 The `step` prop allows you to define the increment and decrement sizes for value adjustments.
 
-### Best practices
+### Best practice
```
joshwooding: I understand why this was changed but this is different to what we do elsewhere

site/src/examples/number-input/ButtonAdornments.tsx
```diff
@@ -4,122 +4,99 @@ import {
...
-              aria-live="polite"
-              style={accessibleTextStyles}
-            >
-              {accessibleText}
-            </span>
-          </>
+          <Button
+            appearance="solid"
+            aria-label={`Reset ${formFieldLabel}`}
+            onClick={() => {
```
joshwooding, line 36: We can use the self reference trick here

site/src/examples/number-input/MinMax.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const MinMax = () => {
+  const [value, setValue] = useState<string>("2");
+
+  const { announce } = useAriaAnnouncer();
+
+  const max = 5;
+  const min = 0;
+
+  const getValidationStatus = () => {
```
joshwooding: NIT: this doesn't need to be a function

packages/lab/src/number-input/NumberInput.css
mark-tate: does not look like it... removed

packages/lab/src/number-input/NumberInput.css
mark-tate: done

packages/lab/src/number-input/NumberInput.tsx
mark-tate: done

packages/lab/src/number-input/NumberInput.tsx
mark-tate: in general no, because the default pattern dis-allows it BUT you could do it via a controlled value or defaultValue.
So for the sake of avoiding an issue will sanitize the value.

packages/lab/stories/date-picker/CustomDatePickerPanel.tsx
mark-tate: doh, removed

packages/lab/stories/number-input/number-input.qa.stories.tsx
mark-tate: A parser can return a number, NaN or null if blank.
null should only be returned by a parse when it's an empty value.
So would require userland to transform the null to "" before it's passed to the formatter.

site/docs/components/number-input/examples.mdx
mark-tate: reverted, consistency is better and then we change everything together.

site/src/examples/number-input/MinMax.tsx
mark-tate: good spot and removed from storybook code

packages/lab/src/number-input/NumberInput.css
mark-tate: not needed, removed

site/src/examples/number-input/ButtonAdornments.tsx
mark-tate: can you point us at an example or we can add later

joshwooding: we can do it later

packages/lab/stories/number-input/number-input.qa.stories.tsx
```diff
@@ -7,52 +7,60 @@ export default {
...
-        decimalScale={2}
+        format={formatWithDecimalScale(2)}
         defaultValue={5}
         max={10}
         min={-5}
         readOnly
         step={0.5}
       />
       <NumberInput
-        decimalScale={2}
```
wemmyo, line 45: We don't have any safe guards against `decimalScale` being less than zero, which will lead to errors

site/src/examples/number-input/StepAndBlock.tsx
```diff
@@ -4,7 +4,15 @@ import { NumberInput } from "@salt-ds/lab";
...
     <FormFieldLabel>Custom steps</FormFieldLabel>
-    <NumberInput defaultValue={10} step={5} stepMultiplier={10} />
+    <NumberInput
+      defaultValue={10}
+      step={5}
+      stepMultiplier={10}
+      onNumberChange={(_event, newValue) => {
+        console.log(`Number changed to ${newValue}`);
+      }}
+      style={{ width: "256px" }}
```
wemmyo:
```suggestion
```

site/src/examples/number-input/ClampBehavior.tsx
```diff
@@ -2,20 +2,78 @@ import {
...
+        onChange={(_event, newValue) => {
+          setValue(newValue);
+        }}
+        onNumberChange={(_event, newValue) => {
+          console.log(`Number changed to ${newValue}`);
+        }}
+        clamp
+        max={max}
+        min={min}
+        style={{ width: "250px" }}
```
wemmyo:
```suggestion
```

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -163,29 +165,66 @@ export interface NumberInputProps
   /**
    * Value of the `NumberInput`, to be used when in a controlled state.
    */
-  value?: number | string;
+  value?: string;
```
tycup: this accepts a string but defaultValue accepts both string and number - assume it should be both?

wemmyo:
Fixed issues I raised here for better visibility https://github.com/jpmorganchase/salt-ds/pull/5689

site/docs/components/number-input/examples.mdx
mark-tate: Maybe not here but will add higher up in the docs.

packages/lab/src/number-input/NumberInput.tsx
mark-tate: done

site/docs/components/number-input/examples.mdx
mark-tate: added

site/docs/components/number-input/examples.mdx
mark-tate: replaced

packages/lab/stories/number-input/number-input.stories.tsx
mark-tate: stale comment

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -265,120 +294,152 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
       }
-      const sanitizedValue = sanitizeInput(event.target.value);
-      const floatValue = toFloat(sanitizedValue);
-      const clampedValue = clampAndFix(floatValue);
-      // Update the value if it has changed
-      if (clampedValue.toString() !== value.toString()) {
-        setValue(clampedValue);
-        onChangeProp?.(event, clampedValue);
+      const validValue = pattern ? pattern(inputValue) : true;
+      console.log("validValue", validValue);
```
joshwooding: This should be removed before we merge

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -418,10 +481,52 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
+      } else if (inputRef.current) {
+        inputRef.current.select();
+      }
+    };
+
+    const handleContainerMouseUp: MouseEventHandler<HTMLDivElement> = (
+      event,
+    ) => {
+      setIsFocused(true);
+      restProps.onMouseUp?.(event);
```
joshwooding: This is not enough, we still need to destructure onMouseUp when it's passed in as it still will be overwritten.

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -265,120 +294,152 @@ export const NumberInput = forwardRef<HTMLDivElement, NumberInputProps>(
...
-      setValue(parsed);
-      onChangeProp?.(event, parsed);
+      adjustedValue += incrementStep;
+      commit(event ?? null, adjustedValue, String(adjustedValue));
     };
 
-    const handleInputKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
-      setIsEditing(true);
-      isAdjustingRef.current = false;
+    useEffect(() => {
```
joshwooding, line 432: This is usually a code smell, it's fine for now but we might need to refactor it out later.

site/allowed-words.txt
```diff
@@ -39,3 +39,4 @@ Tabstrips
 Taurean
 Tickbox
 UITK
+kmbt
```
joshwooding: Should we add things like this or disable checking for the regex on the line?

site/docs/components/number-input/examples.mdx
```diff
@@ -28,6 +29,37 @@ By default, the `NumberInput` is transparent. You can add borders using the `bor
...
+
+## Read-only
+
+The `NumberInput` can be set to a read-only state using the `readOnly` prop, allowing users to highlight and copy the value without editing it.
+
+By default, a dash ('-') serves as the empty marker. You can customize this marker by using the `emptyReadOnlyMarker` prop if your application needs a different symbol.
+
+**Note:** To address browser and screen reader limitations in conveying a read-only state, the `NumberInput`'s role is changed to `textbox` when set to read-only.
+
+### Best practice
```
joshwooding:
```suggestion
### Best practices
```

site/docs/components/number-input/examples.mdx
```diff
@@ -60,28 +94,66 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
 
 ## Decimal scale
 
-By default, the decimal scale is derived from the highest number of decimal places found in the `step` prop, or if not defined, in either the `defaultValue` or `value` prop.
-However, you can set it to any value of your choice using the `decimalScale` prop and that will have priority.
+The `decimalScale` prop allows you to define the number of decimal places displayed in the `NumberInput`. By default, the decimal scale is determined by the greater number of decimal places between the `step` value and the `value` or `defaultValue` provided.
 
-### Best practices
+### Best practice
```
joshwooding:
```suggestion
### Best practices
```

site/docs/components/number-input/examples.mdx
```diff
@@ -60,28 +94,66 @@ The `NumberInput` can display validation states, `error`, `warning`, and `succes
...
+
+<LivePreview componentName="number-input" exampleName="FixedDecimalScale" />
+
 ## Formatting
 
-You can format the value of the `NumberInput` using the `format` callback prop. It typically works alongside a `parse` callback for complex formatting.
+The `NumberInput` component offers the ability to format its value using the `format` callback prop. This feature is typically used in conjunction with a `parse` callback, which can be used to determine the valid input format and extract the numerical value of the formatted input.
 
-Users will be able to see the formatted value in the input, e.g. "5K". When the input gains focus, the value will be parsed into its numerical form, so "5K" becomes "5000". This allows users to edit the values with precision. When the input loses focus, the value will be formatted again.
+### Best practice
```
joshwooding:
```suggestion
### Best practices
```

site/docs/components/number-input/usage.mdx
```diff
@@ -37,7 +39,41 @@ To avoid misleading users, set the width to correlate to the length of the value
...
+
+To render a custom value such as `25K` or partial value `25,` return back `NaN` and the entered value renders as-typed.
+
+### Custom formatters
+
+The default formatter, supports signed decimal numbers.
+
+A custom formatter can be provided through the `format` prop.
+
+- Formatting does not occur, whilst the input has focus and is being edited.
```
joshwooding:
```suggestion
Formatting does not occur, whilst the input has focus and is being edited.
```

site/src/examples/number-input/FixedDecimalScale.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <FormField style={{ width: "256px" }}>
+      <FormFieldLabel>Number input with fixed decimal scale</FormFieldLabel>
+      <NumberInput
+        defaultValue={100.25}
+        decimalScale={4}
+        onNumberChange={(_event, newValue) =>
+          console.log(`Number changed to ${newValue}`)
+        }
+        format={(value: string) => {
```
joshwooding:
```suggestion
        format={(value) => {
```

I'm guessing this isn't needed?

site/src/examples/number-input/FormattingCompactNotation.tsx
```diff
@@ -0,0 +1,53 @@
...
+            return value;
+          }
+          return new Intl.NumberFormat("en-US", {
+            notation: "compact",
+            compactDisplay: "short",
+            maximumFractionDigits: 3,
+          }).format(floatValue);
+        }}
+        step={1000}
+        parse={(raw: string) => {
```
joshwooding:
```suggestion
        parse={(raw) => {
```

same here?

packages/lab/src/number-input/NumberInput.tsx
joshwooding:
```suggestion
```

.changeset/curvy-buckets-build.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+Updated the `NumberInput` API:
+
+1. Simplified `format`, previously took `string` or `number` value, now receives value as a `string` and returns a `string`.
+   Used to render the final value before render.
+
+```diff
+- format?: (value: number | string) => string | number;
++ format?: (value: string) => string;
```
tycup, line 12: `defaultValue` originally took string | number and i aligned `value` to also take number alongside just a string - should this change be reverted for the above diff to be true for `format` to only accept string?

tycup: (i didnt see this when i made the change, but felt like making defaultValue and value the same type made sense)

.changeset/curvy-buckets-build.md
```diff
@@ -0,0 +1,56 @@
...
++ onChange?: (event: ChangeEvent<HTMLInputElement>) => void;
+```
+
+5. Added `onNumberChange`, a callback called when the number value changes, due to increment/decrement or `onBlur`.
+
+```
+  onNumberChange?: (value: number | null) => void;
+```
+
+6. Simplified `parse`, is now parsed a string form of the current value and returns either a `number`, `NaN` or `null` if empty.
```
tycup, line 29: same for this (see comment on format)

.changeset/curvy-buckets-build.md
mark-tate: the value can be passed in as a number or a string BUT, we normalise it to a number internally and so format would always get a string.

One thing I am thinking of whether we should call onNumberChange with a number or a string or, whether we base that decision on the value we were passed.

.changeset/curvy-buckets-build.md
mark-tate: Parse will always take a string and return a number.

packages/lab/src/number-input/NumberInput.tsx
mark-tate: moved

packages/lab/src/number-input/NumberInput.tsx
mark-tate: simplified

tycup (approved):

PR #5498, opened 2025-08-28
Stack dialog action buttons on mobile

wemmyo:


.changeset/shy-eagles-ring.md
```diff
@@ -0,0 +1,8 @@
+---
+"@salt-ds/core": patch
+"@salt-ds/site": patch
```
mark-tate: site does not have changelogs ? should this be here ?

.changeset/shy-eagles-ring.md
```diff
@@ -0,0 +1,8 @@
...
+---
+
+- Updated storybook dialog actions with multiple buttons to use button bar pattern
+- Updated site dialog actions with multiple buttons to use button bar pattern
+- Added `column-reverse` and `row-reverse` to LayoutDirection
```
mark-tate: This is a great story, we improve our documentation but also improve a Core API
The problem is this changelog does not tell you that. Invest time in changelogs, they are a direct line to our stakeholders and help communicate value, which might not otherwise be visible.

How about something like this ?

- Enhanced `FlexLayout` by extending `LayoutDirection` to support both `column-reverse` and `row-reverse` directions, providing more flexible layout options.

Also, we've mixed it with site, so you are going to see this in the core Changelog.
We do not communicate site changes AFAIK.

So I think we  delete lines 3,6,7

packages/core/src/flex-layout/FlexLayout.tsx
```diff
@@ -21,7 +21,11 @@ import { parseSpacing } from "./parseSpacing";
 const withBaseName = makePrefixer("saltFlexLayout");
 
 export type LayoutSeparator = "start" | "center" | "end";
-export type LayoutDirection = "row" | "column";
+export type LayoutDirection =
+  | "row"
+  | "column"
```
mark-tate: Is this worth improving our docs for ?
We already have an example that shows row and column.

The minimum is that it's documented through props but perhaps we could improve our examples ?

The nice thing with the layout examples is they have little controls, that explain the API.
The downside to this, is they become less copy/pasteable.

But you would not really need/want to copy/paste just for one prop, which makes extending the example more reasonable.

.changeset/shy-eagles-ring.md
mark-tate: Actually this affects more than FlexLayout, I might list all the layouts, it improves.

packages/core/src/flex-layout/FlexLayout.tsx
wemmyo: Thanks. I've updated the docs and examples to reflect the changes

.changeset/shy-eagles-ring.md
wemmyo: I've updated the changeset and added a code snippet to show usage. Thanks

joshwooding (commented):
I'd prefer if the layout component changes can be moved into a separate PR.

.changeset/shy-eagles-ring.md
```diff
@@ -0,0 +1,10 @@
+---
+"@salt-ds/core": patch
```
joshwooding: This would be a minor change as it's new functionality.

.changeset/shy-eagles-ring.md
```diff
@@ -0,0 +1,10 @@
...
+---
+
+- Enhanced `FlexLayout`, `StackLayout` and `SplitLayout` components by extending `LayoutDirection` to support both `column-reverse` and `row-reverse` directions, providing more flexible layout options.
+- The Responsive props in Layout components now support reversing the order of items.
```
joshwooding: Was there an actual change that enabled this or is this related to the above point?

packages/core/stories/dialog/dialog.stories.tsx
```diff
@@ -222,16 +270,21 @@ const AlertDialogTemplate: StoryFn<
         status={status}
         open={open}
         onOpenChange={onOpenChange}
-        // focus the ok instead of the cancel button
-        initialFocus={1}
```
joshwooding, line 226: Was removal the right choice here? Should we conditionally change this depending on the button order?

CC @jake-costa

packages/core/stories/dialog/dialog.stories.tsx
```diff
@@ -290,18 +359,24 @@ export const MandatoryAction: StoryFn<typeof Dialog> = ({
         role="alertdialog"
         open={open}
         onOpenChange={onOpenChange}
-        initialFocus={1}
```
joshwooding, line 293: Same question here

packages/core/stories/dialog/dialog.stories.tsx
```diff
@@ -384,13 +480,85 @@ export const StickyFooter: StoryFn<typeof Dialog> = ({
...
+        onOpenChange={onOpenChange}
+        // Set initial keyboard focus to the "Ok" button
+        initialFocus={1}
+      >
+        <DialogHeader header={header} />
+        <DialogContent>{content}</DialogContent>
+        <DialogActions>
+          <StackLayout
+            direction={{
+              xs: "column-reverse",
```
joshwooding: This column-reverse still means the tab order and dom order are different. @jake-costa can you check this please

site/docs/components/flex-layout/examples.mdx
```diff
@@ -18,11 +18,13 @@ Set `wrap={true}` to overflow items onto subsequent rows when they no longer fit
 
 <LivePreview componentName="flex-layout" exampleName="Default" />
 
-## Column direction
+## Layout direction
```
joshwooding: I think we should mention `row` here too

site/docs/components/split-layout/examples.mdx
```diff
@@ -16,11 +16,13 @@ The items drive the height and width of the overall layout.
 
 <LivePreview componentName="split-layout" exampleName="Default" />
 
-## Column direction
+## Layout direction
```
joshwooding: Same here

site/docs/components/stack-layout/examples.mdx
```diff
@@ -16,11 +16,13 @@ The items contained within the layout determine its height and width.
 
 <LivePreview componentName="stack-layout" exampleName="Default" />
 
-## Row direction
+## Layout direction
```
joshwooding: We should probably mention `column` here too

site/src/examples/flex-layout/LayoutDirection.tsx
```diff
@@ -0,0 +1,64 @@
...
+            {index + 1}
+          </FlexItem>
+        ))}
+      </FlexLayout>
+      <div className={layoutDirectionStyles.radioButtonGroups}>
+        <FormField>
+          <FormFieldLabel>Direction</FormFieldLabel>
+          <RadioButtonGroup
+            direction={"horizontal"}
+            aria-label="Direction Controls"
```
joshwooding:
```suggestion
```

I don't think this is needed, we already have the FormFieldLabel.

@jake-costa can you assist here please

site/src/examples/stack-layout/LayoutDirection.module.css
```diff
@@ -0,0 +1,11 @@
+.container {
+  display: flex;
+  flex-direction: column;
+  gap: var(--salt-spacing-300);
+}
+
+.radioButtonGroups {
```
joshwooding: Is this needed? The container only has one child

wemmyo:
Layout changes moved to #5536

.changeset/shy-eagles-ring.md
wemmyo: This is related to the above point

.changeset/shy-eagles-ring.md
wemmyo: I've moved the layout changes to a new PR. Do we still need this changeset?
This PR only has site and storybook changes

packages/core/stories/dialog/dialog.stories.tsx
wemmyo: I was unsure about this but conditionally changing `initialFocus` sounds like a good idea. Waiting for confirmation from @jake-costa

.changeset/shy-eagles-ring.md
joshwooding: We can remove it

packages/core/stories/dialog/dialog.stories.tsx
wemmyo: Updated!

packages/core/stories/dialog/dialog.stories.tsx
wemmyo: In the meantime, I've updated to conditionally change.

navkaur76 (approved):

joshwooding (approved):

PR #5507, opened 2025-09-01
Refactor useChart & density sensitive options

tycup:
Refactor Highcharts theme approach

- Move to default options to useChart hook, so the user doesn't need to do the extra (and not ideal) step of setting via global Options
- Added density values to default options.
- Clean up examples and documentation


.changeset/witty-pumpkins-walk.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+- Move default options to useChart hook pattern
+- Remove global options object
+- Add density-sensitive values
+- Clean up examples and documentation
+- Remove unnecessary theme provider
```
wemmyo: I believe these should be in past tense. Also, is the first and second point related? If so maybe it should be together as a single sentence?

packages/highcharts-theme/src/examples/DonutChart.tsx
```diff
@@ -11,64 +12,13 @@ export interface DonutChartProps {
...
-          { name: "Payments", y: 5 },
-          { name: "Foreign Exchange", y: 5 },
-          { name: "Custody Services", y: 5 },
-          { name: "Insurance", y: 5 },
-          { name: "Compliance & Risk Services", y: 5 },
-        ],
-      },
-    ],
-  };
+  const donutChartOptions: Options = useChart(DonutRef, options);
```
wemmyo:
```suggestion
  const donutChartOptions = useChart(DonutRef, options);
```
Question: Do we need to explicitly set the type?

packages/highcharts-theme/src/useChart.tsx
```diff
@@ -1,30 +1,23 @@
...
-    touch: 16,
-  } as const;
-
-  return {
-    legendIconSize: legendIconSizeMap[density] ?? legendIconSizeMap.medium,
-  };
+  return useMemo(() => {
+    const defaultOptions = getDefaultOptions(density);
+    return Highcharts.merge(defaultOptions, chartOptions);
+  }, [density, chartOptions]);
```
wemmyo: Question: Considering chartOptions is an object, should it be a dependency?

.changeset/witty-pumpkins-walk.md
tycup: Sure - think i wrote this closer to a "commit" style message, will update thanks!

packages/highcharts-theme/src/useChart.tsx
tycup: chartOptions should stay in the dependency array because when it changes, the merged options need to be recalculated to reflect the new configuration

wemmyo (dismissed):

packages/highcharts-theme/src/examples/DonutChart.tsx
```diff
@@ -3,76 +3,30 @@ import { clsx } from "clsx";
...
   patterns?: boolean;
+  options: Options;
 }
 
-const DonutChart: FC<DonutChartProps> = ({ patterns = false }) => {
+const DonutChart: FC<DonutChartProps> = ({
+  patterns = false,
+  options = donutOptions,
+}) => {
   const DonutRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding:
```suggestion
  const donutRef = useRef<HighchartsReact.RefObject>(null);
```
Nit, variables should be camel cased.

packages/highcharts-theme/src/examples/LineChart.tsx
```diff
@@ -3,155 +3,20 @@ import { clsx } from "clsx";
...
   patterns?: boolean;
+  options: Options;
 }
 
-const LineChart: FC<LineChartProps> = ({ patterns = false }) => {
+const LineChart: FC<LineChartProps> = ({
+  patterns = false,
+  options = lineOptions,
+}) => {
   const LineChartRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding:
```suggestion
  const lineChartRef = useRef<HighchartsReact.RefObject>(null);
```

packages/highcharts-theme/src/examples/PieChart.tsx
```diff
@@ -3,71 +3,30 @@ import { clsx } from "clsx";
...
   patterns?: boolean;
+  options: Options;
 }
 
-const PieChart: FC<PieChartProps> = ({ patterns = false }) => {
+const PieChart: FC<PieChartProps> = ({
+  patterns = false,
+  options = pieOptions,
+}) => {
   const PieRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding:
```suggestion
  const pieRef = useRef<HighchartsReact.RefObject>(null);
```

packages/highcharts-theme/src/default-options.ts
```diff
@@ -0,0 +1,58 @@
...
+        margin: tokens["--salt-spacing-200"],
+      },
+    },
+    legend: {
+      layout: "vertical",
+      align: "right",
+      verticalAlign: "top",
+      symbolWidth: tokens["--salt-size-icon"],
+      symbolHeight: tokens["--salt-size-icon"],
+      symbolRadius: 3,
```
joshwooding: Should this change with density too?

packages/highcharts-theme/src/density-token-map.ts
```diff
@@ -0,0 +1,36 @@
+import type { Density } from "@salt-ds/core";
+
+export const getDensityTokenMap = (density: Density) => {
+  let saltSpacing100: number;
+  let saltSizeIcon: number;
+
+  switch (density) {
```
joshwooding: I have reservations about defining token values like this. I think we previously discussed this?

Is it possible to read these from the DOM?

.changeset/witty-pumpkins-walk.md
```diff
@@ -0,0 +1,8 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+- Moved default options to be provided by `useChart`
```
joshwooding: Do we need a diff here?

.changeset/witty-pumpkins-walk.md
```diff
@@ -0,0 +1,8 @@
...
+- Added density-sensitive values to default options
+- Updated usage docs
+- Removed stale implementation across site and package
```
joshwooding: Our changelogs typically only list changes to our consumable assets.

site/src/examples/highcharts-theme/PieChart.tsx
```diff
@@ -9,56 +9,56 @@ import styles from "./index.module.css";
...
+        { name: "Custody Services", y: 5 },
+        { name: "Insurance", y: 5 },
+        { name: "Compliance & Risk Services", y: 5 },
+      ],
+    },
+  ],
+};
+
 export const PieChart = () => {
   const PieRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding:
```suggestion
  const pieRef = useRef<HighchartsReact.RefObject>(null);
```

site/src/examples/highcharts-theme/LineChart.tsx
```diff
@@ -150,10 +159,9 @@ const lineDataOptions: Options = {
 
 export const LineChart = () => {
   const LineChartRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding:
```suggestion
  const lineChartRef = useRef<HighchartsReact.RefObject>(null);
```

site/src/examples/highcharts-theme/LineChart.tsx
```diff
@@ -49,6 +55,9 @@ const lineDataOptions: Options = {
     startOnTick: false,
     maxPadding: 0,
   },
+  legend: {
+    symbolWidth: 32,
```
joshwooding, line 59: Should this be a px value or relative?

site/docs/components/highcharts-theme/usage.mdx
joshwooding, line 1: The casing for titles in this file needs updating to be consistent with our content standards e.g. Line patterns

packages/highcharts-theme/src/useChart.tsx
```diff
@@ -1,30 +1,23 @@
...
-  chartRef: React.RefObject<HighchartsReact.RefObject>;
-}): { legendIconSize: number } => {
+export const useChart = (
+  chartRef: RefObject<HighchartsReact.RefObject>,
+  chartOptions: Options,
+) => {
   const density = useDensity();
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: triggered by density change
+  // biome-ignore lint/correctness/useExhaustiveDependencies: redraw on density change
```
joshwooding: If we add a check to density in the useEffect can we remove this? e.g.

```tsx
if (density) {
  chartRef.current?.chart.redraw()
}
```

packages/highcharts-theme/src/useChart.tsx
joshwooding, line 1: NIT: we dont need to use .tsx as the file extension .ts would be fine.

site/src/examples/highcharts-theme/LineChart.tsx
tycup: Takes `number` and fixed across densities

packages/highcharts-theme/src/useChart.tsx
tycup: removed entirely for new approach

packages/highcharts-theme/src/density-token-map.ts
tycup: Updated - take a look

PR #5513, opened 2025-09-02
Add Column Chart

tycup:
Add Column Chart
- Fix border-radius for bullet in v11+

### DIFF
--- a/packages/highcharts-theme/src/examples/dependencies/columnOptions.ts
+++ b/packages/highcharts-theme/src/examples/dependencies/columnOptions.ts
@@ -1,0 +1,38 @@
+import type { Options } from "highcharts";
+
+export const columnOptions: Options = {
+  chart: {
+    type: "column",
+  },
+  title: {
+    text: "Regional revenue by product",
+  },
+  accessibility: {
+    description:
+      "A column chart comparing revenue across regions. Each category shows a single column representing the total value for that region.",
+  },
+  xAxis: {
+    categories: ["NA", "EMEA", "APAC", "LATAM"],
+    title: {
+      text: "Region",
+    },
+  },
+  yAxis: {
+    min: 0,
+    title: {
+      text: "Revenue ($ millions)",
+    },
+  },
+  tooltip: {
+    headerFormat: "<span>{point.key}</span><br/>",
+    pointFormat:
+      '<span>{series.name}: </span><span class="value">${point.y}M</span>',
+  },
+  series: [
+    {
+      name: "Revenue",
+      type: "column",
+      data: [5, 3, 4, 7],
+    },
+  ],
+};
--- a/site/docs/components/highcharts-theme/examples.mdx
+++ b/site/docs/components/highcharts-theme/examples.mdx
@@ -31,13 +31,13 @@ Use a line chart to show changes in data over time. To ensure accessible present
 
 ## Pie chart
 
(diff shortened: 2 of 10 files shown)


packages/highcharts-theme/src/examples/dependencies/columnOptions.ts
```diff
@@ -0,0 +1,39 @@
+import type { Options } from "highcharts";
+
+export const options: Options = {
+  chart: {
+    type: "column",
+  },
+  title: {
+    text: "Regional Revenue by Product",
```
amritadesmet:
```suggestion
    text: "Regional revenue by product",
```

site/src/examples/highcharts-theme/ColumnChart.tsx
```diff
@@ -0,0 +1,79 @@
...
+  chart: {
+    type: "column",
+  },
+  title: {
+    text: "Regional revenue by product",
+  },
+  accessibility: {
+    description:
+      "A column chart comparing revenue across regions. Each category shows a single column representing the total value for that region.",
+  },
```
wemmyo, line 22:
```suggestion
 point: {
      valueDescriptionFormat:
        "{index}. {xDescription}, {value} million dollars.",
    },
  },
```
This should help screen readers announce the correct amount rather than just the number

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -71,3 +71,9 @@ Use a bullet chart to show performance against a target across categories. Each
 </Callout>
 
 <LivePreview componentName="highcharts-theme" exampleName="BulletChart" />
+
+## Column Chart
+
+Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
wemmyo:
```suggestion
Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
Nitpick: This might be clearer

site/src/examples/highcharts-theme/ColumnChart.tsx
joshwooding: @jake-costa are you able to comment here, guidance on `valueDescriptionFormat` seems like it could extend across chart types. What do we expect to be announced here?

site/docs/components/highcharts-theme/examples.mdx
tycup: Thanks, I like this more because it avoids stacked modifiers and applies the end-weight principle better - applied to all docs.

amritadesmet:
Just realised, is Tooltip meant to appear over Pie + Donut chart when hovering @tycup ? Doesn't seem to work for me in Preview

site/src/examples/highcharts-theme/ColumnChart.tsx
amritadesmet: Circling back to this for @jake-costa to review

tycup: Since there are quite a few accessibility features we could opt in/out of in the Options object, i'll create a separate draft PR with my best effort on this, and then we could review together & tweak @jake-costa ? 

(Which would unblock this PR too)

joshwooding: Makes sense

joshwooding (approved):

PR #5515, opened 2025-09-02
Add Bullet Chart

tycup:
Add Bullet Chart w/ warning on plot bands being a WIP

### DIFF
--- a/packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
+++ b/packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
@@ -1,0 +1,37 @@
+import type { Options } from "highcharts";
+
+export const bulletOptions: Options = {
+  chart: {
+    type: "bar",
+    height: 180,
+  },
+  title: {
+    text: "Revenue performance vs target",
+  },
+  accessibility: {
+    description: "A bullet chart showing revenue performance vs target for Q4",
+  },
+  xAxis: {
+    categories: ["Q4"],
+  },
+  yAxis: {
+    title: {
+      text: "Revenue ($M)",
+    },
+    min: 0,
+    max: 100,
+    tickPositions: [0, 20, 40, 60, 80, 100],
+  },
+  tooltip: {
+    headerFormat: "<span>{point.key}</span><br/>",
+    pointFormat:
+      '<span>{series.name}: </span><span class="value">${point.y}M</span><br/><span>Target: </span><span class="value">${point.target}M</span>',
+  },
+  series: [
+    {
+      name: "Current revenue",
+      type: "bullet",
+      data: [{ y: 60, target: 80 }],
+    },
+  ],
+};
--- a/site/docs/components/highcharts-theme/examples.mdx
+++ b/site/docs/components/highcharts-theme/examples.mdx
@@ -15,7 +15,10 @@ Current issues to be resolved before lab release:
 
   <ul>
     <li>
-      Legend positioning is not consistent with Figma examples due to Highcharts API limitations - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5555).
+      Legend positioning isn't consistent with Figma examples due to Highcharts API limitations - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5548).
+    </li>
+    <li>
+      Plot bands aren't currently supported due to v10 accessibility and token issues. If you need to use plot bands, you need to ensure that appropriate color contrast is maintained - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5548 )
     </li>
   </ul>
 </Callout>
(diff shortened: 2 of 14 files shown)


amritadesmet:
After speaking with Darrin, the background range colors should go in sequential order rather than skipping for best practices. Could you adjust this @tycup? Thanks!

site/src/examples/highcharts-theme/LineChart.tsx
```diff
@@ -0,0 +1,189 @@
...
+const options: Options = {
+  chart: {
+    type: "line",
+  },
+  accessibility: {
+    description:
+      "A summary communicating the trends, insights, or patterns the chart is intended to provide in a couple sentences.",
+  },
+  title: {
+    text: "Currency Performance Trends",
```
amritadesmet:
```suggestion
    text: "Currency performance trends",
```

packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
```diff
@@ -0,0 +1,39 @@
+import type { Options } from "highcharts";
+
+export const bulletOptions: Options = {
+  chart: {
+    type: "bar",
+    height: 180,
+  },
+  title: {
+    text: "Revenue Performance vs Target",
```
amritadesmet:
```suggestion
    text: "Revenue performance vs target",
```

packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
```diff
@@ -0,0 +1,39 @@
...
+  },
+  tooltip: {
+    headerFormat: "<span>{point.key}</span><br/>",
+    pointFormat:
+      '<span>{series.name}: </span><span class="value">${point.y}M</span><br/><span>Target: </span><span class="value">${point.target}M</span>',
+    stickOnContact: true,
+  },
+  series: [
+    {
+      name: "Current Revenue",
```
amritadesmet:
```suggestion
      name: "Current revenue",
```

packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
```diff
@@ -0,0 +1,38 @@
...
+    },
+    min: 0,
+    max: 100,
+    tickPositions: [0, 20, 40, 60, 80, 100],
+  },
+  tooltip: {
+    headerFormat: "<span>{point.key}</span><br/>",
+    pointFormat:
+      '<span>{series.name}: </span><span class="value">${point.y}M</span><br/><span>Target: </span><span class="value">${point.target}M</span>',
+    stickOnContact: true,
```
joshwooding: Random question, do we need to put this in every options or can this be a default we set in useChart?

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -15,7 +15,10 @@ Current issues to be resolved before lab release:
 
   <ul>
     <li>
-      Legend positioning is not consistent with Figma examples due to Highcharts API limitations - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5555).
+      Legend positioning isn't consistent with Figma examples due to Highcharts API limitations - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5548).
+    </li>
+    <li>
+      Plot bands aren't supported as part of this release due to v10 accessibility and token issues. If you need to use plot bands, you need to ensure that appropriate color contrast is maintained - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5548 )
```
joshwooding:
```suggestion
      Plot bands aren't currently supported due to v10 accessibility and token issues. If you need to use plot bands, you need to ensure that appropriate color contrast is maintained - [track issue](https://github.com/jpmorganchase/salt-ds/issues/5548 )
```

packages/highcharts-theme/src/examples/dependencies/bulletOptions.ts
tycup: I leave it in site examples to make it obvious that the tooltip format is highly customizable - i foresee a lot of consumers will look to edit the format of the tooltip quickly and this provides more context on how they might do that (i myself initially struggled to even find it in the Highcharts doc, because pointFormat isn't the most intuitive). It's similar to how donut chart has the 80% innerSize in the examples and not in the defaults.

You could argue that perhaps it needs it's own "Best practices" section like Donut does, but since tooltip formatting is common across all chart types, i think its fine to not include it, or better yet, give it a dedicated section in future docs (legend would also benefit from this).

...and so it's kept in the package examples for consistency

joshwooding: Ah, I only meant the specific option there, `stickOnContact`. I think the tooltip formatting is better in each example.

tycup: oops, yep that can go in default, mb - updated

joshwooding (approved):

amritadesmet (approved):

PR #5541, opened 2025-09-09
fix date input issue recorded in issue 5499

### LINKED ISSUE #5499: DateRange Custom Overlay Options breaks if manually cleared
### Latest version

### Description

DateRange Custom Overlay Options breaks if manually cleared.

### Steps to reproduce

Use date range selector however you want to select two dates. Close overlay and delete date values. Open overlay again and select X years. Note that overlay doesn't close and values aren't selected.

https://www.saltdesignsystem.com/salt/components/date-picker/range-date-picker/usage#custom-overlay-content

### Expected behavior

Overlay should close; values selected.

### Package name(s)

Lab (@salt-ds/lab)

### Package version(s)

_No response_

### Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

mark-tate:
Closes #5499

### DIFF
--- a/packages/lab/src/date-input/DateInputRange.tsx
+++ b/packages/lab/src/date-input/DateInputRange.tsx
@@ -401,9 +401,10 @@ export const DateInputRange = forwardRef<
         parseDateValue(dateValue?.startDate, DateParserField.START) ?? {};
       const { date: endDate = undefined, ...endDateParseDetails } =
         parseDateValue(dateValue?.endDate, DateParserField.END) ?? {};
+
       const updatedDateRange: DateRangeSelection<TDate> = {
-        startDate,
-        endDate,
+        startDate: dateValue?.startDate?.length ? startDate : null,
+        endDate: dateValue?.endDate?.length ? endDate : null,
       };
       if (dateAdapter.isValid(startDate)) {
         updatedDateRange.startDate = dateAdapter.setTimezone(
--- a/packages/lab/src/date-input/DateInputSingle.tsx
+++ b/packages/lab/src/date-input/DateInputSingle.tsx
@@ -282,10 +282,13 @@ export const DateInputSingle = forwardRef<
     const apply = (event: SyntheticEvent) => {
       const parse = parseProp ?? dateAdapter.parse.bind(dateAdapter);
       const parseResult = parse(dateValue ?? "", format);
-      let { date: parsedDate, ...parseDetails } = parseResult;
+      let parsedDate: TDate | null;
+      let parseDetails: DateDetail;
+      ({ date: parsedDate, ...parseDetails } = parseResult);
+      parsedDate = dateValue ? parsedDate : null;
       let formattedValue = "";
       const isDateValid = dateAdapter.isValid(parsedDate);
-      if (isDateValid) {
+      if (isDateValid && parsedDate) {
         parsedDate = dateAdapter.setTimezone(parsedDate, timezone);
         if (preservedTime.current) {
           parsedDate = dateAdapter.set(parsedDate, preservedTime.current);
--- a/.changeset/poor-tips-complain.md
+++ b/.changeset/poor-tips-complain.md
@@ -1,0 +1,9 @@
+---
+"@salt-ds/lab": patch
+---
+
+fix date input selection
+
+empty dates should return null, rather than an invalid date
+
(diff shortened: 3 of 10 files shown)


packages/lab/src/date-input/DateInputRange.tsx
```diff
@@ -401,9 +401,10 @@ export const DateInputRange = forwardRef<
         parseDateValue(dateValue?.startDate, DateParserField.START) ?? {};
       const { date: endDate = undefined, ...endDateParseDetails } =
         parseDateValue(dateValue?.endDate, DateParserField.END) ?? {};
+
       const updatedDateRange: DateRangeSelection<TDate> = {
-        startDate,
-        endDate,
+        startDate: dateValue?.startDate?.length ? startDate : null,
```
joshwooding, line 406: I'm struggling a little to understand this part, so if the unparsed date is not an empty string, null or undefined then use the parsed value?

packages/lab/src/date-input/DateInputSingle.tsx
```diff
@@ -282,10 +282,13 @@ export const DateInputSingle = forwardRef<
     const apply = (event: SyntheticEvent) => {
       const parse = parseProp ?? dateAdapter.parse.bind(dateAdapter);
       const parseResult = parse(dateValue ?? "", format);
-      let { date: parsedDate, ...parseDetails } = parseResult;
+      let parsedDate: TDate | null;
+      let parseDetails: DateDetail;
+      ({ date: parsedDate, ...parseDetails } = parseResult);
+      parsedDate = dateValue ? parsedDate : null;
```
joshwooding, line 288: Is this because dateValue could be set to null when controlled? Right now the types would suggest it would only ever be a string?

packages/lab/src/date-input/DateInputRange.tsx
mark-tate: **How does the parser work?**

The parsing of dates is kept simple... We give you a string you give us a date.
There is no API around supporting null as a special case.

So calling parse, we will always get back a date object.

**What happens if I pass the parser an empty string ?**

If I pass you an empty string though, we are going to get back an invalid date.

....Which is the start of the bug, this change fixes.

**What does apply do ?**

Apply will callback and communicate the change in date but we need to convert from an invalid date created by an empty string to null.

We could "some" through all the details to get UNSET but this is somewhat more direct/simpler/equivalent.

So in short, this code will return a date object, valid or invalid, which was not created by an empty string.

**What happens without this change?**

Currently the input parses an empty date as an invalid date and so in CustomDatePickerPanel this code
tried to add a tenor to an invalid date and silently failed.

```
                  newSelection = selectedDate
                    ? dateAdapter.add(selectedDate, {
                        years: tenor,
                      })
                    : dateAdapter.add(dateAdapter.today(), {
                        years: tenor,
                      });
```

Using null for empty/unset dates, pre-exists this change, so this change, aligns the current code with the API.

packages/lab/src/date-input/DateInputSingle.tsx
mark-tate: DateValue is always a string, but like range, we don't want to return an invalid date created by an empty string.

packages/lab/src/date-input/DateInputRange.tsx
joshwooding: Should we check explicitly for an empty string to make this obvious?

packages/lab/src/date-input/DateInputSingle.tsx
joshwooding: Can we check if the string is empty explicity?

Also is there no way to check if the parsedDate is invalid?

packages/lab/src/date-input/DateInputRange.tsx
joshwooding: Is there no way to check if dateValue.startDate is invalid directly?

mark-tate: yes isValid but it does not tell you what created it.
the choices are "some" over the parser details or to use the string value passed to the parser.

Issue #5467, opened 2025-08-18
Update a11y page structure template


jake-costa:
### Related page(s)

https://www.saltdesignsystem.com/salt/index

### Kind of issue

Other

### Description

Update the [a11y template](https://github.com/jpmorganchase/salt-ds/blob/main/templates/component-pages/component-name/accessibility.mdx ) to re-define the structure of the a11y page for components with any additional headings/sections as well as things to consider

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #5553, opened 2025-09-12
A11y page template changes

jake-costa:
Update the [a11y template](https://github.com/jpmorganchase/salt-ds/blob/main/templates/component-pages/component-name/accessibility.mdx) page for components with any additional sections as well as things to consider.

Closes: https://github.com/jpmorganchase/salt-ds/issues/5467

### DIFF
--- a/templates/component-pages/component-name/accessibility.mdx
+++ b/templates/component-pages/component-name/accessibility.mdx
@@ -17,9 +17,9 @@ Accessibility best practices and things to be aware of for this component.
 
 Bullet points describing what accessibility features are baked into the component or just the note below.
 
-**Note:** Salt handles the foundational accessibility behaviors and attributes for you in 
-the components. Depending on your specific use case, you may need to make additional 
-changes, and some of the information below may not be relevant. Product teams need 
+**Note:** Salt handles the foundational accessibility behaviors and attributes for you in
+the components. Depending on your specific use case, you may need to make additional
+changes, and some of the information below may not be relevant. Product teams need
 to ensure their digital products meet all current accessibility requirements.
 
 ## Keyboard interactions
@@ -62,7 +62,7 @@ Information or bullet points describing known issues and affected combinations.
 - [Link to WCAG success criterion](url)
 - [Link to WCAG technique](url)
 - [Link to ARIA pattern](url)
-- [Link to any other apporprite resources](url)
+- [Link to any other appropriate resources](url)
 
 ## Feedback
 
--- a/site/allowed-words.txt
+++ b/site/allowed-words.txt
@@ -40,4 +40,5 @@ Tabstrips
 Taurean
 Tickbox
 UITK
+frontmatter
 kmbt


aimee-liou (commented):
Looks good. Curious about your thoughts on including a section for most common screen reader considerations? Previously working on a product team, I'd often get flagged by ADA which sometimes resulted in big retroactive changes.

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -39,3 +48,22 @@ You can have as many Keyboard control blocks as you need.
...
+Information or bullet points describing known issues and affected combinations.
+
+## References
+
+- [Link to WCAG success criterion](url)
+- [Link to WCAG technique](url)
+- [Link to ARIA pattern](url)
+- [Link to any other apporprite resources](url)
+
+## Feedback
```
joshwooding, line 67: Across the site we ask for feedback, we probably need a consistent way to do it

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -39,3 +48,22 @@ You can have as many Keyboard control blocks as you need.
 
 </KeyboardControl>
 </KeyboardControls>
+
+## WCAG compliance
```
joshwooding, line 52: What do you envision this containing?

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -13,6 +13,15 @@ data:
 
 Accessibility best practices and things to be aware of for this component.
 
+## What Salt provides
```
joshwooding, line 16: I understand the reasoning behind this but I've got a few thoughts, the note below is an accessibility statement and it feels odd to copy that into multiple pages and not have it in a central location.

Also what do you define as accessibility features

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -39,3 +48,22 @@ You can have as many Keyboard control blocks as you need.
...
 </KeyboardControl>
 </KeyboardControls>
+
+## WCAG compliance
+
+Statement about WCAG conformance level and testing.
+
+## Known issues
+
+Information or bullet points describing known issues and affected combinations.
```
joshwooding, line 58: We should decide if this should be inline on the site or part of the GitHub backlog

templates/component-pages/component-name/accessibility.mdx
jake-costa: I agree an accessibility statement belongs on the site. Two reasons a short note on each component’s Accessibility tab would be helpful:
* We don’t have the accessibility statement on the public site yet.
* People on a component page are more likely to see the tab than a separate policy page.

I understand it may seem odd to repeat the note, but I think it would be helpful. This isn’t a full statement. It sets scope and responsibility: Salt covers baseline behaviors, and product teams still need to validate their specific use cases against current requirements. This counters the common “Salt out-of-the-box = accessible app” assumption. Also, as I mentioned, Visa’s design system felt the need to add something similar on their accessibility pages.

As for accessibility features, I think we could mention items like wrapping `FormFieldLabel` and `Input` in `FormField` allows Salt to handle the programmatic association between the label and the input for assistive technology. Or how we handle changes for aria attributes (e.g. aira-expanded) or any aria announcements we bake in.

Thoughts?

templates/component-pages/component-name/accessibility.mdx
jake-costa: I think it could house the success criteria that the component was tested and validated against. Or possibly Information to clarify some success criteria. (e.g. [3.3.2: Labels or Instructions](https://www.w3.org/WAI/WCAG21/Understanding/labels-or-instructions.html)): 
* Which we could indicate that there are various methods for labeling or providing formatting instructions in case we get more users reaching out saying that some examples that we have on the site are non-compliant. 

Otherwise I have no problem removing it. Thoughts?

templates/component-pages/component-name/accessibility.mdx
jake-costa: Understood, yeah a consistent feedback pattern would be great. However, I still think adding a feedback entry point on each component’s Accessibility tab is useful. It meets users where they are and lowers the effort to report issues or gaps. We could possibly even embed the component in the link so a GitHub issue opens pre-populated with the component name as a label/tag.

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -13,6 +13,15 @@ data:
...
 Accessibility best practices and things to be aware of for this component.
 
+## What Salt provides
+
+Bullet points describing what accessibility features are baked into the component or just the note below.
+
+**Note:** Salt handles the foundational accessibility behaviors and attributes for you in 
+the components. Depending on your specific use case, you may need to make additional 
+changes, and some of the information below may not be relevant. Product teams need 
+to ensure their digital products meet all current accessibility requirements.
```
mark-tate, line 23: I was thinking, what comes for free with the component and what we need a user to add.
ie. wiring up labels to elements is something they would need to do.

We do create examples that demonstrate this but do we need to be clear on what a user needs to do?
I wonder if a checklist would help, each user.
We don't need to re-document a11y, if we are referring them to better docs but we could make it easier for folks to validate their own use-cases.

A checklist could be a simple list of actions or do's and don't for using this component ?

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -39,3 +48,22 @@ You can have as many Keyboard control blocks as you need.
 
 </KeyboardControl>
 </KeyboardControls>
```
mark-tate, line 50: What was onerous with Carousel was having to verbalise user journeys for focus. It must be horrible to read.
How can we describe focusable user journeys better?

Could we annotate the elements in a view and then call them out as action &  effect?

mark-tate:
How about a pivot ?

Ignore the specific in this, more of a general idea, that we reduce text and turn this into a checklist instead.
Easier to read, easier to write and also could be used to drive testing.

For focus journeys, it would be easy to create if we used Mermaid but Mermaid is, or was, expensive to add to the site.
Maybe we could look again at that or alternatives, but a visual guide would be nice.
I could create a POC for this to see how expensive it is now.

# Accessibility Checklist for [Component Name]

## Best Practices Checklist
- [ ] All interactive elements are reachable and operable via keyboard.
- [ ] Clear focus indicators are visible for all focusable elements.
- [ ] Semantic HTML is used for structure and controls.

## Salt Component Usage Checklist
- [ ] Use the component as provided to benefit from built-in accessibility features.
- [ ] Review and, if needed, override ARIA roles and attributes for your specific use case.
- [ ] Ensure labels are correctly associated with controls in your implementation.
- [ ] Confirm that any customizations do not break foundational accessibility behaviors.
- [ ] Test your product for accessibility after integrating the component.

## Keyboard Interaction Checklist
- [ ] `Tab` moves focus to the next interactive element.
- [ ] `Shift + Tab` moves focus to the previous interactive element.
- [ ] `Enter` or `Space` activates the selected item or button.
- [ ] All custom keyboard shortcuts are documented and tested.

## Color Contrast Checklist
- [ ] Text and background colors meet WCAG 2.1 AA minimum contrast ratio (4.5:1 for normal text, 3:1 for large text).
- [ ] Interactive elements (buttons, links) have sufficient contrast against their background.
- [ ] Focus indicators are clearly visible and meet contrast requirements.
- [ ] Color is not used as the only means of conveying information.

## Screen Reader Accessibility Checklist
- [ ] All interactive elements have accessible names (using labels, aria-label, or aria-labelledby).
- [ ] ARIA roles and properties are used appropriately and do not conflict with native semantics.
- [ ] Dynamic updates (e.g., error messages, modal dialogs) are announced to screen readers.
- [ ] The reading order matches the visual order.
- [ ] Non-text content (images, icons) have appropriate alt text or ARIA labels.
- [ ] Test the component with at least one major screen reader (e.g., NVDA, JAWS, VoiceOver).

## WCAG Compliance Checklist
- [ ] Component usage meets **WCAG 2.1 AA** conformance.
- [ ] Accessibility testing is performed (automated and manual).
- [ ] Any exceptions or deviations are documented and justified.

## Known Issues
- In some screen reader/browser combinations, the focus indicator may not be visually prominent.
- VoiceOver on iOS may announce the component label twice in certain contexts.
- [Add any other known issues relevant to your component.]

## References
- [WCAG 2.1 Success Criterion 1.4.3: Contrast (Minimum)](https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html)
- [WCAG 2.1 Success Criterion 4.1.2: Name, Role, Value](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
- [ARIA Authoring Practices Guide: Widget Patterns](https://www.w3.org/WAI/ARIA/apg/patterns/)
- [WCAG Techniques for Keyboard Accessibility](https://www.w3.org/WAI/WCAG21/Techniques/general/G202)

## Feedback
If you encounter accessibility issues or have suggestions, please provide feedback to the Salt team.

templates/component-pages/component-name/accessibility.mdx
mark-tate: Documenting it somewhere is a good first step, which this template allows and then we could extract and move wholesale. I like the Github back log idea as it creates 2 way discussion and an opportunity for folks to help us fix issues. We have this section already, so probably ok to leave in this template for now, until we revisit with a site wide plan.

templates/component-pages/component-name/accessibility.mdx
mark-tate: We should add that accessibility statement page, added a ticket

https://github.com/jpmorganchase/salt-ds/issues/5640

templates/component-pages/component-name/accessibility.mdx
mark-tate: I think https://github.com/jpmorganchase/salt-ds/issues/5640 can speak to this, through the backlog.

templates/component-pages/component-name/accessibility.mdx
mark-tate: I would think that could is covered by the links to other docs.
Anything specific to our own work could be covered more generally through 
https://github.com/jpmorganchase/salt-ds/issues/5640

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -13,6 +13,15 @@ data:
 
 Accessibility best practices and things to be aware of for this component.
 
+## What Salt provides
+
+Bullet points describing what accessibility features are baked into the component or just the note below.
+
+**Note:** Salt handles the foundational accessibility behaviors and attributes for you in
```
aimee-liou, line 20: Thinking we could remove "Note:" - if it's just this body of text in this section (OR bullets), it's implied that it's a note here...

templates/component-pages/component-name/accessibility.mdx
```diff
@@ -39,3 +48,22 @@ You can have as many Keyboard control blocks as you need.
...
+## Known issues
+
+Information or bullet points describing known issues and affected combinations.
+
+## References
+
+- [Link to WCAG success criterion](url)
+- [Link to WCAG technique](url)
+- [Link to ARIA pattern](url)
+- [Link to any other appropriate resources](url)
```
aimee-liou, line 65: Understand this was added for a11y reasons, but removing "Link to..." in each bullet would reduce redundancy/easier to scan

aimee-liou (dismissed):
Left two small notes, but good to approve if implemented or not.

templates/component-pages/component-name/accessibility.mdx
jake-costa: Appreciate the feedback, to clarify yes we would not want to include "Link to" For each link this is just for the template. So for example "Link to WCAG success criterion" Would likely be modified to something like "WCAG 1.4.11 Non-Text Contrast"

PR #5574, opened 2025-09-15
Update Tabs

joshwooding:


jake-costa:
Hi @joshwooding I appreciate your patience. Blow our my findings from my most recent review of tabs. Feel free to let me know if you have any questions or want to jump on a call to discuss. 

# Tab A11y Review:

* Overflow dialog: Recommend adding `aria-modal="true"` if possible.
* Overflow tab `aria-haspopup`: I recommend using `true` instead of `dialog`.
* Closable tabs — **highly recommend**:

  * Remove the close button from the tab order and hide it from AT (drop the `aria-label`).
  * Support the Delete key to close a tab. This aligns with APG, reduces tab stops, and clears the Axe `aria-required-children` flag.
  * Set `aria-description` to: “Press Delete to remove the tab.”
  * Keep the close icon button clickable for mouse users.
  * Note: If we add nested menus, also support “Shift+F10” to open the menu.

## Examples:

I believe these are example-specific, not component-level. Correct me if wrong.

* Add `aria-hidden="true"` to decorative icons (e.g., “With Icon”). **Is this something we should also add to the documentation or is it something we can bake in at the component level**
* Reflow: Storybook tabs reflow at 320×256 px. Vercel site does not.
* Badge “1” accessible name reads “One updates.” Expected: “One update.”
* Tablist name: add an accessible name to the main tablist (`aria-label`). For the overflow tablist, use `aria-labelledby` that points to the dialog label.

  * We can add accessible-naming guidance to the docs. I can work on drafting it.

### Vercel overflow example:

* VoiceOver/Safari: announces “Checks, tab 4 of 4.” Overflow tab not counted. Unsure if this is fixable. Checking if this is a known issue.
* NVDA/Firefox: announces “Checks, tab 4 of 5” (includes overflow).
* JAWS/Chrome: announces “Checks, tab 4 of 5” (includes overflow).

## Comments:

When a tab is chosen from overflow it moves to the main list and gets focus. Selecting another main-list tab then pushes that tab back into overflow. Do we want this, or should surfaced tabs remain in the main list until overflow recalculates?

joshwooding:
>  For the overflow tablist, use aria-labelledby that points to the dialog label.

@jake-costa what is the dialog label in this example?

jake-costa:
> > For the overflow tablist, use aria-labelledby that points to the dialog label.
> 
> @jake-costa what is the dialog label in this example?

I was referring to aria-label="Overflow Menu" on dialog. That said, since it is already announcing correctly, we would need to validate that this works correctly. While it would align with the APG, this is more of an edge case and I am unsure if the label would be announced twice. If it does announce twice I would recommend we keep the aria-label on the dialog but not add aria-labelledby to tablist.

jake-costa:
# A11y Review Items

* Can we rename the overflow tab label to just **“Overflow”** instead of **“13 tabs hidden”**? Right now screen readers announce: **“13 tabs hidden tab collapsed subMenu 5 of 5”**. With “Overflow” it would be: **“Overflow tab collapsed subMenu 5 of 5”**.
* Can we support the **Delete** key to close a tab when the tab itself has focus, while still keeping the close button in the tab order?
* Not sure why, but on **Mac + Safari** in the closable example, after closing a tab it looks like **focus gets lost**. I’m not seeing this with **JAWS/NVDA**. (@karlgoldstraw, @joshwooding Can you reproduce?)
* Should I open a ticket to also create a **scrollable tablist**, or do we have one already?

## Notes

* We’re still going to have aXe flag the closable example because of the close button but we do call this out on the Tabs a11y page.
* VoiceOver announces **“4 tabs”** while **JAWS/NVDA** announce **“5 tabs.”** This isn’t a violation. VoiceOver appears to use heuristics to exclude the overflow control (treating it as a menu trigger), while Windows screen readers count every `role="tab"` literally.

packages/lab/src/tabs-next/TabListNext.tsx
```diff
@@ -98,15 +99,94 @@ export const TabListNext = forwardRef<HTMLDivElement, TabListNextProps>(
...
+      }, 166);
+    });
+
+    useEffect(() => {
+      const handleFocus = () => {
+        if (!tabstripRef.current) return;
+
+        setTimeout(() => {
+          handleTabRemoval();
+        }, 33);
```
joshwooding:
```suggestion
        }, 66);
```

packages/lab/src/tabs-next/TabListNext.tsx
```diff
@@ -98,15 +99,94 @@ export const TabListNext = forwardRef<HTMLDivElement, TabListNextProps>(
...
+      }, 166);
+    });
+
+    useEffect(() => {
+      const handleFocus = () => {
+        if (!tabstripRef.current) return;
+
+        setTimeout(() => {
+          handleTabRemoval();
+        }, 66);
```
joshwooding:
```suggestion
        }, 48);
```

packages/lab/src/tabs-next/TabListNext.tsx
```diff
@@ -98,15 +99,94 @@ export const TabListNext = forwardRef<HTMLDivElement, TabListNextProps>(
...
+      }, 166);
+    });
+
+    useEffect(() => {
+      const handleFocus = () => {
+        if (!tabstripRef.current) return;
+
+        setTimeout(() => {
+          handleTabRemoval();
+        }, 48);
```
joshwooding:
```suggestion
        }, 33);
```

packages/lab/src/tabs-next/TabNext.tsx
```diff
@@ -56,7 +56,7 @@ export const TabNext = forwardRef<HTMLDivElement, TabNextProps>(
 
     const disabled = !!disabledProp;
 
-    const id = useId(idProp);
+    const id = value;
```
tycup: This then gets passed down to a TabNextPanel's `aria-labelledby` - but if the value has spaces, e.g. "Liquidity Checks" thats fine for `id` but aria-labelledby will see it as 2 different id's since its space separated and accepts multiple ids

packages/lab/src/tabs-next/TabListNext.tsx
tycup: TabListNext vs TabstripNext - would have different display names?

packages/lab/src/tabs-next/hooks/useOverflow.ts
```diff
@@ -213,29 +209,28 @@ export function useOverflow({
...
+
+  const hiddenPinnedIndex = hidden.findIndex(
+    (child) =>
+      isValidElement<TabNextProps>(child) && child?.props?.value === pinned,
+  );
 
-  if (selected && hiddenSelectedIndex !== -1) {
-    const removed = hidden.splice(hiddenSelectedIndex, 1);
+  if (hiddenPinnedIndex !== -1) {
+    const removed = hidden.splice(hiddenPinnedIndex, 1);
```
tycup: kinda nit: mutating hidden/visible when they are usMemo cached - could strict mode double render splice the wrong element/in concurrent mode abandon render? i think could be cleaned up by filtering hidden and spreading

packages/lab/src/tabs-next/TabNext.tsx
tycup, line 45: probs part of your clean up - unused idProp

packages/lab/src/tabs-next/TabListNext.tsx
```diff
@@ -1,21 +1,164 @@
...
+
+    const handleFocus = (event: FocusEvent) => {
+      if (!tabstripRef.current) return;
+
+      const wasInTablist =
+        event.target instanceof HTMLElement && isInTabList(event.target);
+      const stillInTablist =
+        event.relatedTarget instanceof HTMLElement &&
+        isInTabList(event.relatedTarget);
+      if ((wasInTablist && !stillInTablist) || event.relatedTarget === null) {
```
tycup: `event.relatedTarget === null` - i think this fires when focus leaves the window entirely so handleTabRemoval() also fires - also if the user was focused on another input eslewhere and then switches windows

that being said, this would only really be a problem in a the pretty rare case of a user switching windows while an async tab dismissal is happening - but could be addressed with:
`if (wasInTablist && (!stillInTablist || event.relatedTarget === null))`

joshwooding: I think this won't work for Safari unfortunately, I'll look into it.

packages/lab/src/tabs-next/TabOverflowList.css
```diff
@@ -8,7 +8,6 @@
   overflow: hidden;
   overflow-y: auto;
   position: absolute;
-  z-index: var(--salt-zIndex-flyover);
```
tycup, line 11: Out of curiosity, why was this removed?

packages/lab/src/tabs-next/TabsNext.tsx
tycup: Should this be `TabsNext`?

tycup, line 73: nit: Typed as ` | null` but `TabsNextContext` `setSelected` doesn't include null

packages/lab/src/tabs-next/TabNextTrigger.tsx
tycup, line 71: `setSelected` gets called in both `handleKeyDown` and in `handleClick`, and since the trigger is a `<button>`, i think the keyboard press also generates a click - so one key press results in two selection updates so two `onChange` calls

PR #5579, opened 2025-09-16
Fix FOUC Flash in SaltProvider Docs

### LINKED ISSUE #5578: SaltProvider: Nested providers remove root classes (FOUC)
### Latest version

### Description

SaltProvider: Root mis-detection causes nested providers (theme="") to remove documentElement classes on unmount leading to FOUC / missing tokens. 

### Steps to reproduce

In production, this happens sporadically but most prone on the SaltProvider doc page. This is because we don't dobule invoke the effect (like in strict mode) so the root re-applies classes quickly.

Locally, since we React have strict mode, we get the nested misidentified provider run its cleanup repeatedly, stripping root classes as it goes. If you navigate to the SaltProvider doc page whilst running locally, you'll see the page with stripped styling.

We get a

```
SaltProvider can only apply CSS classes to the root if it is the root level SaltProvider.
```
warning in both.

### Expected behavior

No warnings or removal of site-level root classes and token loss; no unstyled flashes

### Package name(s)

Core (@salt-ds/core)

### Package version(s)

_No response_

### Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

tycup:
Fixes #5578

### DIFF
--- a/packages/core/src/salt-provider/SaltProvider.tsx
+++ b/packages/core/src/salt-provider/SaltProvider.tsx
@@ -258,7 +258,10 @@ function InternalSaltProvider({
     actionFont: inheritedActionFont,
   } = useContext(ThemeContext);
 
-  const isRootProvider = inheritedTheme === undefined || inheritedTheme === "";
+  const targetWindow = useWindow();
+
+  const isRootProvider =
+    inheritedWindow === undefined || inheritedWindow !== targetWindow;
   const density = densityProp ?? inheritedDensity ?? DEFAULT_DENSITY;
   const themeName =
     themeProp ?? (inheritedTheme === "" ? DEFAULT_THEME_NAME : inheritedTheme);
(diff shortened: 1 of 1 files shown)


packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -258,7 +258,10 @@ function InternalSaltProvider({
     actionFont: inheritedActionFont,
   } = useContext(ThemeContext);
 
-  const isRootProvider = inheritedTheme === undefined || inheritedTheme === "";
+  const targetWindow = useWindow();
```
tycup: determine the browser window this provider targets - need this before computing 'root-ness' so we can base decision on window ownership

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -348,7 +349,11 @@ function InternalSaltProvider({
       }
     }
     return () => {
-      if (applyClassesTo === "root" && targetWindow) {
+      if (
+        applyClassesTo === "root" &&
+        targetWindow &&
+        (inheritedWindow === undefined || inheritedWindow !== targetWindow)
```
wemmyo:
```suggestion
        isRootProvider
```

PR #5584, opened 2025-09-17
initial type rework to make mutable types for some props

mark-tate:


.changeset/add-injected-class-support.md
```diff
@@ -0,0 +1,12 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/core": minor
```

packages/core/src/status-indicator/ValidationStatus.ts
```diff
@@ -1,8 +1,20 @@
...
+
+export type ValidationStatus = keyof ValidationStatuses;
+
+const ValidationStatusValues = [
   "error",
   "warning",
   "success",
   "info",
-] as const;
+] as const satisfies ReadonlyArray<ValidationStatus>;
```
joshwooding: Not sure we can use this. Would need to check the output but satisfies falls outside our supported TS version

packages/core/src/status-indicator/StatusIndicator.tsx
```diff
@@ -36,17 +29,11 @@ export const StatusIndicator = forwardRef<SVGSVGElement, StatusIndicatorProps>(
...
-      error: ErrorIcon,
-      success: SuccessIcon,
-      warning: WarningIcon,
-      info: InfoIcon,
-    };
-
-    const IconComponent = iconMap[status];
-    const ariaLabel = statusToAriaLabelMap[status];
+    const icons = useIcon();
+    const titleCaseStatus = status.charAt(0).toUpperCase() + status.slice(1);
```
joshwooding: Should we use the capitalise util from core?

.changeset/tall-peas-begin.md
```diff
@@ -0,0 +1,25 @@
+---
+"@salt-ds/styles": patch
```
joshwooding:
```suggestion
"@salt-ds/styles": minor
```

packages/core/src/button/Button.tsx
```diff
@@ -97,10 +123,14 @@ function variantToAppearanceAndColor(
 }
 
 export const Button = forwardRef<HTMLButtonElement, ButtonProps>(
-  function Button(
-    {
+  function Button(props, ref?): ReactElement<ButtonProps> {
```
joshwooding:
```suggestion
  function Button(props, ref) {
```

packages/core/src/dropdown/Dropdown.css
```diff
@@ -181,6 +181,7 @@
   overflow: hidden;
   text-overflow: ellipsis;
   text-wrap: nowrap;
+  line-height: normal;
```
joshwooding: Not sure about this change, the previous implementation seems to be correct.

packages/core/src/toggle-button/ToggleButton.tsx
```diff
@@ -45,17 +48,20 @@ export interface ToggleButtonProps extends ComponentPropsWithoutRef<"button"> {
   /**
    * Value of the toggle button, to be used when in a controlled state.
    */
-  value: string | ReadonlyArray<string> | number | undefined;
+  value?: string | ReadonlyArray<string> | number;
```
joshwooding: Why was this needed?

packages/core/src/tooltip/Tooltip.tsx
```diff
@@ -86,20 +91,15 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
...
     } = useFormFieldProps();
 
     const disabled = disabledProp || formFieldDisabled;
-    const status =
-      statusProp ??
-      (formFieldValidationStatus !== undefined &&
-      VALIDATION_NAMED_STATUS.includes(formFieldValidationStatus)
-        ? formFieldValidationStatus
-        : undefined);
+    const status = statusProp ?? formFieldValidationStatus ?? undefined;
```
joshwooding: I think this is a regression https://github.com/jpmorganchase/salt-ds/issues/3868

packages/core/src/index.ts
```diff
@@ -45,6 +45,7 @@ export * from "./parent-child-layout";
 export * from "./pill";
 export * from "./progress";
 export * from "./radio-button";
+export type {} from "./registerClassNameInjection";
```
joshwooding, line 51: Can we add a comment if this is required for the module augmentation

packages/styles/src/useClassNameInjection.tsx
```diff
@@ -0,0 +1,149 @@
...
+export type ClassNameInjectionRegistry<
+  ComponentName extends SupportedComponent = SupportedComponent,
+> = Map<ComponentName, ClassNameInjectorEntry[]>;
+
+const EMPTY_REGISTRY: ClassNameInjectionRegistry = new Map();
+const InjectionContext =
+  createContext<ClassNameInjectionRegistry>(EMPTY_REGISTRY);
+
+export type ClassNameInjectionProviderProps = {
+  children: React.ReactNode;
```
joshwooding: We should use the destructure here

packages/core/src/tooltip/Tooltip.tsx
mark-tate: const status = statusProp ?? formFieldValidationStatus || undefined;
```suggestion
    const status = statusProp ?? formFieldValidationStatus || undefined;
```

packages/core/src/index.ts
mark-tate: add a comment to say what it is for

packages/core/src/dropdown/Dropdown.css
mark-tate: removed this and went for the internal line-height fix

packages/core/src/status-indicator/StatusIndicator.tsx
mark-tate: done

packages/core/src/index.ts
mark-tate: done

packages/styles/src/useClassNameInjection.tsx
mark-tate: done

packages/core/src/toggle-button/ToggleButton.tsx
mark-tate: removed and will re-test snapshot

packages/core/src/status-indicator/ValidationStatus.ts
mark-tate: removed

.changeset/semantic-icon.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/core": minor
```

packages/styles/src/useClassNameInjection.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+export function ClassNameInjectionProvider({
+  children,
+  value,
+}: ClassNameInjectionProviderProps) {
+  const registry = useMemo(() => value ?? EMPTY_REGISTRY, [value]);
+
+  useEffect(() => {
+    if (!hasWarnedExperimentalOnce && process.env.NODE_ENV !== "production") {
+      // eslint-disable-next-line no-console
```
joshwooding: We don't use eslint

joshwooding (dismissed):

packages/styles/src/useClassNameInjection.tsx
```diff
@@ -0,0 +1,154 @@
...
+  );
+
+  // Compute injected classes provided through ClassNameInjectionRegistry
+  const injected = useMemo(() => {
+    const { className: _ignore, ...restProps } = props as any;
+    if (!entries.length) return [];
+    return entries
+      .map((e) => e.fn(restProps))
+      .filter((v): v is string => v != null);
+  }, [entries, deps, props]);
```
wemmyo, line 96:
```suggestion
  }, [entries, props]);
```

packages/core/src/tooltip/Tooltip.tsx
```diff
@@ -48,7 +49,7 @@ export interface TooltipProps
   /**
    * Optional string to determine the status of the Tooltip.
    */
-  status?: ValidationStatus;
+  status?: FormFieldValidationStatus | ValidationStatus;
```
wemmyo, line 52:
```suggestion
  status?: ValidationStatus;
```
Do we need both, ValidationStatus already has all supported statuses

mark-tate: FormFieldValidationStatus are used by FormFields and ValidationStatus has wider status, such as Banner.

So there are subsets of statuses.

If we do don't have FormFieldValidationStatus, then you would extend ValidationStatus beyond FormField which would include Banner which is not what we want.

It's a complexity of ValidationStatus and FormField, where you can use Tooltip with a FormField or on it's own.

I think this is a safe change as FormFieldValidationStatus is a subset of ValidationStatus that omits info

packages/styles/src/useClassNameInjection.tsx
mark-tate: It's safer to go with this.. and optimise later, this is experimental, so will need stakeholder to check.

PR #5586, opened 2025-09-17
Update Site Vertical Navigation

wemmyo:


site/src/components/navigation/VerticalNavigation.tsx
```diff
@@ -82,85 +163,130 @@ const renderNavigationItem = (
...
-            0,
-          ),
-        )}
-      </StackLayout>
-    </nav>
+    <VerticalNavigationComponent
+      aria-label="Sidebar"
+      data-testid="vertical-navigation"
+      appearance="bordered"
+      style={{ paddingLeft: 24 }}
```
joshwooding: This is the wrong way to apply styles, we should always use tokens

site/src/components/navigation/VerticalNavigation.tsx
```diff
@@ -82,85 +163,130 @@ const renderNavigationItem = (
...
+      </VerticalNavigationItem>
+    );
+  }
+
+  return (
+    <VerticalNavigationItem
+      active={isActive}
+      key={id}
+      ref={setItemRef}
+      style={{ scrollMarginTop: 24 }}
```
joshwooding: This should probably use a token too

site/src/components/navigation/VerticalNavigation.tsx
```diff
@@ -82,85 +163,130 @@ const renderNavigationItem = (
...
         active={isActive}
-        orientation="vertical"
-        onExpand={shouldRenderAsParent ? handleExpand : undefined}
-        parent={shouldRenderAsParent}
-        render={renderItem}
-        expanded={isExpanded}
-        level={level}
+        key={id}
+        ref={setItemRef}
+        style={{ scrollMarginTop: 24 }}
```
joshwooding: Same here

site/src/components/navigation/VerticalNavigation.tsx
wemmyo: Thanks. I've updated to use tokens.
Would it be better to use classNames here as well or are inline styles fine?

joshwooding: classes would be better, but I'd also add it where the other padding is (primary sidebar)

site/src/components/navigation/PageNavigation.tsx
tycup, line 13: wondering if, since you're doing a mapping later anyway, you could do it earlier (here?) and then we could get rid of this `//@ts-expect-error` ?

tycup, line 15: Can `selectedGroupIds` be removed since i don't think it's used inside the `VerticalNavigation`?

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+  return (
+    <VerticalNavigationComponent
+      aria-label="Sidebar"
+      data-testid="vertical-navigation"
+      appearance="bordered"
+      className={styles.verticalNavigation}
+    >
+      {mappedNavData.map((item) => (
+        <NestedItem
+          key={item.title}
```
tycup: nit: is title a good key to use? href might be slightly better?

site/src/components/navigation/vertical-navigation/utils/index.ts
```diff
@@ -0,0 +1,67 @@
...
+      let children: Item[] = [];
+      if (item.kind === "group" && Array.isArray(item.childNodes)) {
+        children = mapMenu(item.childNodes as SidebarNodeWithStatus[]);
+      }
+      // Remove children if only one child and its title matches the parent
+      const shouldRemoveChildren =
+        children.length === 1 && children[0].title === item.name;
+      return {
+        title: item.name,
+        href: shouldRemoveChildren ? children[0].href : item.id,
```
tycup, line 61: curious; why is `item.id` preferred as a fallback instead of `item.data?.link`? (could we prefer `item.data.link` otherwise `item.id`?)

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+      </VerticalNavigationItemContent>
+    </VerticalNavigationItem>
+  );
+};
+
+export const VerticalNavigation = ({
+  menu,
+  selectedNodeId,
+}: VerticalNavigationProps) => {
+  const mappedNavData = mapMenu(menu);
```
tycup, line 118: I know we talked about this, but i just realised that since this isn't memoized, it means your useMemo below would recalculate anyway since it has `mappedNavData` as a dependency (its changing reference forces the recomputation).

site/src/components/navigation/vertical-navigation/utils/index.ts
wemmyo: It's pretty much the same value. However `item.id` is always present but `item.data` can be an empty object

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Good point, I don't think the useMemo is required anyway so I've removed it

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Yeah I agree, I'll update thanks

site/src/components/navigation/PageNavigation.tsx
wemmyo: Yup! Thanks

wemmyo: The type returned from `data?.sharedConfig?.sidebar` doesn't match the returned data, not sure why. But I'll remove the ts-expect so we can have some type safety for the VerticalNavigation

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+  const [collapsed, setCollapsed] = useState(isOpen);
+
+  const { title, status, href, children } = item;
+  const isItemSelected = selectedNodeId === href;
+
+  if (Array.isArray(children) && children.length > 0) {
+    return (
+      <VerticalNavigationItem active={isItemSelected && collapsed}>
+        <Collapsible
+          onOpenChange={() => setCollapsed(!collapsed)}
```
joshwooding:
```suggestion
          onOpenChange={() => setCollapsed(old => !old)}
```

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+const NestedItem = ({ item, selectedNodeId }: NestedItemProps) => {
+  const isOpen = containsSelected(item, selectedNodeId);
+  const [collapsed, setCollapsed] = useState(isOpen);
+
+  const { title, status, href, children } = item;
+  const isItemSelected = selectedNodeId === href;
+
+  if (Array.isArray(children) && children.length > 0) {
+    return (
+      <VerticalNavigationItem active={isItemSelected && collapsed}>
```
joshwooding: This logic doesn't look correct, blur active should be when the group is collapsed and an item inside is selected.

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Collapsible
+          onOpenChange={() => setCollapsed(!collapsed)}
+          open={collapsed}
+        >
+          <VerticalNavigationItemContent>
+            <CollapsibleTrigger>
+              <VerticalNavigationItemTrigger>
+                <VerticalNavigationItemLabel>
+                  <FlexLayout justify="space-between">
```
joshwooding: Do you need a flex layout or can you do this how the example does it https://github.com/jpmorganchase/salt-ds/blob/main/site/src/examples/vertical-navigation/WithBadges.tsx#L58

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Good spot, thanks. Updated to use `selectedGroupIds`

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Done

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Removed Flex layout. Thanks

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,147 @@
...
+  item,
+  selectedNodeId,
+  selectedGroupIds,
+}: NestedItemProps) => {
+  const isOpen = containsSelected(item, selectedNodeId);
+  const [expanded, setExpanded] = useState(isOpen);
+
+  const { title, status, href, children } = item;
+  const isItemActive = selectedNodeId === href;
+  const isGroupActive = selectedGroupIds.has(href) ?? false;
```
joshwooding: Why is this nullish operator needed?

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -0,0 +1,147 @@
...
+      </VerticalNavigationItem>
+    );
+  }
+
+  return (
+    <VerticalNavigationItem active={isItemActive}>
+      <VerticalNavigationItemContent>
+        <VerticalNavigationItemTrigger render={<LinkBase href={href} />}>
+          <VerticalNavigationItemLabel>
+            <FlexLayout justify="space-between">
```
joshwooding: Same as before is this flex layout needed too?

wemmyo: Missed that, removed now

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Not needed and removed. Initially made `selectedGroupIds` optional and used that as a fallback

PR #5595, opened 2025-09-19
Split salt themes

### LINKED ISSUE #5316: Refactor the file hierachy of theme to seperate next and legacy values.
### Description

Re-work the file structure of the theme package to clearly seperate the legacy and next theme.

### Expected behavior

We can easily seperate the two, with a view to deprecating/removing the legacy tokens at some point.

### Alternatives and examples

n/a

### Context

Reduce complexity of the theme and make it easier to maintain/consume/explain.

joshwooding:
Closes #5316

### DIFF
--- a/.changeset/eleven-camels-find.md
+++ b/.changeset/eleven-camels-find.md
@@ -2,10 +2,10 @@
 "@salt-ds/theme": minor
 ---
 
-Added `--salt-shadow-lowest`, `--salt-shadow-lower`, `--salt-shadow-low`, `--salt-shadow-mediumLow` and `--salt-shadow-medium`.
+- Added `--salt-shadow-lowest`, `--salt-shadow-lower`, `--salt-shadow-low`, `--salt-shadow-mediumLow` and `--salt-shadow-medium`.
 
-Deprecated `--salt-shadow-100`, `--salt-shadow-200`, `--salt-shadow-300`, `--salt-shadow-400` and `--salt-shadow-500`.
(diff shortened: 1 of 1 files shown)


bhoppers2008 (approved):

pseys (approved):

packages/theme/css/next/palette/shadow.css
```diff
@@ -0,0 +1,15 @@
+.salt-theme.salt-theme-next[data-mode="light"] {
+  --salt-shadow-lowest: 0 1px 3px 0 rgba(0, 0, 0, 0.1);
```
mark-tate, line 2: is this the only ramp we have created that is a partial ramp?

Why not always create a full ramp and add placeholders for the un-supported values.
This would create a more scalable foundation as folks don't need to change the ramp to add new values, if someone wants a more intense shadow, they would modify the current medium token, rather than perhaps using high, which would be the right way to do it.

It's only a couple of tokens and would make things more consistent and easier to document.
Unused values could be set to undefined or a neighbour.
Documentation should explain how to use these tokens and can also say that high values are un-used and a placeholder.

According to LLM, other notable design systems use this approach and it's considered more standard.

joshwooding: We have a lot of "partial" ramps. 

I don't think it's worth adding any more shadows yet, we know we want to change this ramp later on and adding more now would just increase the amount we have to change

bhoppers2008 (approved):

PR #5597, opened 2025-09-19
`AriaAnnouncer` improvements, previously `delay` was used to queue an…

mark-tate:
…nouncements, but it could create overlapping announcements. We have deprecated `delay`, although it is still supported. Instead, we have added a new options API.

- `duration`: provides a sequential delay between announcements, ensuring they do not overlap.
- `ariaLive` determines the importance and urgency of the announcement.

### DIFF
--- a/.changeset/nasty-planes-sell.md
+++ b/.changeset/nasty-planes-sell.md
@@ -1,0 +1,19 @@
+---
+"@salt-ds/core": minor
+---
+
+- `AriaAnnouncer` improvements, previously `delay` was used to queue announcements, but it could create overlapping announcements. We have deprecated `delay`, although it is still supported. Instead, we have added a new options API.
+- `duration`: provides a sequential delay between announcements, ensuring they do not overlap.
+- `ariaLive` determines the importance and urgency of the announcement.
+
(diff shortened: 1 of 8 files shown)


BharateshPoojary:
Hi,

I tested the snapshot @salt-ds/core@0.0.0-snapshot-20250929093937 with our Cypress component tests. The tests fail immediately with:
image
It looks like Cypress is trying to import raw .css from @salt-ds/icons/src, which isn’t compatible in this setup. Previous versions didn’t show this issue.

Should the snapshot be exporting prebuilt CSS instead of raw src for tests, or should we add custom Cypress config to stub CSS imports?

Thanks!

jake-costa:
@mark-tate, Just wanted to add while we can and should still support Delay/Duration I do not believe we should apply values to either of these by default. 

If using polite (which should set by default), I would also highly recommend adding ~300ms to debounce, like for the carousel.  

Ideally, in most cases, we should allow the screen reader to handle any queue/interruptions of messages. We could also add information to the docs to indicate that it is accessibility best practice to avoid, where possible, creating situations where multiple aria live messages are added in rapid succession.

PR #5601, opened 2025-09-22
Update CI to Node 24, update @types/node and update npm token reference

joshwooding:


wemmyo (approved):

.github/workflows/test.yml
```diff
@@ -116,7 +106,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v5
         with:
-          node-version: "20"
+          node-version: "24"
```
tycup: The cypress container image is node-22.14 and then install node 24 - curious if thats intended/is there a cypress 24 image that we could be using so we dont split?

tycup: ah cypress seems to* [edit] prefer using LTS which 24 yet isnt, at least it seems from this thread: https://github.com/cypress-io/cypress-docker-images/issues/1336

joshwooding: We might not need it anymore, I don't think there's a good version since that was mainly to pin the version of Chrome

tycup (approved):

wemmyo (approved):

Issue #5229, opened 2025-06-16
Chart Theme - Area Chart


amritadesmet:
Salt area chart based on UITK + Figma sticker sheet

Planned work for Q2/Q3

Parent issue #3323

amritadesmet:
Reminder: When PR is up, review w/ Jake if overlapping opacities of categorical colors are compliant with Pattern Fills

PR #5618, opened 2025-09-24
Highcharts | Area Chart

tycup:
In relation to #5229 

Also increases the consistency of code patterns (you can see this in a separate commit)


packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -132,6 +137,7 @@
 
 .highcharts-theme-salt .highcharts-legend-item-hidden .highcharts-point {
   stroke: none !important;
+  fill: inherit !important;
```
tycup, line 140: !important because default stylesheet also uses important

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -77,3 +77,9 @@ Use a bullet chart to show performance against a target across categories. Each
 Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="ColumnChart" />
+
+## Area Chart
+
+Use an area chart to show trends over time and the contribution of categories to a total. To ensure accessible presentation, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
wemmyo:
```suggestion
Use an area chart to show trends over time and the contribution of categories to a total. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
Nit: To be consistent with the others

packages/highcharts-theme/src/examples/AreaChart.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+export interface AreaChartProps {
+  patterns?: boolean;
+  options: Options;
+}
+
+const AreaChart: FC<AreaChartProps> = ({
+  patterns = false,
+  options = areaOptions,
+}) => {
```
wemmyo, line 16:
```suggestion
const AreaChart = ({
  patterns = false,
  options = areaOptions,
}: AreaChartProps) => {
```
Question: Do we need to use FC?

site/src/examples/highcharts-theme/AreaChart.tsx
```diff
@@ -0,0 +1,132 @@
...
+    value == null
+      ? null
+      : Math.round(value * scale + (index % offsetEvery) * offsetAmount),
+  );
+
+const corporateBondsData = generateScaledSeries(
+  governmentBondsData,
+  0.65,
+  5,
+  150,
```
wemmyo: Nit: Considering we have a couple of arguments here, might be better to use an object to improve readability

site/src/examples/highcharts-theme/AreaChart.tsx
```diff
@@ -0,0 +1,132 @@
...
+];
+
+const generateScaledSeries = (
+  source: Array<number | null>,
+  scale: number,
+  offsetEvery: number,
+  offsetAmount: number,
+) =>
+  source.map((value, index) =>
+    value == null
```
wemmyo:
```suggestion
    value === null
```

packages/highcharts-theme/src/examples/AreaChart.tsx
tycup: For typing, yep , don't think it detracts - we use this in other places in the repo albeit not entirely consistently (features in e.g. Calendar)

site/src/examples/highcharts-theme/AreaChart.tsx
tycup: Yep good call - I ended up just putting the calculated data directly into the series object to be consistent with other examples (the consumer doesn't care for the calculation since its just an example)

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -77,3 +77,9 @@ Use a bullet chart to show performance against a target across categories. Each
 Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="ColumnChart" />
+
+## Area Chart
+
+Use an area chart to show trends over time and the contribution of categories to a total. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
joshwooding, line 83: We should add a note to review the content before we release charts. I'm not too fussed about making sure it's the best it could be for now.

site/src/examples/highcharts-theme/AreaChart.tsx
```diff
@@ -0,0 +1,122 @@
...
+        10734, 10311, 9812, 8259, 7236, 6604, 5650, 5585, 5194, 5288, 5372,
+        5471, 5480, 4737, 4834, 4929, 5021, 5113, 5147, 5181, 4329, 4317, 2541,
+        2661, 2709, 2762, 2862, 2123, 2086, 2048,
+      ],
+    },
+  ],
+};
+
+export const AreaChart = () => {
+  const AreaChartRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding: This naming is inconsistent with the other examples, also we need to decide if the refs should all just be `chartRef` or specific to the type of chart.

packages/highcharts-theme/src/examples/AreaChart.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const AreaChartRef = useRef<HighchartsReact.RefObject>(null);
```
joshwooding: Same comment about ref naming here

site/src/examples/highcharts-theme/AreaChart.tsx
tycup: Improved the consistency across examples in general in its own commit - take a look

tycup: - opted for chartRef/chartOptions inside the components because they are already scoped by both the component name and the passed options

- renamed the options objects in the site examples to specific charts and then generalised to `chartOptions` inside the component to match generalised `chartRef` and again, because its scoped by the component name

packages/highcharts-theme/src/examples/AreaChart.tsx
wemmyo: Yeah it might be best we move away from using FC because it includes `children` which isn't always necessary. What are your thoughts @joshwooding

wemmyo (dismissed):

.changeset/breezy-grapes-sit.md
joshwooding, line 1: I think a rebase or update went wrong, these changesets already exist right?

amritadesmet:
Small one: I just noticed that some of the chart headers are labelled 'X Chart' and some 'X chart' - shall we choose 1 for consistency? @tycup

Issue #5592, opened 2025-09-18
Aria Announce Provider ADA Issue


jake-costa:
### Latest version

- [x] I have tested the latest version

### Description

Currently the default state for the aria announce provider is assertive so when components use the announcer for announcements that are not assertive, this results in an ADA violation. 

### Steps to reproduce

.

### Expected behavior

Components using aria announce provider need to be able to set it as polite.

### Package name(s)

Core (@salt-ds/core)

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [x] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #5623, opened 2025-09-26
Fix AriaAnnouncer ADA Compliance Issue

BharateshPoojary:
*Fixes Issue #5592*
Resolves ADA violation where AriaAnnounceProvider was hardcoded to aria-live="assertive", preventing components from using appropriate announcement levels.

*Problem*

AriaAnnounceProvider had aria-live hardcoded to "assertive"
This caused ADA violations when components needed "polite" announcements for non-urgent content
Components couldn't control the announcement priority level

 *Solution*
Added ariaAnnounce prop to AriaAnnouncerProvider with the following features:
Changes Made:

New Prop: Added ariaAnnounce?: "off" | "assertive" | "polite" | undefined
Default Value: Maintains "assertive" as default for backwards compatibility
Flexible Usage: Components can now specify appropriate announcement levels

*Testing*
 Added Cypress component tests for all prop values
 Verified backwards compatibility (existing code unchanged)
image


mark-tate:
Thanks for the contribution, it comes at the right time, as we have another open PR that solves the same issue.
https://github.com/jpmorganchase/salt-ds/pull/5597

I've generated a snapshot for the other PR
can you pull this snapshot into your app and feedback whether this solves your issue.
Separately, we are looking whether a pre or post delay works better with polite.
Perhaps you can opine on that as well ?

For now we will combine your requirement with the other PR and close this PR. Your version has commented out code and also does not consider the other code, such as useAriaAnnouncer, so cannot be merged. Happy for us to work together on the other PR, once we get your feedback on the snapshot.

PR #5630, opened 2025-09-30
Highcharts | Scatterplot Chart

tycup:


amritadesmet:
Preview is missing the marker shapes?

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -77,3 +77,9 @@ Use a bullet chart to show performance against a target across categories. Each
 Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="ColumnChart" />
+
+## Scatter plot
+
+Use a scatter plot to show the relationship between two numerical variables. To ensure accessible presentation, fill patterns can be applied to the points (see [Patterns and Fills](./usage#patterns-and-fills) for details).
```
tycup: Don't think this should have fills since the markers are so small?

amritadesmet: I'd agree with this. What do you think @jake-costa?

packages/highcharts-theme/src/default-options.ts
```diff
@@ -14,6 +14,16 @@ export const getDefaultOptions = (
     chart: {
       styledMode: true,
       colorCount: 20,
+      events: {
```
tycup: This is the interesting part of this PR; unless I use :has() i can't target the .`axes-grid-lines .highcharts-yaxis .highcharts-axis-line` as they appear in the DOM before the scoped chart series classes (such as `highcharts-scatter-series`)

this is a styledMode issue because otherwise, you can set these properties in the plotOptions object scoped to a series type (but that's disabled in styledMode)

amritadesmet:
Have we considered re-ordering Column chart to be placed under Donut chart since Stacked + Bullet are variations of a Bar chart? I think scatterplot, area, boxplot etc are slightly more complex so would work well below as you have it now @tycup

tycup:
> Have we considered re-ordering Column chart to be placed under Donut chart since Stacked + Bullet are variations of a Bar chart? I think scatterplot, area, boxplot etc are slightly more complex so would work well below as you have it now @tycup

Done in AreaChart PR!

packages/highcharts-theme/src/examples/ScatterplotChart.tsx
```diff
@@ -3,7 +3,7 @@ import { clsx } from "clsx";
 import Highcharts, { type Options } from "highcharts";
 import HighchartsReact from "highcharts-react-official";
 import { type FC, useRef } from "react";
-import { scatterplotOptions } from "./dependencies/scatterplotOptions";
+import { scatterplotChartOptions } from "./dependencies";
```
tycup: This is a consistency fix for after AreaChart PR gets merged, where this is the new pattern

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -77,3 +77,9 @@ Use a bullet chart to show performance against a target across categories. Each
 Use a column chart to compare values across categories. Each category is represented by a single vertical bar. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="ColumnChart" />
+
+## Scatter chart
```
tycup, line 87: Prior incorrectly capitalised chart titles are updated in the AreaChart PR

site/docs/components/highcharts-theme/examples.mdx
jake-costa: Even though they are small we might be able to. For example, if we have 5 shapes and at least 2 clear patterns (Like a dot in the middle and a diagonal line for example) then we might be able to have up to 15 data points (Shape with no pattern, Shape+Pattern1, Shape+Pattern2). Thoughts? 

Otherwise, I am not sure how we would support more than 5 data points.

site/allowed-words.txt
```diff
@@ -29,6 +29,7 @@ Popout
 Preheader
 Robo
 Selectability
+Scatterplot
```
joshwooding: Is this being used?

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -242,3 +243,22 @@
...
+.highcharts-theme-salt .highcharts-scatter-series .highcharts-point {
+  fill-opacity: 0.75;
+}
+
+/* Remove line from scatter legend symbols */
+.highcharts-theme-salt .highcharts-scatter-series.highcharts-legend-item .highcharts-graph {
+  stroke-width: 0;
+}
+
+/* Show the y-axis line when chart type has the additional .axes-grid-lines*/
```
joshwooding: Are we still keeping this?

tycup: Was speaking to @amritadesmet  about this more after our chat and im swayed more towards keeping;
- I don't like that the alternative would be for `default-options` to run a function on every load that looks for specifically scatter/bubble chart to toggle a class (if the user never has them, that doesnt feel good) - unless there's a better idea?
- If we don't specifically call out that this class can theoretically be applied to any chart to create double grid lines, and have it only and specifically for the scatter/bubble examples, then we imply that these (and only these) charts should be using this class - and if they use it on other charts, well, we don't officially recommend that. If user requests come in for double grid lines for other chart types, we can priortise adding the supplementary Figma variants and provide a 'temporary' solution by informing users who are unaware that well, you _could_ add this class, but we won't have supported stickers in Figma for them just yet (at which point we can prio it)

tycup: Added a line in example docs on what the class is/does

packages/highcharts-theme/src/default-options.ts
```diff
@@ -70,6 +71,12 @@ export const getDefaultOptions = (
       bullet: {
         borderRadius: 0,
       },
+      scatter: {
+        marker: {
+          // This should be a token?
+          radius: 4,
```
tycup, line 77: raw value used in design too

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -245,3 +245,22 @@
   fill-opacity: 1;
   transition: none;
 }
+
+.highcharts-theme-salt .highcharts-scatter-series .highcharts-point {
+  fill-opacity: 0.75;
```
joshwooding, line 250: How are we deciding the opacities? Do they need to align with other parts of the system?

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -245,3 +245,22 @@
...
+}
+
+/* Remove line from scatter legend symbols */
+.highcharts-theme-salt .highcharts-scatter-series.highcharts-legend-item .highcharts-graph {
+  stroke-width: 0;
+}
+
+/* Show the y-axis line when chart type has the additional .axes-grid-lines*/
+.highcharts-theme-salt.axes-grid-lines .highcharts-yaxis .highcharts-axis-line {
```
joshwooding: Should this merge with the selectors below?

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -83,3 +83,11 @@ Use a bullet chart to show performance against a target across categories. Each
 Use an area chart to show trends over time and the contribution of categories to a total. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="AreaChart" />
+
+## Scatter chart
+
+Use a scatter chart to show the relationship between two numerical variables. To ensure accessible presentation, fill patterns can be applied to the scatter points (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
+The `axes-grid-lines` class applied to the container ensures both vertical and horizontal grid lines are displayed.
```
joshwooding:
```suggestion
The `axes-grid-lines` class name applied to the container means both vertical and horizontal grid lines are displayed.
```

I'm not sure if this is the right terminology but I don't think "ensures" is the right word.

tycup: We use ensures in other places in the repo; grammatically 'means' is more descriptive/passive - 'ensures' suggests it 'does' something - so it depends if you want to convey a definition or a descriptive action (i prefer the action via 'ensures')

packages/highcharts-theme/css/salt-highcharts-theme.css
tycup: Decided based on highcharts defaults but those themselves aren't always consistent - 
Screenshot 2025-10-13 at 12 04 50

Area chart has a specified 0.75 opacity default and we align this both here and in bubble chart to make it consistent

site/docs/components/highcharts-theme/examples.mdx
joshwooding: Do we use "ensures" in the same context though? 

To me saying adding a class name ensures both grid lines are displayed seems odd as it really just enables the vertical/horizontal grid lines.  Scanning the docs ensures is mostly used when explaining why options are chosen rather than what the option does

tycup: By definition (as they're not synonyms):
- ensures: make certain that (something) will occur or be the case / guarantees or causes something to happen. Implies a cause and effect relationship

- means: (importantly: as a _noun_)  an action or system by which a result is achieved; a method. In your sentence, youre using it as a _verb_ which indicates that it "signifies / indicates " rather than act as a method (or by the book, "to express or represent something". _It makes it descriptive not causal_. Means can ofc be used as a verb too, but only if paired with "means *that*" or "means + an ing" ending word. It's also a grammatically secondary definition (at least by cambridge's standard).

So since adding this class name is causal (lines appearing or not) i prefer "ensures".

_All that being said_  (and honestly, the above is more just more out of interest than to challenge), there is a wide enough balance of ensures and means being technically misused that i can settle for a nice grammatical "means _that_" even if i have a personal preference for ensures :D so:

**TLDR** ill change it to "means that" which is correct and kinda middle ground

joshwooding: We can also change it to my new suggestion

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -83,3 +83,11 @@ Use a bullet chart to show performance against a target across categories. Each
 Use an area chart to show trends over time and the contribution of categories to a total. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
 
 <LivePreview componentName="highcharts-theme" exampleName="AreaChart" />
+
+## Scatter chart
+
+Use a scatter chart to show the relationship between two numerical variables. To ensure accessible presentation, fill patterns can be applied to the scatter points (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
+The `axes-grid-lines` class name applied to the container means that both vertical and horizontal grid lines are displayed.
```
joshwooding:
```suggestion
The `axes-grid-lines` class name applied to the container enables both vertical and horizontal grid lines.
```

site/docs/components/highcharts-theme/examples.mdx
tycup: done 🤝

PR #5632, opened 2025-09-30
Highcharts | Fix spelling

tycup:


yunjungyeh (approved):

site/allowed-words.txt
```diff
@@ -27,6 +27,7 @@ Overlayable
 Picklist
 Popout
 Preheader
+Robo
```
tycup, line 30: For "Robo-advisory" https://en.wikipedia.org/wiki/Robo-advisor used in donut/pie examples

PR #5641, opened 2025-10-03
Fix disabled Menu opening on hover

wemmyo:


packages/core/src/menu/MenuBase.tsx
```diff
@@ -121,10 +121,15 @@ export function MenuBase(props: MenuBaseProps) {
     }
   }, [getVirtualElement, refs]);
 
+  const isTriggerDisabled = () => {
+    const triggerElement = refs.reference.current as HTMLDivElement | null;
```
joshwooding: This is reading a ref during rendering which doesn't follow React's best practices https://react.dev/learn/referencing-values-with-refs#best-practices-for-refs

wemmyo: Tried another approach using context callback

packages/core/src/menu/MenuItem.tsx
```diff
@@ -48,6 +51,17 @@ export const MenuItem = forwardRef<HTMLDivElement, MenuItemProps>(
...
     const handleRef = useForkRef<HTMLDivElement>(ref, item.ref);
+
+    useEffect(() => {
+      if (!triggersSubmenu) return;
+      setTriggerDisabled(!!disabled);
+      return () => {
+        if (triggersSubmenu && disabled) {
+          setTriggerDisabled(false);
+        }
+      };
```
joshwooding: Can we simplify this? Does this still work:


```suggestion
      setTriggerDisabled(!!disabled);
```

wemmyo: Yeah that works too, added the submenu check as a safeguard. This change will disable all hover events so thought it might be good to check if there are even submenus to open first.
Can remove if you think it's not necessary

joshwooding: I think it's fine to simplify. disabled items shouldn't support hover anyway

.changeset/purple-waves-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix disabled `MenuItem` trigger opening content on hover
```
joshwooding:
```suggestion
Fixed disabled `MenuItem`s opening submenus on hover.
```

packages/core/src/menu/MenuItem.tsx
```diff
@@ -48,6 +51,11 @@ export const MenuItem = forwardRef<HTMLDivElement, MenuItemProps>(
       window: targetWindow,
     });
     const handleRef = useForkRef<HTMLDivElement>(ref, item.ref);
+
+    useEffect(() => {
+      setTriggerDisabled(!!disabled);
+    }, [triggersSubmenu, disabled, setTriggerDisabled]);
```
joshwooding:
```suggestion
    }, [disabled, setTriggerDisabled]);
```

PR #5642, opened 2025-10-05
mark `@salt-ds/embla-carousel` as stable

mark-tate:


.changeset/fancy-falcons-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/embla-carousel": major
+---
+
+`Carousel` now marked as stable.
```
joshwooding, line 5: We should have a more comprehensive changeset, with a summary of features or something similar.

joshwooding:
Moved to draft as #5679 is open

Issue #5190, opened 2025-06-09
Chart Theme - Bubble chart


amritadesmet:
Salt bubble chart based on UITK

Planned work for Q2/Q3

PR #5645, opened 2025-10-06
Highcharts | Bubble chart

tycup:
Add Bubble Chart #5190 
- Tiny patch to consistency in scatter example

### DIFF
--- a/packages/highcharts-theme/css/salt-highcharts-theme.css
+++ b/packages/highcharts-theme/css/salt-highcharts-theme.css
@@ -175,10 +175,6 @@
   fill: var(--salt-content-primary-foreground);
 }
 
-.highcharts-theme-salt .highcharts-bubble-series .highcharts-point {
-  fill-opacity: 0.75;
-}
-
 .highcharts-theme-salt .highcharts-legend-item-hidden > text,
 .highcharts-theme-salt .highcharts-legend-item-hidden span {
   /* Highcharts default CSS uses !important for fill/stroke; mirror to override */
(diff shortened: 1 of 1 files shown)


amritadesmet:
> > Should we make note in docs about editing inner label truncation / behaviour - not present by default - as we discussed or not mention it? @tycup
> 
> Leaning against; goes into territory of redefining the Highcharts API - since we're just theming, i think its the users responsibility to go lean on the Highcharts API (especially as this is an opinionated behaviour and not just a stylistic choice).

Sounds good, thanks. I won't mention this behaviour in the Figma docs either

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -163,6 +169,16 @@
   font-family: var(--salt-text-label-fontFamily);
 }
 
+/* Bubble chart data labels use primary foreground for better visibility */
+.highcharts-theme-salt .highcharts-bubble-series .highcharts-data-label text,
+.highcharts-theme-salt .highcharts-bubble-series text.highcharts-data-label {
+  fill: var(--salt-content-primary-foreground);
+}
+
+.highcharts-theme-salt .highcharts-bubble-series .highcharts-point {
```
joshwooding: should this be combined with the scatter selector?

wemmyo:
Bubble labels aren't readable when patterns are enabled. We might need to change text colour to white or something else. Not sure if this has been raised already

wemmyo (approved):

PR #5656, opened 2025-10-10
Vertical Navigation fixes

joshwooding:


packages/core/src/vertical-navigation/VerticalNavigationItemLabel.css
```diff
@@ -1,3 +1,7 @@
 .saltVerticalNavigationItemLabel {
   flex: 1;
 }
+
+.saltVerticalNavigationItemLabel + .saltVerticalNavigationItemExpansionIcon {
+  padding: 0 var(--salt-spacing-100);
```
wemmyo, line 6: Question: We're getting double padding right from the icon and nav item content. Is this expected?

joshwooding: Yes

joshwooding: The extra padding means when a button is used inside navigation item the icons align

PR #5674, opened 2025-10-15
cleanup some biome warnings

mark-tate:


joshwooding (approved):
Do we need a changeset?

PR #5675, opened 2025-10-15
Highcharts | Add Box Plot Chart

tycup:
Add Box Plot Chart

### DIFF
--- a/.changeset/true-results-fly.md
+++ b/.changeset/true-results-fly.md
@@ -1,0 +1,7 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+Added Box plot chart
+
+To see an example Options object required to configure a basic box plot chart, see the examples section.
--- a/packages/highcharts-theme/css/salt-highcharts-theme.css
+++ b/packages/highcharts-theme/css/salt-highcharts-theme.css
@@ -272,5 +272,18 @@
 .highcharts-theme-salt.axes-grid-lines .highcharts-yaxis .highcharts-axis-line,
 .highcharts-theme-salt.axes-grid-lines .highcharts-xaxis-grid .highcharts-grid-line,
 .highcharts-theme-salt.axes-grid-lines .highcharts-yaxis .highcharts-yaxis-grid .highcharts-grid-line {
-  stroke-width: 1;
+  stroke-width: var(--salt-size-fixed-100);
+}
+
+.highcharts-theme-salt .highcharts-boxplot-box {
+  fill: inherit;
+  fill-opacity: 0.75;
+}
+
+.highcharts-theme-salt .highcharts-point .highcharts-boxplot-box {
+  stroke-width: var(--salt-size-fixed-200);
+}
+
+.highcharts-theme-salt .highcharts-boxplot-median {
+  stroke: var(--salt-content-primary-foreground);
 }
--- a/site/docs/components/highcharts-theme/examples.mdx
+++ b/site/docs/components/highcharts-theme/examples.mdx
@@ -78,6 +78,22 @@ Use a bullet chart to show performance against a target across categories. Each 
 
 <LivePreview componentName="highcharts-theme" exampleName="BulletChart" />
 
+## Box plot chart
+
+Use a box plot chart to show the distribution of data across different categories, displaying minimum, first quartile, median, third quartile, and maximum values. To ensure the presentation is accessible, fill patterns can be applied to the chart (see [Patterns and Fills](./usage#patterns-and-fills) for details).
+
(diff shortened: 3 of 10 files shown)


packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -274,3 +274,16 @@
...
   stroke-width: 1;
 }
+
+.highcharts-theme-salt .highcharts-boxplot-box {
+  fill: inherit;
+  fill-opacity: 0.75;
+}
+
+.highcharts-theme-salt .highcharts-point .highcharts-boxplot-box {
+  stroke-width: 2px;
```
joshwooding: Something that I just thought of, ideally all stroke-widths should be using fixed tokens right?

joshwooding: We don't need to do this now though

site/docs/components/highcharts-theme/examples.mdx
```diff
@@ -78,6 +78,18 @@ Use a bullet chart to show performance against a target across categories. Each
 
 <LivePreview componentName="highcharts-theme" exampleName="BulletChart" />
 
+## Box plot chart
```
joshwooding, line 81: What is determining the order of these?

packages/highcharts-theme/src/default-options.ts
```diff
@@ -77,6 +77,9 @@ export const getDefaultOptions = (
           radius: 4,
         },
       },
+      boxplot: {
+        pointWidth: 40,
```
joshwooding: Was this just a value that looks nice? What was the decision making process?

.changeset/true-results-fly.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/highcharts-theme": minor
+---
+
+Added Box Plot Chart
```
joshwooding:
```suggestion
Added Box plot chart
```

packages/highcharts-theme/css/salt-highcharts-theme.css
amritadesmet: We initially strayed away from the 'size-bar' type tokens due to them changing with density, where we'd like consumers to edit themselves. However, I just checked and didn't realize that 'size-fixed-xxx' remains the same across densities. In this case, we'd need to update the stroke in some other charts too (i.e. bullet chart target bar). TBD @tycup

site/docs/components/highcharts-theme/examples.mdx
amritadesmet: We ordered the charts around them being variations or additions of each other (i.e stacked bar after bar chart since it's a more complex version of it or area, scatter and bubble together due to them covering large areas with an opacity). Do you think box plot should be elsewhere in the order? If so, where?

packages/highcharts-theme/src/default-options.ts
tycup: Moved out of defaults and back into examples w/ a best practices note to show that this can be decided by the consumer and this is just an example value to match figma

site/docs/components/highcharts-theme/examples.mdx
tycup: by loose association with the other examples; 

Line
----- both variants of Pie:
Pie
Donut
------ all some variant of column/bar:
Column
Bar
StackedBar (Bar variant)
Bullet (Bar variant/follows visually from StackedBar)
BoxPlot (also visually related to Bar/Column)
---- all charts with opacity fills:
area
scatter
bubble (which visually feels visually related to scatter)


Definitely open to reorganising (like i would probs organise the bar/column section after Line chart since they seem like more commonly known charts etc.)

joshwooding: Honestly I would just make it alphabetical, any other order is more subjective than we probably need to be. Happy for other opinions though

tycup: Let's leave it for now and get another design/content opinion - i'm a fan of them being grouped by 'similarity' like so, but see that its subjective (maybe something for Ben P?)

joshwooding: Yeah, I don't think it's related to this PR.

packages/highcharts-theme/css/salt-highcharts-theme.css
amritadesmet: @joshwooding  We've updated borders to use the fixed tokens in both Figma and this PR, thanks for raising this.

PR #5680, opened 2025-10-16
Add theme link to color page

amritadesmet:
Quick fix, low risk. 

Requested by @bhoppers2008

### DIFF
--- a/site/docs/foundations/color/semantic-use-of-color.mdx
+++ b/site/docs/foundations/color/semantic-use-of-color.mdx
@@ -7,9 +7,11 @@ sidebar:
 
 Colors are divided into several categories to communicate several concepts across the Salt design system.
 
+## Semantic tokens
+
 For further guidance on how to apply theme and tokens using Salt, please refer to the [Themes page](/salt/themes/index).
 
-### Status
+## Status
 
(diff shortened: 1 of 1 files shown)


bhoppers2008:
@amritadesmet think we're missing a trick here with the anchors. Maybe add a heading for semantic tokens which has your new sentence in it, and then bump the heading up for the others (sentiment, status)... so they all appear as anchor links in the right side bar.

amritadesmet:
> @amritadesmet think we're missing a trick here with the anchors. Maybe add a heading for semantic tokens which has your new sentence in it, and then bump the heading up for the others (sentiment, status)... so they all appear as anchor links in the right side bar.

You're right the headers aren't correct. Will update.

bhoppers2008 (approved):

site/docs/foundations/color/semantic-use-of-color.mdx
```diff
@@ -7,7 +7,11 @@ sidebar:
 
 Colors are divided into several categories to communicate several concepts across the Salt design system.
 
-### Status
+## Semantic tokens
+
+For further guidance on how to apply theme and tokens using Salt, please refer to the [Themes page](/salt/themes/index).
```
mark-tate, line 12: Would be nice to have a consistent naming convention for links.
[xxx page] [xxx documentation], [xxx]
Search for [ to see the many forms we have.
We can sweep these up in one go and align as one.
@bpiven

PR #5684, opened 2025-10-16
Fix NumberInput tests

wemmyo:
Fix failing cypress and lint tests:

- Updates disabled class name in test
- Checks if both class name and disabled attribute are present
- Add `kmbt` to allowed words on site


site/allowed-words.txt
```diff
@@ -42,3 +42,4 @@ UITK
 testid
 datasource
 divs
+kmbt
```
wemmyo, line 45: Temporary fix, we might need to disable `cspell` for code snippets at some point.  `kmbt` used in regex formatting example

packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
```diff
@@ -157,8 +157,9 @@ describe("Number Input", () => {
     cy.findByLabelText("increment value").realClick();
     cy.findByLabelText("increment value").should(
       "have.class",
-      "saltNumberInput-increment-disabled",
```
mark-tate, line 160: makes sense we switched to Button as part of code review feedback. Thanks for fixing

site/allowed-words.txt
mark-tate: Will revisit, have a regexp that ignores code snippets somewhere in our code

mark-tate (approved):

PR #5744, opened 2025-10-30
Highcharts | Candlestick Chart

tycup:
Adds Candlestick Chart w/ sentiment colours

### DIFF
--- a/packages/highcharts-theme/css/salt-highcharts-theme.css
+++ b/packages/highcharts-theme/css/salt-highcharts-theme.css
@@ -227,7 +227,8 @@
 }
 
 /* Override default stroke for chart series segment borders */
-.highcharts-theme-salt .highcharts-area-series.highcharts-legend-item .highcharts-point {
+.highcharts-theme-salt .highcharts-area-series.highcharts-legend-item .highcharts-point,
+.highcharts-theme-salt .highcharts-candlestick-series.highcharts-legend-item .highcharts-point {
   stroke: none;
 }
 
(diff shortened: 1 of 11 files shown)


packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -323,3 +324,13 @@
...
 }
+
+.highcharts-theme-salt .highcharts-candlestick-series .highcharts-point-up {
+  stroke: var(--salt-sentiment-positive-foreground-decorative);
+  fill: none;
+}
+
+.highcharts-theme-salt .highcharts-candlestick-series .highcharts-point-down {
+  stroke: var(--salt-sentiment-negative-foreground-decorative);
+  fill: var(--salt-sentiment-negative-foreground-decorative);
```
joshwooding, line 336: Is there a reason this has a fill but the up doesn't?

amritadesmet: I believe this is OOTB behaviour and consumers can change the candlestick fill preference, based on[ this](https://www.highcharts.com/docs/stock/candlestick-chart?gad_source=1&gad_campaignid=22994297277&gbraid=0AAAAAo4v_OvlWalPc7XF67pOpfctyOE3N&gclid=EAIaIQobChMIxoaz7oj8kAMVJolQBh0YbCenEAAYASAAEgKX3PD_BwE#!). @tycup Correct me if I'm wrong?

tycup: Yep it follows defaults but that being said, users could potentially want fully filled candles - with styledmode true, the user can't set the colours through the options as in the link, which means that users would have to target the class themselves.

thoughts on having a custom class such as "filled-candle" to toggle between and make that easier for the consumer?

amritadesmet: @joshwooding What do you think? Trying to get this answer across the line before we request reviews on PR + Figma branch

joshwooding: I don't think we need to add anything yet, but we should set the default to be what we think users will most likely want e.g. 
image

joshwooding: I feel like we're a bit opinionated here already because we're using sentiments

amritadesmet: Sure, happy to conform to what @tycup thinks is best / easiest in code. I have a Figma variant that shows something v similar to this in addition to a variant that matches Tymon's OOTB default so can do either.

tycup: Went for full fills for candlestick by default - there's also [hollowcandlestick](https://www.highcharts.com/docs/stock/hollow-candlestick-chart) for which i supplied classes to and tested. Haven't added it as an example since a variant of candlestick.

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -323,3 +324,28 @@
...
+}
+
+.highcharts-theme-salt .highcharts-hollowcandlestick-series .highcharts-point-up {
+  stroke: var(--salt-sentiment-positive-foreground-decorative);
+  fill: transparent;
+}
+
+.highcharts-theme-salt .highcharts-hollowcandlestick-series .highcharts-point-down {
```
joshwooding: Should we collapse these styles with the non-hollow kind?

joshwooding (approved):

packages/highcharts-theme/css/salt-highcharts-theme.css
amritadesmet: Thanks, adjusting Figma variants to match

joshwooding (approved):

PR #5765, opened 2025-11-05
Remove hardcoded aria-live in banner

jake-costa:
Remove the hardcoded aria-live="polite" from banner and update docs on role="status" limitations.

### DIFF
--- a/.changeset/purple-rice-hammer.md
+++ b/.changeset/purple-rice-hammer.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Removed the hardcoded aria-live="polite" from Banner to address accessibility concerns.


site/docs/components/banner/accessibility.mdx
```diff
@@ -14,6 +14,8 @@ Check and test any focusable elements included in the banner content.
 
 When using a banner to alert the user, apply `role="alert"`. When using a banner to notify the user, apply `role="status"`.
 
+* **Note:** There are known screen reader announcement issues when dynamically inserting content with `role="status"`. During your testing, if you find that adding `role="status"` to the banner does not properly announce conent using a screen reader, we recommend preloading a container with `role="status"` that will house the banner. That will ensure the content is properly announced via screen reader.
```
wemmyo:
```suggestion
**Note:** There are known screen reader announcement issues when dynamically inserting content with `role="status"`. During your testing, if you find that adding `role="status"` to the banner does not properly announce content using a screen reader, we recommend preloading a container with `role="status"` that will house the banner. That will ensure the content is properly announced via screen reader.
```

PR #5771, opened 2025-11-06
Deprecate separable background

### LINKED ISSUE #5768: Deprecate --salt-separable-background
`--salt-separable-background` was created specifically to support the Splitter component. However, it's not being used and has no other current or planned uses within the design system. 

Deprecating it means one less token to manage in code and Figma as well as help simplify the creation and management of future themes e.g. Chase

joshwooding:
Closes #5768


mark-tate:
we will release this wednesday

PR #5772, opened 2025-11-06
improve precision and add support for decimal.js usage

mark-tate:
Lab version.. (copied from previous Core PR)

### DIFF
--- a/packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
+++ b/packages/lab/src/__tests__/__e2e__/number-input/NumberInput.cy.tsx
@@ -44,7 +44,7 @@ describe("Number Input", () => {
     cy.mount(<Default increment={increment} />);
 
     cy.get(".saltNumberInput-increment").realClick();
-    cy.get("incrementSpy").should("have.been.called");
+    cy.get("@incrementSpy").should("have.been.called");
     cy.findByRole("spinbutton").should("have.value", "42");
   });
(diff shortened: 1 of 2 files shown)


.changeset/rude-jars-leave.md
```diff
@@ -0,0 +1,8 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/lab": patch
```

packages/lab/src/number-input/NumberInput.tsx
```diff
@@ -202,40 +217,95 @@ function getNumberPrecision(num: number | string) {
...
+    return null;
+  }
+  if (
+    sanitizedValue === "." ||
+    sanitizedValue === "+" ||
+    sanitizedValue === "-"
+  ) {
+    return 0;
+  }
+  const floatString = Number.parseFloat(value).toFixed(decimalScale);
```
joshwooding, line 266: Should this be `value` or `sanitizedValue` ?

joshwooding (approved):

wemmyo (approved):

packages/lab/src/number-input/NumberInput.tsx
mark-tate: sanitizedValue is used to return early and only offers a trimmed value, which  parseFloat does not need.

PR #5794, opened 2025-11-11
Number input to core

mark-tate:


site/src/examples/number-input/DecimalJS.tsx
```diff
@@ -55,7 +54,7 @@ export const DecimalJS = () => {
           const sanitized = value.replace(/,/g, "");
           try {
             const decimalValue = new Decimal(sanitized);
-            return decimalValue.toNumber();
+            return decimalValue.toDecimalPlaces(5).toNumber();
```
mark-tate, line 57: Small change from the lab release, it ensures the value that is called to onNumberChange, has the same precision as the final formatted value.

bhoppers2008 (approved):

PR #5808, opened 2025-11-14
Salt provider clean up

joshwooding:


packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -247,35 +137,34 @@ function InternalSaltProvider({
...
     mode: inheritedMode,
-    window: inheritedWindow,
     corner: inheritedCorner,
     headingFont: inheritedHeadingFont,
     accent: inheritedAccent,
     actionFont: inheritedActionFont,
   } = useContext(ThemeContext);
 
-  const targetWindow = useWindow();
+  const isRootProvider = prevProvider === null;
```
joshwooding, line 157: Using a new "provider" context means it's directly checking for nested providers instead of trying to fingerprint them

packages/core/src/salt-provider/SaltProvider.tsx
```diff
@@ -310,90 +199,39 @@ function InternalSaltProvider({
...
+                    applyClassesToProp ?? (isRootProvider ? "root" : "scope")
+                  }
+                  density={density}
+                  theme={clsx(
+                    DEFAULT_THEME_NAME,
+                    { [DEFAULT_THEME_NAME_NEXT]: themeNext },
+                    themeProp,
+                  )}
+                  mode={mode}
+                  accent={themeNext ? accent : undefined}
```
joshwooding, line 228: Theme next attributes are set in the context even in the legacy theme so I conditionally pass them to the applicator

packages/core/src/salt-provider/ThemeApplicator.tsx
```diff
@@ -0,0 +1,195 @@
...
+   * Determines font family of action components, mostly Buttons.
+   * @default "Open Sans"
+   */
+  actionFont?: ActionFont;
+  children?: ReactNode;
+}
+
+const providerSymbol = Symbol.for("salt-provider");
+
+export function ThemeApplicator({
```
joshwooding, line 77: This means anything applying the theme to the DOM is centralised

joshwooding: The aim was also to make the theme applicator know less about the types of themes

PR #5809, opened 2025-11-14
Deprecate unused content tokens

### LINKED ISSUE #5691: Deprecate Unused Interactive Content Foreground Tokens
### Background 
Within the Salt Design System, the content characteristic group includes several interactive tokens: content-foreground-visited, content-foreground-hover, and content-foreground-active. However, with the exception of visited, these tokens are not used by the design system’s components or UI examples. Even the visited token—representing the only use of purple foundation colors for visited states—is not used in Salt’s own documentation sites.

Including these unused tokens can cause confusion for consumers, especially since actionable characteristics already provide foreground values for interactive states. The presence of redundant tokens increases the maintenance burden for the Salt DS team, as each token must be managed and mapped across multiple themes and design tools (such as Figma). Removing these tokens will simplify the design system, making it easier for consumers to choose the correct tokens and for the team to support new themes in the future.

### Acceptance criteria
- content-foreground-hover and content-foreground-active tokens are deprecated and removed from the system.
- content-foreground-visited is also deprecated, as it is not used in Salt’s documentation or components.
- Deprecation of any referenced palette tokens in code made redundant by the change
- Deprecation of any referenced palette variables in Figma made redundant by the change
- All references to these tokens are removed from code, documentation, and Figma libraries.
- No visual regressions occur in supported themes (JPM Brand, Legacy, Lab).
- Documentation is updated to clarify which tokens should be used for interactive states (currently in progress)

joshwooding:
Closes #5691

### DIFF
--- a/.changeset/funny-terms-shake.md
+++ b/.changeset/funny-terms-shake.md
@@ -4,6 +4,8 @@
 
 Deprecated unused content tokens.
 
-- `--salt-content-foreground-highlight` can be replaced with `--salt-selectable-background-selected`
-- `--salt-content-foreground-active` can be replaced with `--salt-content-accent-foreground`
-- `--salt-content-foreground-hover` can be replaced with `--salt-content-accent-foreground`
+| Deprecated token                      | Replacement token                       |
+| ------------------------------------- | --------------------------------------- |
+| `--salt-content-foreground-highlight` | `--salt-selectable-background-selected` |
+| `--salt-content-foreground-active`    | `--salt-content-accent-foreground`      |
+| `--salt-content-foreground-hover`     | `--salt-content-accent-foreground`      |


.changeset/funny-terms-shake.md
```diff
@@ -0,0 +1,9 @@
+---
+"@salt-ds/theme": minor
+---
+
+Deprecated unused content tokens.
+
+- `--salt-content-foreground-highlight`
```
pseys: Can we indicate that if a substitute is needed `--salt-selectable-background-selected` can be used as it results in the same values across all themes.

.changeset/funny-terms-shake.md
```diff
@@ -0,0 +1,9 @@
...
+- `--salt-content-foreground-active`
+- `--salt-content-foreground-hover`
```
pseys: For active and hover can we direct people to use `--salt-content-accent-foreground` instead, as it results in the same values across themes.

joshwooding: @pseys feel free to make inline suggestions. When leaving a comment you can use 
image

or feel free to edit the content via the browser

pseys (approved):

ivan-calderon (approved):

PR #5812, opened 2025-11-17
Fix capitalizations on Date family

bpiven:


site/docs/components/date-picker/range-date-picker/index.mdx
wemmyo, line 11: Does this need to be updated?

site/docs/components/date-picker/index.mdx
wemmyo, line 12: Same here

wemmyo (commented):
Left some comments

PR #5824, opened 2025-11-18
fix right click on Slider and NumberInput

mark-tate:


.changeset/slick-icons-peel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/core": patch
+---
+
+Fix Right click on Slider and NumberInput
```
joshwooding:
```suggestion
Fixed right click behavior on Slider and NumberInput.
```

wemmyo: Maybe with the backticks as well. `Slider` and `NumberInput`

packages/core/src/number-input/internal/useLongPressPointerAction.ts
```diff
@@ -0,0 +1,95 @@
...
+import {
+  type PointerEvent,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react";
+
+// Interval hook in TypeScript
+function useInterval(callback: () => void, delay: number | null) {
```
tycup: I see there's also a `useInterval.tsx` file in `internal` that's very similar - is that intended/can be combined?

mark-tate: good shout.

.changeset/slick-icons-peel.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed right click behavior on `NumberInput` and `Slider`.
+
+Previously, when users right-clicked on the increment/decrement buttons of the `NumberInput` or the thumb of the `Slider`, these components could remain stuck in a pressed state. Update the components to properly handle right-clicks and other pointer events by canceling the pressed state when such interactions occur.
```
joshwooding:
```suggestion
Previously, when users right-clicked on the increment/decrement buttons of the `NumberInput` or the thumb of the `Slider`, these components could remain stuck in a pressed state. The components have been updated to properly handle right-clicks and other pointer events by canceling the pressed state when such interactions occur.
```

PR #5839, opened 2025-11-25
Highcharts | Versioning and clean-up

tycup:
- Applied Josh's suggestions for versioning
- Remove unused biome rule

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -65,17 +65,6 @@
           }
         ]
       }
-    },
-    {
-      "files": ["**/highcharts-theme/css/**/*.css"],
-      "rules": {
-        "salt/correct-theme-token-usage": [
-          false,
-          {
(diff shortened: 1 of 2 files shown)


joshwooding (dismissed):

packages/highcharts-theme/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@salt-ds/highcharts-theme",
-  "version": "0.1.0",
+  "version": "0.0.0",
```
wemmyo, line 3: Curious why the version was changed to `0.0.0`

tycup: So all the changesets in the highcharts branch can 'take over' versioning from 0

wemmyo: Got it. Thanks

PR #5889, opened 2025-12-09
Add Tree Component

tycup:
Adds Tree component.

Link to TreeView pattern: https://www.w3.org/WAI/ARIA/apg/patterns/treeview/

Note: Site docs may change slightly based on feedback received by @navkaur76  so primarily looking for implementation review for now.

### DIFF
--- a/site/docs/components/tree/usage.mdx
+++ b/site/docs/components/tree/usage.mdx
@@ -20,14 +20,12 @@ See the [examples](/salt/components/tree/examples) for feature demonstrations in
 ### When to use
 
 - To display hierarchical data with parent-child relationships, such as file systems, organizational structures or category taxonomies.
-- When users need to navigate through nested levels of content and optionally select items.
-- To allow users to expand and collapse sections to focus on relevant content.
+- When users need to explore a multi-level hierarchy, browse nested levels and optionally select items, while allowing multiple branches to remain open.
 
 ### When not to use
 
 - For navigation between pages or routes. Instead, use [`VerticalNavigation`](../vertical-navigation).
 - For flat, non-hierarchical lists. Instead, use [`ListBox`](../list-box).
-- When the hierarchy is only one level deep. Consider using [`Accordion`](../accordion) or a simple list.
 
 ## Import
 


packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,260 @@
...
+      aria-hidden
+      className={clsx("saltTreeNode-expansionIcon", {
+        "saltTreeNode-expansionIcon-expanded": expanded,
+      })}
+    />
+  );
+}
+
+export interface TreeNodeProps extends ComponentPropsWithoutRef<"li"> {
+  /**
```
mark-tate, line 24: Just by reading the docs.
It's not clear in the docs whether this has to be unique ?

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,260 @@
...
+   * Identifier of the node
```
mark-tate: In the basic/default example, we don't use this ID but it's provided?
Does it have to be provided ?

site/src/examples/tree/Controlled.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <StackLayout style={{ width: "30ch" }}>
+      <Text>
+        Expanded: {expanded.length > 0 ? expanded.join(", ") : "none"}
+      </Text>
+      <Text>
+        Selected: {selected.length > 0 ? selected.join(", ") : "none"}
+      </Text>
+      <Tree
+        aria-label="File browser"
+        expanded={expanded}
```
mark-tate: On Collapsible we have 

onOpenChange, open and defaultOpen, can we align on some common API/naming rather than create a new API.

Thoughts?

site/src/examples/tree/Disabled.tsx
```diff
@@ -0,0 +1,37 @@
+import { StackLayout } from "@salt-ds/core";
+import { Tree, TreeNode } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+export const Disabled = (): ReactElement => (
+  <StackLayout direction="row" gap={4}>
+    <Tree
```
mark-tate: Is there a read-only tree ?

site/src/examples/tree/WithIcons.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Tree, TreeNode } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+export const WithIcons = (): ReactElement => (
+  <Tree
+    aria-label="File browser"
+    defaultExpanded={["documents"]}
+    style={{ width: "30ch" }}
+  >
+    <TreeNode value="documents" label="Documents" icon={<FolderOpenIcon />}>
```
mark-tate: This is a bit clunky?
Icons are closely tied to the language.
Would it be easier to just give an icon name ?
We then drive the icons to have a meaning and they can be holistically applied through the SemanticIconProvider?

So here the example would be to use the SemanticIconProvider and not just add your own icons, without thinking systematically.

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,260 @@
...
+   */
+  value: string;
+  /**
+   * Label for the node
+   */
+  label: ReactNode;
+  /**
+   * Optional icon to display before the label
+   */
+  icon?: ReactNode;
```
mark-tate: This is a ReactNode, so you could pass a string ?

packages/lab/src/tree/TreeContext.ts
```diff
@@ -0,0 +1,111 @@
+import {
+  createContext,
+  type Dispatch,
+  type SetStateAction,
+  type SyntheticEvent,
+  useContext,
+} from "react";
+
+export interface TreeContextValue {
+  /** Set of expanded node values (Set for faster lookup) */
```
mark-tate: is this documented.
First pass of docs this did not come across ?
Should I be doing this ?

packages/lab/src/tree/TreeContext.ts
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Sets if selecting a parent node should also select its descendants
+   * Only applies when multiselect or checkbox prop is enabled
+   */
+  propagateSelect: boolean;
+  /**
+   * Sets if selecting all children should automatically select the parent
+   * Only applies when multiselect or checkbox is enabled
+   */
+  propagateSelectUpwards: boolean;
```
mark-tate: Do we have 2 props which are mutually exclusive, what happens if I set both ?
Should I select both, not sure this was explicit enough in the docs?

packages/lab/src/tree/TreeContext.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Sets if selection can be toggled off in single-select mode */
+  togglableSelect: boolean;
```
mark-tate: FYI a Boolean may soon be deprecated if further behaviours are needed ?
Or will lead to multiple booleans that can conflict ?
Seems this only applies to single select? so I wonder if this prop is the same prop as multi-select ?
Do we have any other props anywhere which describe selection behaviour, maybe this needs to be a union, maybe not? 
But worth a conversation?

packages/lab/src/tree/TreeContext.ts
```diff
@@ -0,0 +1,111 @@
...
+  /** Set the active node */
+  setActiveNode: Dispatch<SetStateAction<string | undefined>>;
+  /** Whether focus is visible */
+  focusVisible: boolean;
+  /** Set focus visible state */
+  setFocusVisible: Dispatch<SetStateAction<boolean>>;
+  /** Disabled state of of tree */
+  disabled: boolean;
+  /** Set of disabled node IDs */
+  disabledIdsSet: Set<string>;
```
mark-tate: We refer I other variables to Node and that is a type string ?
and here we talk about Ids ?
Are these Nodes as well ?
Also, we suffixed with Set, that seems redundant as the types is a Set ?
Maybe this should be disabledNode?

packages/lab/src/tree/TreeNode.tsx
tycup: Not sure what you mean? the `value` prop which identifies the node is used both in storybook and site examples. It is marked as not an optional prop so it does have to be provided

site/src/examples/tree/Controlled.tsx
tycup: Currently:

- `Accordion`
- `NavigationItem`
- `Stepper/Step`

use expanded too. Is `expanded` and `open` the same? To me, tree nodes expand to reveal children in-place within a hierarchical structure. "Open" implies revealing something in a separate layer (like a dialog or dropdown)

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,420 @@
...
+   */
+  onExpandedChange?: (event: SyntheticEvent | null, expanded: string[]) => void;
+  /**
+   * Callback on node expanded or collapsed
+   */
+  onExpand?: (
+    event: SyntheticEvent | null,
+    value: string,
+    expanded: boolean,
+  ) => void;
```
wemmyo: These seem similar, do we need both?

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,420 @@
...
+  /**
+   * Sets if selecting a parent node should also select its descendants
+   * Only applies when multiselect is enabled
+   */
+  propagateSelect?: boolean;
+  /**
+   * Sets if selecting all children should automatically select the parent
+   * Only applies when multiselect is enabled
+   */
+  propagateSelectUpwards?: boolean;
```
wemmyo: Question: Should this be the default behaviour? Does it need to be a prop?

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,420 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Default disabled node IDs (uncontrolled)
+   */
+  defaultDisabledIds?: string[];
+  /**
+   * Disabled node IDs (controlled).
+   */
+  disabledIds?: string[];
```
wemmyo: Would it be better to remove these and just handle disabled at the node?

packages/lab/src/__tests__/__e2e__/tree/Tree.cy.tsx
```diff
@@ -1,2157 +1,1366 @@
...
+            </TreeNode>
+          </Tree>,
+        );
+        cy.findByRole("tree").should("exist");
+        cy.findByRole("treeitem", { name: "Node 1" }).should("exist");
+        cy.realPress("Tab");
+        cy.findByRole("treeitem", { name: "Node 1" }).should("be.focused");
+      });
+
+      it("should update focus target as nested selected nodes become visible through expansion", () => {
```
tycup, line 1933: TODO: Check if this is the correct focus behaviour; tldr, if a deeply nested node is selected, then parents are collapsed, where should the focus go to/return after leaving the tree?

tycup: @karlgoldstraw @jake-costa Your expertise here would be appreciated !

packages/lab/src/tree/TreeContext.ts
tycup: Removed after @bhoppers2008 feedback on preventing toggling this behaviour

packages/lab/src/tree/Tree.tsx
tycup: Yep correct - removed in favour of being default

packages/lab/src/tree/TreeNode.tsx
tycup: Updated in usage.mdx

site/src/examples/tree/Disabled.tsx
tycup: I see a scenario where a user has a 'read-only' permission and is not allowed to make selections but may still want to be aware of a hierarchy..? Let me know if thats too far out...

Also consistent with List box

site/src/examples/tree/WithIcons.tsx
tycup: Done. Users can now pass icons as: `<TreeNode icon={DocumentIcon} value="documents" label="Documents" />`

packages/lab/src/tree/TreeNode.tsx
tycup: Updated to `ComponentType<IconProps>;`

packages/lab/src/tree/TreeContext.ts
tycup: Unclear on my side - I am referring to the Set data structure

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -0,0 +1,211 @@
...
+  disabled?: boolean;
+  /**
+   * Child nodes. Nested TreeNodes create a hierarchy.
+   */
+  children?: ReactNode;
+}
+
+const withBaseName = makePrefixer("saltTreeNode");
+
+// Need to take another look at this because its slightly brittle - alternative could be:
```
tycup, line 48: @joshwooding thoughts on this?

packages/lab/src/tree/TreeNodeTrigger.css
```diff
@@ -0,0 +1,59 @@
...
+  font: inherit;
+  text-align: inherit;
+  margin: 0;
+  outline: none;
+
+  box-sizing: border-box;
+  display: flex;
+  align-items: flex-start;
+  gap: var(--salt-spacing-100);
+  width: 100%;
```
wemmyo, line 6: Wonder if nested nodes taking full width of the container is an accessibility issue

tycup: Can you elaborate? Not sure i understand why it would be

wemmyo: As nested level increases the the trigger will take up more space than is probably needed

tycup: The trigger element occupies the same total width at every level (100% of the container) and what changes is the distribution of that width between the indentation padding and the content area. It's the `padding-left` formula that changes with nesting level. If your concern is with excessive indentation, thats a valid UX consideration but its a deliberate design trade off that (i presume) all treeview patterns have with deeply nested trees

site/src/examples/tree/WithIcons.tsx
```diff
@@ -0,0 +1,37 @@
...
+    aria-label="File browser"
+    defaultExpanded={["documents"]}
+    style={{ width: "30ch" }}
+  >
+    <TreeNode value="documents" label="Documents" icon={FolderOpenIcon}>
+      <TreeNode value="reports" label="Reports" icon={FolderClosedIcon}>
+        <TreeNode
+          value="annual-report"
+          label="Annual Report"
+          icon={DocumentIcon}
```
wemmyo: Does this only support salt icons? If so, do we need to be able to support any icon?

packages/lab/src/tree/TreeContext.ts
```diff
@@ -0,0 +1,99 @@
+import {
+  createContext,
```
wemmyo: We should probably use `createContext` from our utils

tycup: Nice catch:) Fixed!

site/docs/components/tree/examples.mdx
```diff
@@ -0,0 +1,55 @@
...
+## With icons
+
+Use the `icon` prop on `TreeNode` to display an icon alongside the label. Icons help users quickly identify the type of content each node represents.
+
+### Best practices
+
+- Only use icons when they provide meaningful context to help users identify node types.
+- Apply icons to all nodes or none for visual consistency. Mixing nodes with and without icons creates an unbalanced visual hierarchy.
+- Use consistent icons for similar node types (e.g., folder icons for parent nodes, document icons for leaf nodes).
+- Icons in a tree are decorative. Ensure that any information conveyed by icons is also available via the node label.
```
navkaur76, line 26: Following on from what was discussed around icon usage + appropriate indentation levels to clearly indicate hierarchy, could it be useful to add a line on that here? LMK what you think @tycup @joshwooding

site/docs/components/tree/examples.mdx
```diff
@@ -0,0 +1,55 @@
...
+
+<LivePreview componentName="tree" exampleName="WithIcons" />
+
+## Multiselect
```
navkaur76, line 30: To align with 'With icons', should this be 'With checkboxes' ??

site/docs/components/tree/examples.mdx
```diff
@@ -0,0 +1,55 @@
...
+### Best practices
+
+- Use checkbox selection when users need to select multiple items for a bulk action.
+- Selecting a parent node automatically selects all its children, and selecting all children automatically selects the parent.
+
+<LivePreview componentName="tree" exampleName="Multiselect" />
+
+## Disabled nodes
+
+Use the `disabled` prop on individual `TreeNode` components to disable specific nodes. Disabled nodes cannot be selected, expanded, or interacted with.
```
navkaur76, line 43: Tymon's informed me that @karlgoldstraw  is currently carrying out an accessibility review on disabled nodes + focusing etc. Just for visibility, it would be good to confirm whether we're definitely offering disabled nodes OOTB, and if so, it would be good to keep in mind what this looks like for checkbox tree (extending the disabled styling to include the checkbox itself in a disabled state too). Possibly worth supporting with storybook examples?

tycup: Disabled variants are already present in storybook under "Disabled" and "Disabled Nodes" stories

site/docs/components/tree/examples.mdx
tycup: I'm not fully sure if it adds value - we define Tree in the header to be "Tree displays hierarchical data as an expandable and collapsible structure. Users can navigate nested items and optionally select one or more nodes."

Feels to me like the idea of indentation levels is intrinsic to what a tree view pattern is, so im not fully sure if that wouldn't be too verbose

But perhaps the first line could be improved to "..they provide meaningful context to help users identify parent nodes and leaf nodes"? along those lines, if we want to be more direct

site/docs/components/tree/examples.mdx
tycup: Also the multiselect variant (checkbox) already extends the disabled styling - you can see this in storybook on the disabled story and toggling 'multiselect' in the controls:
Screenshot 2026-01-22 at 14 41 31

site/docs/components/tree/examples.mdx
tycup: With checkboxes might be less intuitive on the development side - if the prop name is `multiselect` and its annotated as `With checkboxes`. Could do `With Multiselection`...?

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,211 @@
...
+          )}
+          style={
+            {
+              "--saltTreeNode-level": level,
+            } as CSSProperties
+          }
+          {...rest}
+        >
+          {usesLabelProp ? (
+            <TreeNodeTrigger>
```
wemmyo: Here's the cause of the problem. `aria-expanded` and `aria-selected` are not being passed to the trigger which is why it wasn't announced correctly

packages/lab/src/tree/Tree.tsx
tycup: Currently you can do both - do you reckon having this doesn't add enough value to substantiate keeping it?

site/src/examples/tree/WithIcons.tsx
tycup: As discussed in person, current implementation is valid (for custom icons, users could use the icon typing and Icon component too)

packages/lab/src/tree/Tree.css
```diff
@@ -1,41 +1,11 @@
...
-  max-height: inherit;
+  padding: 0;
   outline: none;
-  overflow-y: auto;
-  position: relative;
-  user-select: none;
-}
-
-.saltTree-child-nodes {
   width: 100%;
```
joshwooding, line 6: We need to add a box-sizing: border-box

packages/lab/src/tree/Tree.tsx
joshwooding: I think it would be confusing offering both, we should probably settle on just one

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,391 @@
...
+      if (activeEl !== element) {
+        element.focus();
+        element.scrollIntoView({ block: "nearest", inline: "nearest" });
       }
-    }
-    if (item.childNodes) {
-      getSelectedItemsFromSource(item.childNodes, selectionStrategy, result);
-      if (isSingleSelection && result.length === 1) {
-        break;
+    }, [activeNode, getElement, targetWindow]);
```
joshwooding: I know we use this in some use-cases but we should see if we can rely on focus in events

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+        } as CSSProperties
+      }
+      onClick={handleClick}
+      onFocus={handleFocus}
+      onBlur={handleBlur}
+      onMouseDown={handleMouseDown}
+      onKeyDown={handleKeyDown}
+      {...rest}
+    >
+      <span ref={triggerContentRef} className={withBaseName()}>
```
joshwooding, line 164: This ref not going to the root (`ref` prop) is not standard.

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+          [withNodeBaseName("expanded")]: expanded,
+          [withNodeBaseName("selected")]: selected && !multiselect,
+          [withNodeBaseName("active")]: isActive,
+          [withNodeBaseName("disabled")]: disabled,
+          [withNodeBaseName("hasChildren")]: hasChildren,
+          [withNodeBaseName("focusVisible")]: focusVisible,
+        },
+        className,
+      )}
+      style={
```
joshwooding, line 151: If `style` is passed in it will replace this

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+  };
+
+  const handleMouseDown = (event: MouseEvent<HTMLLIElement>) => {
+    wasMouseDownRef.current = true;
+    onMouseDown?.(event);
+  };
+
+  const handleKeyDown = (event: KeyboardEvent<HTMLLIElement>) => {
+    onKeyDown?.(event);
+  };
```
joshwooding: I don't think we need this

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+    select(event, value);
+  };
+
+  const handleFocus = (event: FocusEvent<HTMLLIElement>) => {
+    if (event.target !== nodeRef.current) return;
+    if (!wasMouseDownRef.current) {
+      setFocusVisible(true);
+    }
+    wasMouseDownRef.current = false;
+    onFocus?.(event);
```
joshwooding: This is a tricky one, but I think i'd expect onFocus to always be called

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+    if (!disabled) {
+      setActiveNode(value);
+    }
+  };
+
+  const handleBlur = (event: FocusEvent<HTMLLIElement>) => {
+    if (event.target !== nodeRef.current) return;
```
joshwooding: Just curious would comparing it with event.currentTarget work

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  useEffect(() => {
+    if (nodeRef.current) {
+      return registerElement(value, nodeRef.current);
+    }
+  }, [value, registerElement]);
+
+  const handleClick = (event: MouseEvent<HTMLLIElement>) => {
+    onClick?.(event);
+    if (disabled) return;
```
joshwooding, line 94: Let's discuss this, another weird one but if we're trying to simulate disabled onClick shouldn't be called

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  const isTabbable = !disabled && tabbableNodeId === value;
+
+  // Expose the trigger content span for tooltip positioning
+  // The li handles focus and ARIA, but positioning should be relative to the visual row
+  useImperativeHandle(
+    ref,
+    () => triggerContentRef.current as HTMLSpanElement,
+    [],
+  );
```
joshwooding: would useForkRef not work here?

packages/lab/src/tree/TreeNodeTrigger.css
```diff
@@ -0,0 +1,56 @@
+.saltTreeNodeTrigger {
+  box-sizing: border-box;
+  display: flex;
+  align-items: flex-start;
+  gap: var(--salt-spacing-100);
+  width: 100%;
+  min-height: var(--salt-size-base);
+
+  padding-top: calc((var(--salt-size-base) - var(--salt-size-selectable)) / 2);
+  padding-bottom: calc((var(--salt-size-base) - var(--salt-size-selectable)) / 2);
```
joshwooding, line 10: Does this align with Figma?

packages/lab/src/tree/TreeNode.tsx
joshwooding: I'd like it if we could simplify it :)

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,161 @@
...
+
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-tree-node",
+      css: treeNodeCss,
+      window: targetWindow,
+    });
+
+    const generatedId = useId();
+    const id = generatedId ?? value;
```
joshwooding:
```suggestion
    const id = useId(value);
```

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!nodeContext) {
+    throw new Error("TreeNodeExpansionIcon must be used within a TreeNode");
+  }
+
+  const { value, hasChildren, expanded, disabled } = nodeContext;
+  const { toggleExpanded } = useTreeContext();
+  const { ExpandGroupIcon } = useIcon();
+
+  const handleClick = (event: MouseEvent<HTMLSpanElement>) => {
+    event.stopPropagation();
```
joshwooding: I'm a bit worried about using this, I guess this is because it would trigger an onClick from the node?

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+    );
+  }
+
+  return (
+    // biome-ignore lint/a11y/useKeyWithClickEvents: keyboard handled at tree level, same as W3C
+    <span
+      ref={ref}
+      className={clsx(withBaseName(), className)}
+      onClick={handleClick}
+      aria-hidden="true"
```
joshwooding: NIT: aria-hidden is applied 2 different ways in this file

joshwooding: Also we don't need two aria-hiddens nested

packages/lab/src/tree/TreeNodeTrigger.css
```diff
@@ -0,0 +1,56 @@
...
+  width: var(--salt-size-selectable);
+  height: var(--salt-size-selectable);
+  min-width: var(--salt-size-selectable);
+  flex-shrink: 0;
+  color: var(--salt-content-primary-foreground);
+}
+
+.saltTreeNode-disabled .saltTreeNodeTrigger > .saltIcon {
+  color: var(--salt-content-primary-foreground-disabled);
+}
```
joshwooding, line 70: What happens with disabled selected

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,204 @@
...
+    ...rest
+  } = props;
+
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-tree-node-trigger",
+    css: treeNodeTriggerCss,
+    window: targetWindow,
+  });
+  useComponentCssInjection({
```
joshwooding: Usually we'd only have one useComponentCssInjection, do we need two?

packages/lab/src/tree/TreeNodeTrigger.css
joshwooding, line 70: @pseys is it worth trying the simplified disabled styles here first?

packages/lab/src/tree/useTree.ts
```diff
@@ -1,179 +1,545 @@
...
+  };
+
+  const registerTrigger = (value: string, element: HTMLElement) => {
+    triggersRef.current.set(value, element);
+    return () => {
+      triggersRef.current.delete(value);
+    };
+  };
+
+  const getTrigger = (value: string): HTMLElement | undefined => {
```
joshwooding: Is this used?

packages/lab/src/tree/useTree.ts
```diff
@@ -1,179 +1,545 @@
...
+        ? [...expandedArray, value]
+        : expandedArray.filter((v) => v !== value);
+
+      setExpandedArray(newExpanded);
+      onExpandedChange?.(event, newExpanded);
+    },
+    [expandedArray, expandedState, onExpandedChange],
+  );
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: getParent/getChildren only depend on treeModel which is already in deps
```
joshwooding: Still should be included :)

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
```diff
@@ -0,0 +1,75 @@
...
+      {...rest}
+    >
+      <ExpandGroupIcon
+        aria-hidden
+        className={clsx(withBaseName("icon"), {
+          [withBaseName("icon-expanded")]: expanded,
```
joshwooding: This is going to be a weird one, we're currently rotating the expandgroup icon instead of using the collapsegroupicon, which isn't technically semantically correct however, I'm guessing it's for the animation?

joshwooding: cc @navkaur76

site/docs/components/tree/usage.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+- Each `TreeNode` requires a unique `value` prop. This value identifies the node for expansion, selection, and disabled state tracking. Ensure values are unique across the entire tree.
+- Use uncontrolled mode with `defaultExpanded` and `defaultSelected` for simple use cases, or controlled mode with `expanded` and `selected` props when you need to manage state externally.
+
+See the [examples](/salt/components/tree/examples) for feature demonstrations including icons, multiselect, and disabled nodes.
+
+### When to use
+
+- To display hierarchical data with parent-child relationships, such as file systems, organizational structures or category taxonomies.
+- When users need to navigate through nested levels of content and optionally select items.
```
joshwooding: Not sure about using the word "navigate" here, what would this use-case be?

site/docs/components/tree/usage.mdx
```diff
@@ -0,0 +1,60 @@
...
+- To allow users to expand and collapse sections to focus on relevant content.
```
joshwooding: This seems like accordion?

site/docs/components/tree/usage.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+### When not to use
+
+- For navigation between pages or routes. Instead, use [`VerticalNavigation`](../vertical-navigation).
+- For flat, non-hierarchical lists. Instead, use [`ListBox`](../list-box).
+- When the hierarchy is only one level deep. Consider using [`Accordion`](../accordion) or a simple list.
```
joshwooding: I don't think accordion and tree should be related?

site/docs/components/tree/usage.mdx
```diff
@@ -0,0 +1,60 @@
...
+
+### `TreeNodeTrigger`
+
+<PropsTable packageName="lab" componentName="TreeNodeTrigger" />
+
+### `TreeNodeLabel`
+
+<PropsTable packageName="lab" componentName="TreeNodeLabel" />
+
+## References
```
joshwooding: I don't think we need references on all the pages

packages/lab/src/tree/TreeNodeTrigger.tsx
tycup: Agree - as a consumer, if i pass onFocus to treeNodeTrigger id expect it to be called for every focus event inc bubbled ones. Also applying to the onBlur below

packages/lab/src/tree/Tree.tsx
tycup: Updated - can you check if i understood correctly? also aware i could've probs just used a boolean but felt clearer this way imo

packages/lab/src/tree/TreeNodeTrigger.tsx
tycup: Yep, nice :)

packages/lab/src/tree/TreeNode.tsx
tycup: Updated but to
`const id = useId(value) ?? value` since useId() can also return undefined and i wouldnt want to widen contexts's `id` to include `undefined`

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
tycup: Yeah...i removed this and moved to prevent this inside `TreeNodeTriggers` event handler using `closest()` - let me know if thats fine, same pattern as with nested `role=treeitem` elements getting clicked

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
tycup: Tbh i jumped ahead...No other component animates it's chevron e.g. accordion or dropdown so will remove

packages/lab/src/tree/TreeNodeTrigger.tsx
tycup: Could we just move the onClick below the `if(disabled) return` statement?

tycup: Think ToggleButton does something similar (guards against disabled first, and only after calls onClick)

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
jake-costa: @tycup would be nice if they did (rotate the Chevron) :)

packages/lab/src/tree/Tree.tsx
tycup: done - disabledIds is out of pattern w/ rest of library + this surfaced a bug with * key expansion (it incorrectly expanded disabled nodes)

packages/lab/src/tree/TreeNodeTrigger.css
tycup: @navkaur76 What would the expected styles be here? The current API allows to set a default set of selected nodes

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
tycup: @jake-costa Agreed but sadly inconsistent with other similar components :(

packages/lab/src/tree/TreeNodeTrigger.tsx
```diff
@@ -0,0 +1,184 @@
...
+    if (nodeRef.current) {
+      return registerElement(value, nodeRef.current);
+    }
+  }, [value, registerElement]);
+
+  const handleClick = (event: MouseEvent<HTMLLIElement>) => {
+    onClick?.(event);
+    if (disabled) return;
+    const target = event.target as HTMLElement;
+    if (target.closest(".saltTreeNodeExpansionIcon")) return;
```
tycup, line 96: @joshwooding Re using `if(event.target !== event.currentTarget) return`

If we were to use that, most clicks would be 'rejected' because target is almost never the `<li>` itself - basically always the descendant e.g. label span, checkbox or icon. So we need to specifically ignore clicks on the expansion chevron (since it has its own toggle handler) and ignore the clicks that bubble up from nested `treeitem` (since child `<li>`s live inside this `<li>`s `<ul role="group">`) (and we still need to allow clicks on all other descendants like the label and checkbox to trigger selection). 

bit iffy

tycup: (only times it would work is if the user clicked on the very border of the li so basically never)

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,398 @@
...
+
+            const currentIndex = activeNode
+              ? visibleNodes.indexOf(activeNode)
+              : -1;
+            let found = false;
+
+            for (let i = currentIndex + 1; i < visibleNodes.length; i++) {
+              const element = getElement(visibleNodes[i]);
+              if (
+                element?.textContent?.toLowerCase().startsWith(searchString)
```
mark-tate, line 296: If I select I with the invoice tree expanded to cycle between all 3 invoices, it can get a bit laggy...
If I hit I and then wait before I hit it again, then it's ok but
Something to follow up on, in RC state.

site/docs/components/tree/accessibility.mdx
```diff
@@ -0,0 +1,93 @@
...
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Best practices
+
+- Provide an accessible name for the tree using `aria-label` or `aria-labelledby`.
+- Ensure node labels are descriptive and convey the purpose of each item.
+- Avoid deeply nested structures that may be difficult to navigate.
```
mark-tate, line 15: On the other page we say that Tree is good for file structures, you can't really control depth in that case.

That aside though, this statement, does not tell you how deep you can go, that's open to interpretation, is there any recommendations as to an optimum max depth. Beyond that, then you accept a degraded UX.
Giving folk some values, may help them define a more  consistent and better UX.

site/src/examples/tree/WithIcons.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Tree, TreeNode } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+export const WithIcons = (): ReactElement => (
+  <Tree
+    aria-label="File browser"
+    defaultExpanded={["documents"]}
+    style={{ width: "30ch" }}
+  >
+    <TreeNode value="documents" label="Documents" icon={FolderOpenIcon}>
```
mark-tate: Is this the only one with an OpenIcon, if representing a FS, should we be toggling icons to match the state ?

packages/lab/src/tree/Tree.tsx
```diff
@@ -1,367 +1,398 @@
...
+        }
+        case "Enter": {
+          handled = true;
+          if (activeNode) {
+            select(event, activeNode);
+          }
+          break;
+        }
+        case " ": {
+          handled = true;
```
mark-tate, line 235: If I press and hold, then it starts flashing

packages/lab/src/tree/TreeNodeTrigger.css
pseys: As discussed in yesterdays stand-up, let's release as is and the include it as part of the wider `selectable` disabled simplification.

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
```diff
@@ -0,0 +1,71 @@
...
+  if (!nodeContext) {
+    throw new Error("TreeNodeExpansionIcon must be used within a TreeNode");
+  }
+
+  const { value, hasChildren, expanded, disabled } = nodeContext;
+  const { toggleExpanded } = useTreeContext();
+  const { ExpandGroupIcon, CollapseGroupIcon } = useIcon();
+
+  const handleClick = (event: MouseEvent<HTMLSpanElement>) => {
+    onClick?.(event);
```
wemmyo, line 41: Should this onClick go below the disabled check?

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,160 @@
...
+  return { contentChildren, nodeChildren };
+}
+
+export const TreeNode = forwardRef<HTMLLIElement, TreeNodeProps>(
+  function TreeNode(props, ref) {
+    const {
+      value,
+      label,
+      icon: Icon,
+      disabled: disabledProp = false,
```
wemmyo: Nit: Wondering if we should remove the default value, should this default to false or undefined?

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -1,72 +1,160 @@
...
+      () =>
+        usesLabelProp
+          ? { contentChildren: [], nodeChildren: Children.toArray(children) }
+          : separateChildren(children),
+      [children, usesLabelProp],
+    );
+
+    const hasChildren = nodeChildren.length > 0;
+
+    // Forward ref to the <li> element rendered by TreeNodeTrigger
```
wemmyo, line 116: ref was forwarded to the span right?

packages/lab/src/tree/TreeNodeExpansionIcon.tsx
tycup: Nope - also because disabled isnt handled like native html disabled  (which does eat the event and never fire)- it's essentially `aria-disabled` toggle with some supression - onClick should still fire before. From an accessibility pov, we do this so that the user can still 'discover' the aria-disabled element, and then onClick can be used to e.g. communicate _why_ its disabled (And it cant do that if it eats it up)

packages/lab/src/tree/Tree.tsx
tycup: To be followed up on after labs merge

site/docs/components/tree/usage.mdx
tycup: @navkaur76

site/docs/components/tree/usage.mdx
tycup: @navkaur76

site/docs/components/tree/usage.mdx
tycup: @navkaur76

navkaur76: I agree with Josh, we can remove that last point about Accordion.

site/docs/components/tree/usage.mdx
navkaur76: Can this point be replaced with "When users need to explore a multi-level hierarchy, browse nested levels and optionally select items, while allowing multiple branches to remain open."

site/docs/components/tree/usage.mdx
navkaur76: We can remove this point entirely as it is covered by the point I made above ("When users need to explore a multi-level hierarchy, browse nested levels and optionally select items, while allowing multiple branches to remain open.")

packages/lab/src/tree/TreeNodeTrigger.tsx
tycup: Noted for next PR

site/docs/components/tree/accessibility.mdx
tycup: Noted for next PR

PR #5892, opened 2025-12-09
Date controls updates

mark-tate:


.changeset/curly-planes-remain.md
```diff
@@ -0,0 +1,70 @@
+---
+"@salt-ds/lab": patch
+---
+
+Calendar updates
+
+- fixed styling issues with selected range border.
+- improved screenreader support.
+- code hygiene improvements
+  - remove un-used code/types and simplified structure of selection types.
```
wemmyo: These two seem related. Do we need to include code clean up in the changeset?

.changeset/curly-planes-remain.md
```diff
@@ -0,0 +1,70 @@
...
+-  disableNavigateNext={isNextDisabled}
++  NextButtonProps={{
++    onClick: handleNext,
++    disabled: isNextDisabled
++  }}
+/>
+```
+
+- `CalendarNavigation` now adds tooltips to out of range dates.
+- When a Calendar has a min/maxDate and the user navigates to a month outside of that range, the Calendar will now automatically navigate back to the closest valid month.
```
wemmyo:
```suggestion
- When a Calendar has a `minDate` or `maxDate` and the user navigates to a month outside of that range, the Calendar will now automatically navigate back to the closest valid month.
```

.changeset/curly-planes-remain.md
```diff
@@ -0,0 +1,70 @@
...
+};
+
+<Calendar
+  createAnnouncement={customCreateAnnouncement}
+  timezone="UTC"
+/>
+```
+
+- `CalendarNavigation` has new props `PreviousButtonProps` and `NextButtonProps` to pass props to the navigation buttons.
+  These props can be used in place of `onNavigateNext`, `disableNavigateNext` and `onNavigatePrevious`, `disableNavigateprevious`.
```
wemmyo:
```suggestion
  These props can be used in place of `onNavigateNext`, `disableNavigateNext` and `onNavigatePrevious`, `disableNavigatePrevious`.
```

packages/core/src/aria-announcer/AriaAnnounce.tsx
```diff
@@ -1,24 +1,36 @@
...
   const { announce } = useAriaAnnouncer();
 
+  // biome-ignore lint/correctness/useExhaustiveDependencies: ignore rest
   useEffect(() => {
     if (announcement) {
-      announce(announcement);
+      if (delay !== undefined) {
+        announce(announcement, delay);
+      }
+      announce(announcement, rest);
```
wemmyo: Is it intended to call announce twice when `delay !== undefined`?

packages/core/src/spinner/Spinner.tsx
```diff
@@ -108,19 +108,20 @@ export const Spinner = forwardRef<HTMLDivElement, SpinnerProps>(
...
+          announce(ariaLabel, { ariaLive: "assertive" });
         }, announcerInterval);
 
       return () => {
         if (disableAnnouncer) return;
 
         interval && clearInterval(interval);
         if (completionAnnouncement) {
-          announce(completionAnnouncement);
+          announce(completionAnnouncement, { ariaLive: "assertive" });
```
tycup: `Pagination` also needs this updated

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -1,82 +1,119 @@
...
 
-  useEffect(() => {
-    mountedRef.current = true;
-    return () => {
-      mountedRef.current = false;
-    };
+  const value = useMemo(() => ({ announce }), [announce]);
+
+  const handleMounted = useCallback((node: HTMLDivElement | null) => {
+    mountedRef.current = node !== null;
```
tycup: What's `mountedRef.current` used for? I don't see it used as a guard anywhere

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
```diff
@@ -1,82 +1,119 @@
...
-          setTimeout(() => {
-            announceAll();
-          }, ARIA_ANNOUNCE_DELAY);
-        } else {
-          isAnnouncingRef.current = false;
-        }
-      });
-    }
-  }, []);
+  const makeAnnouncement = useCallback(
```
tycup, line 59: I think the legacy delay is being applied twice - once in`useAriaAnnouncer`, via setTimeout and again via `AriaAnnouncerProvider.announce`. Potential fix could be to forward `options` instead of `legacyDelay` after the hook waited

.changeset/curly-planes-remain.md
mark-tate: combined with another line

- type improvements
  - remove un-used code/types and simplified structure of selection types.
  - simplify types, generics for usage of date adapters are no longer essentially required unless you are creating a custom date adapter.

.changeset/curly-planes-remain.md
mark-tate: done

mark-tate:
> Looks like we've introduced a visual regression when date picker calendar icon is clicked and the dialog is open, the icon has a grey background

Intended, part of Button

/* Styles applied when the button triggers a dialog or menu */
.saltButton[aria-expanded="true"][aria-haspopup="menu"],
.saltButton[aria-expanded="true"][aria-haspopup="dialog"] {
  background: var(--saltButton-background-active, var(--button-background-active));
  color: var(--saltButton-text-color-active, var(--button-text-color-active));
  border-color: var(--saltButton-borderColor-active, var(--button-borderColor-active));
}

packages/core/src/aria-announcer/AriaAnnounce.tsx
mark-tate: good spot

packages/core/src/spinner/Spinner.tsx
mark-tate: Maybe not... the default was assertive for all announcements and the new default is polite
So this change above keeps the spinner completion as before the change (assertive).
So I think Pagination is ok, (was assertive, now polite) and I suspect we remove this assertive, so spinner announcements are also polite @jake-costa @karlgoldstraw can you confirm ?

packages/core/src/aria-announcer/AriaAnnouncerProvider.tsx
mark-tate: good shout... redundant code, part of the old implementation
as seen in this draft PR (where this implementation was sketched out)
https://github.com/jpmorganchase/salt-ds/pull/5818/changes#diff-e7f66aec40a0f26fedc50a191bb7182015a4124161aa992f7aa4be336af2cf1b

removed

jake-costa:
- [ ] A few DatePicker examples with a visible label provided by a parent FormField do not properly programmatically associate the label with the input, so the visible label is not announced to screen readers. **Fixed Y/N/Comment:**  
  Current NVDA output: date input, edit, date format DD MMM YYYY (e.g. 09 June 2024), blank  
  Expected: Select a date, edit, date format DD MMM YYYY (e.g. 09 June 2024), blank  
  Notes:  
  From what I can tell, this appears to be because aria-labelledby is being added to the wrapper <div> instead of the input.  
  aria-labelledby should be on the <input>, not the wrapper.  
  An input should contain either aria-label or aria-labelledby.  
  
  **Fixed Y/N ?**

- [ ] A few range DatePicker examples with a visible group label do not differentiate the start and end inputs. Both inputs announce the same accessible name, so users cannot tell which input has focus. **Fixed Y/N/Comment:**  
  Current for both inputs: select a date range, edit, dd mmm yyyy, blank  
  Expected for the start input: select a date range, start date, edit, dd mmm yyyy, blank  
  Expected for the end input: select a date range, end date, edit, dd mmm yyyy, blank  
  Note: The default range example (no visible group label) already differentiates start and end correctly via aria-label. This may be a use case for the aria-labelledby self-reference technique when a visible group label is present.  
  
  **Fixed Y/N ?**

- [ ] The month combobox announces the abbreviated month (Jan) instead of the full name when collapsed, and the aria-label duplicates the combobox role by including the word "Dropdown". **Fixed Y/N/Comment:**  
  Current: Month dropdown, combo box, Jan, collapsed  
  Expected: Month, January, combo box, collapsed  
  Note: The same role-duplication applies to the year combobox.  
  
  Cannot be fixed at present.

- [ ] When two calendars are displayed in a range DatePicker, both pairs of previous/next month buttons share the same accessible name (Previous Month, Next Month) with no indication of which calendar they control. Screen reader users navigating via the elements list, and voice-control users speaking the button name, cannot distinguish the four buttons. **Fixed Y/N/Comment:**  

  **Fixed Y/N ?**

  Notes:  
  Preferred fix: collapse to a single shared pair, as shipped by Adobe React Aria RangeCalendar and MUI X DateRangeCalendar. Place Previous Month at the left edge of the first calendar and Next Month at the right edge of the second calendar, and remove the inner pair between the two grids. Update the live-region announcement on navigation to convey both visible months (for example, April 2026 and May 2026).  
  If two pairs are preserved, each pair needs a unique accessible name identifying the calendar it sits on rather than the month it navigates to (for example, Previous Month, April 2026 on the left pair and Previous Month, May 2026 on the right).

- [ ] Add aria-hidden="true" to the decorative SVG icon next to the error message in the initial-error example. 

  **Fixed Y/N ?**
  Not a Date Picker issue, comes from StatusIndicator.

- [ ] After selecting the start day in a range DatePicker, no prompt is announced to select the end date. The date confirmation fires but the guidance prompt is missing. **Fixed Y/N/Comment:**  
  Current: start date. Tuesday 14 April 2026 selected  
  Expected: Start date Tuesday 14 April 2026 selected. Select end date.  
  
  **Fixed Y/N ?**

- [ ] After the start date is selected in a range DatePicker, the remaining day buttons announce complete new range for each candidate end date, and the selected start day's label needs to be modified as well. **Fixed Y/N/Comment:**  
  Current for a candidate end date: complete new range: Wednesday 15 April 2026, button  
  Expected for a candidate end date: Wednesday 15 April 2026, select as end date, button  
  Current for the selected start date: start range: Tuesday 14 April 2026 selected  
  Expected for the selected start date: start date: Tuesday 14 April 2026, selected, button  
 
  **Fixed Y/N ?**

- [ ] Dial back the verbosity of the previous/next month announcement. **Fixed Y/N/Comment:**  
  Current: Visible Month starts from Friday 1 May 2026  
  Expected: May 2026.  
  Note: The W3C APG reference implementations for both the Date Picker Dialog and Date Picker Combobox consistently use the full Month Year format on every navigation, regardless of whether the year has changed.  
 
  **Fixed Y/N ?**

- [ ] At 320×256 px the range DatePicker overlay renders two calendars side by side and overflows the viewport. Per dev team this is consumer-implementation responsibility; flagging for potential documentation update. **Fixed Y/N/Comment:**  
 
   **No action planned**

jake-costa:
## Accessibility Findings

* The month combobox announces the abbreviated month (`Jan`) instead of the full name when collapsed, and the `aria-label` duplicates the combobox role by including the word "Dropdown".

  * Current: `Month dropdown, combo box, Jan, collapsed`
  * Expected: `Month, January, combo box, collapsed`
  * Note: The same role-duplication applies to the year combobox.

* The decorative SVG icon next to the error message in the initial-error example is still announced. Add `aria-hidden="true"` so screen readers skip it, since the error state is already conveyed through the helper text and `aria-invalid` on the input.

* After selecting the start day in a range `DatePicker`, no prompt is announced to select the end date. The date confirmation fires but the guidance prompt is missing.

  * Current: `start date. Tuesday 14 April 2026 selected`
  * Expected: `start date. Tuesday 14 April 2026 selected.` followed by `Select end date.`
  * Note: The start-date confirmation already announces correctly via the day button's accessible name, only the `Select end date.` prompt needs to fire from the live region on top of it. Putting the full date back into the live-region message would double up on what the button already announces.

* A few of the locale examples are missing the `lang` attribute on the wrapping element. Some examples do this correctly, but `DateInput`'s `SingleWithLocale` and `RangeWithLocale` still need it.

* In the custom-content example, the listbox is not implemented correctly, its options have `role="button"` instead of `role="option"`, and navigating between the options with the screen reader does not work. Initial focus on open also lands in the current day in the calendar rather than on the first option in the listbox.

  * Note: This may be specific to the example rather than the underlying `ListBox` component, but worth flagging since consumers may copy the pattern.
  
* Lastly, both `<h4 class="saltText saltText-primary">Note</h4>` on the example page looks to be getting flagged by aXe. (Should likely remove heading semantics)

jake-costa:
## Accessibility Findings

### Priority

* The month combobox announces the abbreviated month (`Jan`) instead of the full name when collapsed, and the `aria-label` duplicates the combobox role by including the word "Dropdown".

  * Current: `Month dropdown, combo box, Jan, collapsed`
  * Expected: `Month, January, combo box, collapsed`
  * Note: The same role-duplication applies to the year combobox.

### Low priority

* The decorative SVG icon is still not properly hidden.

### Usability enhancements (Not a WCAG violation)

* When the month or year is changed using the Month or Year dropdown, the live region also announces the new month and year (for example, `May 2026.`) on top of the dropdown's own selection feedback. The dropdown already conveys the selected value to AT, so the live-region message is not needed.

* After selecting the start day in a range `DatePicker`, no follow-up guidance is announced to direct the user to select the end date.
  * Recommended fix: announce exactly `Select end date.` from the live region after start-only selection. Do not repeat the date or the word "selected" in the live-region string, since the button's accessible name already covers that and including it would double up.

## Example Cleanup

* A few of the locale examples are missing the `lang` attribute on the wrapping element. Most examples do this correctly, but `DateInput`'s `SingleWithLocale` and `RangeWithLocale` still need it. Setting the `lang` attribute is typically a consumer responsibility in the host HTML and the `LocalizationProvider` does not emit it, so this is not a component defect. Flagging for example consistency, since other locale examples (`SingleWithLocaleZhCN`) do set it and the inconsistency could mislead downstream teams copying the pattern.

* In the custom-content example, the listbox is not implemented correctly. Its options have `role="button"` instead of `role="option"`, and navigating between the options with the keyboard is not possible.

packages/date-adapters/src/date-fns-tz-adapter/index.ts
```diff
@@ -66,26 +84,44 @@ export class AdapterDateFnsTZ extends AdapterDateFns {
...
-      minute: 0,
-      second: 0,
-      millisecond: 0,
-    });
-    return tzdate;
+    // Construct TZDate preserving the wall-clock (local) time in the target timezone.
+    // We extract the local date/time components from the source date and recreate them
+    // in the target timezone so that "Jan 5 midnight" stays "Jan 5 midnight" regardless
+    // of which timezone is selected.
+    return new TZDate(
```
tycup, line 106: does this work with strings that encode an instant (and does this matter here?)

naive date strings work fine, but strings like "2025-01-05T00:00:00.000Z" will produce different results depending on the host timezone. hose getters read in the machine's local zone before the value is reinterpreted in the target zone. E.g.: under TZ=Europe/London it resolves to 2025-01-05T00:00:00.000-05:00, while under TZ=America/New_York it resolves to 2025-01-04T19:00:00.000-05:00. Also conflicts with the behaviour the Day.js adapter tests  which now lock in for "…Z" inputs

packages/date-components/src/date-picker/useDatePicker.ts
```diff
@@ -186,9 +180,18 @@ export function useDatePicker<
   } = props;
 
   const previousSelectedDate = useRef<typeof selectedDateProp>();
+  const incompleteRangeAnnouncementKeyRef = useRef<string | null>(null);
   const datePickerRef = useRef<HTMLDivElement>(null);
   const containerRef = useForkRef(ref, datePickerRef);
 
+  const { announce } = useDateSelectionAnnouncer(
```
tycup, line 194: based on code-review: add AriaAnnouncer to overlay as aria-model commit

if the intention is to move date-picker announcements inside the modal dialog,  useDatePicker() calls useDateSelectionAnnouncer() before DatePickerOverlay renders its new AriaAnnouncerProvider. 

I think that this means the data-selection announcements can continue to go to the live region outside the aria-model dialog (i.e. the new provider is rendered too low in th tree to fix the bug i think you are trying to fix)

packages/core/src/aria-announcer/AriaAnnounce.tsx
```diff
@@ -1,24 +1,36 @@
 import { type ComponentType, useEffect } from "react";
-
+import type { AnnounceFnOptions } from "./AriaAnnouncerContext";
 import { useAriaAnnouncer } from "./useAriaAnnouncer";
 
-export interface AriaAnnounceProps {
+export interface AriaAnnounceProps extends AnnounceFnOptions {
```
tycup, line 5: AriaAnnounceProps still extends AnnounceFnOptions, so the public API still advertises duration, but the implementation now only forwards { ariaLive }. This means <AriaAnnounce duration={//} /> silently falls back to the provider default instead of the requested time

package.json
```diff
@@ -16,8 +16,9 @@
     "build": "yarn workspaces foreach --exclude @salt-ds/site -Apt run build && yarn bundle:css",
     "build:ag-grid-theme": "yarn workspace @salt-ds/ag-grid-theme build",
     "build:highcharts-theme": "yarn workspace @salt-ds/highcharts-theme build",
-    "bundle:css": "yarn bundle:core:css && yarn bundle:lab:css && yarn bundle:embla-carousel:css && yarn copy:icon:css && yarn copy:countries:css",
+    "bundle:css": "yarn bundle:core:css && yarn bundle:lab:css && yarn bundle:date-components:css & yarn bundle:embla-carousel:css && yarn copy:icon:css && yarn copy:countries:css",
```
joshwooding, line 19:
```suggestion
    "bundle:css": "yarn bundle:core:css && yarn bundle:lab:css && yarn bundle:date-components:css && yarn bundle:embla-carousel:css && yarn copy:icon:css && yarn copy:countries:css",
```

packages/core/src/aria-announcer/useAriaAnnouncer.ts
```diff
@@ -33,15 +50,28 @@ export const useAriaAnnouncer: useAriaAnnouncerHook = ({
       }
 
       function makeAnnouncement() {
-        if (mountedRef.current) {
-          if (isReactAnnouncerInstalled) {
-            context.announce(announcement);
+        console.log(announcement);
```
joshwooding:
```suggestion
```

mark-tate: Removed on my branch

package.json
mark-tate: added on my branch

.changeset/chalky-white-board.md
```diff
@@ -0,0 +1,16 @@
+---
+"@salt-ds/lab": patch
```
joshwooding:
```suggestion
"@salt-ds/lab": minor
```

packages/date-components/src/calendar/CalendarNavigation.css
```diff
@@ -1,10 +1,13 @@
...
   grid-template-columns: min-content auto min-content;
   padding-bottom: var(--salt-spacing-100);
-  grid-gap: var(--salt-spacing-25);
+  grid-gap: var(--salt-spacing-fixed-200);
   align-items: center;
   box-sizing: border-box;
+  max-width: var(--calendarNavigation-maxWidth, none);
 }
 
 .saltCalendarNavigation-yearDropdown {
```
joshwooding:
```suggestion
.saltDropdown.saltCalendarNavigation-yearDropdown {
```

packages/date-components/src/calendar/CalendarNavigation.css
joshwooding:
```suggestion
  min-width: min-content;
```

mark-tate: done

packages/date-components/src/calendar/CalendarNavigation.css
mark-tate: done

.changeset/chalky-white-board.md
```diff
@@ -0,0 +1,54 @@
+---
+"@salt-ds/lab": minor
+"@salt-ds/date-components": patch
```
joshwooding:
```suggestion
"@salt-ds/date-components": minor
```

packages/date-components/src/calendar/internal/CalendarDay.css
```diff
@@ -254,18 +254,18 @@
 .saltCalendarDay-highlighted {
   position: absolute;
   left: 0;
-  right: 0;
+  right: -1px;
```
joshwooding, line 257: Hardcoded value?

packages/date-components/src/localization-provider/LocalizationProvider.tsx
```diff
@@ -78,8 +82,8 @@ if (process.env.NODE_ENV !== "production") {
 }
 
 export const LocalizationProvider = function LocalizationProvider<
```
joshwooding, line 84: Not sure if this should be in date controls? If so maybe we rename it to be more specific?

joshwooding (approved):

packages/date-components/src/calendar/internal/CalendarDay.css
mark-tate: agreed to have this hardcoded a while ago.
Can add a token if it needs to change but don't see many folks recreating the calendar grid.

tycup (approved):

packages/date-components/src/date-picker/useDatePicker.ts
mark-tate: fixed

PR #5902, opened 2025-12-15
Add support for read-only Switch

### LINKED ISSUE #3840: Add a read-only state to Switch
### Area

UI Components

### The problem

Most of the other selection-based components have a read-only state, e.g. Checkbox, Radio Button. Read-only is commonly used to show a non-editable state in an application.

### The solution

Add readOnly to Switch.

 - Still requires focus

### Alternatives and examples

Creates a11y challenges, other design options exist

### Related
- #3941

wemmyo:
Closes #3840

### DIFF
--- a/.changeset/two-clouds-hug.md
+++ b/.changeset/two-clouds-hug.md
@@ -1,5 +1,5 @@
 ---
-"@salt-ds/core": patch
+"@salt-ds/core": minor
 ---
 
 Added support for read-only `Switch` via `readOnly` prop.
--- a/packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
+++ b/packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
@@ -178,16 +178,19 @@ describe("GIVEN a Switch", () => {
 
   describe("WHEN readOnly is true", () => {
     it("THEN should not toggle if clicked", () => {
-      cy.mount(<Switch readOnly />);
+      const changeSpy = cy.stub().as("changeSpy");
+      cy.mount(<Switch readOnly onChange={changeSpy} />);
       cy.findByRole("switch").should("have.attr", "aria-readonly", "true");
       cy.findByRole("switch").should("not.be.checked");
 
       cy.findByRole("switch").realClick();
       cy.findByRole("switch").should("not.be.checked");
+      cy.get("@changeSpy").should("not.have.been.called");
     });
 
     it("THEN should not toggle when pressing the Space key", () => {
-      cy.mount(<Switch readOnly />);
+      const changeSpy = cy.stub().as("changeSpy");
+      cy.mount(<Switch readOnly onChange={changeSpy} />);
       cy.findByRole("switch").should("have.attr", "aria-readonly", "true");
       cy.findByRole("switch").should("not.be.checked");
 
@@ -195,6 +198,7 @@ describe("GIVEN a Switch", () => {
       cy.findByRole("switch").should("be.focused");
       cy.realPress("Space");
       cy.findByRole("switch").should("not.be.checked");
+      cy.get("@changeSpy").should("not.have.been.called");
     });
   });
 });


packages/core/src/switch/Switch.css
```diff
@@ -121,6 +147,10 @@
   border: 0;
 }
 
+.salt-density-high .saltSwitch-readOnly .saltSwitch-thumb {
```
joshwooding, line 150: Why is this needed?

wemmyo: Right above sets border to zero for some reason
https://github.com/jpmorganchase/salt-ds/pull/5902/changes/BASE..4a562019d7b63cb4f01e646299925842aa7ae743#diff-5b3ab8c0000859aab2212f3251e991de737f6728c8cc74a2d002e54d14920025L115

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
```diff
@@ -175,4 +175,21 @@ describe("GIVEN a Switch", () => {
...
+
+  describe("WHEN readOnly is true", () => {
+    it("THEN should not be toggle if clicked", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("not.be.checked");
+      cy.findByRole("switch").realClick();
+      cy.findByRole("switch").should("not.be.checked");
+    });
+
+    it("THEN should not be toggle when pressing the Space key", () => {
```
mark-tate: NIT - grammar

either 
THEN should not toggle if clicked ..
or
THEN should not be toggled ...
in both descriptions.

packages/core/src/switch/Switch.css
joshwooding: I assume it's to do with sizing, could you check with the designers if what is there meets the expected outcome please

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
wemmyo: Updated, thanks!

packages/core/src/switch/Switch.css
aimee-liou: Figma doesn't have a stroke on the switch thumb in all the densities. Read-only states look correct though.

site/src/examples/switch/Readonly.tsx
```diff
@@ -0,0 +1,9 @@
+import { StackLayout, Switch } from "@salt-ds/core";
+import type { ReactElement } from "react";
+
+export const Readonly = (): ReactElement => (
+  <StackLayout>
+    <Switch readOnly checked label="Read only + checked" />
```
aimee-liou:
```suggestion
    <Switch readOnly checked label="Read-only + Checked" />
```

site/src/examples/switch/Readonly.tsx
```diff
@@ -0,0 +1,9 @@
...
+    <Switch readOnly label="Read only" />
```
aimee-liou:
```suggestion
    <Switch readOnly label="Read-only" />
```

site/docs/components/switch/accessibility.mdx
```diff
@@ -8,6 +8,10 @@ data:
   $ref: ./#/data
 ---
 
+## Best Practice
+
+Ideally we shouln't use readOnly
```
wemmyo: Do we need to talk more about why we shouldn't? We included a note in the example section, wondering if we should move it here? @karlgoldstraw

jake-costa: @wemmyo I was walking Karl through how to edit the docs. I believe he might add more information to this but I'll take a look as well. Also, I did make a slight change to the read only note and tried to align it with what we are using elsewhere (Checkbox for example). Let me know what you think.

wemmyo: Yeah sounds good thanks!

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
 
 <LivePreview componentName="switch" exampleName="DisabledChecked" />
 
+## Read-only
+
+A read-only switch allows the user to select and copy its text, but does not allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
+
+**Note:** We recommend avoiding read-only switches due to the lack of adequate `aria-readonly` support and conflicting user expectations. If you must use a read-only switch, you can add visual or non-visual labels to make its state clear.
+
```
mark-tate, line 56: Copying and pasting a label does not seem to be of that much value but that's how we are selling read-only.

What is more valuable is a focusable control being announced by the screen reader, whilst a disabled control cannot be focused and so will not be seen by the screenreader.

- Lack of adequate support for aria-readonly, we are not clear enough hereon the impact, I guess you browser and screenreader combination.

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
...
+### Best practices
+
+Unlike disabled switches, read-only switches are focusable and shown in the tab order when using a keyboard, allowing users to interact with its text. Use a read-only switch when its text contains information that is valuable to the user.
+
+<LivePreview componentName="switch" exampleName="Readonly" />
+
```
mark-tate, line 62: No guidance currently on the accessibility tab.
Feels like accessibility info is on this page instead ?

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
```diff
@@ -175,4 +175,21 @@ describe("GIVEN a Switch", () => {
...
+    it("THEN should not toggle if clicked", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("not.be.checked");
+      cy.findByRole("switch").realClick();
+      cy.findByRole("switch").should("not.be.checked");
+    });
+
+    it("THEN should not toggle when pressing the Space key", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("not.be.checked");
```
mark-tate, line 195: no test of focus but it's something we call out as a differentiator between disabled?

mark-tate: read-only attribute as well ?

wemmyo: Good point thanks, I've included them in the tests

site/docs/components/switch/examples.mdx
wemmyo: I believe generally we're not trying to sell the readonly switch because of it's limitations including limited support for screen readers, but it's something we have just to be consistent with similar components. Guidance here is similar to what we have for checkbox but @karlgoldstraw and @jake-costa can help expand on that

site/docs/components/switch/examples.mdx
wemmyo: From @jake-costa's comment earlier @karlgoldstraw might be adding more here but we thought the note in the examples section might be beneficial. We can move it to the accessibility tab if that's better

jake-costa: @wemmyo @mark-tate In addition, we have this kind of language on other read only examples (checkbox, radio, etc.)

packages/core/stories/switch/switch.qa.stories.tsx
```diff
@@ -28,6 +28,13 @@ export const AllExamplesGrid: StoryFn<
         disabled
         label="Checked + Disabled"
       />
+      <Switch
+        className={className}
+        checked
+        readOnly
+        label="Checked + Readonly"
```
joshwooding:
```suggestion
        label="Checked + Read-only"
```

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
```diff
@@ -175,4 +175,32 @@ describe("GIVEN a Switch", () => {
...
+      cy.realPress("Space");
+      cy.findByRole("switch").should("not.be.checked");
+    });
+
+    it("THEN should have aria-readonly attribute", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("have.attr", "aria-readonly", "true");
+    });
+
+    it("THEN should be focusable", () => {
```
joshwooding: Can we fold this check into the other tests?

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
```diff
@@ -175,4 +175,32 @@ describe("GIVEN a Switch", () => {
...
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("not.be.checked");
+      cy.realPress("Tab");
+      cy.realPress("Space");
+      cy.findByRole("switch").should("not.be.checked");
+    });
+
+    it("THEN should have aria-readonly attribute", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("have.attr", "aria-readonly", "true");
```
joshwooding, line 194: same here, it feels a bit wasteful to have a test with only one assertion

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
 
 <LivePreview componentName="switch" exampleName="DisabledChecked" />
 
+## Read-only
+
+A read-only switch allows the user to select and copy its text, but does not allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
```
joshwooding:
```suggestion
A read-only switch allows the user to select and copy its text, but doesn't allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
```

site/docs/components/switch/examples.mdx
joshwooding: I think we should give examples of visual or non-visual labels, it sounds very vague at the moment

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
...
+
+**Note:** We recommend avoiding read-only switches due to the lack of adequate `aria-readonly` support and conflicting user expectations. If you must use a read-only switch, you can add visual or non-visual labels to make its state clear.
+
+### Best practices
+
+Unlike disabled switches, read-only switches are focusable and shown in the tab order when using a keyboard, allowing users to interact with its text. Use a read-only switch when its text contains information that is valuable to the user.
```
joshwooding: I would probably just remove this best practice

wemmyo: I agree, I removed the heading but left the paragraph. Does that work?

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
wemmyo: Yeah merged into the others

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
wemmyo: Done!

site/docs/components/switch/examples.mdx
wemmyo: This is similar to what we have for checkbox and radio. I'm not sure what a non-visual label refers to, is it a standalone switch with an `aria-label`? We don't have an example anywhere else

wemmyo: @karlgoldstraw helped clarify and I've included the examples

site/docs/components/switch/examples.mdx
aimee-liou: This is similar to Read-only checkbox - they have it sectioned by "Best practice".

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,16 @@ Use a disabled checked state for switches that have permissions, dependencies or
 
 <LivePreview componentName="switch" exampleName="DisabledChecked" />
 
+## Read-only
+
+A read-only switch allows the user to select and copy its text, but doesn't allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
+
+Unlike disabled switches, read-only switches are focusable and shown in the tab order when using a keyboard, allowing users to interact with its text. Use a read-only switch when its text contains information that is valuable to the user.
+
+**Note:** We recommend avoiding read-only switches due to the lack of adequate `aria-readonly` support and conflicting user expectations. If you must use a read-only switch, you can add visual or non-visual labels to make its state clear.
```
joshwooding: Technically readOnly attribute has terrible support too, should this be expanded to talk more generally about read-only controls?

site/docs/components/switch/examples.mdx
joshwooding: The content on the checkbox page is better. If we want to keep this as a best practice it should be phrased in a similar way.

wemmyo: Updated to be phrased similar to checkbox

packages/core/src/switch/Switch.tsx
```diff
@@ -159,12 +171,15 @@ export const Switch = forwardRef<HTMLLabelElement, SwitchProps>(
...
           <span className={withBaseName("thumb")}>
-            {checked && (
+            {checked && !readOnly && (
               <CheckmarkSolidIcon
                 aria-hidden
                 className={withBaseName("icon")}
               />
             )}
+            {checked && readOnly && (
+              <CheckmarkIcon aria-hidden className={withBaseName("icon")} />
```
mark-tate, line 180: curious... is there an indeterminate state missing as well as read-only ?

mark-tate: No a switch is binary... but if an indeterminate state is needed they should use a checkbox.

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
 
 <LivePreview componentName="switch" exampleName="DisabledChecked" />
 
+## Read-only
+
+A read-only switch allows the user to select and copy its text, but doesn't allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
```
mark-tate: Read-only and disabled seem to differentiate themselves by saying you can copy text.
(the same goes for checkboxes)

Is this by design, I can only safely copy text in a read-only state.
I cannot copy text in a non read-only state without selecting a new state?

It looks like that to copy the text of a non-read only state the user has actually change the state twice.

mark-tate: Actually confusingly, even on a disabled control I can copy the text.
There are 2 ways to select text, click and drag or double click.

Double clicking selects all the text in one go, but if you do that, then it toggles as well.

joshwooding:
```suggestion
A read-only switch allows the user to focus the control, but doesn't allow them to change its state. You can set a switch as read-only via the prop `readOnly`.
```

site/docs/components/switch/examples.mdx
```diff
@@ -48,6 +48,18 @@ Use a disabled checked state for switches that have permissions, dependencies or
...
+
+**Note:** We recommend avoiding read-only switches due to the lack of adequate `aria-readonly` support and conflicting user expectations. If you must use a read-only switch, you should always add an accessible name that makes it obvious the switch is read-only, be that via a visual label or an `aria-label` if there's no space for a visual label.
+
+### Best practices
+
+Read-only switches are navigable using keyboard shortcuts. This means that, unlike disabled switches, users can interact with the text description. Use a read-only switch instead of a disabled switch when the text description contains information that is valuable to the user.
```
joshwooding:
```suggestion
You can set a switch as read-only via the prop `readOnly`.

A read-only switch allows users to view and focus the switch without changing its checked state. This is useful when the switch and its state is considered relevant to the user’s workflow. If the switch is not considered to be relevant it should either be removed or disabled.

**Note**: We recommend avoiding read-only switches due to the lack of adequate screen reader support and conflicting user expectations. If you must use a read-only switch, you should ensure its read-only state is communicated through its label e.g. `aria-label` or `FormFieldLabel`.
```

joshwooding: @mark-tate @karlgoldstraw are you happy with this updated copy?

packages/core/src/switch/Switch.tsx
```diff
@@ -128,13 +137,15 @@ export const Switch = forwardRef<HTMLLabelElement, SwitchProps>(
...
             [withBaseName("checked")]: checked,
+            [withBaseName("readOnly")]: readOnly,
           },
           className,
         )}
         ref={ref}
         {...rest}
       >
         <input
+          aria-readonly={readOnly || undefined}
```
joshwooding, line 148: @jake-costa @karlgoldstraw How necessary is using aria-readonly and the readOnly attribute? If this is to give us the best chance of support we should add a comment. cc @wemmyo

packages/core/src/switch/Switch.tsx
```diff
@@ -104,14 +109,18 @@ export const Switch = forwardRef<HTMLLabelElement, SwitchProps>(
...
+      readOnly: formFieldReadOnly,
+    } = useFormFieldProps();
 
     const disabled = formFieldDisabled || disabledProp;
+    const readOnly = formFieldReadOnly || readOnlyProp;
 
     const handleChange: ChangeEventHandler<HTMLInputElement> = (event) => {
       // Workaround for https://github.com/facebook/react/issues/9023
-      if (event.nativeEvent.defaultPrevented) {
+      if (event.nativeEvent.defaultPrevented || readOnly) {
```
joshwooding, line 123: The native readOnly attribute would handle this right?

wemmyo: It doesn't. Excluding the 'OR readOnly' will trigger the event call

site/docs/components/switch/examples.mdx
mark-tate: great... we removed copy 

is there still a bug with clicking the label of a switch ?

mark-tate: It looks like many component libraries have the same behaviour (including UITK)
Anyone copying labels needs to drag and not double click (although they can in read-only)
I guess until it's part of any a11y standard we should not consider it an issue.

packages/core/src/switch/Switch.tsx
joshwooding: Ah read-only doesn't work on non-editable controls. Since technically it isn't supported

packages/core/src/switch/Switch.tsx
karlgoldstraw: @joshwooding 
According to MDN, the native readonly attribute only applies to text-editable inputs (e.g., text, search, textarea). It’s ignored on checkboxes and radios, so adding it won’t do anything. 

I'd say from this we can remove the native attribute and leave the aria-readOnly in place. 

The advice I read goes onto reiterate that we shouldn't be using readOnly on this element and instead should use disabled. But that's a fight we're already aware of and one for another day. 

cc @jake-costa @wemmyo

karlgoldstraw (approved):

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
```diff
@@ -175,4 +175,26 @@ describe("GIVEN a Switch", () => {
...
   });
+
+  describe("WHEN readOnly is true", () => {
+    it("THEN should not toggle if clicked", () => {
+      cy.mount(<Switch readOnly />);
+      cy.findByRole("switch").should("have.attr", "aria-readonly", "true");
+      cy.findByRole("switch").should("not.be.checked");
+
+      cy.findByRole("switch").realClick();
+      cy.findByRole("switch").should("not.be.checked");
```
joshwooding, line 187: Can we check that the onChange is not called too in these tests please

.changeset/two-clouds-hug.md
```diff
@@ -0,0 +1,11 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/core": minor
```

aimee-liou (approved):

tycup (approved):

packages/core/src/__tests__/__e2e__/switch/Switch.cy.tsx
wemmyo: Done!

PR #5938, opened 2026-01-07
Add Rating component

MinalBhujade:
Added 
- Rating component
- RatingItem component
- Favorite outline icon
- Examples
   -> Basic
   -> Clear Selection
   -> Visual Label
   -> Form Field Support
   -> Custom Icons
   -> Disabled
   -> Read-only
- Site documentation

### DIFF
--- a/.changeset/dirty-bees-sin.md
+++ b/.changeset/dirty-bees-sin.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": minor
+---
+
+Added RatingIcon, RatingSelectedIcon, RatingUnselectingIcon to SemanticIconProvider.


.changeset/seven-cities-show.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/icons": patch
```
joshwooding:
```suggestion
"@salt-ds/icons": minor
```

packages/icons/src/components/FavoriteOutline.tsx
```diff
@@ -0,0 +1,24 @@
+// WARNING: This file was generated by a script. Do not modify it manually
+
+import { forwardRef } from "react";
+
+import { Icon, type IconProps } from "../icon";
+
+export type FavoriteOutlineIconProps = IconProps;
+
+export const FavoriteOutlineIcon = forwardRef<
```
joshwooding: We're going to need an entry into the synonyms for this

packages/lab/src/rating/index.tsx
```diff
@@ -0,0 +1,2 @@
+export * from "./Rating";
+export * from "./RatingItem";
```
joshwooding: If RatingItem isn't used directly by consumers we shouldn't export it

packages/lab/src/rating/Rating.css
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.saltRating-label {
+  color: var(--salt-content-primary-foreground);
+  font-family: var(--salt-text-fontFamily);
+  font-size: var(--salt-text-fontSize);
+  font-style: normal;
+  font-weight: var(--salt-text-fontWeight);
+  line-height: var(--salt-text-lineHeight);
+  width: 150px;
```
joshwooding: This looks wrong, we shouldn't have a fixed width with magic values.

.changeset/plain-groups-tease.md
```diff
@@ -0,0 +1,16 @@
...
+
+- Examples
+  -> Basic
+  -> Clear Selection
+  -> Visual Label
+  -> Form Field Support
+  -> Custom Icons
+  -> Disabled
+  -> Read-only
+- Site documentation
```
joshwooding: Can we swap this out with a codeblock should the basic rating component please

.changeset/seven-cities-show.md
```diff
@@ -0,0 +1,7 @@
...
+---
+
+Added:
+
+- FavoriteOutlineIcon
```
joshwooding: @pseys we need synonyms for the icons

packages/lab/src/rating/Rating.css
```diff
@@ -0,0 +1,30 @@
...
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
```
joshwooding: truncating text like this might be an issue, I'd suggest we remove it and let the label wrap cc @bhoppers2008

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+export interface RatingProps extends FlexLayoutProps<"div"> {
+  /**
+   * current selected rating. If nothing is selected, 0 is assigned.
+   */
+  value?: number;
+  /**
+   * callback function for rating change.
+   */
+  onValueChange?: (itemValue: number) => void;
```
joshwooding: Most of our change callbacks follow the same format, is there a reason why we can't pass the event as the first parameter?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+import ratingCss from "./Rating.css";
+import { RatingItem, type RatingItemProps } from "./RatingItem";
+
+const withBaseName = makePrefixer("saltRating");
+
+export interface RatingProps extends FlexLayoutProps<"div"> {
+  /**
+   * current selected rating. If nothing is selected, 0 is assigned.
+   */
+  value?: number;
```
joshwooding, line 28: This only allows for a controlled API, all of our components should support uncontrolled and controlled behaviour.

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * callback function for rating change.
+   */
+  onValueChange?: (itemValue: number) => void;
+  /**
+   * To pass additional props to button
+   */
+  ButtonProps?: ComponentPropsWithoutRef<"button">;
```
joshwooding: Is this really needed?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+): ReactElement<RatingProps> {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-rating",
+    css: ratingCss,
+    window: targetWindow,
+  });
+  const [reset, setReset] = useState(false);
+  const [currentHoveredStarIndex, setCurrentHoveredStarIndex] = useState(0);
+  const [selected, setSelectedItem] = useState<number>(value ? value : 0);
```
joshwooding: We should be using `useControlled` here

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+  const groupRef = useRef<HTMLDivElement>(null);
+  useImperativeHandle(ref, () => groupRef.current as HTMLDivElement);
```
joshwooding: This should be `useForkRef`

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  const getLabel = (value: number): string => {
+    if (typeof semanticLabels === "function") {
+      return semanticLabels(value, max); // Generate label dynamically
+    }
+    return semanticLabels?.[value - 1] || "No rating selected";
+  };
+
+  const [label, setLabel] = useState(
```
joshwooding: Why do we need the label in state? Can we derive this from the value instead?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+        setLabel("No rating selected"); // Reset label when cleared
+      }
+    } else {
+      setSelectedItem(value);
+      setReset(false);
+      setLabel(getLabel(value)); // Update label based on selected star
+    }
+  };
+
+  useEffect(() => {
```
joshwooding: Using useEffect to call change handlers isn't recommended. They should be called when the event happens.

I would recommend creating a helper function that allows you to set state and call the change handler together

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+    if (onValueChange) {
+      onValueChange(selected);
+    }
+  }, [selected]);
+
+  useEffect(() => {
```
joshwooding: This can be removed when useControlled is used

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+    setSelectedItem(value);
+  }, [value]);
+
+  return (
+    <FlexLayout
```
joshwooding: Can we avoid using layout components within components? They are quite heavy and we prefer to avoid unnecessary dependencies between components. We can just rely on CSS

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+          </div>
+        )}
+      <FlexLayout
+        role="radiogroup"
+        gap={0.5}
+        ref={groupRef}
+        onKeyDown={handleKeyDown}
+        onFocus={handleFocus}
+        className={clsx(withBaseName(), className)}
+        {...restProps}
```
joshwooding: refs, classnames and "rest" should always be applied to the root element

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+      setLabel(getLabel(value));
+    } else if (_event.type === "mouseleave") {
+      setReset(false);
+      setLabel(selected > 0 ? getLabel(selected) : "No rating selected"); // Reset label
+    }
+    setCurrentHoveredStarIndex(value);
+  };
+
+  const handleFocus = () => {
+    if (selected === 0) {
```
joshwooding: I don't think this is the correct behaviour. But I see it's defined in the spec @jake-costa , @karlgoldstraw  @bhoppers2008 @amritadesmet

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+          >
+            {label}
+          </div>
+        )}
+      <FlexLayout
+        role="radiogroup"
+        gap={0.5}
+        ref={groupRef}
+        onKeyDown={handleKeyDown}
+        onFocus={handleFocus}
```
joshwooding: onKeyDown and onFocus will be removed is a user passes them in

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,56 @@
+.saltRatingItem {
+  height: 14px;
```
joshwooding: We should be avoiding hardcoded values, using pixels here means density doesn't work. This should be 
```suggestion
  height: var(--salt-size-icon);
```

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,56 @@
...
+  width: 14px;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: inherit;
+  border: none;
+  cursor: pointer;
```
joshwooding: We have a set of cursor tokens that should be used instead.

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,56 @@
...
+.saltRatingItem-icon.saltRatingItem-icon-empty {
+  color: var(--salt-content-secondary-foreground);
+}
+
+.saltRatingItem-icon.saltRatingItem-icon-hovered {
+  color: var(--salt-accent-background);
+}
+
+.saltRatingItem-icon:focus-visible {
+  border: 2px dashed var(--salt-focused-outline);
```
joshwooding: We have border style tokens that should be used.

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const iconToRender = isActiveState
+        ? outlinedIcon
+        : isCurrentStarHovered || isSelectedStyle
+          ? filledIcon
+          : emptyIcon;
+      return <div className={iconClass}>{iconToRender}</div>;
+    };
+
+    return (
+      <button
```
joshwooding: We should just rely on the semantic radio inputs, that way our keyboard navigation will be handled for us

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+        aria-label={`Rating${itemValue}`}
+        role="radio"
+        tabIndex={
+          disabled
+            ? -1
+            : (readOnly && itemValue === currentRating) ||
+                itemValue === currentRating ||
+                (itemValue === 1 && currentRating === 0)
+              ? 0
+              : -1
```
joshwooding: This nested tertiary is difficult to read, also we can simplify the values. We're already using the disabled prop so we can avoid applying a tab index, also since buttons are focusable we can avoid applying a tabIndex of 0

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+   * Indicates whether the current rating item is in an active state.
+   * An active state typically means that the item is visually highlighted
+   * or styled differently to indicate that it is part of the current selection
+   * or interaction (e.g., hover or focus).
+   */
+  isActiveState?: boolean;
+}
+
+export const RatingItem = forwardRef<HTMLButtonElement, RatingItemProps>(
+  function RatingItem(props, ref?): ReactElement<RatingItemProps> {
```
joshwooding: We can remove the types applied on this line

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
+import { makePrefixer } from "@salt-ds/core";
+import {
+  FavoriteIcon,
+  FavoriteOutlineIcon,
+  FavoriteSolidIcon,
+} from "@salt-ds/icons";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { clsx } from "clsx";
+import React, {
```
joshwooding: Can we avoid using `React` and instead use the exported values directly.

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
+import { makePrefixer } from "@salt-ds/core";
+import {
+  FavoriteIcon,
+  FavoriteOutlineIcon,
+  FavoriteSolidIcon,
+} from "@salt-ds/icons";
```
joshwooding: We should discuss if these need to be semantic icons

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+          elements[prevIndex]?.focus();
+          setSelectedItem(prevIndex + 1); // Save the selection as the rating
+          setLabel(getLabel(prevIndex + 1)); // Update the label
+        }
+        break;
+    }
+  };
+
+  const handleMouseHover = (
+    _event: MouseEvent<HTMLButtonElement>,
```
joshwooding: _ prefixes are usually used to indicate the parameter isn't used but it's used in the function

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+   */
+  disabled?: boolean;
+  /**
+   * Whether to allow clearing the rating when clicking the same rating again.
+   */
+  allowClear?: boolean;
+  /**
+   * Total number of icons displayed.
+   */
+  max?: number;
```
joshwooding, line 51: @bhoppers2008 @amritadesmet do we want to allow this flexibility or should we only support 5/10

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+  /**
+   * Defines the minimum increment value change allowed.
+   */
+  precision?: number;
```
joshwooding: This isn't used right?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+   */
+  readOnly?: boolean;
+  /**
+   * If true, the rating component will be in a disabled state.
+   */
+  disabled?: boolean;
+  /**
+   * Whether to allow clearing the rating when clicking the same rating again.
+   */
+  allowClear?: boolean;
```
joshwooding: We should consider if this needs to be a prop or if it should be a controlled API.

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+  max?: number;
+  /**
+   * Defines the minimum increment value change allowed.
+   */
+  precision?: number;
+  /**
+   * Custom labels for each rating value. Can be an array of strings or a function
+   * that generates labels dynamically (e.g., numerical values like "1/5").
+   */
+  semanticLabels?: string[] | ((value: number, max: number) => string);
```
joshwooding: using an array has some issues:

what happens if you don't put all the values in, what happens if you put more than the number of items

site/docs/components/rating/usage.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+## Using the components
+
+### When to use
+
+- When you have to provide feedback or indicate your level of satisfaction with a product, service, or experience.
+
+### When not to use
+
+- When you have fewer than 3 increments (such as a binary thumbs up/down or like/dislike) do not provide enough nuance for most rating scenarios. In this scenario, use a [`ToggleButtonGroup`](../button).
```
joshwooding: The link is to "button" but the text says toggle button

packages/lab/stories/rating/rating.stories.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <Rating
+        value={4}
+        semanticLabels={(value, max) => `${value}/${max}`}
+        {...args}
+      />
+    </FlexLayout>
+  );
+};
+
+export const FormFieldSupport: StoryFn<typeof Rating> = (args) => {
```
joshwooding: Although there is an example, there is no form field support. Help text and the form field label should be associated programatically with the control

site/docs/components/rating/accessibility.mdx
```diff
@@ -0,0 +1,30 @@
...
+- Tab moves focus into the component and currently selected icon receives focus.
+- If the user hasn't selected rating, Tab moves focus to the first rating in the group.
+
+</KeyboardControl>
+<KeyboardControl keyOrCombos="Space">
+
+- User can press enter or space to deselect/re-select the current rating.
+
+</KeyboardControl>
+<KeyboardControl keyOrCombos="Up arrow / Down arrow / Left arrow / Right arrow">
```
joshwooding: We should split up/left and down/right

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,72 @@
...
+title:
+  $ref: ./#/title
+layout: DetailComponent
+sidebar:
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Basic
```
joshwooding: The site docs will need updating. Currently they follow the design docs and don't follow our other example docs.

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+**Note:** Label position: The label can be positioned top, right, bottom or left of the Rating component. The default label position is ‘right’. The application can define the semantic value of each icon; it could also be a numerical value (e.g., 1/5 - 5/5)
+
+<LivePreview componentName="rating" exampleName="VisualLabel" />
+
+## Form Field Support
+
+If the component is wrapped in a Form Field and you need to provide additional guidance, you can display help text beneath the rating.
+
+**Note:** Rating (and optional label) are vertically centred inside a Form Field control area with --size-base height.
```
joshwooding: This should be handled by rating

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,72 @@
...
+  $ref: ./#/data
+---
+
+## Basic
+
+The default Rating configuration displays five increments, as stars, from which users can select a rating.
+
+<LivePreview componentName="rating" exampleName="Basic" />
+
+## Clear Selection
```
joshwooding: example titles should be sentence case

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,72 @@
...
+In this scenario, use a Toggle Button Group.
+10 increments is the recommended maximum for highly detailed feedback but can be overwhelming, and is unnecessary for most scenarios.
+
+### Best practice
+
+Display the same number of increments across all Rating components within your application to ensure that the value of each increment remains consistent for users.
+
+<LivePreview componentName="rating" exampleName="CustomIcons" />
+
+### Disabled
```
joshwooding: This should be a h2

site/docs/components/rating/index.mdx
```diff
@@ -0,0 +1,15 @@
+---
+title: Rating
+data:
+  description: "`Rating` component allows users to provide feedback relating to your product or experience."
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/blob/main/packages/lab/src/rating"
+  package:
+    name: "@salt-ds/lab"
+  alsoKnownAs: ["Review", "Star rating"]
+  relatedComponents: [{ name: "Toggle button", relationship: "similarTo" }]
```
joshwooding:
```suggestion
```

site/docs/components/rating/examples.mdx
amritadesmet: Yes, I will work on this and provide to @MinalBhujade

.changeset/seven-cities-show.md
joshwooding: The icons were added in a different PR, @MinalBhujade can we remove the icon addition from this PR please

packages/icons/src/components/index.ts
```diff
@@ -163,6 +163,7 @@ export * from "./Exponentiation";
 export * from "./Export";
 export * from "./ExportSolid";
 export * from "./Favorite";
+export * from "./FavoriteOutline";
```
mark-tate: Without a synonym the icon will show up as deprecated

No other icons have outline in the name, as outline is the default.
We now have  3 icons, 2 are outlines are 1 is solid.
Do we think the naming helps here ?

When does someone use FavouriteOutline and when does someone use Favourite
I don't think the naming helps here, will there be confusion and use of FavouriteOutline when they should use Favourite?

amritadesmet: @pseys could you help answer these questions? Since this work was handed over to me halfway through, I’m not familiar with the reasoning behind the decision-making process for the icon naming.

packages/lab/src/rating/Rating.tsx
mark-tate: P already uses 5, 
` {Array.from({ length: 5 }, (_, index) => (`
so this flexibility has crept into this iteration

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,290 @@
...
+      <FlexLayout
+        role="radiogroup"
+        gap={0.5}
+        ref={groupRef}
+        onKeyDown={handleKeyDown}
+        onFocus={handleFocus}
+        className={clsx(withBaseName(), className)}
+        {...restProps}
+      >
+        {Array.from({ length: max }, (_, index) => (
```
mark-tate: Why did we not follow CheckBoxGroup approach, a data driven component that uses Stars seems very opionionated, it does not look scalable with business demands.

If we had a composable approach this could scale to different variations of rating components, as we saw in the kickoff.

By building the name Star into API nomenclature we are already creating challenges.

If we want to accept that there are only ever Stars and that everyone always uses the same scale and UX then you can build one component, but if you then start saying we need additional behaviour then your overloading the API and then creating complexity.

I see other challenges, "allowClear", if you had this prop, then when you add controlled behaviour, you've created a prop that does nothing or creates confusion for the API.

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,197 @@
...
+   */
+  reset?: boolean;
+  /**
+   * defines the current selected rating.
+   */
+  currentRating?: number;
+  /**
+   *  callback function when feedback item is hovered.
+   */
+  onFeedbackItemHover: (
```
mark-tate: You only need to add specifics to naming when there are more than one..
Otherwise this is just onHover.

packages/lab/src/rating/Rating.tsx
amritadesmet: I'm not sure I understand what flexibility you're referring to - do you mean supporting up to 10 stars? Let me know

.changeset/seven-cities-show.md
MinalBhujade: Reverted newly added icon changes.

packages/icons/src/components/FavoriteOutline.tsx
MinalBhujade: Reverted newly added icon (FavoriteOutlineIcon) changes

packages/lab/src/rating/index.tsx
MinalBhujade: Removed RatingItem from exporting.

.changeset/seven-cities-show.md
MinalBhujade: Reverted newly added icon (FavoriteOutlineIcon) changes

packages/lab/src/rating/Rating.tsx
MinalBhujade: Removed.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated onKeyDown and onFocus.

packages/lab/src/rating/RatingItem.css
MinalBhujade: Updated value with token

packages/lab/src/rating/RatingItem.css
MinalBhujade: Updated value with token.

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: Removed.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Removed un-used props.

site/docs/components/rating/examples.mdx
MinalBhujade: Updated for now. Once site documentation design is ready, this will get updated again.

site/docs/components/rating/index.mdx
MinalBhujade: Removed

packages/icons/src/components/index.ts
MinalBhujade: As new FavoriteStrongIcon is added, I have reverted changes related to newly added FavoriteOutlineIcon.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated with event parameter.

site/docs/components/rating/examples.mdx
MinalBhujade: Once site documentation design is ready, this will get updated.

site/docs/components/rating/examples.mdx
MinalBhujade: Once site documentation design is ready, this will get updated.

packages/lab/src/rating/Rating.css
MinalBhujade: Removed truncation part.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated.

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: Removed.

site/docs/components/rating/usage.mdx
MinalBhujade: Updated to toggle-button.

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+export interface RatingProps extends FlexLayoutProps<"div"> {
+  /**
+   * current selected rating. If nothing is selected, 0 is assigned.
+   */
+  value?: number;
+  /**
+   * Callback function for rating change.
+   * The first parameter is the event, and the second is the selected rating value.
+   */
+  onValueChange?: (
```
mark-tate: make this onChange

packages/lab/src/rating/Rating.tsx
mark-tate: consider whether there is a better name that is consistent.

packages/lab/src/rating/Rating.tsx
mark-tate: keep

set a default of 5

packages/lab/src/rating/Rating.tsx
mark-tate: getLabel

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+  max?: number;
+  /**
+   * Custom labels for each rating value. Can be an array of strings or a function
+   * that generates labels dynamically (e.g., numerical values like "1/5").
+   */
+  semanticLabels?: string[] | ((value: number, max: number) => string);
+  /**
+   * Whether to display the label.
+   */
+  showLabel?: boolean;
```
mark-tate: ask design whether by default we should always have a label..
if so... this is hideLabel else this is fine

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Custom character for the rating icons.
+   */
+  character?: React.ReactNode | ((props: RatingItemProps) => React.ReactNode);
```
mark-tate: we could use semanticIconProvider
favouriteIconEmpty
favouriteIconSolid
favouriteIconStrong

we will add render later if needed

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+  /**
+   * Position of the label relative to the rating component.
+   * Can be "top", "right", "bottom", or "left".
+   */
+  labelPosition?: "top" | "right" | "bottom" | "left";
```
mark-tate: keep

packages/lab/src/rating/Rating.tsx
amritadesmet: Based on Ben's specs, the label is optional, not on by default. Default is without label

packages/lab/src/rating/Rating.tsx
mark-tate: Should have a default.

add "hidden"

default, depends if label is always shown by default

decide on the best name

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+  ref?,
+): ReactElement<RatingProps> {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-rating",
+    css: ratingCss,
+    window: targetWindow,
+  });
+  const [reset, setReset] = useState(false);
+  const [currentHoveredStarIndex, setCurrentHoveredStarIndex] = useState(0);
```
mark-tate: remove star naming

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+          <div
+            className={clsx(
+              withBaseName("label"),
+              withBaseName(`label-${labelPosition}`),
+            )}
+          >
+            {label}
+          </div>
+        )}
+      <FlexLayout
```
mark-tate: No need for FlexLayout.
try using native radio elements for stars and see if that means we simplify the keyboard navigation

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+        onKeyDown={(event) => {
+          handleKeyDown(event);
+          restProps?.onKeyDown?.(event);
+        }}
+        onFocus={(event) => {
+          handleFocus();
+          restProps?.onFocus?.(event);
+        }}
+        className={clsx(withBaseName(), className)}
+        {...restProps}
```
mark-tate: destructure event handlers

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+   */
+  isCurrentStarHovered?: boolean;
+  /**
+   * To specify if the item is selected.
+   */
+  isSelectedStyle?: boolean;
+  /**
+   * resets the current selection.
+   */
+  reset?: boolean;
```
mark-tate: remove this

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+export interface RatingItemProps extends ComponentPropsWithoutRef<"button"> {
+  /**
+   * specifies the value of the feedback item.
+   */
+  itemValue: number;
+  /**
+   * To define if the current star is being hovered.
+   */
+  isCurrentStarHovered?: boolean;
```
mark-tate: simplify/align naming

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * To specify if the item is selected.
+   */
+  isSelectedStyle?: boolean;
```
mark-tate: naming

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+} from "react";
+import ratingItemCss from "./RatingItem.css";
+
+const withBaseName = makePrefixer("saltRatingItem");
+
+export interface RatingItemProps extends ComponentPropsWithoutRef<"button"> {
+  /**
+   * specifies the value of the feedback item.
+   */
+  itemValue: number;
```
mark-tate: value

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+        ? outlinedIcon
+        : isCurrentStarHovered || isSelectedStyle
+          ? filledIcon
+          : emptyIcon;
+      return <div className={iconClass}>{iconToRender}</div>;
+    };
+
+    return (
+      <button
+        data-testid={`salt-${itemValue}`}
```
mark-tate: no need for this... use RTL a11y attributes

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+        aria-label={`Rating${itemValue}`}
+        role="radio"
+        tabIndex={
+          disabled
+            ? -1
+            : (readOnly && itemValue === currentRating) ||
+                itemValue === currentRating ||
+                (itemValue === 1 && currentRating === 0)
+              ? 0
+              : -1
```
mark-tate: we do not need to pass in 2 values from parent to set one attribute on child

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+      window: targetWindow,
+    });
+
+    const handleMouseEnter = (event: MouseEvent<HTMLButtonElement>) => {
+      if (readOnly || disabled) return; // Prevent hover interaction in read-only mode
+      onFeedbackItemHover(event, itemValue);
+    };
+    const handleMouseLeave = (event: MouseEvent<HTMLButtonElement>) => {
+      if (readOnly || disabled) return; // Prevent hover interaction in read-only mode
+      onFeedbackItemHover(event, 0);
```
mark-tate: we maybe not need to pass back a value that parent already knows.

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+              currentHoveredStarIndex > 0 &&
+              index + 1 <= currentHoveredStarIndex
+            }
+            isSelectedStyle={index + 1 > 0 && index + 1 <= selected}
+            isActiveState={
+              currentHoveredStarIndex > 0 && // A star is being hovered
+              index + 1 > currentHoveredStarIndex && // This star is beyond the hovered star
+              index + 1 <= selected
+            }
+            onFeedbackItemHover={handleMouseHover}
```
mark-tate: closure would reduce the props to Rating Item and complexity

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,284 @@
...
+        className={clsx(withBaseName(), className)}
+        {...restProps}
+      >
+        {Array.from({ length: max }, (_, index) => (
+          <RatingItem
+            reset={reset}
+            currentRating={selected}
+            isCurrentStarHovered={
+              currentHoveredStarIndex > 0 &&
+              index + 1 <= currentHoveredStarIndex
```
mark-tate: cleanup index

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,198 @@
...
+   */
+  filledIcon?: React.ReactNode;
+  /**
+   * Custom icon for the empty state.
+   */
+  emptyIcon?: React.ReactNode;
+  /**
+   * Defines the minimum increment value change allowed.
+   */
+  precision?: number;
```
mark-tate: remove precision

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,56 @@
...
+  padding: 0;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: inherit;
+  border: none;
+  cursor: var(--salt-cursor-hover);
+}
+.saltRatingItem-icon {
+  --saltIcon-size: 14px;
```
mark-tate: use the salt token

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,56 @@
...
+
+.saltRatingItem-disabled {
+  cursor: var(--salt-cursor-disabled);
+  opacity: 0.4;
+}
+
+.saltRatingItem .saltRatingItem-icon {
+  transition:
+    color 0.2s ease,
+    transform 0.2s ease;
```
joshwooding: @amritadesmet can you help @MinalBhujade check these please. Also we don't change "transform" so we can remove it

packages/lab/src/rating/RatingItem.css
amritadesmet: Use 'size-icon' which equals to 14px

.changeset/plain-groups-tease.md
MinalBhujade: updated.

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

site/docs/components/rating/accessibility.mdx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: As default is without label this is fine.

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: removed

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: removed

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: done

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: removed

packages/lab/src/rating/RatingItem.css
MinalBhujade: updated

MinalBhujade: updated

packages/lab/src/rating/Rating.css
MinalBhujade: removed fixed width.

site/docs/components/rating/examples.mdx
MinalBhujade: done

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: Updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.css
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.css
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/stories/rating/rating.stories.tsx
MinalBhujade: updated

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/lab/src/rating/Rating.tsx
MinalBhujade: updated

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
```diff
@@ -89,6 +95,9 @@ const defaultIconMap: SemanticIconMap = {
   CompletedIcon: SuccessCircleSolidIcon,
   LockedIcon: LockedIcon,
   InProgressIcon: ProgressInprogressIcon,
+  FavoriteEmptyIcon: FavoriteIcon,
+  FavoriteSolidIcon: FavoriteSolidIcon,
+  FavoriteStrongIcon: FavoriteStrongIcon,
```
joshwooding: @pseys are you happy with these

packages/lab/src/rating/Rating.css
```diff
@@ -0,0 +1,53 @@
...
+  flex-direction: column;
+}
+
+.saltRating-wrapper-left,
+.saltRating-wrapper-right {
+  flex-direction: row;
+}
+
+.saltRating-container {
+  gap: calc(var(--salt-spacing-100) * 0.5);
```
joshwooding:
```suggestion
  gap: var(--salt-spacing-50);
```

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,327 @@
...
+   * Default rating value for uncontrolled mode.
+   * @default 0
+   */
+  defaultValue?: number;
+  /**
+   * Callback function for rating change.
+   * The first parameter is the event, and the second is the selected rating value.
+   */
+  onChange?: (
+    event: React.MouseEvent<HTMLButtonElement>,
```
joshwooding: This should probably be a synthetic event since it could be a keyboard event too?

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
amritadesmet: @pseys

.changeset/plain-groups-tease.md
```diff
@@ -0,0 +1,22 @@
+---
+"@salt-ds/lab": minor
+---
+
+New `Rating` and `RatingItem` components have been added to allow users to provide feedback relating to your product or experience.
```
joshwooding:
```suggestion
`Rating` has been added to allow users to provide feedback relating to your product or experience.
```

packages/lab/src/rating/Rating.css
```diff
@@ -0,0 +1,48 @@
...
+  justify-content: flex-start;
+  align-items: stretch;
+  box-sizing: border-box;
+}
+
+.saltRating-label {
+  color: var(--salt-content-primary-foreground);
+  font-family: var(--salt-text-fontFamily);
+  font-size: var(--salt-text-fontSize);
+  font-style: normal;
```
joshwooding:
```suggestion
```
We shouldn't need to define this

packages/lab/src/rating/Rating.css
```diff
@@ -0,0 +1,48 @@
...
+  font-weight: var(--salt-text-fontWeight);
+  line-height: var(--salt-text-lineHeight);
+  min-width: auto;
```
joshwooding:
```suggestion
```
or this

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+   */
+  max?: number;
+  /**
+   * Function that generates labels dynamically based on the value and max.
+   */
+  getLabel?: (value: number, max: number) => string;
+  /**
+   * Custom character for the rating icons.
+   */
+  character?: React.ReactNode | ((props: RatingItemProps) => React.ReactNode);
```
joshwooding: We had agreed to remove this for now


```suggestion
```

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+   */
+  strongIcon?: React.ReactNode;
+  /**
+   * Custom icon for the filled state.
+   */
+  filledIcon?: React.ReactNode;
+  /**
+   * Custom icon for the empty state.
+   */
+  emptyIcon?: React.ReactNode;
```
joshwooding:
```suggestion
```

Same here

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * The name to be set on each radio button within the group. If not set, then one will be generated for you.
+   */
+  name?: string;
+  /**
+   * Additional props to be applied to the label element.
+   */
+  labelProps?: React.HTMLAttributes<HTMLDivElement>;
```
joshwooding: What is this for?

MinalBhujade: This is to add minWidth for label if the placement is left. Please check below example - 
<Rating
          labelPlacement="left"
          getLabel={(value) => labels[value - 1] || "No rating"}
          labelProps={{ style: { minWidth: "15ch" } }}
        />

packages/lab/src/rating/Rating.tsx
MinalBhujade: done

packages/lab/src/rating/Rating.tsx
MinalBhujade: done

packages/lab/src/rating/Rating.css
MinalBhujade: done

packages/lab/src/rating/Rating.css
MinalBhujade: removed

.changeset/plain-groups-tease.md
MinalBhujade: updated

site/src/examples/rating/Basic.tsx
```diff
@@ -0,0 +1,10 @@
+import { FormField, FormFieldLabel } from "@salt-ds/core";
+import { Rating } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+export const Basic = (): ReactElement => (
+  <FormField>
+    <FormFieldLabel color="secondary">Rate your experience</FormFieldLabel>
```
joshwooding:
```suggestion
    <FormFieldLabel>Rate your experience</FormFieldLabel>
```

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+    labelPlacement = "right",
+    labelProps,
+    onKeyDown,
+    onFocus,
+    "aria-labelledby": ariaLabelledBy,
+    "aria-describedby": ariaDescribedBy,
+    ...restProps
+  },
+  ref?,
+): ReactElement<RatingProps> {
```
joshwooding:
```suggestion
  ref,
) {
```

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+  /**
+   * Custom icon for the filled state.
+   */
+  filledIcon?: React.ReactNode;
+  /**
+   * Custom icon for the empty state.
+   */
+  emptyIcon?: React.ReactNode;
+  /**
+   * The name to be set on each radio button within the group. If not set, then one will be generated for you.
```
joshwooding, line 67: Do we need to generate a name automatically?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,331 @@
...
+    value > 0
+      ? getLabel?.(value, max) || `${value} out of ${max}`
+      : "No rating selected";
+
+  const updateRating = (newValue: number, event: any) => {
+    setSelected(newValue);
+    onChange?.(event, newValue);
+  };
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLDivElement>) => {
```
joshwooding: We should avoid referencing the global "React" and we should prefer destructuring

packages/lab/src/rating/Rating.tsx
MinalBhujade: I have made name prop as optional. If there is no name prop then radio buttons in the same group must share the same name attribute to function correctly.

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
joshwooding: @pseys bump

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,61 @@
+.saltRatingItem {
+  height: var(--salt-size-icon);
+  width: var(--salt-size-icon);
+  display: flex;
+  position: relative;
+  align-items: center;
+  justify-content: center;
+}
+
+.saltRatingItem:not(.saltRatingItem-disabled):not(.saltRatingItem-readOnly) {
```
joshwooding: We typically avoid using :not as it adds a lot of specificity. Can you rewrite this to follow other components where extra selectors are added instead please.

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,61 @@
...
+}
+
+.saltRatingItem-selected .saltRatingItem-icon,
+.saltRatingItem-hovered .saltRatingItem-icon,
+.saltRatingItem-active .saltRatingItem-icon {
+  color: var(--salt-sentiment-accent-dataviz);
+}
+
+.saltRatingItem.saltRatingItem-active .saltRatingItem-icon {
+  transition-duration: var(--salt-duration-perceptible);
```
joshwooding: Has this transition been signed off @amritadesmet

packages/lab/src/rating/RatingItem.css
```diff
@@ -0,0 +1,61 @@
...
+  pointer-events: none;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  color: var(--salt-sentiment-neutral-dataviz);
+  transition: color var(--salt-duration-instant) ease-in-out;
+}
+
+.saltRatingItem-icon .saltIcon {
+  --saltIcon-size: var(--salt-size-icon);
```
joshwooding, line 47: Do we need both this and line 25?

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,313 @@
...
+    <div
+      ref={ref}
+      className={clsx(
+        withBaseName("wrapper"),
+        withBaseName(`wrapper-${labelPlacement}`),
+        className,
+      )}
+      {...restProps}
+    >
+      <div
```
joshwooding: Do we actually need a live region @jake-costa  @karlgoldstraw

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,313 @@
...
+      </div>
+      {isTopLeft && displayLabel}
+      <div
+        role="radiogroup"
+        className={withBaseName("container")}
+        ref={radioGroupRef}
+        onFocus={handleFocus}
+        onKeyDown={handleKeyDown}
+        aria-labelledby={ariaLabelledByValue}
+        aria-label={!ariaLabelledByValue ? "Rating" : undefined}
```
joshwooding: We typically don't add default aria-labels. Can we remove this please

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+          ref={ref}
+          type="radio"
+          name={name}
+          value={value}
+          checked={currentRating === value}
+          onClick={onClick}
+          disabled={disabled}
+          readOnly={readOnly}
+          className={withBaseName("input")}
+          aria-label={`${value} star${value !== 1 ? "s" : ""}, ${ariaLabel ?? ""}`}
```
joshwooding: The aria-labels look odd to me, @jake-costa @karlgoldstraw can you review please

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+          tabIndex={isFocusable ? 0 : -1}
```
joshwooding: We shouldn't need to manage the tabIndex. @MinalBhujade have you tried without these?

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+      return emptyIcon;
+    }, [
+      props,
+      isHovered,
+      isSelected,
+      isActive,
+      filledIcon,
+      strongIcon,
+      emptyIcon,
+    ]);
```
joshwooding: This useMemo is unnecessary. It also gets re-calculated when props changes which is pretty much every render

packages/lab/src/rating/RatingItem.tsx
```diff
@@ -0,0 +1,171 @@
...
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-rating-item",
+      css: ratingItemCss,
+      window: targetWindow,
+    });
+
+    const isInteractive = !readOnly && !disabled;
+
+    const handleHover = useCallback(
```
joshwooding: This useCallback isn't necessary. It's an event handler being passed directly to the DOM.

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,64 @@
...
```
joshwooding:
```suggestion
## Disabled
```

site/docs/components/rating/usage.mdx
```diff
@@ -0,0 +1,38 @@
...
+
+## Using the components
+
+### When to use
+
+- When you have to provide feedback or indicate your level of satisfaction with a product, service, or experience.
+
+### When not to use
+
+- When you have fewer than 3 increments (such as a binary thumbs up/down or like/dislike) do not provide enough nuance for most rating scenarios. In this scenario, use a [`ToggleButtonGroup`](../toggle-button).
```
joshwooding:
```suggestion
- When you have fewer than 3 increments (such as a binary thumbs up/down or like/dislike) don't provide enough nuance for most rating scenarios. In this scenario, use a [`ToggleButtonGroup`](../toggle-button).
```

site/src/examples/rating/VisualLabel.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <Rating defaultValue={4} getLabel={(value, max) => `${value}/${max}`} />
+    </FormField>
+    <FormField>
+      <FormFieldLabel>Select rating</FormFieldLabel>
+      <Rating
+        labelPlacement="left"
+        getLabel={(value) => labels[value - 1] || "No rating"}
+        labelProps={{ style: { minWidth: "15ch" } }}
+      />
+      <Text style={{ maxWidth: "450px" }}>
```
joshwooding: This should probably be ch rather than px?

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+You can display up to 10 increments, reducing or increasing the number to suit your particular use case. To configure the number of increments, use the max prop.
+
+**Note:** There should be at least 3 increments—this allows users to express a basic range of sentiment -
+Poor/ Unsatisfactory
+Average/Neutral
+Good/Satisfactory
+Fewer than 3 increments (such as a binary thumbs up/down or like/dislike) do not provide enough nuance for most rating scenarios.
+In this scenario, use a Toggle Button Group.
+10 increments is the recommended maximum for highly detailed feedback but can be overwhelming, and is unnecessary for most scenarios.
```
joshwooding:
```suggestion
```
This is covered in the usage docs.

site/docs/components/rating/examples.mdx
```diff
@@ -0,0 +1,64 @@
...
+
+You can display Rating with a aligned label. This alignment is particularly useful when your application’s vertical real estate is limited.
+
+**Note:** Label position: The label can be positioned top, right, bottom or left of the Rating component. The default label position is ‘right’. The application can define the semantic value of each icon; it could also be a numerical value (e.g., 1/5 - 5/5)
+
+<LivePreview componentName="rating" exampleName="VisualLabel" />
+
+## Form field support
+
+If the component is wrapped in a Form Field and you need to provide additional guidance, you can display help text beneath the rating.
```
joshwooding:
```suggestion
If the component is wrapped in a [`FormField`](/salt/components/form-field) and you need to provide additional guidance, you can display help text beneath the rating.
```

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,313 @@
...
+  /**
+   * Default rating value for uncontrolled mode.
+   * @default 0
+   */
+  defaultValue?: number;
+  /**
+   * Callback function for rating change.
+   * The first parameter is the event, and the second is the selected rating value.
+   */
+  onChange?: (event: SyntheticEvent, itemValue: number) => void;
```
joshwooding:
```suggestion
  onChange?: (event: SyntheticEvent, value: number) => void;
```

packages/lab/src/rating/Rating.tsx
joshwooding: Typically we'd avoid adding props for something like this. Users could either target the element using a classname on the root or they could add a minWidth on the root

site/src/examples/rating/WithFormField.tsx
```diff
@@ -0,0 +1,15 @@
+import { FormField, FormFieldHelperText, FormFieldLabel } from "@salt-ds/core";
+import { Rating } from "@salt-ds/lab";
+import type { ReactElement } from "react";
+
+const labels = ["Poor", "Fair", "Good", "Very Good", "Excellent"];
+export const WithFormField = (): ReactElement => (
+  <FormField labelPlacement="top">
+    <FormFieldLabel color="secondary">Overall Experience</FormFieldLabel>
```
joshwooding:
```suggestion
    <FormFieldLabel>Overall Experience</FormFieldLabel>
```

packages/lab/src/rating/RatingItem.tsx
MinalBhujade: In read-only mode, the currently selected rating item must be keyboard focusable to ensure screen reader users can discover and navigate to the rating value. This is the reason added tabIndex. If I remove it will not focus selected rating on tab in read-only mode.

packages/lab/src/rating/Rating.tsx
MinalBhujade: This is the scenario when there is no label present or rating is not inside any form field.

packages/lab/src/rating/RatingItem.css
MinalBhujade: This transition is mentioned in characteristic spec figma when user is hovered on lower rating if higher is selected and higher rating (from selected state to state where star is strong outlined) should have perceptible transition. @amritadesmet, please confirm if this is correct.

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
pseys: As this is for the semanticIconProvider I'd expect this to be more specific to the context, something like:

- RatingIcon
- RatingSelectedIcon
- RatingUnselectingIcon

@joshwooding

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
```diff
@@ -95,9 +95,9 @@ const defaultIconMap: SemanticIconMap = {
   CompletedIcon: SuccessCircleSolidIcon,
   LockedIcon: LockedIcon,
   InProgressIcon: ProgressInprogressIcon,
-  ErrorStatusAdornment: ErrorAdornmentIcon,
-  WarningStatusAdornment: WarningAdornmentIcon,
-  SuccessStatusAdornment: SuccessAdornmentIcon,
+  FavoriteEmptyIcon: FavoriteIcon,
```
pseys: Can we change this to `RatingIcon`

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
```diff
@@ -95,9 +95,9 @@ const defaultIconMap: SemanticIconMap = {
...
+  FavoriteSolidIcon: FavoriteSolidIcon,
```
pseys: Can we change this to `RatingSelectedIcon`

packages/core/src/semantic-icon-provider/SemanticIconProvider.tsx
```diff
@@ -95,9 +95,9 @@ const defaultIconMap: SemanticIconMap = {
...
+  FavoriteStrongIcon: FavoriteStrongIcon,
```
pseys: Can we change this to `RatingUnselectingIcon`

packages/lab/src/rating/RatingItem.css
amritadesmet: I've asked @MinalBhujade to see a preview of this with duration-instant instead as perceptible in practice looks too slow. Result to be confirmed

MinalBhujade: As per discussion, removed transition.

packages/lab/src/rating/RatingItem.tsx
karlgoldstraw: @joshwooding can you define what you mean by odd?

packages/lab/src/rating/Rating.tsx
karlgoldstraw: I believe that the live region enhanced my experience when using a screen reader, telling me when the star were updated in a human-readable way.

joshwooding: But this doesn't follow the APG implementation? Do we actually have good reasons to deviate?

packages/lab/src/rating/Rating.tsx
joshwooding: I know, but we don't do this anywhere else. Why is rating different?

packages/lab/src/rating/RatingItem.tsx
joshwooding: The contents of the aria-label don't align to what I'd expect. 

image

packages/lab/src/rating/Rating.tsx
jake-costa: @MinalBhujade @joshwooding @karlgoldstraw 
We do not need a live region if we no longer have a clear example. If we add a live region to announce the selected item when the screen reader will already announce the selected item then we would add a redundant overly verbose announcement which is not ideal and as Josh mentioned does not align with the APG implementation.

Please remove the live region

packages/lab/src/rating/Rating.tsx
```diff
@@ -0,0 +1,305 @@
...
+        } else if (selected === 0) {
+          newValue = 1;
+          targetIndex = 0;
+        }
+        break;
+
+      default:
+        onKeyDown?.(event);
+        return;
+    }
```
joshwooding: This should be handled by the native inputs right?

joshwooding: I think it was missed in a previous comment

packages/lab/src/rating/RatingItem.tsx
jake-costa: @MinalBhujade @joshwooding @karlgoldstraw 
For these labels we need to align with the APG and just indicate "one star, two star, etc.". We should not include the "selected" or "not selected" information in the aria label.

packages/lab/src/rating/Rating.tsx
jake-costa: @MinalBhujade Teams are required to add an accessible name, we should not be adding a failback accessible name.

packages/lab/src/rating/Rating.tsx
MinalBhujade: Default behaviour is to wrap the focus, but as per figma interaction it is mentioned we should not wrap it -
Selection limits

Given the visual focus is on the first (or last star) in the component When the user presses the left/up (or right/down) arrow key Then nothing happens—focus does not wrap to the other end of the component

packages/lab/src/rating/RatingItem.tsx
joshwooding: The tricky part is that we are trying to avoid "star" in the label, I might have a suggestion here

site/docs/components/rating/accessibility.mdx
```diff
@@ -0,0 +1,35 @@
...
+<KeyboardControls>
+<KeyboardControl keyOrCombos="Tab">
+
+- Tab moves focus into the component and currently selected icon receives focus.
+- If the user hasn't selected rating, Tab moves focus to the first rating in the group.
+
+</KeyboardControl>
+<KeyboardControl keyOrCombos="Space">
+
+- User can press enter or space to deselect/re-select the current rating.
```
jake-costa: This needs to be removed, cannot deselect using space or enter

packages/lab/stories/rating/rating.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+};
+
+export const VisualLabel: StoryFn<typeof Rating> = () => {
+  const labels = ["Poor", "Fair", "Good", "Very Good", "Excellent"];
+  return (
+    <FlexLayout direction="column" gap={3}>
+      <FormField>
+        <FormFieldLabel>Product quality</FormFieldLabel>
+        <Rating getLabel={(value) => labels[value - 1] || "No rating"} />
```
jake-costa: The options in the radio button group should not be labeled using getLabel={(value) => labels[value - 1] || "No rating"} and  getLabel={(value, max) => ${value}/${max}}, these should be replaced with just the label, otherwise it creates verbose, redundant announcements. The semantic radio group itself will already convey position and handle when an option is not selected.

packages/lab/src/rating/Rating.tsx
jake-costa: We should allow the focus to wrap so that if it is on the first star and the left arrow is pressed it goes to the last star and vice versa. This is the expected keyboard interaction for a semantic radio button group. In addition to allowing up and down arrows

packages/lab/src/rating/Rating.tsx
MinalBhujade: removed as per APG

PR #5953, opened 2026-01-09
Highcharts | Add a stroke around Bubble Chart label for contrast

tycup:
Accessibility feedback

- Change the label color to "white" (across both light/dark) to be consistent with Waterfall Chart
- When fill-patterns are applied, add a "black" outline around the text to ensure contrast is met against opacity background.
- We don't do this anywhere else unless opacity causes a contrast fail

### DIFF
--- a/site/docs/foundations/density.mdx
+++ b/site/docs/foundations/density.mdx
@@ -5,7 +5,7 @@ sidebar:
   priority: 8
 ---
 
-Density relates to the amount of content that can fit on a screen based on the size and spacing of components. When building a product with Salt, select the density that best suits your use case. All components, spacing, and typographic elements will align and remain proportionate regardless of the density you choose. Density can be changed at a page or region level.
+Density relates to the amount of content that can fit on a screen based on the size and spacing of components. When building a product with Salt, select the density that best suits your use case. All components, spacing, and typographic elements will align and remain proportionate regardless of the density you choose.
 
 Read about implementing density in code on the [SaltProvider](/salt/components/salt-provider/examples) documentation.
 


.changeset/mighty-trains-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/highcharts-theme": minor
```
joshwooding:
```suggestion
"@salt-ds/highcharts-theme": patch
```

.changeset/mighty-trains-jam.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Add stroke to Bubble Chart data labels when pattern fills are applied via `.salt-fill-patterns`
```
joshwooding: Can we mention it's due to contrast requirements please

tycup: Updated

packages/highcharts-theme/css/salt-highcharts-theme.css
```diff
@@ -174,10 +174,18 @@
...
 .highcharts-theme-salt .highcharts-bubble-series text.highcharts-data-label {
-  fill: var(--salt-content-primary-foreground);
+  fill: var(--salt-content-bold-foreground);
   font-weight: var(--salt-text-label-fontWeight-strong);
 }
 
+.highcharts-theme-salt.salt-fill-patterns .highcharts-bubble-series .highcharts-data-label text,
+.highcharts-theme-salt.salt-fill-patterns .highcharts-bubble-series text.highcharts-data-label {
+  /* stylelint-disable-next-line salt/correct-theme-token-usage -- permissable due to lack of better token */
+  stroke: var(--salt-color-black);
```
wemmyo, line 184: Will this still meet contrast requirement without the black stroke?

joshwooding (approved):

packages/highcharts-theme/css/salt-highcharts-theme.css
amritadesmet: You're right, it doesn't meet color contrast when it's only white. @tycup and I are discussing possible solutions for this which can either be addressed in this PR or when merging to core.

Issue #5910, opened 2025-12-16
Date Picker minDate/maxDate selection issue


mark-tate:
Steps to reproduce

1. Using a min/max date date range example
2. Select a valid date within range
3. Change the year to go to date prior to the min date

ACTUAL Result

The visible month jumps to the start of the current year before the selection

EXPECTED Result

The visible month jumps to the minDate

mark-tate:
fixed in release @salt-ds/date-components 0.1.0

PR #5976, opened 2026-01-17
Fix DatePicker calendar visible month when user enters out of range date

origami-z:
cc #5910


mark-tate:
This change was absorbed into our last release.

PR #5978, opened 2026-01-19
Improve Table reflow and add accessible name to scroll container

wemmyo:
Accessibility changes to table to improve reflow and add scrollable region

- Added a `TableContainer` to handle scrollable region and overflow  
- Replaced `root table` with `propTable` class name on site CSS as it was interfering with table styling
- Added accessible name using caption to tables on site
- Added Accessibility docs

### DIFF
--- a/site/src/layouts/DetailComponent/DetailComponent.tsx
+++ b/site/src/layouts/DetailComponent/DetailComponent.tsx
@@ -103,6 +103,7 @@ export const DetailComponent = ({ children }: LayoutProps) => {
   const isOverview = route?.endsWith("components/index");
 
   const currentTab = tabs.find(({ name }) => route?.includes(name));
+  const currentTabName = currentTab?.name ?? tabs[0].name;
 
   useEffect(() => {
     // Default to first tab, "Examples"
@@ -153,7 +154,7 @@ export const DetailComponent = ({ children }: LayoutProps) => {
         {!isOverview && (
           <TabsNext
             className={styles.content}
-            value={currentTab?.name ?? tabs[0].name}
+            value={currentTabName}
             onChange={handleTabChange}
           >
             <TabBar divider>
@@ -167,7 +168,7 @@ export const DetailComponent = ({ children }: LayoutProps) => {
             </TabBar>
             {tabs.map(({ name }) => (
               <TabNextPanel className={styles.tabPanel} key={name} value={name}>
-                {children}
+                {name === currentTabName ? children : null}
               </TabNextPanel>
             ))}
           </TabsNext>


site/src/examples/table/Dividers.tsx
```diff
@@ -68,57 +72,60 @@ export const Dividers = (): ReactElement => {
...
-                })}
-              </TR>
-            );
-          })}
-        </TBody>
-        <TFoot divider={footer}>
-          <TR>
-            {Array.from({ length: 3 }, (_arrItem, i) => {
+      <TableContainer aria-labelledby={id}>
+        <caption id={id}>Table with dividers</caption>
```
tycup: The `<caption>` element is placed as a direct child of `<TableContainer>` instead of inside `<Table> `which isnt semantically correct - can it be moved inside the Table component so its a child of `<table>`?

tycup: (i see you've done this elsewhere so might be just an omission)

packages/core/src/table/Table.css
```diff
@@ -223,3 +222,8 @@ table.saltTable td.saltTable-td-align-left {
 table.saltTable td.saltTable-td-align-right {
   text-align: end;
 }
+
+.saltTable-tableContainer {
```
tycup: Wondering if this could be .saltTable-container to avoid double 'table' ?

wemmyo: Yup! Updated

site/src/examples/table/Dividers.tsx
wemmyo: Good spot thanks. Fixed!

site/docs/components/table/accessibility.mdx
```diff
@@ -8,4 +8,46 @@ data:
   $ref: ./#/data
 ---
 
-## Coming soon
+## Best practices
```
wemmyo, line 11: @karlgoldstraw @jake-costa I've created this a11y docs based on what I've gathered from conversations so far. Do let me know if any changes are required

.changeset/giant-mugs-change.md
```diff
@@ -0,0 +1,19 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/core": minor
```

.changeset/giant-mugs-change.md
```diff
@@ -0,0 +1,19 @@
...
+---
+
+Added `TableContainer` component to provide accessibility features for scrollable tables
+
+- Introduced new TableContainer wrapper component for tables with horizontal scroll
+- Implemented ARIA attributes (role="region", aria-labelledby) for screen reader support
+- Added keyboard navigation support for scrollable content (tab focus on container)
+- Ensured focus indicators are visible when navigating scrollable areas
```
joshwooding: Are these actually individual points?

#1 repeats the first line of the changelog, #2 is an implementation detail and we should avoid focusing too much on them, #3 is I guess part of the justification of doing the work and #4 seems like a result of the entire work

packages/core/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { useWindow } from "@salt-ds/window";
+import { clsx } from "clsx";
+import { forwardRef, type HTMLAttributes, useRef } from "react";
+import { useElementScrollable, useForkRef } from "../utils";
+import { withTableBaseName } from "./Table";
+import tableCss from "./Table.css";
+
+export const TableContainer = forwardRef<
+  HTMLDivElement,
+  HTMLAttributes<HTMLDivElement>
```
joshwooding: We should have TableContainerProps to be consistent

packages/core/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <div
+      ref={handleRef}
+      className={clsx(withTableBaseName("container"), className)}
+      role={role ?? (isScrollable ? "region" : undefined)}
+      tabIndex={tabIndex ?? (isScrollable ? 0 : undefined)}
+      {...(ariaLabelledby && isScrollable
+        ? { "aria-labelledby": ariaLabelledby }
+        : {})}
+      {...(ariaLabel && isScrollable ? { "aria-label": ariaLabel } : {})}
```
joshwooding: This needs a discussion, elsewhere if you add aria-labelledby or aria-label it's always applied. This breaks that convention. Also if this is decided we should clearly document that fact.

wemmyo: In this case it was to resolve a biome warning where `aria-label` and `aria-labelledby` not supported by a div, but when the region role is true the aria labels become valid.
Also requested by @jake-costa to conditionally apply the labels when scrollable

.changeset/giant-mugs-change.md
wemmyo: Good point, I can just merge them all into a single sentence

.changeset/giant-mugs-change.md
```diff
@@ -0,0 +1,14 @@
+---
+"@salt-ds/core": minor
```
wemmyo:
```suggestion
"@salt-ds/lab": minor
```

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,53 @@
...
+      <div
+        ref={handleRef}
+        className={clsx(withTableBaseName("container"), className)}
+        role={role ?? (isScrollable ? "region" : undefined)}
+        tabIndex={tabIndex ?? (isScrollable ? 0 : undefined)}
+        {...(ariaLabelledby && isScrollable
+          ? { "aria-labelledby": ariaLabelledby }
+          : {})}
+        {...(ariaLabel && isScrollable ? { "aria-label": ariaLabel } : {})}
+        {...rest}
```
mark-tate: can we not make this magical?

If someone has passed in props, they disappear if it's not scrollable.
I would say any props we add based on scrolling are defaults that can be overriden through props.

sp we should not destructure `ariaLabel` and `ariaLabelledby` and instead refer to them as `props?.ariaLabel` instead

wemmyo: Can do, thanks. I'll make those changes

packages/lab/src/__tests__/__e2e__/table/Table.cy.tsx
```diff
@@ -0,0 +1,94 @@
...
+  NonScrollableTable,
+} = composedStories;
+
+describe("GIVEN a Table", () => {
+  checkAccessibility(composedStories);
+});
+
+describe("GIVEN a Table inside a TableContainer", () => {
+  it("THEN should have region role, be focusable and have accessible name when scrollable", () => {
+    cy.mount(<ScrollableVertically />);
```
mark-tate, line 23: Do we need to check horizontal as well ?
(or is that covered by one of the other tests)

packages/lab/src/utils/useElementScrollable.ts
```diff
@@ -0,0 +1,77 @@
...
+  const [scrollable, setScrollable] = useState({ x: false, y: false });
+  const contextWindow = useWindow();
+  const targetWindow = options?.targetWindow ?? contextWindow;
+
+  useEffect(() => {
+    const element = ref?.current ?? null;
+    if (!element) return;
+
+    const checkScrollable = () => {
+      const x = element.scrollWidth > element.clientWidth;
```
mark-tate: I guess we are not worried about 

overflowX/Y = "hidden" ?

which may set things are scrollable.

Don't need to add if there is not a use-case but if we are making this available from our barrel file, it may be used by ways that do.

packages/lab/src/utils/useElementScrollable.ts
```diff
@@ -0,0 +1,77 @@
...
+        return prev;
+      });
+    };
+
+    // Initial check
+    checkScrollable();
+
+    // Set up ResizeObserver to monitor size changes
+    const ResizeObserverCtor =
+      (targetWindow as WindowWithResizeObserver | null)?.ResizeObserver ??
```
mark-tate: Probably enough for the mo... the ways to get scrollbars are the window changes size but the content can dynamically change size as well.

That might be overkill at this stage, would wait to see if it is an issue but highlighting for the radar.

packages/lab/src/__tests__/__e2e__/table/Table.cy.tsx
wemmyo: Yes the other stories have horizontal scroll. This is the only story with vertical scroll

packages/lab/src/utils/useElementScrollable.ts
wemmyo: That's a good point thanks. I'll make a note of this and the comment above, so we can keep track of it

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,64 @@
...
+export interface TableContainerProps extends HTMLAttributes<HTMLDivElement> {
+  /**
+   * Accessible label for the table when it is scrollable.
+   */
+  label?: string;
+  /**
+   * ID of the element that contains the accessible label for the table when it is scrollable.
+   */
+  labelId?: string;
+}
```
wemmyo: @mark-tate @joshwooding I've introduced these props to address previous issues raised. `label` maps to `aria-label` and `labelId` maps to `aria-labelledby` which will be used to provide the defaults when scrollable. This prevent direct mutation of the aria-labels allowing users easily override when necessary

packages/lab/src/table/TableContainer.tsx
wemmyo: This should be fixed now, thanks

.changeset/giant-mugs-change.md
wemmyo: Updated this to lab as well

packages/core/src/table/TableContainer.tsx
wemmyo: Done!

packages/core/src/table/TableContainer.tsx
wemmyo: Taken a new approach which should resolve this by adding label and labelId props to table container

packages/lab/src/utils/useElementScrollable.ts
```diff
@@ -0,0 +1,86 @@
...
+      (typeof ResizeObserver !== "undefined" ? ResizeObserver : undefined);
+
+    if (!ResizeObserverCtor) {
+      return;
+    }
+
+    const resizeObserver = new ResizeObserverCtor(checkScrollable);
```
tycup: As discussed in person - is there an opportunity to use the existing `useResizeObserver.ts` either now or perhaps as a note for future refactor? (non-blocking point)

packages/lab/src/utils/useElementScrollable.ts
```diff
@@ -0,0 +1,86 @@
...
+    resizeObserver.observe(element);
+
+    return () => {
+      resizeObserver.disconnect();
+    };
+  }, [ref, targetWindow]);
```
tycup: Might be fine, but just a placeholder to double check if it's ok to use this ref in the dependency array

packages/lab/src/utils/useElementScrollable.ts
wemmyo: Good shout, this has been updated thanks

packages/lab/src/utils/useElementScrollable.ts
wemmyo: This has now been refactored 👍

tycup (dismissed):

packages/lab/src/__tests__/__e2e__/table/Table.cy.tsx
joshwooding, line 1: If this change is landing in lab, the tests should stay in lab too

packages/lab/src/table/Table.tsx
```diff
@@ -26,39 +27,56 @@ export interface TableProps extends ComponentPropsWithoutRef<"table"> {
...
+      variant = "primary",
+      divider = "tertiary",
+      zebra = false,
+      ...rest
+    } = props;
+
+    const generatedId = useId();
+    const { setId, setLabelledBy } = useTable();
+    const tableId = props.id ?? generatedId;
+    const tableLabelledBy = props["aria-labelledby"];
```
joshwooding: Can we clean this code up by destructuring the props we use e.g. id and aria-labelledby

packages/lab/src/table/Table.tsx
```diff
@@ -26,39 +27,56 @@ export interface TableProps extends ComponentPropsWithoutRef<"table"> {
...
+    const {
+      children,
+      className,
+      variant = "primary",
+      divider = "tertiary",
+      zebra = false,
+      ...rest
+    } = props;
+
+    const generatedId = useId();
```
joshwooding: The id prop can be passed in here rather than what's being done on line 50

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+    const targetWindow = useWindow();
+    useComponentCssInjection({
+      testId: "salt-table-container",
+      css: tableCss,
+      window: targetWindow,
+    });
+
+    const ariaLabelledBy = props["aria-labelledby"];
+    const ariaLabel = props["aria-label"];
```
joshwooding: Can we destructure these please

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,88 @@
...
+      () => ({ id: tableId, setId: setTableId, labelledBy, setLabelledBy }),
+      [tableId, labelledBy],
+    );
+
+    return (
+      <TableContext.Provider value={contextValue}>
+        <div
+          ref={handleRef}
+          className={clsx(withTableBaseName("container"), className)}
+          role={role ?? (isOverflowing ? "region" : undefined)}
```
joshwooding: Is there a way to group the props applied when isOverflowing is true? We've got these lines here and then ariaProps above

packages/lab/src/table/Table.tsx
wemmyo: Good to know, thanks! Updated

packages/lab/src/table/TableContainer.tsx
wemmyo: Looks a bit messy but it's grouped now, let me know what you think

packages/lab/src/table/Table.tsx
wemmyo: Done!

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,112 @@
...
+  useRef,
+  useState,
+} from "react";
+import { withTableBaseName } from "./Table";
+import tableCss from "./Table.css";
+import { TableContext } from "./TableContext";
+
+export interface TableContainerProps
+  extends Omit<
+    HTMLAttributes<HTMLDivElement>,
```
joshwooding:
```suggestion
    ComponentPropsWithoutRef<"div">,
```

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+    useResizeObserver({ ref: scrollRef, onResize: checkOverflow });
+
+    useIsomorphicLayoutEffect(() => {
+      checkOverflow();
+    }, [checkOverflow]);
+
+    const overflowProps = isOverflowing
+      ? {
+          role: "region",
```
joshwooding: Should the role prop be ignored in the scenario?

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,112 @@
...
+          tabIndex: 0,
```
joshwooding: Same for the tabIndex here

packages/lab/src/table/TableContainer.tsx
wemmyo: That's a good question, wasn't sure of the best approach between reinforcing our best practice and giving more freedom. 
 I guess it makes sense to include it, if they're overriding these props even with the type warnings then we should probably just allow it

packages/lab/src/table/TableContainer.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const scrollRef = useRef<HTMLDivElement | null>(null);
+    const handleRef = useForkRef<HTMLDivElement>(ref, scrollRef);
+
+    const checkOverflow = useCallback(() => {
+      const element = scrollRef.current;
+      if (!element) return;
+      const overflowing =
+        element.scrollHeight > element.clientHeight ||
+        element.scrollWidth > element.clientWidth;
+      setIsOverflowing((prev) => (prev !== overflowing ? overflowing : prev));
```
joshwooding: NIT: I don't think this optimisation makes sense? It's dealing with booleans

joshwooding (dismissed):

packages/lab/src/table/TableContainer.tsx
wemmyo: Good point, I'll update. Thanks!

PR #6028, opened 2026-02-06
Update foundation pages

joshwooding:


site/docs/foundations/density.mdx
```diff
@@ -22,9 +26,9 @@ Read about implementing density in code on the [SaltProvider](/salt/components/s
   background="secondary"
 />
 
-## High density
+### High density
```
wemmyo, line 29: I previously suggested h2 to make the sections easier to find on the right side navigation

joshwooding: The structure of the page shouldn't be decided by what gets shown in the TOC. It's decided by the logical grouping of the content.

The TOC was standardised to shown what it shows.

PR #6031, opened 2026-02-09
Fix vertical navigation active state for usage and accessibility tabs

### LINKED ISSUE #5678: Add scrolling to the active navigation item on site
### Description

On site, vertical navigation should scroll to active item on page load so users can see which nav item is open

### Expected behavior

When active nav item is not in view, the vertical nav should scroll to active nav link/item

### Alternatives and examples

.

### Context

Previous ticket: https://github.com/jpmorganchase/salt-ds/issues/5562#issuecomment-3408178938
Context on the issue: https://github.com/jpmorganchase/salt-ds/pull/5586#issuecomment-3328739822

### LINKED ISSUE #6074: Vertical navigation is only active on examples
### Related page(s)

https://www.saltdesignsystem.com/salt/components/accordion/usage

### Kind of issue

Other

### Description

Vertical navigation should stay active for a components doc, when in the examples, usage and accessibility tabs

### Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

wemmyo:
Closes #6074
Fixes the vertical navigation not highlighting the active tab when navigating to /usage, /accessibility routes and improve active nav item discoverability on site.

- Extends the route normalization function to include /usage, /accessibility, and /examples paths
- Ensures the vertical navigation correctly reflects the active state for these routes
- Auto scroll to active nav item only on initial load. Closes #5678

### DIFF
--- a/site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
+++ b/site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
@@ -126,10 +126,7 @@ export const VerticalNavigation = ({
   const hasScrolledToActiveRef = useRef(false);
 
   useEffect(() => {
-    if (
-      hasScrolledToActiveRef.current ||
-      !normalizedSelectedNodeId
-    ) {
+    if (hasScrolledToActiveRef.current || !normalizedSelectedNodeId) {
       return;
     }
 


site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
```diff
@@ -122,12 +122,37 @@ export const VerticalNavigation = ({
...
+    if (!navigationContainer) return;
+
+    const activeItem = navigationContainer.querySelector(
+      '[aria-current="page"]',
+    );
+    if (activeItem instanceof HTMLElement) {
+      activeItem.scrollIntoView({ block: "center" });
+      hasScrolledToActiveRef.current = true;
+    }
+  }, [mappedNavData, normalizedSelectedNodeId]);
```
tycup: Does `mappedNavData` need to be in the dependancy array - i think its created fresh on each render via `mapMenu(menu)`?. I see its guarded by `hasScrolledToActiveRef `which makes it work but a bit misleading perhaps. 

Worth memoizing mappedNavData?

site/src/components/navigation/vertical-navigation/utils/index.ts
tycup: Does this comment need to be updated to include the routes added?

site/src/components/navigation/vertical-navigation/utils/index.ts
```diff
@@ -23,8 +23,10 @@ export function statusToBadgeValue(status: string) {
  This function normalizes the selectedNodeId to match the href in navData
  **/
 export function normalizeSelectedNodeId(link: string, navData: Item[]): string {
```
tycup, line 27: nit - do you need regex here? `endsWith` already confirms the suffix so could use `slice` to remove it once instead of creating two RegExp objects

```
const base = link.slice(0, -matchedRoute.length);
const indexLink = `${base}/index`;
```

wemmyo: Good point thanks. Updated

site/src/components/navigation/vertical-navigation/utils/index.ts
wemmyo: Technically its still the case because /index will redirect to /examples by default but I've rewritten the comment to explain what the function does

site/src/components/navigation/vertical-navigation/VerticalNavigation.tsx
wemmyo: Yeah `mappedNavData` isn't a necessary dependency and has been removed. Thanks

PR #6033, opened 2026-02-09
Add keyboard navigation by default to dialog content overflow

### LINKED ISSUE #4308: Modals with long content does not accept keyboard navigation by default
### Latest version

- [X] I have tested the latest version

### Description

When showing long content Overlay does not allow user to TAB into the scrollable content to scroll with the keyboard.  This can be rectified by setting a tabIndex on OverlayPanelContent. 

### Steps to reproduce

https://salt-template-yxnzmn.stackblitz.io

### Expected behavior

Automatically enable tab into content when content becomes scrollable 

### Package name(s)

Core (@salt-ds/core)

### Package version(s)

1.37.0

### Browser

- [X] Chrome
- [ ] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [X] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

wemmyo:
Closes #4308

**Fix accessible name reference**
- Fix issue where the dialog's accessible name incorrectly included interactive elements (e.g., Close button)
- Point `aria-labelledby` to the header text element instead of the outer container to ensure only the title is announced

**Refactor header-to-dialog relationship**
- Refactor ID generation so the header creates and owns its ID
- Apply `aria-labelledby` conditionally, only when a header exists and provides an ID

**Add keyboard navigation for scrollable content**
- Add keyboard navigation support for scrollable dialog content
- Use header text as the primary accessible name, with a fallback when no header is provided

### DIFF
--- a/.changeset/twenty-singers-taste.md
+++ b/.changeset/twenty-singers-taste.md
@@ -2,5 +2,5 @@
 "@salt-ds/core": patch
 ---
 
-- Fixed dialog accessible name to exclude interactive elements and added keyboard navigation for scrollable dialog content
-- Deprecated `idProp` prop in Dialog component. IDs are now auto-generated to ensure proper ARIA labeling
+- Fixed dialog accessible name to exclude interactive elements and added keyboard navigation for scrollable dialog content.
+- Deprecated `idProp` prop in Dialog component.


packages/core/src/dialog/DialogContext.tsx
```diff
@@ -3,10 +3,16 @@ import type { ValidationStatus } from "../status-indicator";
...
+  setHeaderId: (id: string) => void;
+  contentScrollId?: string;
+  setContentScrollId?: (id: string) => void;
 }>({
   status: undefined,
-  id: "",
+  headerId: undefined,
+  setHeaderId: () => {},
+  contentScrollId: undefined,
+  setContentScrollId: () => {},
```
tycup: The whole of `DialogContext` is exposed via index.tsx (can it be renamed to .ts? - nit). Would consumers use setContentScrollId?

packages/core/src/dialog/DialogHeader.tsx
```diff
@@ -79,7 +91,7 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
       >
         {status && <StatusIndicator status={status} />}
         <div className={withBaseName("container")}>
-          <H2 className={withBaseName("header")}>
+          <H2 id={id} className={withBaseName("header")}>
```
tycup: I guess we assume the consumer wouldn't add an `id` on `DialogHeader`? Technically we could get two identical `id`s if so

packages/core/src/dialog/DialogContext.tsx
wemmyo: Yeah I'll rename it. They shouldn't need use setContentScrollId, only needed internally to set defaults

packages/core/src/dialog/DialogHeader.tsx
wemmyo: That's a good point thanks. let me look into it

wemmyo: Should be good now

packages/core/stories/dialog/dialog.qa.stories.tsx
```diff
@@ -27,7 +27,15 @@ export default {
 
 function FakeDialog({ children, status, id }: DialogProps) {
   return (
-    <DialogContext.Provider value={{ status, id }}>
+    <DialogContext.Provider
+      value={{
+        status,
+        headerId: id,
```
joshwooding: Unfortunately this is a breaking change (rename from id to headerId) since DialogContext is exported.

wemmyo: Updated

site/docs/components/dialog/accessibility.mdx
```diff
@@ -23,12 +23,15 @@ You can control which element receives focus when the dialog is opened by passin
 - For dialogs that prompt for user confirmation, the call to action button takes focus first.
 - For dialogs with destructive interactions, the cancel button takes focus first.
 
+### Scrollable Content
```
wemmyo:
```suggestion
### Scrollable content
```

packages/core/src/dialog/Dialog.tsx
```diff
@@ -93,6 +93,8 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
       disableScrim,
       idProp,
       initialFocus,
+      id,
+      "aria-labelledby": ariaLabelledBy,
```
joshwooding, line 99: So, there seems to be a bug here initially `idProp` instead of `id` but we shouldn't have them do 2 different things

packages/core/src/dialog/Dialog.tsx
```diff
@@ -102,11 +104,11 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
       window: targetWindow,
     });
 
-    const id = useId(idProp);
-
+    const contentScrollId = useId(id);
     const currentBreakpoint = useCurrentBreakpoint();
 
     const [showComponent, setShowComponent] = useState(false);
+    const [headerId, setHeaderId] = useState(idProp);
```
joshwooding, line 113: I found the logic flow here to be confusing, when the idProp changes it gets passed to header which then sets this state?

packages/core/src/dialog/Dialog.tsx
```diff
@@ -135,16 +137,20 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
...
     return (
       <DialogContext.Provider value={contextValue}>
         <ConditionalScrimWrapper condition={showComponent && !disableScrim}>
           <FloatingComponent
+            id={contentScrollId}
             open={showComponent}
             role="dialog"
             aria-modal="true"
-            aria-labelledby={id}
+            aria-labelledby={ariaLabelledBy ?? headerId}
```
joshwooding: I'm not too sure about making ariaLabelledBy overwrite headerId, is there a downside to concatenating it

packages/core/src/dialog/DialogContent.tsx
```diff
@@ -58,9 +60,13 @@ export const DialogContent = forwardRef<HTMLDivElement, DialogContentProps>(
...
     const checkOverflow = useCallback(() => {
       if (!divRef.current) return;
-      setIsOverflowing(
+      setIsOverflowingVertically(
         divRef.current.scrollHeight > divRef.current.offsetHeight,
       );
+      setIsOverflowing(
+        divRef.current.scrollWidth > divRef.current.offsetWidth ||
+          divRef.current.scrollHeight > divRef.current.offsetHeight,
+      );
```
joshwooding: Is there a reason we need to handle these separately? DialogContent shouldn't overflow horizontally

packages/core/src/dialog/DialogHeader.tsx
```diff
@@ -62,10 +66,16 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
     });
 
     const status = statusProp ?? statusContext;
+    const headerId = useId(contextHeaderId);
+
+    useIsomorphicLayoutEffect(() => {
```
joshwooding: I don't think we need to use a layout effect here, we can probably get away with using useEffect.

packages/core/stories/dialog/dialog.qa.stories.tsx
```diff
@@ -27,7 +27,13 @@ export default {
 
 function FakeDialog({ children, status, id }: DialogProps) {
   return (
-    <DialogContext.Provider value={{ status, id }}>
+    <DialogContext.Provider
+      value={{
+        status,
+        id,
+        setId: () => {},
```
joshwooding: Did you decide whether this should be optional?

joshwooding:
Sorry, for the slow review. I've left a few comments.

packages/core/src/dialog/Dialog.tsx
wemmyo: Previously we had the Dialog always creating the `id` and passing it to the Header. However, if the Header isn't provided then the dialog has `aria-labelledby` pointing to nothing. I reversed the flow to allow the header create the id and the Dialog checks if the header has provided the name before assigning the `aria-labelledby`. From what I understand, the `idProp` let's the user provide a custom id for naming the Dialog, because the Header now creates the id, if the idProp is provided we set it as the initial value sent the header to use as the `id`

packages/core/src/dialog/Dialog.tsx
joshwooding: Will deprecate idProp

packages/core/stories/dialog/dialog.qa.stories.tsx
wemmyo: Yes, this has been made optional

packages/core/src/dialog/DialogContent.tsx
wemmyo: Single setter checking for x and y overflow. We shouldn't have horizontal overflow but left for consistency

packages/core/src/dialog/DialogHeader.tsx
wemmyo: Updated to use useEffect

packages/core/src/dialog/Dialog.tsx
wemmyo: Done

packages/core/src/dialog/Dialog.tsx
```diff
@@ -135,16 +139,20 @@ export const Dialog = forwardRef<HTMLDivElement, DialogProps>(
...
     return (
       <DialogContext.Provider value={contextValue}>
         <ConditionalScrimWrapper condition={showComponent && !disableScrim}>
           <FloatingComponent
+            id={dialogId}
             open={showComponent}
             role="dialog"
             aria-modal="true"
-            aria-labelledby={id}
+            aria-labelledby={clsx(ariaLabelledBy ?? headerId) || undefined}
```
joshwooding:
```suggestion
            aria-labelledby={clsx(ariaLabelledBy, headerId) || undefined}
```

wemmyo: Oops, thanks. Don't know how I missed that

packages/core/src/dialog/Dialog.tsx
wemmyo: Also included in changeset

packages/core/src/dialog/DialogHeader.tsx
```diff
@@ -62,10 +67,16 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
     });
 
     const status = statusProp ?? statusContext;
+    const headerId = useId(contextHeaderId);
+
+    useEffect(() => {
+      if (headerId) {
+        setHeaderId?.(headerId);
```
joshwooding, line 74: @wemmyo not sure if I'm misremembering but you mentioned ids were going to be supplied via Header, but I don't think that's currently possible?

wemmyo: That's correct, header creates an id then passes it to the Dialog through context. Dialog also creates an id which acts as a backup in the absence of the header id. This ensures the scrollable region has two options for the accessible name

packages/core/src/dialog/DialogContext.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const DialogContext = createContext<DialogContextValue>(
+  "DialogContext",
+  {
+    status: undefined,
+    id: undefined,
+    setId: () => {},
+    dialogId: undefined,
+    setDialogId: () => {},
```
tycup, line 20: Does `setDialogId` get used anywhere?

packages/core/src/dialog/DialogHeader.tsx
```diff
@@ -62,10 +67,16 @@ export const DialogHeader = forwardRef<HTMLDivElement, DialogHeaderProps>(
     });
 
     const status = statusProp ?? statusContext;
+    const headerId = useId(contextHeaderId);
+
+    useEffect(() => {
```
tycup, line 72: nit/maybe unrealistic/non-blocking: does this need a cleanup function? i guess DialogHeader would never unmount while the dialog remains open

packages/core/src/dialog/DialogContext.ts
wemmyo: No, its not used anywhere. However, since context is exported might be useful to also include the setter function.

packages/core/src/dialog/DialogHeader.tsx
wemmyo: Yeah like you said, it won't unmount when dialog is open and there's no listener or subscription setup so shouldn't be necessary

Issue #2386, opened 2023-09-12
Announcement Dialog Pattern


bhoppers2008:
## Summary

Create documentation and Storybook/Figma assets for the Announcement Dialog pattern to achieve UITK parity and support Salt Design System adoption.

---

## Description

- Review UITK’s documented use cases and identify any additional cases relevant to Salt.
- Engage with MTK (known stakeholder), and use BLM group feedback and Figma analytics to identify further use cases and stakeholders.
- Schedule and facilitate a kickoff meeting with all identified stakeholders to align on requirements, deliverables, and timelines.
- Spec pattern flows using the new spec format (currently a work in progress, but sufficient to begin), ensuring consistency with Salt Design System standards.
- Create and publish pattern documentation page.
- Spec and build pattern examples, to link to docs page. 
- Create and organize Figma assets for the Announcement Dialog.

---

## Context

This work is part of the ongoing effort to close the UITK parity gap for Salt Design System patterns. The Announcement Dialog is a UITK component which enables products to highlight new features and releases. The pattern requires comprehensive documentation and assets are required to ensure consistency and usability across platforms. Stakeholder engagement is critical to capture all relevant use cases and requirements.

---

## Additional information

- UITK Announcement Dialog docs can be found [here](https://uitk.gaiacloud.jpmchase.net/build/components/announcement-dialog).
- Salt Announcement Dialog should consider the new Salt Dialog component, in addition to the new Salt Typography and Spacing foundations.
- Figma analytics and BLM group feedback

**Notes:**
- Please flag any blockers or gaps in the new spec format as you proceed.
- Collaborate with documentation and accessibility leads as needed.

mark-tate:
@bhoppers2008 can you allocate to a provisional sprint in Q4, feels like a 1 sprint effort ?

origami-z:
SourceGraph search around 50 instances

origami-z:
Mocha: kick-off, EOS ready for dev

bhoppers2008:
Scheduled for Q2.

aimee-liou:
- Reviewed and merged component to main Figma library with Yun
- Created and merged stickers for Stickersheet

PR #6037, opened 2026-02-10
Add announcement dialog pattern

tycup:
TODO:

- [x] Add images from Figma


packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,499 @@
...
+  SplitLayout,
+  StackLayout,
+  type StackLayoutProps,
+  Text,
+  useResponsiveProp,
+} from "@salt-ds/core";
+import { CloseIcon } from "@salt-ds/icons";
+import type { Meta, StoryFn } from "@storybook/react-vite";
+import { type ElementType, useState } from "react";
+import stockPhoto from "../../assets/stockPhoto.png";
```
tycup: Let me know if there's a better image to use (doesn't scale that well, could use the exampleImage pngs?)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,499 @@
...
+          preheader="New feature"
+          header="Trade across markets"
+          actions={<CloseButton onClick={() => setOpen(false)} />}
+          disableAccent
+        />
+        <DialogContent>
+          <SplitLayout
+            startItem={
+              <StackLayout gap={1} className="announcementContent">
+                <H3 style={{ margin: "0 0 var(--salt-spacing-100) 0" }}>
```
tycup: If we don't change the margin, we get double margin from the DialogHeader

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,499 @@
...
+                  Get started today and discover how easy it is to build
+                  institutional-grade portfolios with just a few clicks. Your
+                  team can collaborate on shared portfolios and track
+                  performance across all your investments.
+                </Text>
+              </StackLayout>
+            }
+            endItem={
+              <img
+                alt="Clouds in the sky"
```
tycup: Should this be better alt or good enough?

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,499 @@
...
+
+interface AnnouncementContent {
+  preheader: string;
+  header: string;
+  subheader?: string;
+  body: string;
+  imageAlt: string;
+}
+
+const multiSlideAnnouncementContent: AnnouncementContent[] = [
```
tycup, line 33: I kept the mock data in the file and placed it in one array only for multi because i presumed its better to be a bit more verbose in example? Like can ofc iterate through the array, but felt more like an example to in-line where it made sense? lmk if incorrect assumption

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,96 @@
+---
+title: Announcement dialog
+description: "An announcement dialog promotes new features and can provide users with access to onward action, such as installing version updates."
+layout: DetailPattern
+tags:
+  - pattern
+aliases:
+  - /salt/patterns/announcement-dialog
+data:
+  resources:
```
tycup, line 10: Should there be a Figma link here too?

aimee-liou: i think its just storybook

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,136 @@
...
+- The content in the announcement description should be short and concise. This allows users to quickly grasp the necessary information, reducing cognitive load and preventing abandonment.
+- By default, the entire image or infographics should be visible without scrolling. To achieve this, adjust the size of the dialog or check that the image is appropriately cropped.
+</Callout>
+
+## Responsive behavior
+
+Once the announcement dialog hits a certain breakpoint, you can configure the layout to either:
+
+1. Stack the announcement description and image.
+2. Change the orientation to stack the button bar, following the [Button bar](/salt/patterns/button-bar).
```
aimee-liou:
```suggestion
2. Change the orientation to stack the button bar.
```
lets actually remove this to align with preference dialog

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,136 @@
...
+### Stacked button bar
+
+<Diagram
+  src="/img/patterns/announcement-dialog/stacked-button.png"
+  srcDark="/img/patterns/announcement-dialog/stacked-button-dark.png"
+  alt="Announcement dialog with a stacked button bar"
+  border
+/>
+
+For more information, refer to the stacked [Button bar](/salt/patterns/button-bar) guidance.
```
aimee-liou:
```suggestion
For more information, refer to the [stacked button bar](/salt/patterns/button-bar) guidance.
```

wemmyo: 'button bar' matches the pattern name so the original might be better. Not sure about the preferred casing though, maybe all lowercase?

site/docs/patterns/announcement-dialog.mdx
wemmyo: Nice to have if available

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
aimee-liou: is this for all the images or a specific one (sry not sure what image this is referring to) - i can scale all the images down to fit the frames better lmk

site/docs/patterns/announcement-dialog.mdx
aimee-liou: fair point - im copying preference dialog and they link '[stacked button bar] guidance'. maybe because 'guidance' is used so it's not fully referring to the Button bar, but the button bar guidance? either is fine with me though

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,494 @@
...
+            endItem={
+              <img
+                alt="Clouds in the sky"
+                src={stockPhoto}
+                className="announcementImage"
+              />
+            }
+          />
+        </DialogContent>
+        <DialogActions>
```
wemmyo, line 514: Bar button pattern is incorrect on desktop but correct on mobile. Should be [transparent, bordered, solid]

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,494 @@
...
+            endItem={
+              <img
+                alt={currentSlide.imageAlt}
+                src={stockPhoto}
+                className="announcementImage"
+              />
+            }
+          />
+        </DialogContent>
+        <DialogActions>
```
wemmyo, line 186: If I remember correctly, similar to wizard. We remove previous on first page rather than disabling it

site/docs/patterns/announcement-dialog.mdx
tycup: think i prefer keeping the link to mach the pattern name but undecided on lowercase - I kept it first letter capitalised also for consistency throughout

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+description: "An announcement dialog promotes new features and can provide users with access to onward action, such as installing version updates."
+layout: DetailPattern
+tags:
+  - pattern
+aliases:
+  - /salt/patterns/announcement-dialog
+data:
+  resources:
+    [
+      { href: "https://storybook.saltdesignsystem.com/?path=/story/patterns-announcement-dialog--announcement-dialog-example", label: "Examples" }
```
wemmyo:
```suggestion
      { href: "https://storybook.saltdesignsystem.com/?path=/story/patterns-announcement-dialog", label: "Examples" }
```
This way it would redirect to the first story regardless of the story name

site/docs/patterns/announcement-dialog.mdx
aimee-liou: i think this is a good thing to align with the team on, im looking at the docs and they all do something different. im also unsure when we use the block styling when referring to components. ill ask tm - standby!

wemmyo: @aimee-liou ignore my previous comment, the most important thing is that the link conveys where it goes to rather than just matching the pattern page name

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+  srcDark="/img/patterns/announcement-dialog/anatomy-dark.png"
+  alt="Announcement dialog anatomy"
+  border
+/>
+
+### Layout
+
+- Use the [Header block](/salt/patterns/header-block) to display a header and pre-header.
+- Use the [Dialog](/salt/components/dialog) and it's components as the base structure of the announcement dialog.
+- Use a [Button bar](/salt/patterns/button-bar) to display any actions at the bottom of the dialog. See the [Button bar](/salt/patterns/button-bar) pattern to understand how to implement actionable interactions within dialogs across your applications.
```
aimee-liou:
```suggestion
- Use a [Button bar](/salt/patterns/button-bar) to display any actions at the bottom of the dialog. See the button bar pattern to understand how to implement actionable interactions within dialogs across your applications.
```

only link once

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+<Diagram
+  src="/img/patterns/announcement-dialog/spacing.png"
+  srcDark="/img/patterns/announcement-dialog/spacing-dark.png"
+  alt="Announcement dialog spacing and layout"
+  border
+/>
+
+## Multi-announcement dialog
+
+If there are multiple features that need to be announced, include a progress label showing the current and total number of slides in the [Button bar](/salt/patterns/button-bar). The slide indicator uses `--salt-spacing-200` from the core buttons.
```
aimee-liou:
```suggestion
If there are multiple features that need to be announced, include a progress label showing the current and total number of slides in the button bar. The slide indicator uses `--salt-spacing-200` from the core buttons.
```
removed link based on content guidance + our Button bar doesn't have a progress label example so just want to avoid confusion

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+## Disclaimers
+
+If you need to provide legal information for compliance purposes, you can include disclaimer text beneath the [Button bar](/salt/patterns/button-bar).
```
aimee-liou:
```suggestion
If you need to provide legal information for compliance purposes, you can include disclaimer text beneath the button bar.
```
same reasons as above

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+<Diagram
+  src="/img/patterns/announcement-dialog/full-image.png"
+  srcDark="/img/patterns/announcement-dialog/full-image-dark.png"
+  alt="Announcement dialog with a stacking full image and content area"
+  border
+/>
+
+## Content area scrolling
+
+In most cases, scrolling shouldn't be required. If additional content does exceed the available space, only the content area and image should scroll together - keeping the header and button bar visible at all times.
```
aimee-liou:
```suggestion
In most cases, scrolling shouldn't be required. If additional content does exceed the available space, only the content area and image should scroll together—keeping the header and button bar visible at all times.
```

site/docs/patterns/announcement-dialog.mdx
aimee-liou: let's follow wemmy's suit in the thread: link 'stacked button bar guidance' (lowercase) and link to the stacked button bar section

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+tags:
+  - pattern
+aliases:
+  - /salt/patterns/announcement-dialog
+data:
+  resources:
+    [
+      { href: "https://storybook.saltdesignsystem.com/?path=/story/patterns-announcement-dialog", label: "Examples" }
+    ]
+  components: ["Dialog", "Parent-child-layout", "header-block"]
```
wemmyo: ParentChildLayout wasn't used in the examples or mentioned in the docs. Should we exclude it here?

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+---
+
+<Diagram
+  src="/img/patterns/announcement-dialog/announcement-dialog.png"
+  srcDark="/img/patterns/announcement-dialog/announcement-dialog-dark.png"
+  alt="Basic announcement dialog"
+  border
```
wemmyo, line 21:
```suggestion
  border
  background="secondary"
```
@aimee-liou Adding secondary background can help show better separation between the background and foreground for these images. What do you think?

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+
+## When to use
+
+Use an announcement dialog to focus the user's attention on new product features or the highlights of a recent release with a follow-up action, such as "Download update" or "Upgrade". They can also be used to display a notification that doesn't require a response from the user, such as a welcome message for new users or an upcoming site-maintenance notification.
+
+## How to build
+
+### Anatomy
+
+An announcement dialog comprises four key features:
```
wemmyo: Should this be five key features? or the pre-header doesn't count because its optional? Its a bit confusing

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+  alt="Announcement dialog with content area and image scroll"
+  border
+/>
+
+<Callout
+  status="success"
+  title="Best practices"
+>
+- The content in the announcement description should be short and concise. This allows users to quickly grasp the necessary information, reducing cognitive load and preventing abandonment.
+- By default, the entire image or infographics should be visible without scrolling. To achieve this, adjust the size of the dialog or check that the image is appropriately cropped.
```
wemmyo, line 98: Based on this point, should scrolling be removed for the full image example on storybook?

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,494 @@
...
+          <Button sentiment="accented">Try It Now</Button>
+        </DialogActions>
+      </Dialog>
+    </>
+  );
+};
+
+export const ContentScrolling: StoryFn = () => {
```
wemmyo, line 302: Might need to add more content or reduce the height, content isn't scrolling

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,494 @@
...
+              <Button sentiment="accented">Try it now</Button>
+              <Button sentiment="accented" appearance="bordered">
+                Go to dashboard
+              </Button>
+              <Button sentiment="accented" appearance="transparent">
+                Remind me later
+              </Button>
+            </StackLayout>
+          ) : (
+            <FlexLayout gap={1}>
```
wemmyo: Will SplitLayout be more appropriate here? Also to be consistent with the other stories

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: Should the same true with the last slide and the "Next" button?

wemmyo: I'd assume so, disabled buttons aren't accessible anyway

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: Referring to the image used by the dialog itself (updated now to use a more generic example image)

site/docs/patterns/announcement-dialog.mdx
tycup: ty, from Preferences Dialog, removed

site/docs/patterns/announcement-dialog.mdx
tycup: Perhaps "An announcement dialog is comprised of the following components" if we want to keep 'optional'?

site/docs/patterns/announcement-dialog.mdx
tycup: Yeah i've improved this but it also varies by viewport and the sb control bar reduces the height and adds scrollbars (it should be better now)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: Improved

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: Not entirely sure it would be - SplitLayout is a wrapper of FlexLayout so would prefer to use it + only used here for button stacking - i did simplify it though for responsiveness via props

site/docs/patterns/announcement-dialog.mdx
aimee-liou: You're right - we forgot to update from 4 to 5. I like what @tycup said though

site/docs/patterns/announcement-dialog.mdx
aimee-liou: I can share the Stickersheet link for internal users once it's available. Meeting with Yun today to double check - Preferences dialog seems to be linking to something different.

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
wemmyo: Not quite sure I understand. What I mean is, with split layout the 'Remind me later' would be on the left and the remaining two button on the right similar to 'Multi announcement dialog' on desktop. Also, with your current approach the tab order is wrong on mobile. 'reverse' property won't change the element order

site/docs/patterns/announcement-dialog.mdx
aimee-liou: true, but let's stick with primary background to match preferences dialog (assuming this is referring to the background of the image frame and the foreground of the dialog)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
wemmyo: Is this spacing correct? Figma says it should be spacing-100. But here we have spacing-100 + gap=1

site/public/img/patterns/announcement-dialog/spacing.png
wemmyo, line 1: Spacing key is missing from the screenshot

aimee-liou: oh oops, didnt realize i should include that in the export. will re-export now ty

wemmyo: No problem, same for dark mode as well

site/docs/patterns/announcement-dialog.mdx
wemmyo: FYI - 'Announcement dialog' can also be included as a related pattern for Dialog

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: I see what you mean - should be fixed now

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+<Diagram
+  src="/img/patterns/announcement-dialog/stacked-content.png"
+  srcDark="/img/patterns/announcement-dialog/stacked-content-dark.png"
+  alt="Announcement dialog with a stacked content area"
+  border
+/>
+
+### Stacked button bar
+
+For more information, refer to the [stacked button bar](/salt/patterns/button-bar#stacked-button-bar) guidance.
```
aimee-liou:
```suggestion
Refer to the [stacked button bar](/salt/patterns/button-bar#stacked-button-bar) guidance for more information. For a multi-announcement dialog, the progress label appears above the buttons. 
```

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,137 @@
...
+- Use a [Button bar](/salt/patterns/button-bar) to display any actions at the bottom of the dialog. See the button bar pattern to understand how to implement actionable interactions within dialogs across your applications.
+
+<Diagram
+  src="/img/patterns/announcement-dialog/spacing.png"
+  srcDark="/img/patterns/announcement-dialog/spacing-dark.png"
+  alt="Announcement dialog spacing and layout"
+  border
+/>
+
+## Multi-announcement dialog
```
tycup, line 60: Should we keep "dialog" here? Might it be worth simplifying to "Multiple announcements"?

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,496 @@
...
+  {
+    preheader: "New feature",
+    header: "Risk management tools",
+    subheader: "Protection",
+    body: "Monitor and manage risk exposure with advanced analytics, scenario modeling, and automated alerts that keep your portfolio within defined parameters.",
+  },
+];
+
+const CloseButton = ({ onClick }: { onClick: () => void }) => (
+  <Button aria-label="Close" appearance="transparent" onClick={onClick}>
```
wemmyo, line 55: Nitpick: "Close dialog" is more descriptive for a screen reader

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,496 @@
...
+                <H3 className="announcementHeading">Builder</H3>
+                <Text>
+                  Create your own optimised corporate bond portfolios targeting
+                  specific characteristics using a wide range of parameters and
+                  constraints including Yield, Risk, churn, costs and more.
+                </Text>
+              </StackLayout>
+            }
+            endItem={
+              <img alt="" src={exampleImage} className="announcementImage" />
```
wemmyo, line 94: Do we need an alt text?

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,496 @@
...
+            }
+          />
+        </DialogContent>
+        <DialogActions>
+          <Button
+            sentiment="accented"
+            style={direction === "column" ? { width: "100%" } : undefined}
```
wemmyo, line 101: Nitpick: We can simplify with clsx

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
wemmyo: Same with the rest

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,496 @@
...
+export const FullImage: StoryFn = () => {
+  const [open, setOpen] = useState(false);
+
+  const direction: StackLayoutProps<ElementType>["direction"] =
+    useResponsiveProp({ xs: "column", sm: "row" }, "row");
+
+  return (
+    <>
+      <Button onClick={() => setOpen(true)}>Announcement Trigger</Button>
+      <Dialog open={open} onOpenChange={setOpen} style={{ maxWidth: 400 }}>
```
wemmyo, line 271: Can we just use size instead of maxWidth? `size="small"`

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,496 @@
...
+      <Button onClick={() => setOpen(true)}>Announcement Trigger</Button>
+      <Dialog open={open} onOpenChange={setOpen}>
+        <DialogHeader
+          preheader="Product update"
+          header="Change the viewport to see how the content stacks"
+          actions={<CloseButton onClick={() => setOpen(false)} />}
+          disableAccent
+        />
+        <DialogContent>
+          <SplitLayout
```
wemmyo: I think FlexLayout might be more appropriate for the content since they're side by side. SplitLayout according to our docs is to arrange at opposite ends of a container

site/public/img/patterns/announcement-dialog/scrolling-dark.png
wemmyo, line 1: The scrollbar shouldn't overlap with the background image in this screenshot. The light mode screenshot is correct

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: See @karlgoldstraw [A11y review in this thread](https://github.com/jpmorganchase/salt-ds/issues/2386) - makes the read-out slightly nicer on NVDA (i do see that other components say "Close Dialog" but based on review)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: Nope - since it's an example image, it's decorative in this case so alt="" is fine (https://www.w3.org/WAI/tutorials/images/decorative/)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
wemmyo: Oh that's interesting. Its weird it would say the dialog twice

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: size="small" can introduce scroll - we want to keep it full image w/o scroll and mainly exemplary

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: We don't really seem to use clsx that often in storybook (i mean, also likely because we don't need it) and since it's an example, i don't mind it being more explicit in code (same reason why not everything is necessarily a css class in the example)

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
tycup: The docs do lean more towards using SplitLayouts for controls (e.g. buttons) but also:

[When to use](https://www.saltdesignsystem.com/salt/components/layouts/split-layout/usage#when-to-use)
> If you need two defined areas at the opposite horizontal ends of a container, for example, if you need two button bar components in a header.

I also _do_ want my regions at opposite ends of a container, so might be semantically sound?

Using `FlexLayout` also adds more complexity - I'd have to add justify-between and flexGrow

wemmyo: Using FlexItem with FlexLayout might help simplify

tycup: Ahh forgot about that - cheers - done

site/public/img/patterns/announcement-dialog/scrolling-dark.png
aimee-liou: resent thx!

site/docs/patterns/announcement-dialog.mdx
aimee-liou: that's what the figma pattern is called at least: "single announcement dialog" and "multi-announcement dialog". not sure if we usually keep it consistent in docs

site/docs/patterns/announcement-dialog.mdx
```diff
@@ -0,0 +1,131 @@
...
+<Diagram
+  src="/img/patterns/announcement-dialog/announcement-dialog.png"
+  srcDark="/img/patterns/announcement-dialog/announcement-dialog-dark.png"
+  alt="Basic announcement dialog"
+  border
+/>
+
+## When to use
+
+Use an announcement dialog to focus the user's attention on new product features or the highlights of a recent release with a follow-up action, such as "Download update" or "Upgrade". They can also be used to display a notification that doesn't require a response from the user, such as a welcome message for new users or an upcoming site-maintenance notification.
```
joshwooding:
```suggestion
Use an announcement dialog to focus the user's attention on new product features or the highlights of a recent release with a follow-up action, such as **Download update** or **Upgrade**. They can also be used to display a notification that doesn't require a response from the user, such as a welcome message for new users or an upcoming site-maintenance notification.
```

packages/core/stories/patterns/announcement-dialog/announcement-dialog.stories.tsx
```diff
@@ -0,0 +1,517 @@
...
+  subheader?: string;
+  body: string;
+}
+
+const multiSlideAnnouncementContent: AnnouncementContent[] = [
+  {
+    preheader: "New feature",
+    header: "Trade across markets",
+    subheader: "Builder",
+    body: "Create your own optimised corporate bond portfolios targeting specific characteristics using a wide range of parameters and constraints including Yield, Risk, churn, costs and more.",
```
joshwooding:
```suggestion
    body: "Create your own optimized corporate bond portfolios targeting specific characteristics using a wide range of parameters and constraints including yield, risk, churn, costs and more.",
```

PR #6058, opened 2026-02-13
Fix holding left or right arrow key not navigating correctly when using Carousel

ThusharaJ07:
Bug fix: on long  arrow key press when focussed on the slide should move to first or last slide based of left or right arrow key pressed.

### DIFF
--- a/.changeset/famous-adults-talk.md
+++ b/.changeset/famous-adults-talk.md
@@ -2,4 +2,4 @@
 "@salt-ds/embla-carousel": patch
 ---
 
-Fixed: Holding left or right arrow key now jumps to the first or last slide in the carousel.
+Fixed holding left or right arrow key not navigating to the first or last slide in the carousel when using the carousel slides.


packages/embla-carousel/src/CarouselSlides.tsx
```diff
@@ -171,31 +167,30 @@ export const CarouselSlides = forwardRef<HTMLDivElement, CarouselSlidesProps>(
...
+          newSnap = Math.min(currentSnap + 1, numberOfSnaps - 1);
           break;
         }
+        default:
+          return;
       }
+
+      emblaApi.scrollTo(newSnap);
+      focusOnSettle.current = true;
       onKeyDown?.(event);
```
joshwooding: Nearly missed this! The addition of the return in the default switch case means onKeyDown won't be called. Can you move it to top of the handler please.

joshwooding:
```suggestion
onKeyDown?.(event);
      if (!emblaApi) return;

      const numberOfSnaps = emblaApi.scrollSnapList().length ?? 1;
      const numberOfSlidesPerSnap = slideRefs.current.length / numberOfSnaps;
const currentSnap = Math.floor(focusedSlideIndex / numberOfSlidesPerSnap);
      let newSnap = currentSnap;

switch (event.key) {
case "ArrowLeft": {
event.preventDefault();
          newSnap = Math.max(currentSnap - 1, 0);
break;
}
case "ArrowRight": {
event.preventDefault();
          newSnap = Math.min(currentSnap + 1, numberOfSnaps - 1);
break;
}
        default:
          return;
}

      emblaApi.scrollTo(newSnap);
      focusOnSettle.current = true;
```

ThusharaJ07: done

ThusharaJ07: move the keydown to the top

.changeset/famous-adults-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/embla-carousel": patch
+---
+
+Fixed: Holding left or right arrow key now jumps to the first or last slide in the carousel.
```
joshwooding:
```suggestion
Fixed holding left or right arrow key not navigating to the first or last slide in the carousel when using the carousel slides.
```

PR #6067, opened 2026-02-18
Update test cases for carousel arrow key navigation to support React 16/17

MinalBhujade:
Update test cases for carousel arrow key navigation to support React 16/17

### DIFF
--- a/packages/embla-carousel/src/__tests__/__e2e__/Carousel.cy.tsx
+++ b/packages/embla-carousel/src/__tests__/__e2e__/Carousel.cy.tsx
@@ -176,6 +176,7 @@ describe("Given a Carousel", () => {
     });
 
     it("should navigate back to first slide", () => {
+      cy.findAllByRole("tab").eq(3).focus();
       verifySlide("4", false);
 
       cy.findByLabelText(/Previous slide/).click();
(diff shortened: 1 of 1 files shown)


joshwooding (approved):

packages/embla-carousel/src/__tests__/__e2e__/Carousel.cy.tsx
```diff
@@ -180,12 +175,8 @@ describe("Given a Carousel", () => {
...
     });
 
-    it("should navigate back to first slide", function () {
-      // TODO : Skipping for React 16 and 17 (https://github.com/jpmorganchase/salt-ds/issues/5933)
-      if (reactVersion.startsWith("16") || reactVersion.startsWith("17")) {
-        this.skip();
-      }
-
+    it("should navigate back to first slide", () => {
+      cy.findAllByRole("tab").eq(3).focus();
```
joshwooding: @MinalBhujade why were these new commands needed? I would have expected the tests to "just work" when removing the skip?

MinalBhujade: Failing test cases without this as focus is on 1st slide but expecting on 4th slide.

joshwooding: I think this is highlighting a bug and the focus you added is hiding that, the initial index might not be working correctly

PR #6088, opened 2026-02-23
Add Ratings tests

### LINKED ISSUE #6073: Add tests to Rating
In order to move to core Rating needs to have Cypress tests

wemmyo:
Closes #6073

- Added e2e tests for Rating
- Rating component now properly inherits disabled and readOnly states when wrapped in a FormField with these properties enabled

### DIFF
--- a/.changeset/happy-tools-doubt.md
+++ b/.changeset/happy-tools-doubt.md
@@ -2,4 +2,4 @@
 "@salt-ds/lab": patch
 ---
 
-Fixed `Rating` not inheriting readOnly and disabled states from form fields
+Fixed `Rating` not inheriting readOnly and disabled states from form fields.


packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        .should("be.checked")
+        .and("have.value", "4");
+    });
+
+    describe("AND using a mouse", () => {
+      it("THEN should call onChange when clicked", () => {
+        cy.mount(<Controlled />);
+        cy.findAllByRole("radio").should("not.be.checked");
+        cy.findByRole("radio", { name: /2 Stars/i }).realClick();
+        cy.findByRole("radio", { name: /2 Stars/i }).should("be.checked");
```
joshwooding: This doesn't actually check onChange

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+    });
+
+    describe("AND using keyboard", () => {
+      it("THEN should call onChange when space is pressed", () => {
+        cy.mount(<Controlled />);
+        cy.findAllByRole("radio").should("not.be.checked");
+        cy.realPress("Tab");
+        cy.findByRole("radio", { name: /1 Star/i }).should("not.be.checked");
+        cy.realPress("Space");
+        cy.findByRole("radio", { name: /1 Star/i }).should("be.checked");
```
joshwooding: same for here

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        cy.findByRole("radio", { name: /2 Stars/i }).should("be.checked");
+      });
+    });
+
+    describe("AND using keyboard", () => {
+      it("THEN should call onChange when space is pressed", () => {
+        cy.mount(<Controlled />);
+        cy.findAllByRole("radio").should("not.be.checked");
+        cy.realPress("Tab");
+        cy.findByRole("radio", { name: /1 Star/i }).should("not.be.checked");
```
joshwooding: Just curious why do we need a regex to check for the name here?

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        cy.realPress("Space");
+        cy.findByRole("radio", { name: /1 Star/i }).should("be.checked");
+        cy.get("@onChangeSpy").should(
+          "have.been.calledWith",
+          Cypress.sinon.match.any,
+          1,
+        );
+      });
+
+      it("THEN should increase value with down and right arrow keys", () => {
```
joshwooding: can you add checks for focus too please

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        cy.findByRole("radio", { name: /2 Stars/i }).should("not.be.checked");
+        cy.findByRole("radio", { name: /3 Stars/i }).should("be.checked");
+        cy.get("@onChangeSpy").should(
+          "have.been.calledWith",
+          Cypress.sinon.match.any,
+          3,
+        );
+      });
+
+      it("THEN should decrease value with up and left arrow keys", () => {
```
joshwooding: it would be good to add focus checks too

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+    it("THEN should use form field label", () => {
+      cy.mount(<FormFieldSupport />);
+      cy.findByRole("radiogroup", { name: "Form field label" }).should(
+        "be.visible",
+      );
+    });
+  });
+
+  describe("WHEN custom icons are provided", () => {
+    it("THEN should render the custom icons", () => {
```
joshwooding, line 352: This is probably better done in Visual QA than a Cypress test but it's fine to keep

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        <Default getLabel={(value) => labels[value - 1]} defaultValue={3} />,
+      );
+
+      cy.findAllByRole("radio").each((radio, index) => {
+        const value = index + 1;
+        cy.wrap(radio).should("have.attr", "aria-label", labels[value - 1]);
+      });
+    });
+
+    it("THEN should update aria-label for each rating value", () => {
```
joshwooding: This seems to be the same as the test above?

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+          getVisibleLabel={(value, max) => `${value}/${max}`}
+        />,
+      );
+
+      cy.findByText("2/5").should("be.visible");
+      cy.findByRole("radio", { name: /4 Stars/i }).realHover();
+      cy.findByText("4/5").should("be.visible");
+    });
+
+    it("THEN should show custom text labels", () => {
```
joshwooding: NIT: fine test but overlaps with the one above

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,465 @@
...
+        <Default
+          defaultValue={3}
+          getVisibleLabel={(value) => labels[value - 1] || "No rating"}
+        />,
+      );
+
+      cy.findByText("Good").should("be.visible");
+    });
+
+    it("THEN should update visible label when hovering different ratings", () => {
```
joshwooding: This is the same as a previous test?

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
joshwooding, line 1: What we have here is fine but just some general feedback, when writings tests we should try and copy a user's typical flow and prefer longer tests than small tests that test small bits. There are a few reasons for this:

1. You get a more realistic test
2. You pay less overhead of tests spinning up

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
wemmyo: Just thought string check would be restrictive, lowercase label would also be valid. Can change to string if preferred

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
wemmyo: Sure, thanks. I was initially treating them like unit tests when they're really not.
Will make the changes

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
```diff
@@ -0,0 +1,391 @@
...
+        name: "4 Stars",
+      })
+        .should("be.checked")
+        .and("have.value", "4");
+    });
+
+    describe("AND using a mouse", () => {
+      it("THEN should handle selection", () => {
+        const onChangeSpy = cy.stub().as("onChangeSpy");
+        cy.mount(<Controlled onChange={onChangeSpy} value={4} />);
```
tycup: `value={4}` is passed but the story's internal useState sets it back to 0 because `value={value}` is placed after the args get spread - would remove `value={4}` so the test isn't accidentally misleading

.changeset/happy-tools-doubt.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fixed `Rating` not inheriting readOnly and disabled states from form fields
```
joshwooding:
```suggestion
Fixed `Rating` not inheriting readOnly and disabled states from form fields.
```

packages/lab/src/__tests__/__e2e__/rating/Rating.cy.tsx
wemmyo: Good shout, thanks. Removed!

PR #6094, opened 2026-02-24
Fix Carousel on pre React 18

joshwooding:
Replaces #6067


packages/embla-carousel/src/CarouselSlides.tsx
tycup: Unrelated -  is this ?? 1 ever reached? if this returns a number[]

joshwooding: Nope

PR #6147, opened 2026-03-17
Tree | Update spacing between tree elements

tycup:


.changeset/puny-taxis-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Change spacing between tree node elements to fixed-1200
```
wemmyo: This should be past tense, also do we need to include implementation detail?

.changeset/puny-taxis-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Changed spacing between tree node elements to --salt-spacing-100
```
joshwooding:
```suggestion
Updated spacing between tree node elements.
```

packages/lab/src/tree/TreeNodeTrigger.css
```diff
@@ -64,7 +60,6 @@
   color: var(--salt-content-primary-foreground);
 }
 
-.saltTreeNodeTrigger > .saltTreeNode-checkbox {
-  margin-left: calc(var(--salt-spacing-150) - var(--salt-spacing-100));
-  margin-right: calc(var(--salt-spacing-150) - var(--salt-spacing-100));
+.saltTreeNode-disabled .saltTreeNodeTrigger > .saltIcon {
+  color: var(--salt-content-primary-foreground-disabled);
 }
```
joshwooding: I think this is incorrect? We use opacity for disabled styles so this would be layered opacity

Issue #6104, opened 2026-02-26
Rating site docs improvements


wemmyo:
### Related page(s)

https://www.saltdesignsystem.com/salt/components/rating

### Kind of issue

Missing information

### Description

- `Space` should be included in keyboard interaction on accessibility section of site docs
- Live region announcement example on site isn't working as expected
- Custom increments site docs might need to be rewritten to discuss about max number of options, we don't support custom increments which can be confused with precision

### Browser

- [x] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

wemmyo:
**Screen Reader Issue Feedback:**

When using Mac VoiceOver with the "Clear selection" example:

- Pressing the "Clear" button the first time announces "was cleared." This message could be more descriptive to help users understand what was cleared. Consider updating the announcement to something like "Selection was cleared" or "Your selection has been removed."
- After making a new selection and pressing "Clear" a second time, nothing is announced. The clear action only works the first time.

amritadesmet:
Completed and merged  #6201 requested site doc content updates. Removing myself from this ticket as Mark's covering the Live region update.

PR #6201, opened 2026-04-01
Quick fix for Rating page wording 

amritadesmet:
Requested to replace "increment" wording for clarity and adds the existing "Space" function in the Accessibility tab 

As requested in #6104

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11188,9 +11188,9 @@ __metadata:
   linkType: hard
 
 "lodash-es@npm:^4.17.15, lodash-es@npm:^4.17.21, lodash-es@npm:^4.17.23":
-  version: 4.17.23
-  resolution: "lodash-es@npm:4.17.23"
-  checksum: 10/1feae200df22eb0bd93ca86d485e77784b8a9fb1d13e91b66e9baa7a7e5e04be088c12a7e20c2250fc0bd3db1bc0ef0affc7d9e3810b6af2455a3c6bf6dde59e
+  version: 4.18.1
+  resolution: "lodash-es@npm:4.18.1"
+  checksum: 10/8bfad225ef09ef42b04283cdaf7830efcc2ba29ae41b56501c74422155ee1ccaa1f0f6e8319def3451a1fe54dec501c8e4bee622bae2b2d98ac993731e0a5cce
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


site/docs/components/rating/accessibility.mdx
```diff
@@ -19,6 +19,12 @@ data:
 
 </KeyboardControl>
 
+<KeyboardControl keyOrCombos="Space">
+
+- If the focused rating item isn't selected, Space changes the state to selected.
```
joshwooding, line 24: Is this actually true? I thought rating had selection follow focus behaviour

joshwooding: Actually nevermind, it's only really used in one scenario. When the rating isn't set.

site/docs/components/rating/examples.mdx
```diff
@@ -44,13 +44,13 @@ Rating component is compatible with [`FormField`](/salt/components/form-field).
 
 <LivePreview componentName="rating" exampleName="WithFormField" />
 
-## Custom increments
+## Custom number of items
```
joshwooding: Items works probably to describe the anatomy, but I think it's clunky here. Since it's really describing the denominator

site/docs/components/rating/accessibility.mdx
amritadesmet: Yes @wemmyo spotted it and showed me

amritadesmet: In the last example, when you have a button and you clear selection and then go back and click Space, it selects the star. Is it worth specifying that this is in conjunction with a button?

site/docs/components/rating/examples.mdx
amritadesmet: What would you suggest instead of 'item'? Not sure denominator sounds right either

mark-tate: ## Number of rating icons
## Custom number of rating icons
## Maximum size of scale
## Number of ratings
## Scale length

site/docs/components/rating/accessibility.mdx
wemmyo, line 24:
```suggestion
If the focused rating item isn't selected, Space changes the state to selected.
```
Nitpick: Single list item not necessary

site/docs/components/rating/examples.mdx
```diff
@@ -10,7 +10,7 @@ data:
 
 ## Default
 
-By default, the Rating displays five increments, as stars, from which users can select a rating.
+By default, the Rating displays five items, as stars, from which users can select a rating.
```
wemmyo:
```suggestion
By default, the Rating displays five stars, from which users can select a rating.
```
What do you think? Might be clearer this way

site/docs/components/rating/accessibility.mdx
amritadesmet: Other component pages on our site use one bullet for this such as Radio Button and Checkbox which is why I copied it

wemmyo: Not a big issue. Just that you wouldn't need a list for a single item since each bullet point represents a list item

amritadesmet: Spoke with @joshwooding, we've agreed to keep the bullet for consistency with rest of page. Thanks for your input

PR #6205, opened 2026-04-02
Added fixed width slide example for carousel

ThusharaJ07:


packages/embla-carousel/src/Carousel.tsx
```diff
@@ -51,6 +52,11 @@ export interface CarouselProps extends ComponentPropsWithoutRef<"section"> {
    * Disable screenreader announcing slide updates, defaults to false.
    */
   disableSlideAnnouncements?: boolean;
+  /**
+   * Fixed width for each slide(e.g. "300px", "20rem")
+   * When not set, the default responsive behavior is preserved.
+   */
+  fixedSlideWidth?: string | number;
```
mark-tate: I think this prop adds API surface for something already achievable through the existing CSS variable.
I'll push up an branch which shows how it can be done without a API change.

mark-tate: Pushed up a commit (and squashed your previous commits), commits can always be removed and your back to your version.

Embla does not have any API props for sizing, it's all managed through CSS and computed from the slide sizes.

mark-tate (dismissed):

site/docs/components/carousel/examples.mdx
```diff
@@ -47,6 +47,14 @@ The carousel can showcase multiple slides simultaneously, offering flexibility i
 
 <LivePreview componentName="carousel" exampleName="MultipleSlides" />
 
+## Fixed width slides
+
+The carousel can display slides at a constant width regardless of viewport size. Set the required Embla props, then set `--carousel-slide-size` to a value (e.g., `550px`) to prevent the slides from scaling with the viewport.
+
+When `--carousel-slide-size` is not overridden, the default responsive behavior is preserved.
```
joshwooding, line 54: Is this a new public API @mark-tate ?

mark-tate: It's not new, it was there before,  and was set to 
```
  --carousel-slide-size: 100%;
```
which created a responsive slide (the default)

this is just setting an example that sets the slide to a specific size to make things fixed size

Similarily multislide examples previously did

```
.carousel.carouselMultipleSlides {
  /** Allow 4% of the next slide to be visible by reducing each slide by 2% */
  --carousel-slide-size: 48%;
}
```

mark-tate (approved):

joshwooding (approved):

PR #6220, opened 2026-04-07
Refactor the Highcharts theme and add Heatmap support

tycup:
[ The following is an AI generated summary given the prompt: Write a github issue description that describes the changes in the highcharts-refactor branch which would be useful for a code reviewer looking for context] - generated by GPT-5.4 High.

## Context

This branch (`highcharts-refactor`) is a broad refactor of `@salt-ds/highcharts-theme` to move styling/theming behavior into the `useChart` hook and token-driven option builders, while adding first-class heatmap + color-axis support and cleaning up docs/examples.

Compared with `main`, this is **66 files changed** (`+3056 / -2357`) and includes API behavior changes that are important to review carefully.

### 1) Theming model moved from external CSS classes to hook-driven options + runtime CSS injection
- Removed packaged CSS bundle files and static entrypoint CSS (`index.css`, `css/*`, build script for CSS packaging).
- `useChart` now injects a small internal stylesheet with `useComponentCssInjection` and computes chart defaults from Salt tokens.
- Site and Storybook no longer globally import Highcharts CSS / Salt highcharts CSS wrappers.

### 2) `useChart` expanded with new config options
- New optional third arg:
  - `fillPatterns?: boolean`
  - `saltColorAxis?: SaltColorAxis`
- Existing direct `colorAxis` in `chartOptions` intentionally takes precedence over `saltColorAxis`.

### 3) Theme defaults refactored into composable builders
- `default-options.ts` now delegates to `theme-option-builders.ts`.
- New option merge behavior in `merge-chart-options.ts` to handle axis arrays correctly (workaround for Highcharts merge behavior with array axes).
- Token map greatly expanded (`density-token-map.ts`) from mostly numeric spacing values to full typography/color/dataviz token resolution.

### 4) Fill pattern implementation rewritten
- Removed legacy `saltPatternDef` export and old class-based pattern toggles.
- Added `fill-patterns.ts` and updated `patterns.ts` to generate pattern objects directly in options.
- Fill behavior now applied programmatically per series type (area/column/bubble/scatter/boxplot/waterfall/candlestick/line handling differs by type).

### 5) New color-axis system for heatmaps
- Added `color-axis.ts` with Salt-specific color-axis config types and resolvers:
  - continuous ramps
  - threshold/divergent ramps
  - data classes (including optional opacity ramp)
- Added heatmap examples + dependency option sets (continuous, threshold, discrete ranges, two-color ranges, data-class variants, explicit override).

### 6) Docs/examples/story updates
- Chart docs now describe module loading by Highcharts major version (v10/v11 vs v12+ auto-init).
- Usage docs switched from wrapper class approach to `useChart(..., { fillPatterns })` + `saltColorAxis`.
- Added dual-axis and heatmap examples to package stories and site docs.
- Version messaging updated to support `>=10 <13` (including v12).

### 7) Packaging/build and linting updates
- `@salt-ds/highcharts-theme` package metadata changed:
  - `sideEffects: false`
  - removed CSS `style/files` packaging fields
  - added `@salt-ds/styles` dependency
  - widened `highcharts` peer range to `>=10.0.0 <13`
- Root stylelint target for highcharts-theme moved from `packages/highcharts-theme/css/**/*.css` to `packages/highcharts-theme/src/**/*.css`.
- `css-inline-plugin` include/exclude globs expanded to support js/ts variants.
- Declarations updated for `*?inline` imports.

## Reviewer focus areas (highest risk)

1. **Runtime styling correctness across Highcharts versions**
- Confirm charts still render correctly in v10/v11/v12 with required module loading paths.
- Validate no regressions from removing global Highcharts CSS usage in docs/story environments.

2. **Option merge semantics**
- Validate axis array merge behavior for dual-axis and custom per-axis overrides.
- Verify that user-supplied chart options still override defaults predictably.

3. **Pattern/color behavior by series type**
- Check per-series fill/marker/stroke logic and hover behavior for edge cases (especially scatter/bubble/boxplot/waterfall/candlestick).

4. **Color-axis behavior**
- Verify continuous/threshold/data-class outputs and opacity ramps.
- Confirm precedence: explicit `chartOptions.colorAxis` over hook-level `saltColorAxis`.

## Complex / nuanced implementation details

A few parts of this refactor are more subtle than the high-level diff suggests and are worth extra reviewer attention:

- **Axis merging has custom logic because native Highcharts merging is insufficient.** `packages/highcharts-theme/src/merge-chart-options.ts` adds a dedicated merge path for `xAxis`/`yAxis`, especially when the consumer passes axis arrays. The code comments call out that Highcharts does not reliably merge object defaults into axis arrays, so defaults are normalized per axis entry before the final merge. This is particularly relevant for the new dual-axis support and for any chart with multiple axes.

- **Pattern handling is no longer just “turn on a CSS class”.** In `packages/highcharts-theme/src/fill-patterns.ts` and `packages/highcharts-theme/src/patterns.ts`, pattern behavior is now applied programmatically per series type. The implementation preserves explicit consumer overrides where possible and only fills in defaults when series/plot options have not already specified colors. Different chart types are treated differently:
  - line charts keep Salt dataviz colors and use dash patterns rather than fills
  - bubble/scatter synthesize marker fill + stroke behavior because Highcharts handles those inconsistently
  - boxplot/candlestick/waterfall receive custom fill/stroke logic rather than a generic pattern pass
  This means regressions, if any, are more likely to be chart-type-specific than global.

- **The new `saltColorAxis` API is additive, but only when Highcharts `colorAxis` is absent.** In `packages/highcharts-theme/src/useChart.ts`, hook-level `saltColorAxis` is only resolved when `chartOptions.colorAxis == null`. That precedence rule is important: explicit Highcharts `colorAxis` config wins. Reviewers should verify that this matches intended library behavior and is clearly documented for consumers.

- **`color-axis.ts` is doing more than simple token substitution.** `packages/highcharts-theme/src/color-axis.ts` resolves several Salt-specific config shapes into Highcharts-ready `colorAxis` objects:
  - continuous ramps
  - threshold/divergent ramps
  - data classes
  It also normalizes raw token names like `--salt-category-1-dataviz`, parses `var(...)` references, resolves them against live token values, and applies opacity via `Highcharts.color(...).setOpacity(...)`. The opacity ramp behavior for `dataClasses` is especially nuanced because it can distribute opacity across ordered classes unless a class explicitly sets its own `opacity`.

- **Heatmap support is not just a new example; it drove theme-option changes.** `packages/highcharts-theme/src/theme-option-builders.ts` now includes heatmap-specific behavior such as axis line/grid decisions, heatmap plot options, and stronger text outlines for data labels. The axis builders also special-case scatter/bubble charts to show both grid directions. Those conditional defaults are easy to miss because they are now centralized rather than spread across CSS.

- **Tooling had to change to support runtime CSS injection.** `packages/highcharts-theme/src/declarations.d.ts` and `tooling/css-inline-plugin/src/index.ts` add support for `?inline` CSS imports so styles can be imported as strings and injected by the hook rather than emitted as external CSS. This is an implementation detail, but it underpins the packaging/runtime behavior shift and is worth checking in Storybook/site builds.

## Suggested validation checklist

- [ ] Run package stories for all chart types with `fillPatterns` on/off.
- [ ] Validate dual-axis defaults and custom axis overrides.
- [ ] Validate heatmap variants (continuous, threshold, data classes, explicit colorAxis override).
- [ ] Smoke test docs examples in site (light/dark theme + density changes).
- [ ] Confirm no regressions for existing consumers using only `useChart(chartRef, options)` (without new config).
- [ ] Confirm migration guidance is adequate for any removed CSS/class-based usage.


packages/highcharts-theme/src/useChart.ts
```diff
@@ -1,26 +1,81 @@
-import { useDensity } from "@salt-ds/core";
+import { useDensity, useTheme } from "@salt-ds/core";
+import { useComponentCssInjection } from "@salt-ds/styles";
 import { useWindow } from "@salt-ds/window";
 import type { Options } from "highcharts";
 import Highcharts from "highcharts";
+import patternFillModule from "highcharts/modules/pattern-fill";
```
tycup: This needs to be moved or imported by the consumer themselves since v11 its auto initalized

packages/highcharts-theme/src/theme-option-builders.ts
```diff
@@ -0,0 +1,366 @@
...
+        },
+        inactive: {
+          opacity: 0.2,
+        },
+      },
+    },
+    scatter: {
+      marker: {
+        radius: 4,
+        lineWidth: 1,
```
tycup, line 310: use token here

packages/highcharts-theme/src/color-axis.ts
```diff
@@ -0,0 +1,353 @@
+import Highcharts, { type Options } from "highcharts";
+import type { SaltChartTokenMap } from "./density-token-map";
+
+const DEFAULT_MAX_OPACITY = 1;
+const DEFAULT_MIN_OPACITY = 0.12;
```
tycup, line 5: Subject to change pending design @amritadesmet

packages/highcharts-theme/src/examples/BarChart.tsx
```diff
@@ -1,35 +1,32 @@
...
+import patternFill from "highcharts/modules/pattern-fill";
 import HighchartsReact from "highcharts-react-official";
 import { type FC, useRef } from "react";
 import { barOptions } from "./dependencies";
 
+patternFill(Highcharts);
+
 export interface BarChartProps {
-  patterns?: boolean;
+  fillPatterns?: boolean;
```
tycup, line 11: renamed all this because it felt like it aligned better with the module you need to import but think this might be an unnecessary change(?)

packages/highcharts-theme/src/color-axis.ts
amritadesmet: Decided together on 0.15 for now, resolving thread

packages/highcharts-theme/src/declarations.d.ts
```diff
@@ -2,3 +2,8 @@ declare module "*.css" {
   const content: string;
   export default content;
 }
+
```
joshwooding: Are we actually using this?

joshwooding (commented):
Changeset?

packages/highcharts-theme/src/index.ts
```diff
@@ -1,2 +1,5 @@
-export { saltPatternDef } from "./patterns";
+export {
```
joshwooding: @tycup

packages/highcharts-theme/src/index.ts
```diff
@@ -1,2 +1,5 @@
...
+  heatmapOptions,
+  heatmapSaltColorAxis,
+} from "./examples/dependencies/heatmapOptions";
 export { useChart } from "./useChart";
```
joshwooding, line 2: Consider exporting SaltColorAxis type

packages/highcharts-theme/src/useChart.ts
```diff
@@ -1,26 +1,85 @@
...
+  saltColorAxis?: SaltColorAxis;
+}
 
 export const useChart = (
   chartRef: RefObject<HighchartsReact.RefObject | null>,
   chartOptions: Options,
+  { fillPatterns = false, saltColorAxis }: UseChartConfig = {},
 ) => {
   const density = useDensity();
+  const { mode } = useTheme();
```
joshwooding, line 30: add comment saying we won't support theme/style options for now

packages/highcharts-theme/src/useChart.ts
```diff
@@ -1,26 +1,85 @@
...
   const targetWindow = useWindow();
 
+  useComponentCssInjection({
+    testId: "salt-highcharts-theme",
+    css: highchartsThemeCss,
+    window: targetWindow,
+  });
+
   const hostElementRef = useRef<Element | null>(null);
```
joshwooding: We might not need this ref anymore

joshwooding (changes requested):

packages/highcharts-theme/src/useChart.ts
tycup: hostElementRef is defensive -  the chart container should normally be available when the effect runs, so could use `chartRef.current?.chart?.container ?? documentElement` directly. 

this ref only helps if the container was available once, then temporarily isnt during a later recalculation which lets the hook keep reading scoped Salt tokens from the last known chart element instead of falling back to root tokens

it verges on over defensive but id lean to leave it in for now

tycup: oh or do you mean as in, we could just const elementUsed = container ?? targetWindow?.document?.documentElement;

joshwooding: Yeah, we don't need it as a ref

joshwooding (dismissed):

joshwooding (commented):
Do we need a changeset for the heatmap?

joshwooding (dismissed):

joshwooding (approved):

mark-tate (approved):

Issue #3734, opened 2024-07-05
Create RC for Side Panel component (previously Inlaid Drawer) in Labs


origami-z:
Follow design in #3306

https://www.figma.com/design/RdkPXIvSRyzIVwqThtILE5/Side-panel?m=auto&t=LrmBXE5jp6Ete6tL-6

mark-tate:
Goal: RC by EOS

origami-z:
Pushing out a few sprints. Not priority from Pepper, but rather Drawer targeting part of page

mark-tate:
Moving due to reduced capacity, if you need this feature then please contact to help us define the priority

origami-z:
+1 from SO #92394

bhoppers2008:
Awaiting validation on how focus management is handled by accessibility. Proposed model from accessibility is unique–we should be sure about the solution before implementing.

mark-tate:
@wemmyo @navkaur76  can you review and  either tick or return for re-work

PR #6221, opened 2026-04-07
Add side panel 

mark-tate:
cc #3734


site/src/examples/side-panel/ContentExample.tsx
```diff
@@ -0,0 +1,42 @@
...
+      }}
+    >
+      {Array.from({ length: 6 }, (_, i) => (
+        <div
+          key={i}
+          style={{
+            backgroundColor: "var(--salt-container-secondary-background)",
+            borderRadius: "var(--salt-palette-corner-weak)",
+            border:
+              "var(--salt-size-fixed-100) dashed var(--salt-container-primary-borderColor)",
```
joshwooding: We should be using border style tokens but also not sure if we should be using dashed?

site/src/examples/side-panel/Scrollable.tsx
```diff
@@ -0,0 +1,133 @@
...
+            </H2>
+            <Button
+              aria-label="Close"
+              appearance="transparent"
+              onClick={() => setOpen(false)}
+            >
+              <CloseIcon aria-hidden />
+            </Button>
+          </FlexLayout>
+          <div
```
joshwooding: We're going to need to create a SidePanelContent component, it needs more for accessibility than this

joshwooding: It will also fix your styling issues

site/src/examples/side-panel/Variants.tsx
```diff
@@ -0,0 +1,102 @@
...
+          </FlexLayout>
+          <Text>Side panel content goes here.</Text>
+        </StackLayout>
+      </SidePanel>
+    </>
+  );
+};
+
+export const Variants = () => (
+  <SidePanelProvider>
```
joshwooding: I think this should be open by default

packages/lab/src/side-panel/SidePanel.css
```diff
@@ -0,0 +1,77 @@
+.saltSidePanel {
+  overflow: hidden;
+  background-color: var(--saltSidePanel-background, var(--salt-container-primary-background));
+  --saltSidePanel-width: 300px;
+  --saltSidePanel-border: var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-container-primary-borderColor);
```
joshwooding: The container border colors should change with variants and the token naming should be updated to make it "internal"

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,123 @@
...
+   * Edge the panel is anchored to; controls animation direction and divider side.
+   * @default "right"
+   */
+  position?: "right" | "left";
+  /**
+   * Which element receives focus when the panel opens.
+   * Pass a number for the tabbable element index (0 = first), or a ref to a specific element.
+   * @default 0
+   */
+  initialFocus?: number | MutableRefObject<HTMLElement | null>;
```
joshwooding: Could we re-use the types provided by floating-ui here?

packages/lab/src/side-panel/SidePanelProvider.tsx
```diff
@@ -0,0 +1,79 @@
...
+  });
+
+  const handleOpenChange = (newOpen: boolean) => {
+    setOpenState(newOpen);
+    onOpenChange?.(newOpen);
+  };
+
+  const [reference, setReference] = useState<HTMLElement | null>(null);
+  const [floating, setFloating] = useState<HTMLDivElement | null>(null);
+  const panelId = useId();
```
joshwooding: This can be removed now

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,123 @@
...
+      panelId,
+    } = useSidePanelContext();
+
+    useComponentCssInjection({
+      testId: "salt-side-panel",
+      css: sidePanelCss,
+      window: targetWindow,
+    });
+
+    const id = useId(idProp || panelId);
```
joshwooding: We can remove this too I think

packages/lab/src/side-panel/SidePanelTrigger.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const handleRef = useForkRef(setReference, ref);
+
+  if (!children || !isValidElement<{ ref?: unknown }>(children)) {
+    return <>{children}</>;
+  }
+
+  const mergedProps = mergeProps(
+    getReferenceProps({
+      "aria-controls": panelId,
+      onClick: () => setOpen(!openState),
```
joshwooding: onClick will get replaced by onClick being passed in

site/src/examples/side-panel/ContentExample.tsx
mark-tate: vibed placeholders. @navkaur76 to spec.

For now updated to 
```         
border: "var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-separable-tertiary-borderColor)",
```

site/src/examples/side-panel/Scrollable.tsx
mark-tate: if Design want a 80% use-case, drop in layout that creates a header, close action and content area, then we could do that.
It would serve as an a11y example, but not block others creating their own from scratch and owning a11y.

@navkaur76  @jake-costa  to spec what that would look like ?

site/src/examples/side-panel/Variants.tsx
mark-tate: done

packages/lab/src/side-panel/SidePanel.css
mark-tate: styles not changed in this PR - @navkaur76 to opine.

packages/lab/src/side-panel/SidePanel.tsx
mark-tate: done

packages/lab/src/side-panel/SidePanelProvider.tsx
mark-tate: yes... can use floating ui ids

packages/lab/src/side-panel/SidePanelTrigger.tsx
mark-tate: done

packages/lab/src/side-panel/SidePanel.tsx
mark-tate: yup

packages/lab/src/side-panel/SidePanel.css
mark-tate: @navkaur76  ping

mark-tate: updated example

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -40,13 +40,19 @@ export interface SidePanelProps extends ComponentPropsWithRef<"div"> {
    * @default "primary"
    */
   variant?: "primary" | "secondary" | "tertiary";
+  /**
+   * Change the panel's appearance.
```
mark-tate: Could polish this, see how document the other similar props in Button and ComboBox
```
The appearance of the button. Options are 'solid', 'bordered', and 'transparent'.
'solid' is the default value.
```

Could keep @default tbh but would improve the description.

wemmyo: Done!

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+- To view supporting details alongside the primary content (e.g., record details, metadata, history, related items) while keeping the main view visible and interactive.
+- To edit or update information in-context (e.g., update attributes, adjust settings, annotate) while referencing the main view for confirmation.
+- To apply and refine filters (e.g., facets, query parameters) while keeping results visible and interactive next to the filtering controls.
+- To support multi-step workflows (e.g., review → decide → update → confirm) where context and supporting controls must remain available across steps.
+- To keep supplementary information and actions persistently available during longer parallel tasks (e.g., compare items, validate fields, cross-check data).
+
+### When not to use
+
+For a temporary, focused workspace that lets users handle a small, self-contained action. Instead, use [Drawer](/salt/components/drawer).
```
navkaur76: Can the "When not to use" section be formatted to use bullet points please :)

navkaur76 (changes requested):

site/docs/components/side-panel/usage.mdx
mark-tate: as these are docs, you could make those changes ?

navkaur76 (dismissed):

site/docs/components/side-panel/usage.mdx
navkaur76: done!

tycup:
Is the consumer allowed to place things in `SidePanel` directly or do they have to use `SidePanelContent`?  I see cypress examples as such, but i suspect that having children directly inside the `SidePanel` prevents `Esc` key dismissal (unless the user uses SidePanelContent)

Should it be handled at the SidePanel/provider level?

packages/lab/src/side-panel/SidePanelContent.css
```diff
@@ -0,0 +1,25 @@
...
+
+.saltSidePanelContent-header {
+  display: flex;
+  align-items: center;
+  flex-shrink: 0;
+  gap: var(--salt-spacing-100);
+  padding-right: var(--salt-spacing-300);
+}
+
+.saltSidePanelContent-header > *:first-child {
```
tycup: Given that `header` is optional, could this cause unintended layouts?

wemmyo:
> Is the consumer allowed to place things in `SidePanel` directly or do they have to use `SidePanelContent`? I see cypress examples as such, but i suspect that having children directly inside the `SidePanel` prevents `Esc` key dismissal (unless the user uses SidePanelContent)
> 
> Should it be handled at the SidePanel/provider level?

We've made some changes to how the side panel handles the Esc key so this should be resolved now. But valid feedback, thanks

packages/lab/src/side-panel/SidePanelContent.css
wemmyo: header also renders dynamically so this should only apply when header is provided

packages/lab/src/side-panel/SidePanelContent.tsx
```diff
@@ -0,0 +1,128 @@
...
+import sidePanelContentCss from "./SidePanelContent.css";
+import { useSidePanelContext } from "./SidePanelContext";
+
+const withBaseName = makePrefixer("saltSidePanelContent");
+
+export interface SidePanelContentProps extends ComponentPropsWithRef<"div"> {
+  /**
+   * Content rendered in the header area next to the close button.
+   */
+  header?: ReactNode;
```
wemmyo: No changes for now, but I think we should remove header prop at some point and instead have a separate header component with a close button component and title component that provides the accessible name out of the box

site/docs/components/side-panel/examples.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+## Left side panel
+
+Consider using a left‑anchored side panel for structure and exploration—such as hierarchical filters, folders, outlines, or multi‑facet filtering.
+
+<LivePreview componentName="side-panel" exampleName="LeftPanel" />
+
+## Panel variants
+
+By default the side panel styling uses `variant=”primary”`. Set `variant=”secondary”` for a secondary side panel and `variant=”tertiary”` for a tertiary side panel.
```
navkaur76: Is it worth adding a minor note suggesting the purpose of the 'none' variant? E.g. something along the lines of "Use variant="none" when the side panel may inherit styling from its surrounding layout." 

Just wary that it's listed as a variant but not mentioned in the description. LMK your thoughts @wemmyo @mark-tate

site/docs/components/side-panel/examples.mdx
```diff
@@ -0,0 +1,68 @@
...
+- Place the trigger and the panel in different layout areas.
+- Manage the panel’s open state with application state.
+- Handle multiple panels with custom rules.
+- Integrate with other components that control the panel state.
+
+This approach trades convenience for flexibility: it requires more wiring, but enables more control over placement and behavior.
+
+<LivePreview componentName="side-panel" exampleName="ManualTrigger" />
+
+## Side panels between content
```
navkaur76: Would suggest renaming heading to "Side panels between content areas" to match how it's referenced in the body text?

site/docs/components/side-panel/examples.mdx
```diff
@@ -0,0 +1,68 @@
...
+
+Side Panels can sit between content areas to create a multi-panel layout. When using multiple panels, ensure that the layout remains clear and that users can easily distinguish between different panels and their purposes.
+
+<LivePreview componentName="side-panel" exampleName="WithNav" />
+
+Side Panels can also be nested within each other to create multi-level drill-in experiences. When nesting, ensure that each panel has a clear purpose and that the navigation between panels is intuitive.
```
navkaur76: this nesting description feels like it could be it's own example/section, supported by a visual? Is there risk of users interpreting “nesting” as either multiple adjacent panels or an overlay pattern? Could perhaps benefit from an example showing Panel A with Panel B opened inside it, to clarify intended behaviour

wemmyo: We have an example on storybook but the opening animation is a bit buggy so maybe we remove any mention of 'nested side panel' in the site docs for now

navkaur76: I agree with removing it for now, it feels a bit random without a supporting example.

site/docs/components/side-panel/examples.mdx
wemmyo: Done!

site/docs/components/side-panel/examples.mdx
wemmyo: Removed!

site/docs/components/side-panel/examples.mdx
wemmyo: For now, I've added that 'none' removes default styling

wemmyo (dismissed):

tycup (dismissed):

packages/lab/src/__tests__/__e2e__/side-panel/SidePanel.cy.tsx
```diff
@@ -0,0 +1,623 @@
...
+                "have.attr",
+                "aria-controls",
+                panelId,
+              );
+            });
+
+          cy.findByRole("region", { name: "Section Title" }).should(
+            "be.visible",
+          );
+          cy.findByRole("region").should("have.class", "saltSidePanel-left");
```
joshwooding: Nit: We should avoid testing for classes in tests

packages/lab/src/__tests__/__e2e__/side-panel/SidePanel.cy.tsx
```diff
@@ -0,0 +1,623 @@
...
+  Left,
+  Default,
+  ManualTrigger,
+  Variants,
+  WithTable,
+  Nested,
+  Scrollable,
+  WithNav,
+} = composedStories;
+describe("GIVEN a SidePanel component", () => {
```
joshwooding, line 16: Another Nit but it can be updated later, we should avoid deeply nested tests, it makes them harder to read.

packages/lab/src/__tests__/__e2e__/side-panel/SidePanel.cy.tsx
```diff
@@ -0,0 +1,623 @@
...
+          cy.realPress("Escape");
+
+          cy.get("@onOpenChange").should("have.been.calledWith", false);
+        });
+      });
+    });
+  });
+
+  describe("Variants", () => {
+    describe("GIVEN the Variants story", () => {
```
joshwooding: We shouldn't really be writing tests based on stories and more functionality

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,151 @@
...
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Using the component
+
+A side panel spans the full height of its container and is always non-modal. This means the content beside it remains interactive when the panel is expanded.
+
+Always choose a width that suits the content in the side panel. Refer to Salt's Forms pattern for information on logical forms width, and the [typography foundation](/salt/foundations/typography) for guidance on paragraph text width.
```
joshwooding: I think we're missing a link to the form pattern here?

joshwooding: We also don't need to clarify it's Salt's form pattern

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+### When to use
+
+- To view supporting details alongside the primary content (e.g., record details, metadata, history, related items) while keeping the main view visible and interactive.
+- To edit or update information in-context (e.g., update attributes, adjust settings, annotate) while referencing the main view for confirmation.
+- To apply and refine filters (e.g., facets, query parameters) while keeping results visible and interactive next to the filtering controls.
+- To support multi-step workflows (e.g., review → decide → update → confirm) where context and supporting controls must remain available across steps.
```
joshwooding: Not sure if we should be using unicode arrows here. @jake-costa ?

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,151 @@
...
+- To keep supplementary information and actions persistently available during longer parallel tasks (e.g., compare items, validate fields, cross-check data).
+
+### When not to use
+
+- For a temporary, focused workspace that lets users handle a small, self-contained action. Instead, use [Drawer](/salt/components/drawer).
+- For a background that organizes content in an application. Instead, use [Panel](/salt/components/panel).
+
+## Import
+
+To import `SidePanel`, `SidePanelContent`, `SidePanelProvider` and `SidePanelTrigger` from the lab Salt package, use:
```
joshwooding: I don't think we typically list all of the components here. Usually we just have a few and then say related components

joshwooding:
Leaving some general feedback but it doesn't need to be addressed

site/src/examples/side-panel/ContentExample.tsx
```diff
@@ -0,0 +1,41 @@
...
+    >
+      {Array.from({ length: 6 }, (_, i) => (
+        <div
+          // biome-ignore lint/suspicious/noArrayIndexKey: In this case, using index as key is acceptable
+          key={`content-${i}`}
+          style={{
+            backgroundColor: "var(--salt-container-secondary-background)",
+            borderRadius: "var(--salt-palette-corner-weak)",
+            border:
+              "var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-separable-tertiary-borderColor)",
```
joshwooding: We shouldn't be using separable tokens for full borders

packages/lab/src/side-panel/SidePanel.css
```diff
@@ -0,0 +1,91 @@
+.saltSidePanel {
+  background-color: var(--saltSidePanel-background, var(--salt-container-primary-background));
+  --saltSidePanel-width: 300px;
+  --saltSidePanel-internal-border: var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--saltSidePanel-borderColor, var(--salt-container-primary-borderColor));
+}
+
+.saltSidePanel-primary {
+  --saltSidePanel-background: var(--salt-container-primary-background);
```
joshwooding: We've not really used this casing for variables to declare internal variables they have always previously been lowercase without the salt prefix

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+const withBaseName = makePrefixer("saltSidePanel");
+
+export interface SidePanelProps extends ComponentPropsWithRef<"div"> {
+  /**
+   * Whether the panel animates its own open/close transitions.
+   * Set to `false` when the parent controls sizing and animation (e.g. inside a splitter).
+   * @default true
+   */
+  animated?: boolean;
```
joshwooding: This naming seems inconsistent, we've always said disableAnimation or something similar

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,205 @@
...
+      window: targetWindow,
+    });
+
+    const { context } = useFloatingUI({
+      rootContext: floatingRootContext,
+    });
+
+    const handleRef = useForkRef<HTMLDivElement>(setFloating, ref);
+
+    const handleAnimationEnd = useCallback(() => {
```
joshwooding: useCallback isn't needed here

packages/lab/src/side-panel/SidePanelContent.tsx
joshwooding: I'm curious why we didn't build it this way initially too

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,151 @@
...
+    <SidePanel position="right" aria-labelledby={headingId}>
+      <SidePanelContent header={<H2 id={headingId}>Details</H2>}>
+        <Text>Side panel content goes here.</Text>
+      </SidePanelContent>
+    </SidePanel>
+  </FlexLayout>
+</SidePanelProvider>;
+```
+
+### Custom triggers
```
bhoppers2008: What is the difference between this section and the Manual Trigger example? If they're saying the same thing, let's align on language and merge into the example.

packages/lab/src/side-panel/SidePanelContent.tsx
mark-tate: we can do it now, it's a simple change.

packages/lab/src/side-panel/SidePanelTrigger.tsx
```diff
@@ -0,0 +1,56 @@
...
+  HTMLButtonElement,
+  SidePanelTriggerProps
+>(function SidePanelTrigger(props, ref) {
+  const { children, onClick, ...rest } = props;
+  const { setReference, getReferenceProps, openState, setOpen, panelId } =
+    useSidePanelContext();
+
+  const handleRef = useForkRef(setReference, ref);
+
+  const handleClick = useCallback(
```
joshwooding: I don't think we need this useCallback

packages/lab/src/side-panel/SidePanelTrigger.tsx
```diff
@@ -0,0 +1,56 @@
...
+  isValidElement,
+  type MouseEvent,
+  type ReactNode,
+  useCallback,
+} from "react";
+import { useSidePanelContext } from "./SidePanelContext";
+
+export interface SidePanelTriggerProps
+  extends ComponentPropsWithoutRef<"button"> {
+  children: ReactNode;
```
joshwooding: Not sure if we need to make this a required prop

site/docs/components/side-panel/index.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Side panel
+data:
+  description: "A side panel is a persistent, side-by-side workspace that keeps supporting information and controls available while users continue their main task. It’s intended for ongoing parallel work—such as referencing details, inspecting data, filtering results, or editing attributes—where maintaining visibility and context improves accuracy and efficiency."
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/tree/main/packages/lab/src/side-panel"
+  package:
+    name: "@salt-ds/lab"
+    initialVersion: ""
+  alsoKnownAs: ["Flyout", "Utility panel", "Details panel"]
```
joshwooding: I'm not sure if a flyout is a good alias?

site/docs/components/side-panel/usage.mdx
wemmyo: I've updated the language to be consistent using 'manual trigger'. The usage docs focuses on the code for just the trigger which is more specific and technical than in the examples where it gives a more general overview. Probably fine to have both

packages/lab/src/side-panel/SidePanelContent.tsx
wemmyo: This has been updated

packages/lab/src/side-panel/SidePanel.css
wemmyo: Renamed the internal variables. I've left the external width css variable as is, do let me know if there are any recommendations for that

packages/lab/src/side-panel/SidePanel.tsx
wemmyo: Updated

site/src/examples/side-panel/ContentExample.tsx
wemmyo: Updated

packages/lab/src/side-panel/SidePanelContent.tsx
```diff
@@ -0,0 +1,108 @@
...
+      bodyAriaLabelledBy = explicitLabelledBy;
+    } else if (headerId) {
+      bodyAriaLabelledBy = `${headerId} ${contentSuffixId}`;
+    } else {
+      bodyAriaLabel = explicitLabel ?? "Content";
+    }
+  }
+
+  return (
+    <div ref={ref} className={clsx(withBaseName(), className)} {...rest}>
```
joshwooding, line 96: I'm pretty sure if users pass an aria-label or aria-labelledby in it will still be applied here?

packages/lab/src/side-panel/SidePanelProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+    open: openState,
+    onOpenChange: handleOpenChange,
+    elements: {
+      reference,
+      floating,
+    },
+  });
+
+  const { getReferenceProps, getFloatingProps } = useInteractions([
+    useDismiss(floatingRootContext, {
```
joshwooding: I don't think useDismiss is doing anything with everything turned off?

packages/lab/src/side-panel/SidePanelProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+        activeElement != null &&
+        (floating?.contains(activeElement) ||
+          reference?.contains(activeElement));
+
+      if (!focusOwnedByPanel) {
+        return;
+      }
+
+      event.preventDefault();
+      event.stopPropagation();
```
joshwooding, line 74: Do we still need this? If the active element is in the panel do we still need to stop propagation

packages/lab/src/side-panel/SidePanelTrigger.tsx
```diff
@@ -0,0 +1,52 @@
...
+>(function SidePanelTrigger(props, ref) {
+  const { children, onClick, ...rest } = props;
+  const { setReference, getReferenceProps, openState, setOpen, panelId } =
+    useSidePanelContext();
+
+  const handleRef = useForkRef(setReference, ref);
+
+  const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+    onClick?.(event);
+    if (!event.defaultPrevented) {
```
joshwooding: We don't usually do this, is there a reason why we need it?

site/src/examples/side-panel/LeftPanel.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <SidePanelProvider>
+      <LeftPanelContent />
+    </SidePanelProvider>
+  );
+};
+
+const LeftPanelContent = () => {
+  const { CloseIcon } = useIcon();
+  const { openState, setOpen } = useSidePanelContext();
```
joshwooding: I'm not a big fan of exporting hooks but I also don't think we should export everything in the context file. Can we create a hook like useSidePanel that is the public version of this please. Adding this can make internals a public API.

wemmyo: Most of these need to be exposed for the Manual trigger use-case where users have complex layouts and will need to handle focus order and aria labels themselves

wemmyo: Let me give it a go and see what needs to be public

wemmyo: [Separate internal and public context](https://github.com/jpmorganchase/salt-ds/pull/6221/commits/cccdc9a5d55e1787b9f9c504cce5f3d1ead6ef78)
@joshwooding @mark-tate let me know what you think

packages/lab/src/side-panel/SidePanelProvider.tsx
wemmyo: I agree, removed!

packages/lab/src/side-panel/SidePanelTrigger.tsx
```diff
@@ -0,0 +1,52 @@
...
+      "aria-controls": openState ? panelId : undefined,
+      ...rest,
+      onClick: handleClick,
+    }) as Record<string, unknown>,
+    children.props,
+  );
+
+  return cloneElement(children, {
+    ...mergedProps,
+    ref: handleRef,
```
joshwooding, line 49: ref's already applied on children are being overwritten

packages/lab/src/side-panel/SidePanel.tsx
```diff
@@ -0,0 +1,204 @@
...
+    });
+
+    const { context } = useFloatingUI({
+      rootContext: floatingRootContext,
+    });
+
+    const handleRef = useForkRef<HTMLDivElement>(setFloating, ref);
+
+    const handleAnimationEnd = () => {
+      setAnimating(false);
```
joshwooding, line 100: This will be called for any nested animations too, we should guard this so it only runs the logic with the panel animation e.g. event.currentTarget === event.target

site/docs/components/side-panel/usage.mdx
```diff
@@ -0,0 +1,204 @@
...
+
+If you need a custom layout or close behavior, you can skip `SidePanelContent` and compose your own body using `useSidePanel`.
+
+### Putting it together
+
+A minimal side panel combines all parts inside a flex container:
+
+```jsx
+const { CloseIcon } = useIcon();
+const { setOpen } = useSidePanel();
```
joshwooding, line 127: This is outside the provider

packages/lab/src/side-panel/SidePanelContent.tsx
mark-tate: yes, fixed in the follow up commit

packages/lab/src/side-panel/SidePanelProvider.tsx
mark-tate: this code was added to support nested panels but I can see the evolution of this code, has broken it..
Fixed in follow up

packages/lab/src/side-panel/SidePanelProvider.tsx
mark-tate: now we have removed this, we can simplify the API
Fixed in follow up PR

packages/lab/src/side-panel/SidePanelTrigger.tsx
mark-tate: correct, if a user wanted to prevent opening/closing they make it controlled, they do not need to go so low level as to use events

mark-tate: fixed in follow up

site/src/examples/side-panel/LeftPanel.tsx
mark-tate: with the other changes, we can simplfy the API, refer to follow up

packages/lab/src/side-panel/SidePanelTrigger.tsx
mark-tate: good spot,  picked this up in the follow up

site/docs/components/side-panel/usage.mdx
mark-tate: folks like to copy and paste examples so we should make all our examples  valid code.

Fixed in follow up

wemmyo (dismissed):

joshwooding (approved):

wemmyo (approved):

joshwooding (approved):

PR #6230, opened 2026-04-08
Mega Menu 

### LINKED ISSUE #2405: Create RC for Mega Menu component in Labs
## Description
The navigation item should include the option of a mega-menu drop-down. Design produced in #3260.

## Tasks

```[tasklist]
### Checklist
- [ ] Density (HD, MD, LD, TD)
- [ ] Mode (Light, Dark)
- [ ] Fill (Primary, Secondary, Tertiary)
- [ ] State (Default, Hover, Selected/Active, Disabled, Read-only, Other)
- [ ] Controlled API
- [ ] UITK accessibility parity (screen reader + keyboard interactions)
```
```[tasklist]
### Deliverables
- [ ] Characteristics spec
- [ ] Interactions spec, where necessary
- [ ] In Code & Figma (Component + Guide)
- [ ] Supporting site documentation
```

ThusharaJ07:
Adding Mega Menu


packages/lab/src/mega-menu/MegaMenuContainer.tsx
```diff
@@ -0,0 +1,301 @@
...
+          });
+        });
+      }
+    },
+    [megaMenu],
+  );
+
+  const handleKeyDown = useCallback(
+    (event: KeyboardEvent<HTMLElement>) => {
+      const isArrowUp = event.key === "ArrowUp";
```
joshwooding, line 127: Just curious did you test relying on floating-ui for the keyboard navigation?

ThusharaJ07: So i had checked. Basically it cannot handle complex navigation between multiple region(like the up/down & right/left arrow key we have).

packages/lab/src/mega-menu/MegaMenu.tsx
```diff
@@ -0,0 +1,129 @@
...
+  children,
+  open,
+  onOpenChange,
+  selectedItem: selectedItemProp,
+  defaultSelectedItem,
+  onSelectedItemChange,
+  placement = "bottom",
+  interactions,
+}: MegaMenuProps) => {
+  const [openState, setOpenState] = useControlled({
```
mark-tate, line 45: we need to provide defaultOpen as well, as per other component APIs.

packages/lab/src/mega-menu/MegaMenu.tsx
```diff
@@ -0,0 +1,129 @@
...
+      getFloatingProps,
+      getReferenceProps,
+      setFloating,
+      setReference,
+      setOpen,
+      selectedItem,
+      setSelectedItem,
+    }),
+    [
+      openState,
```
mark-tate, line 98: imagine you get a biome warning on dependencies here

can leave until the final merge into the main branch where others will see the warnings.

> yarn biome ci

packages/lab/src/mega-menu/MegaMenuContainer.tsx
```diff
@@ -0,0 +1,293 @@
...
+  }, [megaMenu]);
+
+  const handleFocus = useCallback((event: FocusEvent<HTMLElement>) => {
+    const index = elementsRef.current.indexOf(event.target as HTMLElement);
+    if (index !== -1) {
+      setActiveIndex(index);
+    }
+  }, []);
+
+  const closeAndFocusNextAfterReference = useCallback(
```
mark-tate: this container has lot of different responsibilities, I wonder if you can extract out the focus behaviour and turn it into custom middleware/interaction, to simplify this file's responsibilities?

packages/core/src/navigation-item/NavigationItem.tsx
```diff
@@ -50,6 +50,10 @@ export interface NavigationItemProps extends ComponentPropsWithoutRef<"div"> {
    * Href to be passed to the Link element.
    */
   href?: string;
+  /**
+   * ID of the element controlled by this navigation item.
+   */
+  "aria-controls"?: string;
```
joshwooding: aria-controls is already supported, we don't need to add the type

.changeset/add-navigation-item-aria-controls.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
```
joshwooding: This is more a bug fix around the prop being applied to the wrong element, can we update it to that please

packages/lab/src/mega-menu/MegaMenu.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const floatingRootContext = useFloatingRootContext({
+    open: openState,
+    onOpenChange: setOpen,
+    elements: { reference, floating },
+  });
+
+  const megaMenuKeyboard = useMegaMenuKeyboard(floatingRootContext);
+
+  const defaultInteractions = [
+    useClick(floatingRootContext, { toggle: true }),
```
joshwooding: toggle is true by default?

packages/lab/src/mega-menu/MegaMenu.tsx
```diff
@@ -0,0 +1,128 @@
...
+    useDismiss(floatingRootContext, { bubbles: true }),
```
joshwooding: Just curious why do we need bubbles?

packages/lab/src/mega-menu/MegaMenuContext.ts
```diff
@@ -0,0 +1,52 @@
...
+  undefined,
+);
+
+export interface MegaMenuCustomInteractions {
+  /**
+   * Override the default floating-ui interaction hooks.
+   * Receives the floating root context and should return an array of interaction props.
+   * When provided, replaces the default `useClick` and `useDismiss` interactions.
+   */
+  interactions?: (context: FloatingRootContext) => Array<ElementProps>;
```
joshwooding: We should simplify and remove this for now

packages/lab/src/mega-menu/MegaMenuCustomRegion.css
```diff
@@ -0,0 +1,14 @@
+.saltMegaMenuCustomRegion {
+  padding: var(--salt-spacing-300);
+}
+.saltMegaMenuCustomRegion-primary {
+  background: var(--saltMegaMenuCustomRegion-background, var(--salt-container-primary-background));
```
joshwooding: This would be cleaner if we set the background once and set a variable in the classes that gets applied

packages/lab/src/mega-menu/MegaMenuHeader.tsx
```diff
@@ -0,0 +1,38 @@
+import { Divider, makePrefixer } from "@salt-ds/core";
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { clsx } from "clsx";
+import { forwardRef, type HTMLAttributes, type ReactNode } from "react";
+import megaMenuHeaderCss from "./MegaMenuHeader.css";
+
+const withBaseName = makePrefixer("saltMegaMenuHeader");
+
+export interface MegaMenuHeaderProps extends HTMLAttributes<HTMLDivElement> {
```
joshwooding: Can we use ComponentPropsWithoutRef instead


```suggestion
export interface MegaMenuHeaderProps extends ComponentPropsWithoutRef<"div"> {
```

packages/lab/src/mega-menu/MegaMenuHeader.tsx
```diff
@@ -0,0 +1,38 @@
...
+    useComponentCssInjection({
+      testId: "salt-mega-menu-header",
+      css: megaMenuHeaderCss,
+      window: targetWindow,
+    });
+
+    return (
+      <div className={clsx(withBaseName(), className)} ref={ref} {...rest}>
+        <div className={clsx(withBaseName("content"))}>{children}</div>
+        <Divider variant="tertiary" aria-hidden />
```
joshwooding: Would it be simpler to just use a bottom border?

packages/lab/src/mega-menu/MegaMenuGroup.tsx
```diff
@@ -0,0 +1,67 @@
...
+      testId: "salt-mega-menu-group",
+      css: megaMenuGroupCss,
+      window: targetWindow,
+    });
+
+    const headerId = useId();
+    let header: ReactNode = null;
+    const items: ReactNode[] = [];
+
+    Children.forEach(children, (child) => {
```
joshwooding, line 42: Not a bit fan of doing this type of logic, is there a way we can avoid messing with passed in children?

packages/lab/src/mega-menu/MegaMenuHeader.tsx
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    );
+  },
+);
+
+MegaMenuHeader.displayName = "MegaMenuHeader";
```
joshwooding: The display name should already be set? Did we need to do this explicitly?

packages/lab/src/mega-menu/MegaMenuCustomRegion.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { clsx } from "clsx";
+import { forwardRef, type HTMLAttributes, type ReactNode } from "react";
+import megaMenuCustomRegionCss from "./MegaMenuCustomRegion.css";
+
+const withBaseName = makePrefixer("saltMegaMenuCustomRegion");
+
+export interface MegaMenuCustomRegionProps
+  extends HTMLAttributes<HTMLDivElement> {
```
joshwooding: Same type change would be needed here

packages/lab/src/mega-menu/MegaMenuCustomRegion.tsx
```diff
@@ -0,0 +1,50 @@
...
+    testId: "salt-mega-menu-custom-region",
+    css: megaMenuCustomRegionCss,
+    window: targetWindow,
+  });
+
+  return (
+    <div
+      className={clsx(
+        withBaseName(),
+        variant && withBaseName(variant),
```
joshwooding: Does this align with design? Not having a default variant?

packages/lab/src/mega-menu/MegaMenuCustomRegion.tsx
```diff
@@ -0,0 +1,50 @@
...
+        className,
+      )}
+      data-mega-menu-column=""
+      ref={ref}
+      {...rest}
+    >
+      {children}
+    </div>
```
joshwooding:
```suggestion
/>
```

We could simplify this and not destruct children

packages/lab/src/mega-menu/MegaMenuHeader.css
```diff
@@ -0,0 +1,15 @@
+.saltMegaMenuHeader-content {
+  display: flex;
+  align-items: center;
+  padding: var(--salt-spacing-50) var(--salt-spacing-100);
+  min-height: var(--salt-size-base);
+  font-family: var(--salt-text-label-fontFamily);
+  font-size: var(--salt-text-label-fontSize);
+  font-weight: var(--salt-text-fontWeight-strong);
+  line-height: var(--salt-text-label-lineHeight);
+  min-height: var(--salt-size-base);
```
joshwooding, line 9: min-height is present twice

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+export interface MegaMenuItemProps extends HTMLAttributes<HTMLLIElement> {
+  /**
+   * The content of the mega menu item.
+   */
+  children?: ReactNode;
+  /**
+   * Whether selecting this item closes the mega menu.
+   * @default true
+   */
+  closeOnSelect?: boolean;
```
joshwooding: I don't think we need this, the default should either be yes and they can handle it via a controlled component or something else

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const targetWindow = useWindow();
+    const megaMenu = useMegaMenu();
+
+    useComponentCssInjection({
+      testId: "salt-mega-menu-item",
+      css: megaMenuItemCss,
+      window: targetWindow,
+    });
+
+    const baseProps = {
```
joshwooding: Why not apply them directly?

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+      className: clsx(withBaseName(), className),
+      tabIndex: 0,
```
joshwooding: links don't need tab indexes

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  type MouseEvent,
+  type ReactNode,
+} from "react";
+import megaMenuItemCss from "./MegaMenuItem.css";
+import { useMegaMenu } from "./useMegaMenu";
+
+const withBaseName = makePrefixer("saltMegaMenuItem");
+
+export interface MegaMenuItemProps extends HTMLAttributes<HTMLLIElement> {
+  /**
```
joshwooding, line 25: Can we add a render prop please

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    const baseProps = {
+      className: clsx(withBaseName(), className),
+      tabIndex: 0,
+      "data-mega-menu-item": "",
+    };
+
+    const handleClick = (event: MouseEvent<HTMLLIElement>) => {
+      rest.onClick?.(event);
```
joshwooding: Can we destructure this, so the order of the props being applied is less important

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+      if (closeOnSelect) {
+        megaMenu.setOpen(false);
+      }
+    };
+
+    const handleKeyDown = (event: KeyboardEvent<HTMLLIElement>) => {
+      rest.onKeyDown?.(event);
+      if (event.key === " " || event.key === "Enter") {
```
joshwooding: Handling space and enter here seems strange, plus can't we just rely on the native link behaviour?

packages/lab/src/mega-menu/MegaMenuItemContent.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export interface MegaMenuItemContentProps
+  extends ComponentPropsWithoutRef<"div"> {
+  /**
+   * The content of Mega Menu Item.
+   */
+  children?: ReactNode;
+}
+
+export const MegaMenuItemContent = forwardRef<
```
joshwooding, line 22: What benefit does this component provide?

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -0,0 +1,14 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  inline-size: var(--saltMegaMenuPanel-inlineSize, fit-content);
+  max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth, calc(100vw - var(--salt-spacing-600))));
+  max-height: 90%;
+  overflow-x: hidden;
+  overflow-y: auto;
+  box-sizing: border-box;
+  border-radius: var(--salt-palette-corner);
+  border: var(--salt-size-fixed-100) solid var(--salt-container-primary-borderColor);
```
joshwooding: we should be using --salt-borderStyle-solid here

packages/lab/src/mega-menu/MegaMenuPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const {
+      openState: isOpen,
+      floatingRootContext,
+      placement,
+      getFloatingProps,
+      setFloating,
+      focusFirstItemOnOpen,
+      setPanelId,
+    } = megaMenu;
```
joshwooding: we should just destructure this immediately

packages/lab/src/mega-menu/MegaMenuPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const id = useId(rest.id);
```
joshwooding: since id isn't destructured it would overwrite it when rest is spread below which is confusing

packages/lab/src/mega-menu/MegaMenuPanel.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    const floatingProps = getFloatingProps();
+
+    // floating-ui adds aria-orientation for listbox-style widgets,
+    // but mega menus are not listboxes — strip it to avoid confusing AT.
+    const {
+      "aria-orientation": _ariaOrientation,
+      style: _style,
+      ...interactionProps
+    } = floatingProps;
```
joshwooding: Just curious why does floating-ui think it's a listbox style widget?

packages/lab/src/mega-menu/MegaMenuItemContent.tsx
ThusharaJ07: this was added to replicate the padding: var(--salt-spacing-75) 0 we have applied on the text content so that icon aligns to the start.

packages/lab/src/mega-menu/MegaMenu.tsx
ThusharaJ07: was not being used. stale code.

packages/lab/src/mega-menu/MegaMenuCustomRegion.tsx
ThusharaJ07: checking with design on this.

ThusharaJ07: Thanks for catching this Josh,
Design requirement is: Users can configure the Custom Region styling however they want. In our example, it’s helpful to add a background color like this to clearly demonstrate customized content.As a default, the Custom Region doesn’t have its own styling, so it inherits the container’s background color. If the entire mega menu container is set to the secondary background color, the Custom Region will also appear with the secondary background by default.
So in code i will remove the variant implementation and put the bg colours at example levels.
CC: @honey-chang

packages/lab/src/mega-menu/MegaMenuItem.tsx
ThusharaJ07: space is not handled by the link. i have asked this question to design if we want to have space to select the menuItem.

packages/lab/src/mega-menu/MegaMenuPanel.tsx
ThusharaJ07: Stale comment.

packages/lab/src/mega-menu/MegaMenuItem.tsx
ThusharaJ07: Design said it should react same as [Salt menu item.](https://www.saltdesignsystem.com/salt/components/menu/examples) By clicking space user should able to select the item. 
CC: @honey-chang

packages/lab/src/mega-menu/MegaMenuGroup.tsx
ThusharaJ07: How about i remove the MegaMenuHeader component and pass this as label prop in MegaMenuGroup? like we are doing for MenuGroup. @joshwooding

joshwooding (approved):
This is a good start

dplsek (approved):

PR #6249, opened 2026-04-10
Fix useEventCallback

joshwooding:
useEventCallback was re-assigning the value to the ref every render.

### DIFF
--- a/packages/core/src/utils/useEventCallback.ts
+++ b/packages/core/src/utils/useEventCallback.ts
@@ -4,13 +4,20 @@ import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
 /**
  * https://github.com/facebook/react/issues/14099#issuecomment-440013892
  */
-export function useEventCallback<Args extends unknown[], Return>(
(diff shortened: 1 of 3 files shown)


packages/core/src/utils/useEventCallback.ts
```diff
@@ -4,13 +4,20 @@ import { useIsomorphicLayoutEffect } from "./useIsomorphicLayoutEffect";
 /**
  * https://github.com/facebook/react/issues/14099#issuecomment-440013892
  */
-export function useEventCallback<Args extends unknown[], Return>(
-  fn: (...args: Args) => Return,
-): (...args: Args) => Return {
-  const ref = useRef(fn);
+export function useEventCallback<const T extends (...args: any[]) => void>(
```
tycup, line 7: void feels a bit misleading but i guess with ts its basically saying that the caller promises to 'ignore' the return value

(but i see why it still works etc, and not really a problem)

tycup (approved):

wemmyo (approved):

PR #6358, opened 2026-04-22
Add Toolbar

tycup:
Add ToolbarNext

See ARCHITECTURE.md for notes on implementation; temporary file just to support review and getting up to speed


aimee-liou:
VQA complete - notes can be found in Toolbar Figma.

tycup:
Some open questions -
- how do we feel about horizontal scrolling in the overflow menu?

accessibility - @jake-costa let's connect but here are the questions alongside a general accessibility review
- Check if ok to remove Home and End (because what if controls that also internally use this are focused? e.g. tab?). Mentioned as optional in APG. Adds complexity.
- Check if ok to not include Up Arrow and Down Arrow can duplicate Left Arrow and Right Arrow. Adds complexity with components like Dropdown that use this (also noted in APG: “…or can be reserved for operating controls”.
- How do we want to handle: “Typically, disabled elements are not focusable when navigating with a keyboard. However, in circumstances where discoverability of a function is crucial, it may be helpful if disabled controls are focusable so screen reader users are more likely to be aware of their presence. For additional guidance, see [Focusability of disabled controls](https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_disabled_controls).”
- Toolbar Overflow Menu roles ARIA -  only allows aria-haspopup to name menu, listbox, tree, grid, or dialog; there is no toolbar value. Since the panel is neither a menu nor a dialog today, the truthful accessible contract is a perhaps a button with aria-expanded/aria-controls that opens a labelled toolbar. Converting to a menu would require changing child roles, keyboard interaction, close-on-activation behavior, and the public content expectations for overflowed tooltrays.

packages/lab/src/toolbar-next/useToolbarNextOverflow.ts
```diff
@@ -0,0 +1,1028 @@
...
+
+      continue;
+    }
+
+    groupMap.set(item.overflowGroupKey, {
+      id: `${item.overflowGroupKey}-${sequence}`,
+      key: item.overflowGroupKey,
+      label:
+        item.overflowGroup === "shared"
+          ? "More"
```
tycup, line 164: Im unsure about this on the accessibility front and would like opinions  - it is used as a generic catch-all and reads "More overflow" - open to suggestions. @jake-costa - let me know if this is enough context for an opinion, otherwise can explain further

site/docs/components/toolbar/index.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: Toolbar
+data:
+  description: "`Toolbar` is a container that organizes and presents a set of frequently used actions, commands, or search. It enables users to quickly access key functions relevant to the current context, improving efficiency and discoverability."
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/blob/main/packages/lab/src/toolbar-next"
+  package:
+    name: "@salt-ds/lab"
+  alsoKnownAs: ["Action Bar"]
```
aimee-liou:
```suggestion
  alsoKnownAs: ["Action bar"]
```

site/docs/components/toolbar/examples.mdx
```diff
@@ -0,0 +1,59 @@
...
+layout: DetailComponent
+sidebar:
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Appearance
+
+Toolbar supports two appearances: Bordered or Transparent. Transparent toolbars are flush with the edges of the toolbar container.
```
aimee-liou:
```suggestion
Toolbar supports two appearances: bordered or transparent. Transparent toolbars are flush with the edges of the toolbar container.
```

this is my bad, let's lowercase these

packages/lab/src/toolbar-next/ToolbarNext.tsx
```diff
@@ -0,0 +1,473 @@
...
+      >
+        {mode === "invalid" ? (
+          children
+        ) : (
+          <ToolbarNextOverflowFloatingBoundaryProvider>
+            <ToolbarNextOverflowOwners
+              hostNodes={itemOwnerHostNodes}
+              items={allItems}
+            />
+            <div aria-hidden className={clsx(withBaseName("measurements"))}>
```
joshwooding:
```suggestion
            <div aria-hidden className={withBaseName("measurements")}>
```

packages/lab/src/toolbar-next/useToolbarNextOverflow.ts
```diff
@@ -0,0 +1,1028 @@
...
+
+      const callback = (node: HTMLButtonElement | null) => {
+        updateObservedNode(triggerMeasureRefs.current, groupKey, node, {
+          groupKey,
+          kind: "shared-trigger-measure",
+        });
+      };
+
+      triggerMeasureCallbacks.current.set(groupKey, callback);
+      return callback;
```
joshwooding, line 1021: Ref callbacks were only added in React 19 🤔 

But the tests pass 🤷🏽

packages/lab/src/toolbar-next/ToolbarNext.tsx
```diff
@@ -0,0 +1,473 @@
...
+    const getItemHostRef = useCallback(
+      (id: string, kind: ToolbarNextItemHostKind) => {
+        const callbackKey = `${id}:${kind}`;
+        const existing = itemHostRefCallbacks.current.get(callbackKey);
+
+        if (existing) {
+          return existing;
+        }
+
+        const callback = (node: HTMLDivElement | null) => {
```
joshwooding, line 190: Same question here

packages/lab/src/toolbar-next/ToolbarContent.css
```diff
@@ -0,0 +1,24 @@
...
+  justify-content: center;
+}
+
+.saltToolbarContent[data-position="end"] {
+  justify-content: flex-end;
+}
+
+.saltToolbarContent[data-implicit] > .saltTooltrayNext[data-align="center"],
+.saltToolbarContent[data-implicit] > .saltTooltrayNext[data-align="end"] {
+  margin-inline: 0;
```
joshwooding: We should avoid this shorthand

packages/lab/src/toolbar-next/ToolbarNextOverflow.css
```diff
@@ -0,0 +1,82 @@
...
+  gap: var(--salt-spacing-100);
+  min-width: 0;
+}
+
+.saltToolbarNextOverflow-slot[data-align="end"] {
+  margin-inline-start: auto;
+}
+
+.saltToolbarNextOverflow-slot[data-align="center"] {
+  margin-inline: auto;
```
joshwooding: Same here

site/docs/components/toolbar/usage.mdx
```diff
@@ -0,0 +1,47 @@
...
+- To organize and group frequently used actions and commands together.
+- To improve efficiency and discoverability of related actions.
+- To provide a consistent location for key actions, typically at the top of a page, panel, or section.
+
+### When not to use
+
+- As primary navigation between major sections or pages. Instead, use [`Navigation item`](../navigation-item).
+- To display only one control.
+- For placing actions that are not directly related to the current context or content.
+- To display a brand logo, navigation, and utility actions. Instead, use [App header](/salt/patterns/app-header).
```
aimee-liou:
```suggestion
- To display a brand logo, navigation, and utility actions. Instead, use [`App header`](/salt/patterns/app-header).
```

site/docs/components/toolbar/accessibility.mdx
```diff
@@ -0,0 +1,47 @@
...
+sidebar:
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Best practices
+
+- Avoid controls that utilize arrow keys, like a text input, in the start or middle of a toolbar. These controls disrupt arrow-key item navigation and make Tab move to the next item instead of exiting, breaking expected toolbar behavior.
+  - If unavoidable, include only one and place it last to preserve arrow-key navigation and normal Tab exit.
```
aimee-liou:
```suggestion
  - If unavoidable, include only one and place it last in the toolbar to preserve arrow-key navigation and normal Tab exit.
```

site/docs/components/toolbar/accessibility.mdx
```diff
@@ -0,0 +1,47 @@
...
+layout: DetailComponent
+sidebar:
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Best practices
+
+- Avoid controls that utilize arrow keys, like a text input, in the start or middle of a toolbar. These controls disrupt arrow-key item navigation and make Tab move to the next item instead of exiting, breaking expected toolbar behavior.
```
aimee-liou:
```suggestion
- Avoid controls that utilize arrow keys, like a text input, in the start or middle of a toolbar. These controls disrupt arrow-key navigation and make Tab move to the next item instead of exiting, breaking expected toolbar behavior.
```

packages/lab/src/toolbar-next/useToolbarNextOverflow.ts
tycup: spoke about this already - for others who are interested: 
the return callback works because the ref callback itself doesn't return a cleanup function which is the react 19 pattern - these here are factories that create stable callback refs for the keyed nodes. They're cached like this to avoid detaching and reattaching the refs on every render, which in turn prevents unnecessary observer work

packages/lab/src/toolbar-next/TooltrayNext.css
```diff
@@ -0,0 +1,19 @@
...
+.saltTooltrayNext > * {
+  flex-shrink: 0;
+}
+
+.saltTooltrayNext[data-align="end"] {
+  margin-inline-start: auto;
+}
+
+.saltTooltrayNext[data-align="center"] {
+  margin-inline: auto;
```
joshwooding: Same here

packages/lab/src/toolbar-next/ToolbarNextOverflow.css
```diff
@@ -0,0 +1,82 @@
...
+  flex-wrap: wrap;
+  gap: var(--salt-spacing-100);
+  max-width: inherit;
+  min-width: max-content;
+  padding: var(--salt-spacing-100);
+}
+
+.saltToolbarNextOverflow-panel .saltTooltrayNext[data-align="center"],
+.saltToolbarNextOverflow-panel .saltTooltrayNext[data-align="end"] {
+  margin-inline: 0;
```
joshwooding: and here

packages/lab/src/toolbar-next/ToolbarNextOverflow.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.saltToolbarNextOverflow-panel .saltTooltrayNext > * {
+  inline-size: auto;
+}
+
+.saltToolbarNextOverflow-content > .saltToolbarNextOverflow-slot .saltTooltrayNext[data-align="center"],
+.saltToolbarNextOverflow-content > .saltToolbarNextOverflow-slot .saltTooltrayNext[data-align="end"] {
+  margin-inline: 0;
```
joshwooding: +1

packages/lab/src/toolbar-next/ToolbarNext.tsx
```diff
@@ -0,0 +1,477 @@
...
+      testId: "salt-toolbar-next",
+      css: toolbarNextCss,
+      window: targetWindow,
+    });
+
+    const { mode, content } = normalizeToolbarChildren(children);
+    const overflowContent = useMemo(
+      () => normalizeToolbarChildren(children).content,
+      [children],
+    );
```
wemmyo, line 95: Do we need to call `normalizeToolbarChildren` twice? Can we consolidate?

packages/lab/src/toolbar-next/ToolbarContent.tsx
wemmyo, line 1: Should the file name be `ToolbarNextContent`?

site/docs/components/toolbar/accessibility.mdx
```diff
@@ -0,0 +1,47 @@
...
+data:
+  $ref: ./#/data
+---
+
+## Best practices
+
+- Avoid controls that utilize arrow keys, like a text input, in the start or middle of a toolbar. These controls disrupt arrow-key navigation and make Tab move to the next item instead of exiting, breaking expected toolbar behavior.
+  - If unavoidable, include only one and place it last in the toolbar to preserve arrow-key navigation and normal Tab exit.
+- Use an `aria-label` when there is no visible label on the toolbar element and `aria-labelledby` when there is a visible label.
+- With transparent toolbar, avoid spanning the full width of a page unless it contains enough controls to be recognized as a toolbar. Otherwise, use a bordered toolbar.
```
aimee-liou, line 16:
```suggestion
- With transparent toolbar, ensure close spacing between controls and groups of controls so it's clearly recognizable as a toolbar. Otherwise, use a bordered toolbar to reinforce the toolbar container.
```
@jake-costa thoughts on this callout for transparent toolbar?

site/docs/components/toolbar/examples.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+## Content sections
+
+A toolbar is divided into three sections: `start content`, `center content`, and `end content`. Each section can contain groups of components, also known as tool trays. In code, every component is wrapped in a tool tray by default so keyboard navigation can be handled in an accessible, consistent way. Each section can be hidden.
+
+<LivePreview componentName="toolbar" exampleName="ContentSections" />
+
+## Dividers
+
+A divider can be used to separate and group like-content together. Alternatively, applying `spacing-300` between these groups can provide the same effect.
```
wemmyo, line 31: Nitpick but the example uses more than `spacing-300` for the spacing between groups. `spacing-300`  + gap (`spacing-100`)

site/docs/components/toolbar/accessibility.mdx
```diff
@@ -0,0 +1,47 @@
...
+
+## Keyboard interactions
+
+<KeyboardControls>
+<KeyboardControl keyOrCombos="Tab">
```
wemmyo, line 21: Not sure if its worth mentioning focus moving to last focused element in the toolbar

packages/lab/src/toolbar-next/ToolbarNext.tsx
tycup: its not pretty but its defensible - two consumers need different reference behaviour; `content` from the direct call needs to be fresh each render because it feeds `allItems`, which feeds keyboard nav and focus restoration. otherwise we get funky timing issues with the keyboard

`overflowContent` is memoized because it feeds `useToolbarNextOverflow` where the stable references help avoid unnecessary overflow measurement calculations (or rather, re calculations). 

could dedupe with a ref cache keyed by the children maybe but i havent thought through if that could cause other issues and this works atm

site/docs/components/toolbar/examples.mdx
tycup: good spot, updated

packages/lab/src/toolbar-next/ToolbarContent.tsx
tycup: 😭 yep, updated

packages/lab/src/toolbar-next/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./ToolbarContentNext";
```
wemmyo, line 1: The naming isn't consistent with the rest. You've used `ToolbarNext-*` for the rest

tycup: :facepalm:

tycup: although, for now, `ToolbarNext-*` is for utilities related to the component `ToolbarNext` and since `ToolbarContentNext` is also a component, its related files would be called `ToolbarContentNextKeyboardNav` (as an example, ToolbarContentNext doesn't have any related utilities). but ill take another look

site/allowed-words.txt
```diff
@@ -48,6 +48,12 @@ Tabstrips
 Taurean
 Telemedicine
 Tickbox
+Tooltray
+UITK
+frontmatter
+kmbt
+affordances
+trackpad
```
joshwooding, line 53: I think these a duplicates?

site/docs/components/toolbar/index.mdx
```diff
@@ -0,0 +1,22 @@
+---
+title: Toolbar
+data:
+  description: "`Toolbar` is a container that organizes and presents a set of frequently used actions, commands, or search. It enables users to quickly access key functions relevant to the current context, improving efficiency and discoverability."
+  sourceCodeUrl: "https://github.com/jpmorganchase/salt-ds/blob/main/packages/lab/src/toolbar-next"
+  package:
+    name: "@salt-ds/lab"
+  alsoKnownAs: ["Action bar"]
+  relatedComponents:
+    [
```
joshwooding: They aren't baked in right? Probably not worth including them here

site/docs/components/toolbar/index.mdx
```diff
@@ -0,0 +1,22 @@
...
+      { name: "Button", relationship: "contains" },
+      { name: "Dropdown", relationship: "contains" },
+      { name: "Input", relationship: "contains" },
+      { name: "Switch", relationship: "contains" },
+    ]
+  relatedPatterns: ["App header"]
```
joshwooding: I don't think app header would be related

site/docs/components/toolbar/examples.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+## Variants
+
+Bordered toolbars are available in three variants: primary, secondary, and tertiary. The primary variant is set by default.
+
+<LivePreview componentName="toolbar" exampleName="Variants" />
+
+## Content sections
+
+A toolbar is divided into three sections: `start content`, `center content`, and `end content`. Each section can contain groups of components, also known as tool trays. In code, every component is wrapped in a tool tray by default so keyboard navigation can be handled in an accessible, consistent way. Each section can be hidden.
```
wemmyo: `start content`, `center content`, and `end content` feels off, it doesn't mean anything in code. This is about the toolbar content position which maps to the position prop which wasn't mentioned in this section

site/docs/components/toolbar/examples.mdx
```diff
@@ -0,0 +1,59 @@
...
+layout: DetailComponent
+sidebar:
+  exclude: true
+data:
+  $ref: ./#/data
+---
+
+## Appearance
+
+Toolbar supports two appearances: bordered or transparent. Transparent toolbars are flush with the edges of the toolbar container.
```
wemmyo:
```suggestion
Toolbar supports two appearances: `bordered` or `transparent`. Transparent toolbars are flush with the edges of the toolbar container.
```
Do you want to mention what it defaults to?

site/docs/components/toolbar/examples.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+<LivePreview componentName="toolbar" exampleName="Appearance" />
+
+## Variants
+
+Bordered toolbars are available in three variants: primary, secondary, and tertiary. The primary variant is set by default.
```
wemmyo:
```suggestion
Bordered toolbars are available in three variants: `primary`, `secondary`, and `tertiary`. The `primary` variant is set by default.
```

joshwooding (dismissed):
Doc changes can go in tomorrow, code is in a good enough state for labs

site/docs/components/toolbar/examples.mdx
tycup: this is to do with the slot naming on going conversations - originally it was regions; can change per RC feedback

wemmyo (approved):

joshwooding (approved):

Issue #6362, opened 2026-04-23
Tree | Incorrect Tooltip positioning and behavior when used with TreeNodeTrigger


tycup:
### Latest version

- [x] I have tested the latest version

### Description

TreeNodeTrigger forwards React focus and blur events even when they originated from a descendant tree node rather than the trigger itself. 

Because an expanded tree node contains nested focusable treeitems, focus moving into a child still bubbled through the parent trigger, and consumers such as Tooltip interpreted that bubbled event as if the parent node had gained focus. As a result, tooltips attached to ancestor nodes could open when focus was actually on a child, remain visible at the same time as the child’s tooltip, or briefly flash and disappear as focus moved through the tree

### Steps to reproduce

1. Go to Tree With Tooltip story in storybook
2. Tab into the Tree
3. ArrowDown from Documents node to Reports (change focus)
4. ArrowUp back to Documents.
5. Documents Tooltip not showing & Tooltip flash

### Expected behavior

The Tooltip appears next to its respective treenode on focus

### Package name(s)

Lab (@salt-ds/lab)

### Browser

- [x] Chrome
- [x] Safari
- [x] Firefox
- [ ] Microsoft Edge

### Operating system

- [x] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

### Are you a JPMorgan Chase & Co. employee?

- [x] I am an employee of JPMorgan Chase & Co.

PR #6363, opened 2026-04-23
Fix TreeNodeTrigger Tooltip bug

tycup:
Addresses #6362 

- TreeNodeTrigger is not a typical flat trigger element: it renders a focusable treeitem, but when expanded it also contains a nested group of descendant treeitems, so native React focus/blur bubbling does not map cleanly to the semantic meaning of “this trigger was focused.” 
- In practice, forwarding bubbled descendant focus through onFocus/onBlur caused parent triggers to behave as though they themselves had gained focus, which is what made parent tooltips open, flash, or persist incorrectly when focus actually moved to a child node. 
- That makes TreeNodeTrigger a special case compared with other trigger components, which usually wrap a single interactive element and therefore can safely expose raw focus events.
- In this case, restricting forwarded focus and blur to event.target === event.currentTarget keeps the external contract aligned with the components real trigger semantics; consumers like Tooltip are notified only when that specific tree node receives focus and not when focus moves anywhere inside its subtree.

### DIFF
--- a/.changeset/rude-rocks-shave.md
+++ b/.changeset/rude-rocks-shave.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/lab": patch
+---
+
+Fixed incorrect Tooltip positioning and behavior when used with TreeNodeTrigger.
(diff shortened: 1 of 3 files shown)


joshwooding:
Do we need a test?

joshwooding (dismissed):

dplsek (dismissed):

mark-tate (approved):

joshwooding (approved):

PR #6427, opened 2026-04-28
Comments pattern

vanessawpws:


site/public/img/patterns/comments/comments-avatar.png
wemmyo, line 1: The image quality isn't great for these screenshots, we mostly use x2 image size for better quality.

wemmyo: Dark mode images are also a 'nice to have'

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
+          <StackLayout padding={1} gap={1}>
+            <div>
+              <Text styleAs="h4">{c.name}</Text>
+              <Text styleAs="label" color="secondary">
+                {c.role} • {c.date}
+              </Text>
+            </div>
+            <Text>{c.text}</Text>
+          </StackLayout>
+          {i < comments.length - 1 && <Divider variant="tertiary" />}
```
wemmyo: StackLayout already supports a `separators` props

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+## How to build
+
+### Anatomy
+
+At the top level, Comments consists of an input for adding new comments and a running list that shows the conversation history, so users can review prior messages.
+
+Each comment tile can include:
+
+1. Author: Display name
```
wemmyo:
```suggestion
1. **Author**: Display name 
```
Helps improve readability

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+- Avatar: Show a photo; fall back to initials. Keep size and placement consistent.
+- Role: Use a short, standardized role or title. 
+
+<Diagram
+  src="/img/patterns/comments/comments-avatar.png"
+  alt="comments with avatars"
+  border
+/>
+
+Use the Salt [`Avatar`](/salt/components/avatar) component’s categorical colors to differentiate authors; don’t use avatar color to imply status or severity.
```
wemmyo:
```suggestion
Use the [`Avatar`](/salt/components/avatar) component’s categorical colors to differentiate authors; don’t use avatar color to imply status or severity.
```

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+- Newest-first (default for ongoing threads): Surfaces the latest activity immediately; ideal for long‑running discussions over days or weeks.
+- Oldest-first: Preserves chronological reading; best for short, time‑boxed reviews where sequence matters.
+
+<Diagram
+  src="/img/patterns/comments/comments-display-order.png"
+  alt="Comments display order"
+  border
+/>
+
+## Submission error
```
wemmyo, line 146: Question: Is any guidance needed for successful submission?

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+<GuidanceCallout type="positive" customPillText="Best practices">
+  - Use a multiline field only if a meaningful portion of users will benefit.
+  - Auto‑grow to a sensible maximum height, then scroll within the field so the thread stays visible.
+</GuidanceCallout>
+
+## Display order
+
+Choose the order that best fits the task.
+
+- Newest-first (default for ongoing threads): Surfaces the latest activity immediately; ideal for long‑running discussions over days or weeks.
```
wemmyo, line 136: Should this be regarded as 'default' or 'recommeded'?

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
+      <FormField style={{ padding: "var(--salt-spacing-100)", width: "auto" }}>
+        <FormFieldLabel>Write a comment</FormFieldLabel>
+        <MultilineInput
+          bordered
+          placeholder="Add a comment..."
+          endAdornment={
+            <>
+              <Label>
+                {`${inputValue.length}/${MAX_CHARS}`}
+                <strong>{`${inputValue.length}/${MAX_CHARS}`}</strong>
```
wemmyo: I believe this was accidentally duplicated

wemmyo: There's more guidance on this in the character count docs https://www.saltdesignsystem.com/salt/components/multiline-input/examples#character-count

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+
+<Diagram
+  src="/img/patterns/comments/comments-display-order.png"
+  alt="Comments display order"
+  border
+/>
+
+## Submission error
+
+Offer recovery without losing work by showing an inline error. Preserve the failed comment text.
```
wemmyo, line 148: I believe there should be a screen reader announcement that comes with an error banner. Might be worth highlighting that somewhere

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
+---
+title: Comments pattern
```
wemmyo:
```suggestion
title: Comments
```

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,154 @@
...
+description: "The Comments pattern enables users to post, view, and interact with user-generated content, such as feedback, discussions, or reviews."
+layout: DetailPattern
+tags:
+  - pattern
+aliases:
+  - /salt/patterns/comments
+data:
+  resources:
+    [
+      { href: "https://storybook.saltdesignsystem.com/?path=/story/patterns-comments--default", label: "Comments examples" },
```
wemmyo:
```suggestion
      { href: "https://storybook.saltdesignsystem.com/?path=/story/patterns-comments", label: "Comments examples" },
```
Removing the story name in the link allows a natural redirect to the first story

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
+    ]);
+    setInputValue("");
+  };
+
+  return (
+    <StackLayout gap={0} style={{ width: "420px" }}>
+      <FormField style={{ padding: "var(--salt-spacing-100)", width: "auto" }}>
+        <FormFieldLabel>Write a comment</FormFieldLabel>
+        <Input
+          style={{ height: "36px" }}
```
joshwooding:
```suggestion
```

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
+        text: inputValue,
+      },
+      ...comments,
+    ]);
+    setInputValue("");
+  };
+
+  return (
+    <StackLayout gap={0} style={{ width: "420px" }}>
+      <FormField style={{ padding: "var(--salt-spacing-100)", width: "auto" }}>
```
joshwooding:
```suggestion
      <FormField style={{ width: "auto" }}>
```

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
```
joshwooding:
```suggestion
      <FormField style={{ width: "auto" }}>
```

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,393 @@
...
+        <FormFieldLabel>Write a comment</FormFieldLabel>
+        <Input
+          style={{ height: "36px" }}
```
joshwooding:
```suggestion
```

packages/core/stories/patterns/comments/comments.stories.tsx
joshwooding: We should put the padding on a div wrapping this

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,397 @@
...
+      <div style={{ padding: "var(--salt-spacing-100)" }}>
+        <FormField>
+          <FormFieldLabel>Write a comment</FormFieldLabel>
+          <Input
+            bordered
+            placeholder="Add a comment..."
+            endAdornment={
+              <Button onClick={handleSubmit} disabled={!inputValue.trim()}>
+                <SendIcon />
+              </Button>
```
joshwooding: The icon buttons need to be updated to follow our accessibility guidance: Add aria-hidden to the icon and the Button needs an aria-label

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,397 @@
...
+            }
+            value={inputValue}
+            onChange={(e) =>
+              setInputValue((e.target as HTMLInputElement).value)
+            }
+            onKeyDown={(e) => {
+              if (e.key === "Enter") handleSubmit();
```
joshwooding: Should we be using a form which would get us this automatically?

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,397 @@
...
+        date: new Date()
+          .toLocaleString("en-GB", {
+            day: "2-digit",
+            month: "short",
+            year: "numeric",
+            hour: "2-digit",
+            minute: "2-digit",
+            hour12: true,
+          })
+          .replace(",", " •"),
```
joshwooding: This is a visual concern so should probably move to rendering, we should store dates in an unix timestamp

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,397 @@
...
+    <StackLayout gap={0} style={{ width: "420px" }}>
+      <div style={{ padding: "var(--salt-spacing-100)" }}>
+        <FormField>
+          <FormFieldLabel>Write a comment</FormFieldLabel>
+          <MultilineInput
+            bordered
+            placeholder="Add a comment..."
+            endAdornment={
+              <>
+                <Label>{`${inputValue.length}/${MAX_CHARS}`}</Label>
```
joshwooding: I'm guessing we're missing some UX tied to this? @ivan-calderon

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,397 @@
...
+        </FormField>
+      </div>
+      <StackLayout gap={3} align="center">
+        <InfoIcon
+          style={{
+            color: "white",
+            backgroundColor: "var(--salt-status-info-foreground-decorative)",
+          }}
+          size={2}
+        />
```
joshwooding: It's likely this should just be using StatusIndicator

site/docs/components/comments/accessibility.mdx
joshwooding, line 1: we don't need this file

site/docs/components/comments/examples.mdx
joshwooding, line 1: or this one

site/docs/components/comments/index.mdx
joshwooding, line 1: or this

site/docs/components/comments/usage.mdx
joshwooding, line 1: or this

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,164 @@
...
+- Role: Use a short, standardized role or title. 
+
+<Diagram
+  src="/img/patterns/comments/comments-avatar.png"
+  srcDark="/img/patterns/comments/comments-avatar-dark.png"
+  alt="comments with avatars"
+  border
+/>
+
+Use the [`Avatar`](/salt/components/avatar) component’s categorical colors to differentiate authors; don’t use avatar color to imply status or severity.
```
joshwooding: @ivan-calderon Do we have to mention status or severity? Not sure how it fits in here

site/docs/patterns/comments.mdx
```diff
@@ -0,0 +1,164 @@
...
+<Diagram
+  src="/img/patterns/comments/comments-user-role.png"
+  srcDark="/img/patterns/comments/comments-user-role-dark.png"
+  alt="comments with user roles"
+  border
+/>
+
+## Timestamp
+
+Always show a timestamp and the last edit time. Place it next to the author by default.
```
joshwooding: Pattern's don't have default so we should update this page following that guidance @ivan-calderon

ivan-calderon:
```suggestion
Always show a timestamp and the last edit time.
```

site/docs/patterns/comments.mdx
ivan-calderon:
```suggestion
Use the [`Avatar`](/salt/components/avatar) component’s categorical colors to differentiate authors.
```

packages/core/stories/patterns/comments/comments.stories.tsx
ivan-calderon: I'm using [Content Status pattern](https://www.saltdesignsystem.com/salt/patterns/content-status) here. I can update example for better alignment + we can point to Content Status for reference. LMK if that works for more clarity.

site/docs/patterns/comments.mdx
joshwooding: I think it's fine to not mention the announcement, that behaviour belongs to Banner.

site/docs/patterns/comments.mdx
joshwooding: Similar to my comment, nothing should be called default

jake-costa:
# Accessibility Findings

1. **No list semantics for the comment list.** The comment list is not semantically coded as a list. StackLayout renders `<div>` elements with no list structure, so there are no programmatic boundaries for the list group or between individual comments. 
    * **Fix:** Use `<ul>` for the comment collection (along with an aria label to name the list for the examples) and `<li>` for each comment. Native HTML list elements are preferred over ARIA `role="list"` where possible.

2. **No status announcement after comment submission.** When a comment is submitted, the input clears and the comment appears visually, but AT receives no confirmation that the action succeeded. Focus remains in the input, which is correct, but there is no accompanying announcement. 
      * **Fix:** Use the AriaAnnouncer provider polite to announce a confirmation message after successful submission.

3. **Character counter not associated with the textarea.** In the WithMultilineInput variant, the character counter is visually positioned inside the input but has no programmatic association. AT users cannot discover the character limit or their current count. 
      * **Fix:** Associate the counter with the textarea via `aria-describedby`. Use the AriaAnnouncer provider polite to announce when the character count changes, and assertive when the character count hits the limit.
   - Note: This association pattern should likely be baked into the MultilineInput component itself to avoid every consumer needing to wire it manually.

4. **Retry and Close button labels are too generic.** The error recovery buttons use `aria-label="Retry"` and `aria-label="Close"`, which don't describe what is being retried or dismissed. 
      * **Fix:** Use labels that convey the action in context, e.g., "Retry posting your comment" and "Dismiss submission error."
   - Note: When the error message appears while focus is on the input, focus should remain on the input and the error message should be announced using the AriaAnnouncer provider assertive. After pressing Retry, focus should return to the input on success or remain on the Retry button on failure. After pressing Dismiss, focus should return to the input.

5. **Decorative images are not properly hidden.** Several images across the pattern are decorative but still exposed to AT.       
    * **Fix:** Remove `aria-label` and add `aria-hidden="true"` on each:
       - **Avatar** (WithAvatar variant): renders `role="img"` with `aria-label="Alex Rivera"` while the author name is already displayed as visible text.
       - **StatusIndicator** (WithEmptyState variant): renders with `aria-label="info"` while the "No comments yet" text already conveys the state.
       - **StatusIndicator** (WithSubmissionError variant): renders with `aria-label="error"` inside the Banner while the error message text already conveys the state.

## Considerations

- **Long comment lists are not currently captured in the examples.** The current examples show a small number of comments with no guidance on handling long or growing threads. Consider if we should demonstrate a scrollable region, pagination, or a "load more" mechanism. If we are unable to incorporate this into the initial release, this could be considered for a future release.

- **Interactive elements within individual comments.** Are there plans to incorporate interactive elements within individual comment containers (e.g., reply, edit, delete actions)? If so, we would need to incorporate the appropriate the keyboard model and focus management for those interactions.

- **Storybook examples do not support reflow at 320x256.** It looks like the examples are using hardcoded widths on the container and do not properly reflow when the viewport is set to 320x256. (Applies to all densities.)

@vanessawpws Let me know if you have any questions on any of these items above and once the changes above have been made, as well as the changes to the disabled button, please let me know so I can review the updated component.

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+      month: "short",
+      year: "numeric",
+      hour: "2-digit",
+      minute: "2-digit",
+      hour12: true,
+    })
+    .replace(",", " •");
+
+export const Default = () => (
+  <AriaAnnouncerProvider>
```
joshwooding: We shouldn't need to do this, one is included with the SaltProvider

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+      >
+        <FormField>
+          <FormFieldLabel>Write a comment</FormFieldLabel>
+          <Input
+            bordered
+            placeholder="Add a comment..."
+            validationStatus={validationStatus}
+            endAdornment={
+              inputValue && (
+                <Button aria-label="Send comment" type="submit">
```
joshwooding, line 116: @jake-costa with the send button disappearing should we handle focus?

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+            }}
+          >
+            <StackLayout padding={1} gap={1}>
+              <StackLayout gap={0.5}>
+                <Text styleAs="h4">{c.name}</Text>
+                <Text styleAs="label" color="secondary">
+                  {c.role} • {formatDate(c.date)}
+                </Text>
+              </StackLayout>
+              <Text>{c.text}</Text>
```
joshwooding: Can we add `line-break: anywhere` to the text content

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+              </FlexLayout>
+            </FormFieldHelperText>
+          )}
+        </FormField>
+      </form>
+      <ul
+        aria-label="Comments"
+        style={{ listStyle: "none", margin: 0, padding: 0 }}
+      >
+        {comments.map((c, index) => (
```
joshwooding: Nit: Can we avoid single letter variables where possible

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+          <li
+            key={`${c.name}-${c.date}`}
+            style={{
+              borderTop:
+                index > 0
+                  ? "1px solid var(--salt-separable-secondary-borderColor)"
```
joshwooding:
```suggestion
                  ? "var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-separable-secondary-borderColor)"
```

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+          e.preventDefault();
+          handleSubmit();
+        }}
+      >
+        <FormField>
+          <FormFieldLabel>Write a comment</FormFieldLabel>
+          <Input
+            bordered
+            placeholder="Add a comment..."
+            validationStatus={validationStatus}
```
joshwooding: We should be using validationStatus on FormField

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+              }
+            }}
+          />
+          {validationStatus === "error" && (
+            <FormFieldHelperText color="error">
+              <FlexLayout gap={0.75} align="center">
+                <StatusIndicator status="error" aria-hidden="true" />
+                Comment can't be blank
+              </FlexLayout>
+            </FormFieldHelperText>
```
joshwooding: If we use validation status on form field this can be heavily simplified:


```suggestion
            <FormFieldHelperText>
                Comment can't be blank
            </FormFieldHelperText>
```

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,590 @@
...
+  return (
+    <StackLayout gap={0} style={{ width: "100%", maxWidth: "420px" }}>
+      <div style={{ padding: "var(--salt-spacing-100)" }}>
+        <FormField>
+          <FormFieldLabel>Write a comment</FormFieldLabel>
+          <Input
+            bordered
+            placeholder="Add a comment..."
+            endAdornment={
+              <Button aria-label="Send comment" disabled>
```
joshwooding: This shouldn't be disabled

packages/core/stories/patterns/comments/comments.stories.tsx
jake-costa: If it's going to be conditionally displayed, focus should go back to the input

jake-costa:
# Accessibility Findings - Updated

1. **Focus is lost when the submit button is removed after comment submission.** The submit button is conditionally rendered and only appears when the input has a value. When a user Tabs to the submit button and presses Enter, the comment submits, the input clears, and the button is removed from the DOM. Focus falls to the document body. 
    -  **Fix:** When the submit button is removed after submission, focus must be placed back on the input.

2. **`aria-invalid` is not applied to the input when validation fails.** When the field is in an error state, the Input and MultilineInput components apply CSS error styling and render a visual error indicator, but `aria-invalid="true"` is not set on the `<input>` or `<textarea>` element. AT users navigating to the field receive no programmatic indication that it is in an error state. 
    - **Fix:** Set `aria-invalid="true"` on the form control when the field is in an error state and removed when error state is removed.

3.  **Error message is not re-announced on repeated blank submissions.** When the user submits an empty comment, the validation error appears and is announced. If the user types whitespace and submits again, the error message is not re-announced because the error state has not changed. AT users receive no feedback that their second submission also failed. 
    - **Fix:** Each time a submission is attempted and fails, the error message must be announced to AT regardless of whether the previous submission also failed. (clear the live region then re-inject)

4. **Decorative StatusIndicators are not hidden from AT.** The StatusIndicator is exposed to AT in two places where the adjacent text already conveys the state:
   - **Fix:** Both StatusIndicators are decorative in these contexts and must be hidden from AT.
       - **WithEmptyState variant**: renders with `aria-label="info"` while the "No comments yet" text already conveys the state.
       - **WithSubmissionError variant**: renders with `aria-label="error"` inside the Banner while the error message text already conveys the state.

5. **AriaAnnouncer is using assertive for all announcements.** All announcements in the pattern seem to use `aria-live="assertive"`, which interrupts the screen reader immediately. Non-error announcements such as "Comment posted" and character count updates must use `aria-live="polite"` so they do not interrupt the user. Error announcements such as "Failed to post comment" and "Character limit reached" are appropriate as assertive. 
    - **Fix:** Non-error announcements must use polite. Error and limit-reached announcements should use assertive. (Might want to connect with Mark on this)

## Considerations

- **Unable to test Retry and Dismiss flow.** The Retry and Dismiss buttons in the WithSubmissionError variant do not appear to be functional in the example.

mark-tate:
Can we review the copilot feedback

Here's my forensic review of the staged comments pattern changes:
 
Issues & Inconsistencies
1. Inconsistent border separator tokens (Stories)
Default story uses the full token: "var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-separable-secondary-borderColor)"
All other stories (WithAvatar, WithMultilineInput, WithEmptyState, WithSubmissionError) use hardcoded "1px solid var(--salt-separable-secondary-borderColor)"
Fix: Use the same token-based approach everywhere (the Default approach is more correct).
2. Massive code duplication across stories
The comment data, handleSubmit, form markup, and comment list rendering are nearly identical across all 5 stories. Consider extracting shared pieces (e.g., a CommentForm helper, a CommentList helper, shared initial data) to reduce duplication and maintenance burden.
3. WithAvatar inconsistency — Default and WithEmptyState lack avatars but WithSubmissionError has them
WithSubmissionError includes <Avatar> in the comment list, but it's not the "WithAvatar" story. This seems unintentional — the submission error story should match Default (no avatar) unless deliberately showing avatar + error together.
4. Banner text in WithSubmissionError contains stray content
The error message reads: "You are offline. Check your connection and resubmit I'll check if saved views are discoverable enough." — The second sentence ("I'll check if saved views are discoverable enough") looks like leftover copy/notes, not real error message text.
5. Dismiss button in WithSubmissionError has no onClick handler
The Retry and Dismiss buttons inside the Banner have no onClick, so they do nothing. At minimum, Dismiss should call setSubmissionError(false).
6. Retry button has sentiment="neutral" but Dismiss does not
Inconsistent button styling within the same Banner actions group.
7. WithSubmissionError always shows the error on mount (submissionError defaults to true)
This means the error banner is visible before the user even tries to submit. It should default to false and only appear after a failed submission.
8. first comment separator — index > 0 but first comment also has no top border from the form
There's no visual separator between the form and the first comment. Consider adding a border-top on the <ul> or the first <li> for consistency.
9. Type union syntax error in diff display
The diff shows "error" "success" undefined — this should be "error" | "success" | undefined. This may just be the diff stripping pipe characters, but worth verifying the actual file is correct.
10. MDX doc: double period in anatomy
"an optional user role, or an 'Edited' indicator.." — has two periods.
11. MDX doc: inconsistent alt text casing
"Comments anatomy" vs "comments layout" vs "comments with avatars" — inconsistent capitalization. Should be consistent (e.g., sentence case throughout).
12. MDX doc: Timestamp section says "Always show… the last edit time"
But none of the stories demonstrate an "edited" indicator or edit time. The docs and stories are out of sync.
13. MDX doc: Display order section has no corresponding story
There's no story demonstrating oldest-first ordering.
14. formatDate uses hardcoded "en-GB" locale
The docs recommend "locale-aware formatting and the viewer's time zone" but the code hardcodes en-GB.
15. Comment timestamps are hardcoded future dates (year ~2026)
1775035560000 = mid-2026. Not a real issue, but worth noting for realism.
16. No aria-live region on the comment list
When a new comment is added, announce() fires, which is good. But the list itself could benefit from being a live region so screen readers pick up additions naturally.
17. lineBreak: "anywhere" should probably be overflowWrap: "anywhere" or wordBreak: "break-word"
line-break: anywhere affects CJK line-breaking rules. For wrapping long unbroken strings (URLs, etc.), overflow-wrap: anywhere is more appropriate.
 
Summary of recommended fixes (prioritized)
Priority
Issue
🔴 Critical
#4 — Stray copy in error message
🔴 Critical
#9 — Verify union type syntax
🟡 Medium
#1 — Inconsistent border tokens
🟡 Medium
#3 — Unintended avatar in submission error story
🟡 Medium
#5 — Missing onClick handlers on Banner buttons
🟡 Medium
#7 — submissionError defaults to true
🟡 Medium
#10 — Double period in MDX
🟡 Medium
#11 — Inconsistent alt text casing
🟡 Medium
#12 — Docs mention "edited" but no story shows it
🟢 Minor
#2 — Code duplication
🟢 Minor
#6, #8, #14, #17 — Small inconsistencies

mark-tate:
Worth checking this ai based feedback with Jake

Looking at the staged comments pattern code for accessibility:

A11y Issues Found
✅ Good: useAriaAnnouncer — announces "Comment posted" on submission.
✅ Good: aria-label="Comments" on <ul> — gives the list an accessible name.
✅ Good: aria-hidden="true" on avatars — decorative avatars are hidden from AT.
✅ Good: aria-label on icon-only buttons — "Send comment", "Retry posting your comment", "Dismiss submission error".
⚠️ Issue: No form error announcement — When validation fails and the helper text appears, there's no aria-live or role="alert" ensuring the error is announced. Salt's FormField with validationStatus="error" may handle this via aria-describedby linking to the helper text, but only if focus remains in the field. If focus is on the submit button, the error won't be announced.
⚠️ Issue: Comment list not a live region — When comments are added/removed, only the manual announce() call notifies AT. This is acceptable but fragile — if announce is forgotten in a future code path, additions are silent.
⚠️ Issue: WithMultilineInput character counter — The aria-describedby={counterId} on the textarea is good, but the live character count updates aren't in an aria-live region. The announce() with a 1-second debounce partially compensates, but the <Label> element showing "X/1000" isn't live-announced natively.
⚠️ Issue: No role="status" or aria-live on empty state — When all comments are cleared or on initial load with no comments, the "No comments yet" message has no live region attribute, so a screen reader user navigating programmatically would find it, but it won't be announced on state change.
✅ Good: <form> with onSubmit — Supports Enter-to-submit natively.
⚠️ Issue: Banner in WithSubmissionError lacks role="alert" — Salt's Banner component likely provides this, but worth verifying. If it does, this is fine.
🔴 Issue: Send button only appears when inputValue is truthy — This is a conditional render. Screen reader users won't know a send button exists until they type. Consider always rendering it (disabled when empty) or providing clear instructions.
⚠️ Issue: Comment <li> elements have no accessible time semantics — The timestamp is just text inside a <Text>. Using a <time datetime="..."> element would give AT richer semantics.
Verdict
The pattern is reasonably accessible — it uses semantic HTML (<ul>, <li>, <form>), labels lists, announces submissions, and hides decorative elements. The main gaps are:
No <time> element for timestamps
Conditional send button visibility (could confuse AT users)
Reliance on manual announce() rather than live regions for dynamic content
These are moderate issues, not blockers, but worth addressing for a reference pattern.

ivan-calderon:
@vanessawpws I think you have resolved some of the below. But if useful on:

1. **Inconsistent border separator tokens (Stories):** This is the [divider component ](https://www.saltdesignsystem.com/salt/components/divider/examples)using the correct tokens. We should use tertiary variant.
6. **Retry button has sentiment="neutral" but Dismiss does not:** You can refer to [Error banner ](https://www.saltdesignsystem.com/salt/components/banner/examples#error)buttons 
8. **First comment separator — index > 0:** First comment was meant not to use divider on top to look for UITK parity. Non critical but valid if we decide to add it.

dplsek:
Use consistent border tokens for all separators: 
var(--salt-size-fixed-100) 
var(--salt-borderStyle-solid) 
var(--salt-separable-tertiary-borderColor)

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -0,0 +1,670 @@
...
+        <div style={{ padding: "var(--salt-spacing-100)" }}>
+          <Banner status="error" variant="secondary">
+            <BannerContent>
+              <StackLayout gap={1}>
+                <Text>
+                  <strong>Couldn’t post your comment</strong>
+                </Text>
+                <Text>
+                  You are offline. Check your connection and resubmit I'll check
+                  if saved views are discoverable enough.
```
joshwooding:
```suggestion
                  You are offline. Check your connection and resubmit
                  "I'll check if saved views are discoverable enough."
```

dplsek:
please set a min width in the [empty state](https://pr-6427.saltdesignsystem-storybook.pages.dev/?path=/story/patterns-comments--with-empty-state) example so it matches the width of the other examples.

dplsek:
@ivan-calderon can we have teh first example on the site match the styling of the [default example](https://pr-6427.saltdesignsystem-storybook.pages.dev/?path=/story/patterns-comments--default) in storybook, ie without avatars?

dplsek:
@vanessawpws can we add a categorical avatar examples like on the [site docs](https://saltdesignsystem-git-feature-pepper-2297-fed-team.vercel.app/salt/patterns/comments#avatar-and-user-role)?

dplsek:
@vanessawpws can we order the storybook examples to match the order of the site doc examples?

ivan-calderon (dismissed):

jake-costa:
> Worth checking this ai based feedback with Jake
> 
> Looking at the staged comments pattern code for accessibility:
> 
> A11y Issues Found ✅ Good: useAriaAnnouncer — announces "Comment posted" on submission. ✅ Good: aria-label="Comments" on
> 
> — gives the list an accessible name.
> ✅ Good: aria-hidden="true" on avatars — decorative avatars are hidden from AT.
> ✅ Good: aria-label on icon-only buttons — "Send comment", "Retry posting your comment", "Dismiss submission error".
> ⚠️ Issue: No form error announcement — When validation fails and the helper text appears, there's no aria-live or role="alert" ensuring the error is announced. Salt's FormField with validationStatus="error" may handle this via aria-describedby linking to the helper text, but only if focus remains in the field. If focus is on the submit button, the error won't be announced.
> ⚠️ Issue: Comment list not a live region — When comments are added/removed, only the manual announce() call notifies AT. This is acceptable but fragile — if announce is forgotten in a future code path, additions are silent.
> ⚠️ Issue: WithMultilineInput character counter — The aria-describedby={counterId} on the textarea is good, but the live character count updates aren't in an aria-live region. The announce() with a 1-second debounce partially compensates, but the element showing "X/1000" isn't live-announced natively.
> ⚠️ Issue: No role="status" or aria-live on empty state — When all comments are cleared or on initial load with no comments, the "No comments yet" message has no live region attribute, so a screen reader user navigating programmatically would find it, but it won't be announced on state change.
> ✅ Good: with onSubmit — Supports Enter-to-submit natively.
> ⚠️ Issue: Banner in WithSubmissionError lacks role="alert" — Salt's Banner component likely provides this, but worth verifying. If it does, this is fine.
> 🔴 Issue: Send button only appears when inputValue is truthy — This is a conditional render. Screen reader users won't know a send button exists until they type. Consider always rendering it (disabled when empty) or providing clear instructions.
> ⚠️ Issue: Comment
> * elements have no accessible time semantics — The timestamp is just text inside a . Using a  element would give AT richer semantics.
>   Verdict
>   The pattern is reasonably accessible — it uses semantic HTML (, * , ), labels lists, announces submissions, and hides decorative elements. The main gaps are:
>     No  element for timestamps
>     Conditional send button visibility (could confuse AT users)
>     Reliance on manual announce() rather than live regions for dynamic content
>     These are moderate issues, not blockers, but worth addressing for a reference pattern.

@vanessawpws let me know if you would like to discuss this. As I would not recommend most of these that the AI provided feedback on and there are a number of reasons we do not want to do a few of these.

jake-costa:
# A11y Review - Third Evaluation

## Accessibility Findings
- **All announcements are using the same type of announcement.** Salt's `useAriaAnnouncer` defaults to `polite`, which is correct for success and informational messages. However, error announcements, "Comment can't be blank", "Failed to post comment", "Comment is too long", and "Character limit reached", must use assertive so they interrupt the screen reader immediately and are not lost behind queued speech. Non-error announcements such as "Comment posted" and periodic character count updates are correctly polite by default. Additionally, announcements must fire after focus has settled, currently "Comment posted" fires before the focus move to the input, which means the focus change speech can swallow the polite announcement.
  - **Fix:** Error and limit-reached announcements must explicitly use assertive. Success and informational announcements should remain polite. Ensure announcements fire after focus management, not before.

- **The submission error announcement text does not match the Banner visual text.** The `announce()` call says "Failed to post comment" but the Banner displays "Couldn't post your comment, You are offline. Check your connection and resubmit." AT users do not receive the same information as sighted users,  they hear a summary but miss the reason for the failure and the recovery instructions. 
  - **Fix:** The announcement text must include the full error content displayed "verbatim" in the Banner so AT users receive equivalent information.

- **Focus is not managed correctly in WithSubmissionError.** Focus management must account for three distinct user actions:
   - **User activates the submit button:** Focus must stay on the submit button. The current implementation moves focus to the input, which is incorrect for a system-level error, the user's input is not at fault, and moving focus away from the trigger is disorienting.
   - **User presses Enter on the input:** Focus should remain on the input. This is the current behavior and is correct.
   - **User activates Retry or Dismiss:** Focus must return to the input after the error banner is removed. Retry does this correctly. Dismiss does not, focus falls to the document body because the handler does not explicitly manage focus.

  - **Fix:** When the submit button triggers the error, focus must remain on the submit button. When the error banner is dismissed via Retry or Dismiss, focus must return to the input.

## Considerations
- **WithSubmissionError recovery flow is inconsistent.** The error Banner tells the user to "Check your connection and resubmit," but resubmitting via the form re-triggers the error while the Retry button succeeds. A subsequent submission attempt should behave the same as activating the Retry button.


 ## Feedback on Additional comments
 >  Issue: Send button only appears when inputValue is truthy — This is a conditional render. Screen reader users won't know a send button exists until they type. Consider always rendering it (disabled when empty) or providing clear instructions.
 * Dynamically displaying the submit button when the user has typed a message is intentional and is one valid way to handle this. Alternatively, the button could remain visible and continue to trigger validation. Either approach is preferable to a disabled button.
 
 > Issue: Comment list not a live region — When comments are added/removed, only the manual announce() call notifies AT. This is acceptable but fragile — if announce is forgotten in a future code path, additions are silent.
 * The comments list does not need to be a live region. The targeted `announce("Comment posted")` is a valid approach.
 
 >  Issue: WithMultilineInput character counter — The aria-describedby={counterId} on the textarea is good, but the live character count updates aren't in an aria-live region. The announce() with a 1-second debounce partially compensates, but the element showing "X/1000" isn't live-announced natively.
 * The character count is already announced via the AriaAnnouncer. The counter does not need to be added as a live region since this is already being handled.
 
 >  Issue: No role="status" or aria-live on empty state — When all comments are cleared or on initial load with no comments, the "No comments yet" message has no live region attribute, so a screen reader user navigating programmatically would find it, but it won't be announced on state change.
 * Currently, do not have the ability to clear and the "Be the first to comment" message On page load is not considered a status message, so it should not have `role="status"` or be a live region. It is static content on initial load, not a status change.
 
 >  Issue: Comment elements have no accessible time semantics — The timestamp is just text inside a . Using a element would give AT richer semantics.
 * Not a violation. `<time>` elements could be added as an enhancement but is not required. The other elements in the comment do not have an appropriate semantic element type. Visually hidden accessible text could be added to denote each section, but that would also be an enhancement.
 
 > Issue: Banner in WithSubmissionError lacks role="alert" — Salt's Banner component likely provides this, but worth verifying. If it does, this is fine.
 * Adding `role="alert"` to a dynamically rendered Banner is not recommended. Live region containers must be present in the DOM before content injection for consistent screen reader behavior. This is documented in Salt's Banner component accessibility section. The pattern correctly uses the AriaAnnouncer, which maintains persistent live regions in the DOM.
 
 @vanessawpws

jake-costa:
@vanessawpws 
The announce() call and the Banner content are maintained as separate hardcoded strings. The Banner also appears to contain stray text, "I'll check if saved views are discoverable enough," which doesn't seem to align to the actual error being displayed.

I'd recommend removing the stray text and extracting the error message into a shared variable or constant that is used for both the Banner content and the announcement. That way they can't get out of sync.

As mentioned in previous reviews and the team chat, there are a number of items we haven't implemented/accounted for or validated for this pattern yet, including nested interactive elements, deleting or editing comments, scrollable regions, and others. I can provide my approval after the item above has been addressed, but there are still items we'll need to account for shortly after the initial release.

One additional note: the Avatar instances correctly have aria-hidden="true", but the component is still setting an aria-label from the name prop. Since aria-hidden removes the element from the accessibility tree, the aria-label is redundant and not ideal (Not a violation). This is likely something we'll need to address at the Avatar component level so that aria-label is not set when aria-hidden="true" is present.

ivan-calderon (approved):

ivan-calderon (approved):

jake-costa (approved):
As mentioned in previous reviews and the team chat, there are a number of items we haven't implemented/accounted for or validated for this pattern yet, including nested interactive elements, deleting or editing comments, scrollable regions, and others. I can provide my approval after the item above has been addressed, but there are still items we'll need to account for shortly after the initial release.

PR #6448, opened 2026-05-01
Polish side panel follow up

wemmyo:


docs/css/salt-date-components.css
joshwooding, line 1: I think this was accidentally added

site/src/examples/side-panel/WithNav.tsx
```diff
@@ -62,13 +62,10 @@ const WithNavContent = () => {
       <Nav />
       <SidePanel position="left">
         <SidePanelHeader>
-          <SidePanelTitle>
-            <H2 id={headerId}>Section Title</H2>
-          </SidePanelTitle>
+          <SidePanelTitle id={titleId}>Section Title</SidePanelTitle>
           <Button
-            id={closeButtonId}
             aria-label="Close"
```
joshwooding, line 67: I think we should make this simpler even though this is a cool way to do it I'm concerned it will confuse consumers @jake-costa do you have any thoughts?

site/src/examples/side-panel/WithTable.tsx
```diff
@@ -152,21 +158,16 @@ const SidePanelExample = () => {
       </div>
 
       <SidePanel position="right" style={panelStyle} key={selectedRow?.id}>
-        {selectedRow && (
-          <>
+        {formValues && (
+          <Fragment>
```
joshwooding: Just curious was there a reason we changed this from <> ?

site/src/examples/side-panel/WithTable.tsx
```diff
@@ -152,21 +158,16 @@ const SidePanelExample = () => {
...
             <SidePanelHeader>
-              <SidePanelTitle>
-                <H2 id={headerId}>
-                  <span className="salt-visuallyHidden">
-                    {selectedRow.name}
-                  </span>
-                  Employee Details
-                </H2>
+              <SidePanelTitle id={titleId}>
+                <span className="salt-visuallyHidden">{formValues.name}</span>
```
joshwooding: Can we add this as a style, it's a bit weird but salt-visuallyHidden was never really meant to be a public API

site/src/examples/side-panel/WithNav.tsx
joshwooding: The alternative would be just an aria-label

site/src/examples/side-panel/WithTable.tsx
wemmyo: There was a `key` which I removed. Can change back to <> now

docs/css/salt-date-components.css
wemmyo: Thanks. Not sure how that got in

site/src/examples/side-panel/WithNav.tsx
wemmyo: Yeah I can just hardcode the full name 'Close section title', the only example that needs the dynamic close button name is the table example. I can leave that one if it helps

site/src/examples/side-panel/WithTable.tsx
wemmyo: Sure. Done!

mark-tate (approved):

Issue #6455, opened 2026-05-05
Tree functionality regression when using TreeNodes wrapped in Fragments


tycup:
### Latest version

- [x] I have tested the latest version

### Description

Users may want to conditionally render TreeNodes inside Tree. This fails to be registered correctly inside Tree, causing users to be unable to navigate through TreeNodes wrapped in React Fragments using the keyboard. Users also face regression across other modes inc. multiselection propagation.

### Steps to reproduce

Create a Tree with nested TreeNodes that are wrapped in a React.Fragment:

```
const SecretFiles = ({ show = true }: { show?: boolean }) =>
  show ? (
    <>
      <TreeNode value="project/config/.env.local" label=".env.local" />
      <TreeNode value="project/config/.npmrc" label=".npmrc" />
      <TreeNode value="project/config/credentials" label="credentials">
        <TreeNode value="project/config/credentials/api-key" label="api.key" />
        <TreeNode
          value="project/config/credentials/service-account"
          label="service-account.json"
        />
      </TreeNode>
      <TreeNode value="project/config/secrets" label="secrets.json" />
    </>
  ) : null;

export const WithConditionalFragment: StoryFn<typeof Tree> = (args) => (
  <Tree
    {...args}
    aria-label="Project files"
    defaultExpanded={[
      "project",
      "project/config",
      "project/config/credentials",
      "project/config/public",
      "project/src",
      "project/src/components",
      "project/tests",
    ]}
  >
    <TreeNode value="project" label="project">
      <TreeNode value="project/config" label="config">
        <SecretFiles />
        <TreeNode value="project/config/public" label="public">
          <TreeNode value="project/config/public/app" label="app.config.ts" />
          <TreeNode value="project/config/public/theme" label="theme.json" />
        </TreeNode>
      </TreeNode>
    </TreeNode>
    <TreeNode value="project/src" label="src">
      <TreeNode value="project/src/components" label="components">
        <TreeNode value="project/src/components/tree" label="Tree.tsx" />
      </TreeNode>
    </TreeNode>
    <TreeNode value="project/tests" label="tests">
      <TreeNode value="project/tests/tree" label="tree.test.tsx" />
    </TreeNode>
  </Tree>
);
```

### Expected behavior

Tree should honour keyboard navigation despite the TreeNode being wrapped in a Fragment inc. arrow keys for focus movement.

In multiselect mode, selecting a parent TreeNode should also select Fragment rendered children.

### Package name(s)

[...]

PR #6456, opened 2026-05-05
Tree | Ensure Tree can be used with Fragment TreeNodes

tycup:
Addresses #6455 

There's a bit of nuance to accessibility, so recommend checking behaviour against the current merged Tree (certain readouts, like the selected state, behave inconsistently across platforms and browsers e.g. use of aria-checked instead of aria-selected, and aria-selected not being read out).


packages/lab/src/tree/TreeNode.tsx
joshwooding: This is the same issue you flagged to me in tabs btw, value could have spaces or something else that would make it an invalid id

packages/lab/src/tree/TreeNode.tsx
```diff
@@ -106,12 +109,15 @@ export const TreeNode = forwardRef<HTMLLIElement, TreeNodeProps>(
...
+            }
           : separateChildren(children),
       [children, usesLabelProp],
     );
 
-    const hasChildren = nodeChildren.length > 0;
+    const hasChildren = nodeChildren.some(isTreeNodeElement);
 
     // Forward ref to the <li> element rendered by TreeNodeTrigger
     useImperativeHandle(ref, () => getElement(value) as HTMLLIElement, [
```
joshwooding: NIT: couldn't we have passed the ref through context?

joshwooding (approved):

mark-tate (approved):

PR #6495, opened 2026-05-13
Add formatted input examples

MinalBhujade:
-> formatted input examples like phone number, phone number with dynamic preview, credit card, currency and postal code

### DIFF
--- a/packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
+++ b/packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
@@ -9,12 +9,27 @@ import {
   useBreakpoint,
 } from "@salt-ds/core";
 import type { Meta, StoryFn } from "@storybook/react-vite";
+import type { ReactNode } from "react";
 import { useFormattedInput } from "./useFormattedInput";
 
 export default {
   title: "Patterns/Formatted Input",
-  component: Input,
 } as Meta;
+
+const PhoneNumberLabelWithPreview = ({
+  previewText,
+}: {
+  previewText: ReactNode;
+}) => (
+  <FlexLayout direction="row" align="center" justify="space-between" gap={1}>
+    <FormFieldLabel>Phone number</FormFieldLabel>
+    {previewText && (
+      <Text styleAs="label" color="secondary">
+        {previewText}
+      </Text>
+    )}
+  </FlexLayout>
+);
 
 export const PhoneNumber: StoryFn = () => {
   const formatPhoneNumber = (cleaned: string) => {
@@ -90,14 +105,18 @@ export const PhoneNumberWithPreview: StoryFn = () => {
   const defaultHelperText = "Enter a US phone number, e.g. +1 (555) 000-0000.";
 
   const { matchedBreakpoints } = useBreakpoint();
-  const isMobile = matchedBreakpoints.indexOf("sm") === -1;
+  const isBelowSm = !matchedBreakpoints.includes("sm");
 
   const formatPhoneNumber = (cleaned: string) => {
     if (cleaned.length === 0) {
       return "";
     }
     if (cleaned.length <= 4) {
-      return `+${cleaned.slice(0, 1)} (${cleaned.slice(1)}`;
+      // Keep the partial format balanced while typing (close the paren when
+      // we have at least 4 digits, otherwise show just the country code).
+      return cleaned.length < 4
+        ? `+${cleaned.slice(0, 1)} ${cleaned.slice(1)}`.trimEnd()
+        : `+${cleaned.slice(0, 1)} (${cleaned.slice(1, 4)})`;
     }
     if (cleaned.length <= 7) {
       return `+${cleaned.slice(0, 1)} (${cleaned.slice(1, 4)}) ${cleaned.slice(4)}`;
@@ -116,7 +135,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
     return "";
   };
 
-  const phoneInputOptions = {
+  const buildPhoneOptions = () => ({
     formatValue: (cleaned: string) =>
       cleaned.length === 11 ? formatPhoneNumber(cleaned) : cleaned,
     normalizedValue: (value: string) => value.replace(/\D/g, ""),
@@ -128,7 +147,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
     invalidFormatMessage:
       "Please enter a valid US phone number: country code (1) + area code + number.",
     generatePreview,
-  };
+  });
 
   const {
     displayValue,
@@ -139,7 +158,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
     handleChange,
     handleBlur,
     handleFocus,
-  } = useFormattedInput(phoneInputOptions);
+  } = useFormattedInput(buildPhoneOptions());
 
   const {
     displayValue: displayValue2,
@@ -150,25 +169,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
     handleChange: handleChange2,
     handleBlur: handleBlur2,
     handleFocus: handleFocus2,
-  } = useFormattedInput(phoneInputOptions);
-
-  const getFormFieldLabel = (preview: string) => {
-    return (
-      <StackLayout
-        direction="row"
-        align="center"
-        gap={1}
-        style={{ justifyContent: "space-between" }}
-      >
-        <FormFieldLabel>Phone number</FormFieldLabel>
-        {preview && (
-          <Text styleAs="label" color="secondary">
-            {preview}
-          </Text>
-        )}
-      </StackLayout>
-    );
-  };
+  } = useFormattedInput(buildPhoneOptions());
 
   return (
     <FlexLayout direction="row" align="start" gap={2} wrap>
@@ -177,7 +178,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
           style={{ width: "300px" }}
           validationStatus={validationStatus}
         >
-          {getFormFieldLabel(preview)}
+          <PhoneNumberLabelWithPreview previewText={preview} />
           <Input
             value={displayValue}
             onChange={handleChange}
@@ -205,8 +206,8 @@ export const PhoneNumberWithPreview: StoryFn = () => {
           style={{ width: "300px" }}
           validationStatus={validationStatus2}
         >
-          {isMobile ? (
-            getFormFieldLabel(preview2)
+          {isBelowSm ? (
+            <PhoneNumberLabelWithPreview previewText={preview2} />
           ) : (
             <FormFieldLabel>Phone number</FormFieldLabel>
           )}
@@ -218,14 +219,14 @@ export const PhoneNumberWithPreview: StoryFn = () => {
               onFocus={handleFocus2}
               placeholder="+1 (000) 000-0000"
               bordered
-              style={{ width: isMobile ? "100%" : "210px" }}
+              style={{ width: isBelowSm ? "100%" : "210px" }}
               inputProps={{
                 "aria-invalid":
                   validationStatus2 === "error" ? true : undefined,
                 autoComplete: "tel",
               }}
             />
-            {!isMobile && (
+            {!isBelowSm && (
               <Text
                 styleAs="label"
                 color="secondary"
@@ -239,7 +240,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
             )}
           </FlexLayout>
           <FormFieldHelperText
-            style={{ maxWidth: isMobile ? "100%" : "210px" }}
+            style={{ maxWidth: isBelowSm ? "100%" : "210px" }}
           >
             {validationMessage2 || defaultHelperText}
           </FormFieldHelperText>
@@ -257,7 +258,7 @@ export const PhoneNumberWithPreview: StoryFn = () => {
 export const CreditCard: StoryFn = () => {
   const formatCreditCard = (cleaned: string) => {
     const match = cleaned.match(/.{1,4}/g);
-    return match ? match.join("-") : cleaned;
+    return match ? match.join(" ") : cleaned;
   };
   const {
     displayValue,
@@ -270,8 +271,7 @@ export const CreditCard: StoryFn = () => {
   } = useFormattedInput({
     formatValue: formatCreditCard,
     normalizedValue: (value) => value.replace(/[\s-]/g, ""),
-    validateNormalized: (cleaned) =>
-      cleaned.length === 16 && /^\d{16}$/.test(cleaned),
+    validateNormalized: (cleaned) => /^\d{16}$/.test(cleaned),
     hasInvalidChars: (value: string) => /[^\d\s-]/.test(value),
     invalidCharMessage: "Only numbers, spaces and hyphens are allowed.",
     invalidCharBlurMessage:
@@ -288,7 +288,7 @@ export const CreditCard: StoryFn = () => {
           onChange={handleChange}
           onBlur={handleBlur}
           onFocus={handleFocus}
-          placeholder="5555-5555-5555-4444"
+          placeholder="5555 5555 5555 4444"
           bordered
           inputProps={{
             "aria-invalid": validationStatus === "error" ? true : undefined,
@@ -318,7 +318,9 @@ export const Currency: StoryFn = () => {
   const validateCurrency = (cleaned: string) => {
     const parts = cleaned.split(".");
     if (parts.length > 2) return false;
-    return parts[0].length !== 0;
+    if (parts[0].length === 0) return false;
+    // Reject a trailing dot with no fractional digits (e.g. "123.").
+    return !(parts.length === 2 && parts[1].length === 0);
   };
 
   const {
@@ -370,7 +372,9 @@ export const Currency: StoryFn = () => {
 };
 
 export const PostalCode: StoryFn = () => {
-  // UK postal code patterns: various formats like M1, CR2, W1A, EC1A, etc.
+  // UK full postal codes: outward + inward, e.g. M1 1AA, CR2 6XH, W1A 1HQ,
+  // EC1A 1BB, SW1A 1AA. The regex matches the full code (no space) after
+  // normalization.
   const UK_POSTALCODE_REGEX = /^[A-Z]{1,2}\d[A-Z\d]?\d[A-Z]{2}$/;
 
   const formatPostalCode = (cleaned: string) => {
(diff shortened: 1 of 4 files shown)


packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
```diff
@@ -0,0 +1,686 @@
...
+  return (
+    <FormField style={{ width: "300px" }} validationStatus={validationStatus}>
+      <FormFieldLabel>Phone number</FormFieldLabel>
+      <Input
+        value={displayValue}
+        onChange={handleChange}
+        onBlur={handleBlur}
+        onFocus={handleFocus}
+        placeholder="+1 (000) 000-0000"
+        aria-describedby="phone-helper-text"
```
joshwooding: Helper text already is associated and shouldn't need this

packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
```diff
@@ -0,0 +1,686 @@
...
+        bordered
+      />
+      <FormFieldHelperText
+        id="phone-helper-text"
+        aria-live="assertive"
```
joshwooding: I think it only needs to be assertive for validation only?

packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
```diff
@@ -0,0 +1,686 @@
...
+  );
+};
+
+export const PostalCode: StoryFn = () => {
+  const [postalCode, setPostalCode] = useState("");
+  const [displayValue, setDisplayValue] = useState("");
+  const [validationStatus, setValidationStatus] = useState<"error" | undefined>(
+    undefined,
+  );
+  const defaultHelperText = "Enter your postal code";
```
joshwooding: Rather than this we should probably have this in the helper text and have the validation message in the state. Then we can swap it out and we no longer need to keep the default in status

packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
MinalBhujade: added assertive only when validation status.

packages/core/stories/patterns/formatted-input/formatted-input.stories.tsx
MinalBhujade: updated

jake-costa:
# Accessibility Review Items
- **`aria-invalid` is not set on the input when validation fails.** The stories set `validationStatus="error"` on `FormField`, but the native `<input>` itself never receives `aria-invalid`, so the control's invalid state is not exposed to assistive technology.
  - **Fix:** Apply `aria-invalid` to the input when in the error state.
  
- **Success and warning messages are announced assertively.** `FormFieldHelperText` uses `aria-live={validationStatus ? "assertive" : "polite"}`, which makes any defined status (including `success` and `warning`) announce assertively. In the PhoneNumber story, "Phone number is valid." interrupts the user every time a valid phone is entered.
  - **Fix:** Scope assertive to `error` only. Warning and success are not time-sensitive and should remain polite.

- **PhoneNumberWithPreview fails reflow at 320 by 256.** The second FormField in this story places the input and the dynamic preview side by side at fixed widths. At the WCAG 1.4.10 reflow viewport (320 by 256), the preview is cut off.
  - **Fix:** Decide on the best way to meet the reflow requirement for this example, possibly reverting to placing the dynamic preview next to the label (Variant A's layout) at small viewports.
  
## Considerations
- **The error announcement currently relies on toggling `aria-live` on the helper text.** This can work, but for consistency with other salt-ds patterns, consider calling `useAriaAnnouncer` as assertive for the error message and using the default polite state for success and warnings.

- **The examples should add `autocomplete` for inputs with a recognized purpose.** Because this is a pattern rather than a component, missing `autocomplete` is not a violation here, but the stories are the recipe consumers will copy. Without `autocomplete`, real implementations will fail WCAG 1.3.5 Identify Input Purpose. Add `autoComplete="tel"` to the phone stories, `autoComplete="cc-number"` to `CreditCard`, and `autoComplete="postal-code"` to `PostalCode`.

- **Consider updating the Currency helper text to mention the currency.** Screen reader users do not have a programmatic indication of which currency is expected for the pattern example. Updating it to "Enter an amount in US dollars (numbers only)." puts the currency context into the description.

- **Consider removing the "Phone number is valid." success message.** For inputs without external validation, the success signal is simply the absence of an error state on tab out, the user does not need a spoken confirmation that no error fired. Unless the pattern is meant to model a flow with real validation work (server check, lookup), the additional "Phone number is valid." announcement is redundant.

@MinalBhujade

mark-tate:
Here's my deep forensic review of both files:
 
---
 
## Storybook Stories (`formatted-input.stories.tsx`)
 
### 🐛 Bugs
 
1. **`handleFocus` restores raw input, not cleaned digits (lines 106–110, 228–233, 422–426, 518–522, 637–641):** `phoneNumber`/`cardNumber`/`amount`/`postalCode` stores the *raw* user input (set on blur from `e.target.value`), but after a successful blur the `displayValue` is already the *formatted* value. On re-focus, you set `displayValue` back to the raw value—but `phoneNumber` was set to `e.target.value` *before* formatting happened, so it could be the raw OR formatted value depending on what the user typed. This is inconsistent. For example, in `CreditCard`, if the user types `5555555555554444`, on blur `cardNumber` = `"5555555555554444"` and `displayValue` = `"5555-5555-5555-4444"`. On re-focus it correctly reverts. But if the user types `5555-5555-5555-4444` directly, `cardNumber` = `"5555-5555-5555-4444"` and on re-focus it shows the same formatted string—defeating the purpose of "revert to raw."
 
2. **Currency: commas in input are stripped inconsistently (line 504):** `hasInvalidChars` allows commas (`/[^\d.,]/`), but the cleaning step uses `/[^\d.]/g` which strips commas. So `1,000.50` becomes `1000.50`—this is fine as normalization, but the user typed commas which are "valid" chars, yet they disappear. Then on re-focus, `amount` holds `"1,000.50"` (the raw input), but `displayValue` was set to `"1,000.50"` (the formatted output which re-adds commas). This works by coincidence but is fragile.
 
3. **Currency: `.50` fails validation (line 467):** `validateCurrency` rejects values where `parts[0].length === 0`, meaning `.50` is invalid. This is arguably correct but might surprise users—worth a comment or helper text clarification.
 
4. **Postal code `hasInvalidChars` is case-sensitive (line 592):** The regex `/[^A-Z0-9\s]/` is tested against `value` in `handleBlur` (line 614) but the value hasn't been uppercased yet at that point—`e.target.value` could have lowercase letters which would be flagged as invalid even though they're valid letters. **Wait**—looking again, `handleChange` uppercases (`line 596: setDisplayValue(uppercase)`), so by the time blur fires the displayed value should be uppercase. But `e.target.value` on blur reads the DOM, which should reflect the controlled `displayValue`. So this is likely fine, but fragile if the controlled value isn't synced in time.
 
### ⚠️ Code Quality / DX Issues
 
5. **Massive duplication in `PhoneNumberWithPreview` (lines 133–359):** The second phone number field (`*2` variables, `handleChange2`, `handleBlur2`, `handleFocus2`) is a near-complete copy-paste of the first. This should be extracted into a reusable component or custom hook. For a *pattern* that users will copy, this sets a bad example.
 
6. **No custom hook extraction:** As a pattern meant for copy-paste reuse, the stories would benefit enormously from a shared `useFormattedInput` hook that encapsulates the `displayValue`/`rawValue`/`validationStatus`/`validationMessage` state machine. Every story repeats the same `handleChange → handleBlur → handleFocus` structure.
 
7. **Unused import (line 10):** `@salt-ds/countries/saltCountries.css` is imported but no country-related component is used anywhere in the file.
 
8. **Inconsistent validation status types:** `PhoneNumber` uses `"error" | "warning" | "success" | undefined` (line 28), while all other stories use `"error" | undefined`. This inconsistency in the pattern examples is confusing.
 
### ♿ Accessibility Issues
 
9. **`aria-live="assertive"` on every validation change:** Using `assertive` for errors means any error message will interrupt the screen reader immediately, even during typing (since `handleChange` sets errors in real-time). For character-level validation during typing, `"polite"` would be less disruptive. Consider using `assertive` only on blur.
 
10. **Preview text (lines 301–305, 338–346) has no `aria-live` region:** The dynamic preview that shows formatted output while typing is purely visual—screen reader users won't know it exists. It should have `aria-live="polite"` or be associated with the input via `aria-describedby`.
 
11. **No `inputMode` or `autoComplete` attributes:** For phone numbers, `inputMode="tel"` and `autoComplete="tel"` would improve mobile UX. For credit cards, `inputMode="numeric"` and `autoComplete="cc-number"`. For currency, `inputMode="decimal"`. These are missing from all stories.
 
12. **No `type` attribute guidance:** All inputs use the default `type="text"`, which is correct for formatted inputs, but it's worth a comment explaining *why* (e.g., `type="tel"` would prevent formatting characters).
 
### 🔧 Improvements
 
13. **`formatPhoneNumber` only handles 11-digit US numbers (line 35):** The helper text says "including the country code and area code" but the format `+1 (xxx) xxx-xxxx` is US-specific. International numbers of different lengths will just show raw digits.
 
14. **Credit card always formats as 4-4-4-4 (line 371):** Amex is 4-6-5, Diners is 4-6-4. The pattern only handles 16-digit cards. This is fine for a basic example but should be noted.
 
15. **No `maxLength` on any input:** Users can type unlimited characters. Adding `maxLength` (e.g., 19 for formatted phone, 19 for formatted credit card) would improve UX.
 
---
 
## Site Docs (`formatted-input.mdx`)
 
### 📝 Documentation Issues
 
16. **Internal link on line 66:** `"CIB Digital Content Style Guide"` links to `[https://go/contentguide/page/502`—this](https://go/contentguide/page/502%60%E2%80%94this) is an internal short-link that won't work for external/open-source consumers. Same issue on **line 184** (`go/contentguide/page/605`). These should either be removed, marked as internal, or replaced with public URLs.
 
17. **Line 3 (description):** Very long single-line description. Not a bug, but hard to maintain.
 
18. **No code examples in the docs:** The MDX file is purely visual (diagrams/images). It references the Storybook examples via a link (line 12) but doesn't embed any code snippets. For a pattern page, inline code examples would significantly improve DX—users shouldn't have to navigate to Storybook to see implementation.
 
19. **No mention of accessibility in the "How to build" section:** While there's a "Best practices" callout about FormField (line 63–67), there's no dedicated accessibility section covering `aria-live`, `inputMode`, `autoComplete`, keyboard behavior, or screen reader announcements.
 
20. **"Copy & paste behavior" section (lines 296–301) has no code example:** It describes expected behavior but doesn't show how to implement it. The stories also don't have a dedicated paste-handling example.
 
21. **No mention of mobile/touch considerations:** The pattern docs don't discuss `inputMode` for mobile keyboards, which is critical for formatted inputs (numeric keyboard for phone/credit card vs. alphanumeric for postal codes).
 
22. **Image-only content:** All behavioral descriptions rely on images (`<Diagram>` components). If images fail to load, the page loses most of its informational value. Alt text is present but generic (e.g., "User types of the formatted input pattern" on line 209 isn't very descriptive).
 
23. **Line 251:** "Input normalization is the practice of converting user-entered values into a consistent, expected representation—without forcing users to type in one exact format." — This is excellent guidance but isn't reflected well in the stories. The phone number story *does* normalize, but there's no explicit story named "Normalization" to match this docs section.
 
---
 
## Summary of Priority Issues
 
| Priority | Issue | Location |
|----------|-------|----------|
| 🔴 High | Unused CSS import | Stories L10 |
| 🔴 High | Massive code duplication (PhoneNumberWithPreview) | Stories L133–359 |
| 🔴 High | Preview text not accessible to screen readers | Stories L301–346 |
| 🔴 High | Internal-only links in public docs | Docs L66, L184 |
| 🟡 Medium | Missing `inputMode`/`autoComplete` on all inputs | Stories |
| 🟡 Medium | `aria-live="assertive"` during typing is too aggressive | Stories |
| 🟡 Medium | No code examples in docs MDX | Docs |
| 🟡 Medium | No accessibility section in docs | Docs |
| 🟢 Low | No `maxLength` on inputs | Stories |
| 🟢 Low | Inconsistent validation status types across stories | Stories |
| 🟢 Low | Generic alt text on diagrams | Docs |
 
Here's a cross-reference analysis of the stories vs. the docs:
 
## Do the Examples Match the Docs?
 
### ✅ Matches
 
| Docs Section | Story Coverage |
|---|---|
| Format on blur (not during typing) | ✅ All stories format in `handleBlur`, not `handleChange` |
| Re-entering reverts to raw value | ✅ All stories restore raw value in `handleFocus` |
| Invalid character detection in real-time | ✅ All stories check `hasInvalidChars` in `handleChange` |
| Error clears when user corrects | ✅ All stories clear error in `handleChange` when chars are valid |
| Error persists on blur if uncorrected | ✅ All stories keep error in `handleBlur` |
| Helper text + placeholder guidance | ✅ All stories have both |
| Dynamic preview | ✅ `PhoneNumberWithPreview` story covers this |
 
### ❌ Mismatches & Gaps
 
1. **Docs describe "Warning" state (lines 157–167) but only `PhoneNumber` implements it.** The other 3 stories (`CreditCard`, `Currency`, `PostalCode`) only use `"error" | undefined`. The docs imply warning is a general pattern capability, but only one contrived example (unusual area code) demonstrates it.
 
2. **Docs describe "Success" state (lines 169–182) but only `PhoneNumber` implements it** (and inconsistently—it sets `"success"` status but `FormField`'s `validationStatus` type in Salt core may not support `"success"` the same way as `"error"`/`"warning"`). The other stories never show success. The docs say "optional success indicator" but the stories don't consistently demonstrate this.
 
3. **Docs section "Input normalization" (lines 249–269) has no dedicated story.** The docs describe accepting flexible input (hyphens vs. spaces) and normalizing on blur. The phone number story does some normalization (stripping non-digits), but there's no story explicitly named or annotated to match this docs section. The caption on the docs diagram says "Hyphens are allowed and will be normalized on blur"—the credit card story does this but it's not called out.
 
4. **Docs section "Copy & paste behavior" (lines 296–301) has no story.** The docs explicitly describe paste behavior but no story demonstrates or tests it.
 
5. **Docs mention "postal/ZIP codes" in the description (line 3) and the `PostalCode` story exists, but the docs diagrams never show a postal code example.** All diagrams appear to show phone numbers. The story exists but isn't visually represented in the docs.
 
6. **Docs say "Currency amounts" in the description but the currency story has no corresponding docs section or diagram.**
 
7. **Dynamic preview docs (lines 271–294) describe two placements:** "near the label" and "next to the input." The `PhoneNumberWithPreview` story does show both (lines 293–321 for label-adjacent, lines 322–356 for input-adjacent) — ✅ this matches.
 
### Consistency Issues Between Stories
 
| Issue | Details |
|---|---|
| **Validation status types differ** | `PhoneNumber`: `"error" \| "warning" \| "success" \| undefined`. Others: `"error" \| undefined`. |
| **`handleValidation` helper differs** | `PhoneNumber` accepts 4 status values; others accept only 2. Not reusable across stories. |
| **Error message style inconsistent** | Blur messages sometimes start with "Remove…" (imperative), sometimes "Please enter…" (polite). Both styles appear but aren't consistently applied. |
| **Placeholder format inconsistent** | Phone: `+1 (000) 000-0000`. Credit card: `5555-5555-5555-4444` (uses real test card number as placeholder—potentially confusing). Postal: `12345 or E14 5JP or SW1A 1AA` (multiple examples in placeholder). |
| **Field width varies arbitrarily** | Phone: `300px`, Credit card: `400px`, Currency: `300px`, Postal: `300px`. No design-token or consistent sizing rationale. |
 
### Summary
 
The stories **broadly follow** the docs' behavioral spec (format-on-blur, real-time invalid char detection, re-enter-reverts). However, **3 of the 6 docs sections have no corresponding story** (normalization, copy/paste, success state shown generically), and the **warning/success states are only in one story** despite being documented as general capabilities. The stories would benefit from 1:1 naming alignment with docs sections and consistent validation status types across all examples.
 
We may not need to address all this but there are some clear issues that need to be addressed.

mark-tate (dismissed):

jake-costa:
>handleFocus restores raw input, not cleaned digits (lines 106–110, 228–233, 422–426, 518–522, 637–641): phoneNumber/cardNumber/amount/postalCode stores the raw user input (set on blur from e.target.value), but after a successful blur the displayValue is already the formatted value. On re-focus, you set displayValue back to the raw value—but phoneNumber was set to e.target.value before formatting happened, so it could be the raw OR formatted value depending on what the user typed. This is inconsistent. For example, in CreditCard, if the user types 5555555555554444, on blur cardNumber = "5555555555554444" and displayValue = "5555-5555-5555-4444". On re-focus it correctly reverts. But if the user types 5555-5555-5555-4444 directly, cardNumber = "5555-5555-5555-4444" and on re-focus it shows the same formatted string—defeating the purpose of "revert to raw."

This is not a bug and what should happen.

>aria-live="assertive" on every validation change: Using assertive for errors means any error message will interrupt the screen reader immediately, even during typing (since handleChange sets errors in real-time). For character-level validation during typing, "polite" would be less disruptive. Consider using assertive only on blur.
>Preview text (lines 301–305, 338–346) has no aria-live region: The dynamic preview that shows formatted output while typing is purely visual—screen reader users won't know it exists. It should have aria-live="polite" or be associated with the input via aria-describedby.

This is incorrect. We should not be announcing the dynamic preview or the change to formatting on tab out

jake-costa:
@mark-tate 
We also need to be very careful about having AI conduct reviews (Especially for accessibility related items) as AIs are a great tool but need very specific context. And without that context, it will have a high chance at flagging incorrect items like what is above.

mark-tate:
> @mark-tate We also need to be very careful about having AI conduct reviews (Especially for accessibility related items) as AIs are a great tool but need very specific context. And without that context, it will have a high chance at flagging incorrect items like what is above.

Agreed++, it's not a list of work, it's a set of un-validated observations, which may or may not be relevant. 

However, at the moment, PRs are being raised and manually reviewed/approved whilst missing issues that AI is able to put a spotlight on.  Some maybe smoking guns, some may lack context, some are quick fixes that can prevent future defects or create better consistency or examples for our users. As you say, some might not be correct assumptions. Design/Dev/A11y/Product should have visibility of this list and decide what needs fixing, it's not just the developer's responsibility to review the deliverable or the list.

Typically we would write a test to ensure that such requirements do not become a regression but patterns don't have tests. In this pattern's case we are pushing the boundaries of what a pattern should be. What we can do, in such cases, is add comments to the code, to make it clear the intent. Consumers have the same tools as we do and so it's highly likely they would let the tools fix the code as they see fit and all the work we have done is lost.

joshwooding (approved):

mark-tate (approved):

PR #6520, opened 2026-05-18
Mega Menu clean up

ThusharaJ07:
More test, clean up after ai review.


packages/lab/src/mega-menu/MegaMenuItem.tsx
tycup: should we be setting href here to a default?

ThusharaJ07: Initially i had added a default to keep the item focusable and keyboard-accessible even if the user hasn't wired up a real link in the example. 

But on second thought, I can drop the default and let the consumer pass href explicitly.Items without href would render as a non-navigating link.

thoughts?

joshwooding (changes requested):
Can we have a setup similar to VerticalNavigation with a browser router instead of the # links or any other work arounds please

ThusharaJ07:
> Can we have a setup similar to VerticalNavigation with a browser router instead of the # links or any other work arounds please

updated

ThusharaJ07:
> The changeset says `MegaMenuItem` falls back to rendering a `<button>` when no `render` prop or `href` is provided, but the implementation always calls `renderProps("a", props)`. Since `renderProps` renders the provided default type when `render` is absent, a no-`href` item still becomes an `<a>` rather than a button.
> 
> That mismatch matters for the new API because keyboard semantics and typing are still anchor-based (`AnchorHTMLAttributes<HTMLAnchorElement>`, `KeyboardEvent<HTMLAnchorElement>`). Either the implementation should switch the default type based on `href`, or the changeset/docs should avoid promising the button fallback.

corrected the changeset.

joshwooding (approved):

kriskoder (approved):

PR #6522, opened 2026-05-18
Comments pattern followup

vanessawpws:


packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -624,6 +479,32 @@ export const WithMultilineInput = () => {
...
+                      `Character limit reached. ${value.length} of ${MAX_CHARS} characters used.`,
+                      { ariaLive: "assertive" },
+                    );
+                  }
+                } else {
+                  prevAtLimitRef.current = false;
+                  setValidationStatus(undefined);
+                  setErrorMessage("");
+                  if (value.length > 0) {
+                    announce(
```
joshwooding, line 498: Pretty sure this gets announced every keystroke?

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -545,31 +422,9 @@ export const WithMultilineInput = () => {
   >(undefined);
```
joshwooding, line 418: Is there a reason we don't use useCommentForm in the multiline example?

packages/core/stories/patterns/comments/comments.stories.tsx
joshwooding: This button being disabled is inconsistent

joshwooding: This should use a functional setState

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -37,46 +71,79 @@ const getAvatarColor = (
...
-  ]);
+  const [comments, setComments] = useState<Comment[]>(initialData);
 
   const handleSubmit = () => {
     if (!inputValue.trim()) {
       setValidationStatus("error");
       inputRef.current?.focus();
       announce("Comment can't be blank", { ariaLive: "assertive" });
-      return;
+      return false;
```
joshwooding: Why are we returning false here?

packages/core/stories/patterns/comments/comments.stories.tsx
```diff
@@ -93,8 +160,43 @@ export const Default = () => {
     requestAnimationFrame(() => {
       announce("Comment posted");
     });
+    return true;
```
joshwooding: and why is this true here?

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: I don't see it used, will remove all return values handleSubmit..

Suspect this is being confused with onSubmit/prevent default type behaviour, which is not required.

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: fixed

mark-tate: a few of these in the code addressed.

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: removed , as before

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: the hook would need to be made generic, 

Extra state — multiline adds errorMessage, MAX_CHARS, counterId, prevAtLimitRef

Maybe clearer to leave the hooks as-it-is for now.

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: added a debounce

packages/core/stories/patterns/comments/comments.stories.tsx
mark-tate: made consistent with conditional rendering

mark-tate (approved):

mark-tate (approved):

joshwooding (approved):

PR #6554, opened 2026-05-28
Mark `@salt-ds/date-components` and `@salt-ds/date-adapters` as stable

wemmyo:
- Marked date components and date adapter as stable
- Removed date components and utilities from lab


.changeset/slick-canyons-pay.md
```diff
@@ -0,0 +1,10 @@
+---
+"@salt-ds/date-adapters": major
+"@salt-ds/date-components": major
+"@salt-ds/lab": minor
```
wemmyo:
```suggestion
"@salt-ds/lab": major
```

packages/date-adapters/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@salt-ds/date-adapters",
-  "version": "0.1.0-alpha.7",
+  "version": "1.0.0",
```
joshwooding: Versions shouldn't be changed in this PR, those are managed by changesets. We'd need to test how changesets deals with alpha packages having a major update but right now this will result in a 2.0 release

packages/date-components/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@salt-ds/date-components",
-  "version": "0.1.0",
+  "version": "1.0.0",
```
joshwooding: Same here

joshwooding (changes requested):
@mark-tate can you review the public API for the date control packages too please

packages/date-adapters/package.json
wemmyo:
```suggestion
  "version": "0.1.0-alpha.7",
```

packages/date-components/package.json
wemmyo:
```suggestion
  "version": "0.1.0",
```

packages/date-adapters/package.json
wemmyo: Updated, thanks

packages/date-components/package.json
wemmyo: Updated as well

packages/date-adapters/package.json
mark-tate: reverted

packages/date-components/package.json
mark-tate: reverted

mark-tate (approved):

packages/date-components/src/date-picker/index.ts
```diff
@@ -1,6 +1,16 @@
...
+// Re-export only the public API from DatePickerContext.
+// SingleDateSelectionContext and DateRangeSelectionContext are intentionally
+// kept module-private (used internally by DatePicker); consumers should use
+// the `useDatePickerContext` hook instead.
+export {
+  type DatePickerState,
+  type RangeDatePickerState,
+  type SingleDatePickerState,
+  type UseDatePickerContextProps,
+  useDatePickerContext,
```
joshwooding, line 12: I'm assuming this hook exposes everything @mark-tate ? Do we want to only provide a cut down version or are we happy with this? It would still mean everything in that context is a public API

joshwooding (approved):

PR #6558, opened 2026-05-29
Toolbar | Update documentation

tycup:


site/docs/components/toolbar/examples.mdx
```diff
@@ -10,7 +10,7 @@ data:
 
 ## Appearance
 
-Toolbar supports two appearances: `bordered` or `transparent`. Transparent toolbars are flush with the edges of the toolbar container. The default appearance is `bordered`.
+Toolbar supports two appearances: `bordered` or `transparent`. Transparent toolbars remove the default inset padding, so their contents align with the toolbar container edges. The default appearance is `bordered`.
```
tycup, line 13: @bhoppers2008 let me know if this is clearer. Essentially the colloquial message is 'the toolbar has no inner padding when they're transparent'.

To your second point on "why is it default" -> since its an accessibility topic, its in the accessibility section. I prioritised it as the first point in accessibility  page to match it being the first topic in the example page.

I'm weary of cross referencing one doc tab (e.g. examples -> accessibility) because its out of pattern (we only do it on occasion).

site/docs/components/toolbar/examples.mdx
```diff
@@ -22,40 +22,47 @@ Bordered toolbars are available in three variants: `primary`, `secondary`, and `
 
 ## Content position
 
-Toolbar supports two positioning models. Use `align` on `TooltrayNext` to place individual trays at the `start`, `center`, or `end` of the toolbar. Use `ToolbarContentNext` with the `position` prop to define a content section that holds one or more `TooltrayNext` components and optional dividers.
+Controls must be placed into `TooltrayNext` trays to participate in grouping and overflow behavior.
```
tycup, line 25: On defining Tooltrays - I do this here implicitly by expressing their function. Then the subsequent positioning models section builds upon their nuance and contextualises them to Toolbar.

site/docs/components/toolbar/examples.mdx
```diff
@@ -22,40 +22,47 @@ Bordered toolbars are available in three variants: `primary`, `secondary`, and `
...
+
+- Use `align` on `TooltrayNext` to place individual trays at the `start`, `center`, or `end` of the toolbar.
+- Use `ToolbarContentNext` with the `position` prop to define a content section that holds one or more `TooltrayNext`.
 
 <LivePreview componentName="toolbar" exampleName="ContentSections" />
 
 ## Dividers
 
-Use a [`Divider`](../divider) to separate groups of controls within a toolbar. Alternatively, apply `spacing-300` between groups when visual separation is enough.
+Use a [`Divider`](../divider) to separate groups of controls within a toolbar.
```
tycup, line 36: @dplsek - simplified this to be blunt and explicit lmk if sufficient.

site/docs/components/toolbar/examples.mdx
```diff
@@ -22,40 +22,47 @@ Bordered toolbars are available in three variants: `primary`, `secondary`, and `
...
 
 <LivePreview componentName="toolbar" exampleName="Dividers" />
 
 ## Overflow
 
+When the toolbar width is reduced and there isn't enough space for every control, trays move into overflow. Each tray collapses as a whole, keeping its controls grouped together.
```
tycup, line 42: @bhoppers2008 This builds upon the Tooltray definition. It is split from the aforementioned to contextualise it with the concept of Overflow.

I'm struggling a bit with the "into overflow". Overflow menu would sound better but its wrong; its still a toolbar, not a menu, but "tray moves into an overflow toolbar" sounds confusing - sub-toolbar also feels off.. let me know if you have a suggestion

tycup: This also serves to help define what an Overflow is in the context of Toolbar very plainly and directly

site/docs/components/toolbar/examples.mdx
```diff
@@ -22,40 +22,47 @@ Bordered toolbars are available in three variants: `primary`, `secondary`, and `
...
+
 ### Collapse order
 
-Use `overflowPriority` to define a custom collapse order. `TooltrayNext` components with higher `overflowPriority` values move into overflow before those with lower values.
+Use `overflowPriority` to define a custom collapse order. Higher priority values collapse sooner. Tooltrays with higher `overflowPriority` values move into the overflow before those with lower values.
 
 ### Independent collapse
 
-By default, `overflowMode` is `"independent"`, which moves each `TooltrayNext` into an overflow individually. When the toolbar's available width shrinks and the content no longer fits, `TooltrayNext` components move into an overflow menu.
+By default, `overflowMode` is `"independent"`, so each `TooltrayNext` collapses on its own. As the toolbar width is reduced, trays move into the overflow one at a time, following their `overflowPriority`.
```
tycup, line 50: @bhoppers2008 removed the generic use of the word "component". Simplified the language

site/docs/components/toolbar/examples.mdx
```diff
@@ -22,40 +22,47 @@ Bordered toolbars are available in three variants: `primary`, `secondary`, and `
...
 
 <LivePreview componentName="toolbar" exampleName="IndependentCollapse" />
 
-### Grouped collapse
+### Named overflows
 
-When `overflowMode` is set to `"grouped"`, controls in the same `TooltrayNext` move into overflow together. Use this when controls are interdependent and need to remain together to work effectively.
+By default, trays collapse into a single shared overflow. Assign a named `overflowGroup` to route trays into their own overflow, each with its own overflow trigger. Use `overflowLabel` when the overflow trigger label should differ from the group name.
```
tycup: Pushing back on 'trigger'. It _is_ a trigger and we use this vocabulary elsewhere too - can you suggest an edit if you disagree? @bhoppers2008

site/docs/components/toolbar/examples.mdx
aimee-liou:
```suggestion
Toolbar supports two appearances: `bordered` or `transparent`. The default appearance is `bordered`.
```
Wondering if simplifying could be an option here, Tabs and Tags are following suit.

site/docs/components/toolbar/examples.mdx
bhoppers2008: This reads much better now. Only one minor suggestion... "move" rather than "route"

bhoppers2008: Also... you can move the last part of the text "Use 'overflowLabel'... group name" into the following Best practice. It's a continuation of the same point.

site/docs/components/toolbar/examples.mdx
tycup: i agree in principle but to @bhoppers2008 feedback, this sentence also serves to help the user make a better informed decision on why they might use transparent over bordered (especially since appearance transparent actually does more than just make the container transparent)

site/docs/components/toolbar/examples.mdx
tycup: I've removed the old best practice point since it was a bit too obvious (and thus a tad redundant) and moved the last line to that section instead; it conveys the same point but with functional vocab

joshwooding (approved):

bhoppers2008 (approved):

PR #6632, opened 2026-06-05
[MEGA MENU][REFACTOR]- refactor, fixes, enhancements 

kriskoder:


tycup:
Should focus go to the first item in the megamenu on open?

tycup:
- If I open a menu with enter,
- then move focus into the menu using the tab/down arrow, such that im on "Digital Banking",
- and then press the right key till im on "document management"
- thenclick the right key one more time
- the menu closes and focus moves to the "Services" megamenu trigger

Is that expected? i assume focus should remain inside the menu and move to the next row

tycup:
Can't move focus to the items in the example with adornments (i presume because they don't contain a link) - should we expect this to be focusable?

tycup:
Screenshot 2026-06-08 at 11 29 39

Might be on your Todo; i presume the container with the links should be aligned

tycup:
Screenshot 2026-06-08 at 11 35 58
Can we make this example responsive too?

.changeset/mega-menu-api-updates.md
wemmyo, line 1: Some of the component names don't align with what we have on Figma

wemmyo: Is this changeset update to date? I don't see some of the components mentioned used anywhere such as `MegaMenuMain`, `MegaMenuFooter` and it's not being exported

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // Wear the id shared by the group so the list can reference it.
+  const { headingId } = useMegaMenuGroup() ?? {};
+
+  return (
+    <div
+      className={clsx(withBaseName(), className)}
+      ref={ref}
+      {...rest}
+      id={headingId}
```
tycup: Since this is a publicly exposed component, i'd want to be able to provide an id as a consumer

packages/lab/src/mega-menu/MegaMenuPanel.tsx
tycup: not sure how i feel about appending the probe into referenceEl which could be a  button- so you get a temporary div in a button - i guess its harmless but feels a bit off

tycup:
I can open a menu with a mouse and if i then Shift+Tab, it remains open

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,54 @@
...
+    window: targetWindow,
+  });
+
+  // Labelled by the group's heading via the shared id.
+  const { headingId } = useMegaMenuGroup() ?? {};
+
+  return (
+    <ul
+      className={clsx(withBaseName(), className)}
+      aria-labelledby={headingId}
```
wemmyo: When the group heading isn't provided we'll have an a11y violation since `headingId` wouldn't exist. We need better fallbacks

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,54 @@
...
+>(function MegaMenuItemList({ children, className, ...rest }, ref) {
+  const targetWindow = useWindow();
+  useComponentCssInjection({
+    testId: "salt-mega-menu-item-list",
+    css: megaMenuItemListCss,
+    window: targetWindow,
+  });
+
+  // Labelled by the group's heading via the shared id.
+  const { headingId } = useMegaMenuGroup() ?? {};
```
wemmyo:
```suggestion
  const { headingId } = useMegaMenuGroup();
```
This isn't a common approach we use, might be better to provide defaults instead

packages/lab/src/mega-menu/MegaMenuItemList.tsx
wemmyo:
```suggestion
      aria-labelledby={clsx(headingId, ariaLabelledBy) || undefined}
```
You can consider this change

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -12,3 +15,12 @@
   background: var(--salt-container-primary-background);
   box-shadow: var(--salt-overlayable-shadow-popout);
 }
+
+/* Small viewport: stack side content and the navigation area top-to-bottom in
+   source order. The keyboard engine detects the stacked layout and degrades
+   arrows to a linear walk. */
+@media (max-width: 600px) {
```
wemmyo: Maybe we take a mobile first approach instead using `min-width`? That's more consistent with other components. I don't see `max-width` being used anywhere else

site/src/examples/mega-menu/index.module.css
wemmyo: BTW I don't think this aligns with our breakpoints

wemmyo: https://www.saltdesignsystem.com/salt/foundations/responsiveness/breakpoints

site/docs/components/mega-menu/accessibility-content-draft.md
wemmyo, line 1: BTW the site accessibility docs can be improved.
- No mention of `Space` and `Enter` keyboard interaction, but it opens the mega menu when focus is on the trigger
- "Up arrow / Down arrow / Left arrow / Right arrow" keyboard controls can most likely be split up

.changeset/mega-menu-api-updates.md
kriskoder: - rename implemented, apart from MegaMenuGroups - will update when updating mega-menu panel layout
- changeset updated

kriskoder:
@tycup 

> Should focus go to the first item in the megamenu on open?

Yes, that's expected. Additionally, space and enter open the panel and do not focus the first item.

> * If I open a menu with enter,
> * then move focus into the menu using the tab/down arrow, such that im on "Digital Banking",
> * and then press the right key till im on "document management"
> * thenclick the right key one more time
> * the menu closes and focus moves to the "Services" megamenu trigger
>
> Is that expected? i assume focus should remain inside the menu and move to the next row

Good catch — only the last item from the last column should perform that action. Fixed.

> Can't move focus to the items in the example with adornments (i presume because they don't contain a link) - should we expect this to be focusable?

Fixed.

> Screenshot 2026-06-08 at 11 29 39
> Might be on your Todo; i presume the container with the links should be aligned

Fixed.

> I can open a menu with a mouse and if i then Shift+Tab, it remains open

Fixed.

site/docs/components/mega-menu/examples.mdx
honey-chang, line 29: On the Solutions mega menu, some items include icons while others do not. Could you remove all icons so that all mega menu items are consistently aligned.

site/docs/components/mega-menu/examples.mdx
```diff
@@ -66,12 +66,12 @@ You can display adornments such as tags, badges, or icons. This allows you to in
 
 ## With link
 
-You can add a link to give users a quick way to access a broader section or view all related items. Use `MegaMenuSupportingActions` to group one or more supporting links beneath the menu sections.
+You can add a link to give users a quick way to access a broader section or view all related items. Place a `MegaMenuSupportingActions` inside `MegaMenuBody`, after the `MegaMenuGroup`s, to render one or more supporting links as a full-width band beneath the menu groups.
 
 <LivePreview componentName="mega-menu" exampleName="WithLink" />
```
honey-chang, line 64: Could you check the width of mega menu of this example? I believe it should hug the content.

site/docs/components/mega-menu/examples.mdx
```diff
@@ -66,12 +66,12 @@ You can display adornments such as tags, badges, or icons. This allows you to in
...
 
 ## With supporting content
 
-Content can be positioned on the left or right, allowing for flexible layouts alongside mega menu items. In small viewports, it reflows into the stacked mega menu in its original order, appearing at the top or bottom of the stack.
+Use `MegaMenuSupportingContent` for side content (it renders to the left of `MegaMenuBody` when placed before it, and to the right when placed after) and `MegaMenuSupportingActions` for full-width content, which sits inside `MegaMenuBody` beneath the menu groups. Position is driven by source order alone, with no wrapper scaffolding required. In small viewports, it reflows into the stacked mega menu in its original order, appearing at the top or bottom of the stack.
 
 <LivePreview componentName="mega-menu" exampleName="WithContent" />
```
honey-chang, line 70: Could you check the mega menu width? On wide screens, it appears that the mega menu expands to fill the full screen width.

site/docs/components/mega-menu/examples.mdx
honey-chang, line 41: The triggers are not centered within the live preview box.

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -1,6 +1,9 @@
 .saltMegaMenuPanel {
+  /* Single horizontal row: optional side content, the center navigation area,
+     optional side content. Position is derived from source order alone, so
+     authors need no wrapper scaffolding. */
   display: flex;
-  flex-wrap: wrap;
+  align-items: stretch;
```
tycup, line 7: Since we're chrome 86 and later, we can't use this :( https://caniuse.com/?search=stretch

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -1,6 +1,9 @@
...
   inline-size: var(--saltMegaMenuPanel-inlineSize, fit-content);
   max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth, calc(100vw - var(--salt-spacing-600))));
   max-height: 90%;
```
tycup: percentage height against an `auto` heigth containing block computes to `none`

packages/lab/src/mega-menu/MegaMenuPanel.tsx
tycup: do we mind that the consumer could pass their own onKeyDown and break things here? could we put `rest` into getFloatingProps() and spread that maybe?

tycup: Also the dom node is created on every resize event - could we use RAF to guard it ?

packages/lab/src/mega-menu/MegaMenuPanel.css
wemmyo: Is this correct? `align-items` has been supported since Chrome 29 https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-items#browser_compatibility

tycup: oh oops yeah confused with stretch applied to things like width

site/docs/components/mega-menu/accessibility.mdx
```diff
@@ -33,36 +33,82 @@ data:
 
 ## Keyboard interactions
 
+### Opening and closing
```
wemmyo: We usually just have a single keyboard interaction table so we shouldn't need to separate it here

site/docs/components/mega-menu/accessibility.mdx
wemmyo: Do we already offer this out of the box? If so, we may not need to include it here.
The Best Practices section is usually best for recommendations that aren’t available out of the box but are still advised, especially items that require users to make updates or configuration changes.

.changeset/mega-menu-api-updates.md
kriskoder: updated names to match design

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuPanel.tsx
kriskoder: fixed buy moving the probe to sibling level, but general logic stayed mostly the same

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: fixed, if no heading or any other element providing label, there'll be no aria-labelledby attribute rendered

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: implemented

packages/lab/src/mega-menu/MegaMenuPanel.css
kriskoder: fixed, switched to mobile first

site/src/examples/mega-menu/index.module.css
kriskoder: aligned with out docs

site/docs/components/mega-menu/accessibility-content-draft.md
kriskoder: improved, added more details

site/docs/components/mega-menu/examples.mdx
kriskoder: removed

site/docs/components/mega-menu/examples.mdx
kriskoder: fixed

site/docs/components/mega-menu/examples.mdx
kriskoder: fixed

site/docs/components/mega-menu/accessibility.mdx
kriskoder: fixed

site/docs/components/mega-menu/examples.mdx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuItem.css
```diff
@@ -32,21 +31,21 @@
...
 }
 
-.saltMegaMenuItem > a:focus-visible {
+.saltMegaMenuItem:focus-visible {
   outline: var(--salt-focused-outline);
   outline-offset: calc(var(--salt-size-fixed-100) * -2);
 }
 
-.saltMegaMenuItem > a:hover {
+.saltMegaMenuItem:hover {
```
tycup: This competes with :focus-visible above (so if my mouse just so happens to be over an item when its focused, we lose focus styling)

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -33,9 +26,9 @@ export interface MegaMenuItemProps
   render?: RenderPropsType["render"];
 }
 
-export const MegaMenuItem = forwardRef<HTMLLIElement, MegaMenuItemProps>(
+export const MegaMenuItem = forwardRef<HTMLAnchorElement, MegaMenuItemProps>(
```
tycup: Typed as anchor element, but can be rendered as a button

packages/lab/src/mega-menu/MegaMenuPanel.tsx
kriskoder: fixed - removed the resize listener and moved the logic into usePageMargin, which only re-measures on reference/density change. Since --salt-layout-page-margin is density-driven, not viewport-driven, resize never changed the value anyway. let me know what you think

kriskoder: fixed by using `rest` with `getFloatingProps()` - let me know your thoughts

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -1,14 +1,18 @@
...
-  flex-wrap: wrap;
+  align-items: stretch;
   inline-size: var(--saltMegaMenuPanel-inlineSize, fit-content);
-  max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth, calc(100vw - var(--salt-spacing-600))));
+  max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth));
   max-height: 90%;
   overflow-x: hidden;
   overflow-y: auto;
   box-sizing: border-box;
   border-radius: var(--salt-palette-corner);
```
tycup: I've seen that often (but not always), our components have a `0` fallback here - i guess because of corner radius option in salt-theme-next? is it just an inconsistency that some don't have it ? @joshwooding

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,79 @@
...
+`MegaMenuPanel` lays its children out as a single horizontal row. `MegaMenuBody`
+is the center navigation area; an optional `MegaMenuSupportingContent` renders
+alongside it — to the left when placed before `MegaMenuBody`, to the right when
+placed after. Position is derived from source order alone — there are no
+placement props.
+
+- **`MegaMenuBody`** — the center navigation area. It stacks a `MegaMenuGroups`
+  and an optional `MegaMenuSupportingActions` band beneath it.
+- **`MegaMenuGroups`** — arranges the `MegaMenuGroup` columns in a grid that hugs
+  its content. Column width is set with the `--saltMegaMenuGroups-columnWidth`
```
tycup: Could it be worth mentioning `--saltMegaMenuPanel-availableWidth` here too

packages/lab/src/mega-menu/MegaMenuPanel.css
kriskoder: stretch sizes the columns to the tallest one without needing a definite container height, should be good here, let me know what do you think

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
```diff
@@ -0,0 +1,54 @@
...
+  }, [id, setHeadingId]);
+
+  return (
+    <div
+      className={clsx(withBaseName(), className)}
+      ref={ref}
+      {...rest}
+      id={id}
+    >
+      <div className={clsx(withBaseName("content"))}>{children}</div>
```
tycup:
```suggestion
      <div className={withBaseName("content")}>{children}</div>
```

site/docs/components/mega-menu/examples.mdx
tycup: would avoid abbreviation here "nav" - "use an alternative navigation component" (but also, is a drawer or accordion intended to be used for navigation in the first place..?)

site/docs/components/mega-menu/usage.mdx
```diff
@@ -30,6 +30,15 @@ data:
...
 - If any top-level item requires a mega menu, use the mega menu for all top-level items in that navigation system.
 
+### Opening and closing
+
+- The mega menu opens on click, not on hover, so it does not expand unexpectedly as the pointer moves across the navigation.
+- It closes when the trigger is clicked again, an item is selected, the user clicks outside the panel, or the user presses `Escape`.
+
+### Grouping items
+
+- Give each `MegaMenuGroup` a single `MegaMenuGroupHeading` and keep it to around 10 items at most; split larger sets into additional groups so the menu stays scannable.
```
tycup: Is the around 10 items more of a "best practice"?

packages/lab/src/mega-menu/MegaMenuPanel.css
tycup: yeah, the stretch part is fine (for handling columns without a definite height) - i mean more that the panels containing block (the floating wrapper) has auto height, so the percentage resolves to `none` and overflow-y: auto doesn't do much (tall panels would overflow the viewport).

basically:
- max-height 90% on the panel tries to cap itself against the wrapper
- wrapper height is 'whatever the panel content needs'
- so the percent doesnt do anything

you might want to bring back the size middleware but for height, put that in a css variable and use max-block-size instead

site/docs/components/mega-menu/accessibility.mdx
kriskoder: good to know, implemented

packages/lab/src/mega-menu/MegaMenuPanel.css
kriskoder: good catch, I used different approach to solve it: Merged the panel into the floating element so now it resolves against the viewport; I've set it to `90dvh`. Extra win: removed the inner .salMegaMenuPanel wrapper `<div>`. let me know what you think

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: fixed, used the same approach as in `VerticalNavigationItemTrigger `

packages/lab/src/mega-menu/MegaMenuPanel.css
kriskoder: good catch, added fallback that's consistent with dominant core convention

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuItem.css
kriskoder: fixed

.changeset/mega-menu-api-updates.md
kriskoder: good advice, added

site/docs/components/mega-menu/usage.mdx
kriskoder: changed to be less specific about numbers

site/docs/components/mega-menu/examples.mdx
kriskoder: updated

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+   * wrapped in its own `<li>`.
+   */
+  children?: ReactNode;
+  /**
+   * Render prop to customize the underlying list element. Defaults to a `<ul>`;
+   */
+  render?: RenderPropsType["render"];
+}
+
+export const MegaMenuItemList = forwardRef<
```
mark-tate: We changed the naming used elsewhere.
Elsewhere we have ListItem and List (floating-ui hook and also FileUpload.

So can this / should this  be MegaMenuList & MegaMenuListItem ?

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  HTMLUListElement,
+  MegaMenuItemListProps
+>(function MegaMenuItemList(
+  { children, className, "aria-labelledby": ariaLabelledBy, render, ...rest },
```
mark-tate: HTMLAttributes omits ref/key/dataset, should this be ComponentPropsWithoutRef?

Also, MegaMenuItemList types as HTMLUListElement but exposes a render prop that can swap the element — the type then lies

packages/lab/src/mega-menu/MegaMenu.tsx
mark-tate: Layout and the page-margin probe assume horizontal flow under the trigger. top, left*, right* will visually break: the panel is sized via inline-size/max-inline-size against viewport width and a flip middleware with horizontal padding. Either:
Narrow the type to "bottom" | "bottom-start" | "bottom-end" (and possibly top-*), or
Drop the prop entirely (mega menus are conventionally bottom-anchored).

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -47,37 +44,30 @@ export const MegaMenuItem = forwardRef<HTMLLIElement, MegaMenuItemProps>(
...
-          )}
-        </ItemAction>
-      </li>
+      <MegaMenuItemAction
+        className={clsx(withBaseName(), className)}
+        ref={ref}
+        href={href}
+        onClick={handleClick}
+        render={render ?? (isLink ? undefined : <button type="button" />)}
+        {...rest}
```
mark-tate: do we support a disabled item ?

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const id = useId(idProp);
+  const { setHeadingId } = useMegaMenuGroup();
+  useIsomorphicLayoutEffect(() => {
+    setHeadingId(id);
+    return () => setHeadingId(undefined);
+  }, [id, setHeadingId]);
+
+  return (
+    <div
```
mark-tate: Co Pilot reporting

A. MegaMenuGroupHeading is a `<div>`. It should expose a heading role. Today the wrapper is `<div>` containing another `<div>`. Even with aria-labelledby on the list, screen-reader users lose heading navigation (H quick-keys). Render as `<h2>/<h3>` (configurable level) or at minimum role="heading" aria-level={n}. Also: the double `<div>` (saltMegaMenuGroupHeading > saltMegaMenuGroupHeading-content) is purely a styling nest — could collapse to one if the heading became semantic.

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -47,37 +44,30 @@ export const MegaMenuItem = forwardRef<HTMLLIElement, MegaMenuItemProps>(
       window: targetWindow,
     });
 
-    const handleClick = (event: MouseEvent<HTMLAnchorElement>) => {
+    const handleClick = (event: MouseEvent<never>) => {
       onClick?.(event);
       megaMenu.setOpen(false);
```
mark-tate: Also the MouseEvent<never> type is wrong; should be MouseEvent<HTMLAnchorElement | HTMLButtonElement>.

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -17,12 +15,11 @@ import { useMegaMenu } from "./useMegaMenu";
 const withBaseName = makePrefixer("saltMegaMenuItem");
 
 // biome-ignore lint/suspicious/noExplicitAny: We don't know the exact type here
-function ItemAction(props: ComponentPropsWithoutRef<any>) {
+function MegaMenuItemAction(props: ComponentPropsWithoutRef<any>) {
   return renderProps("a", props);
 }
 
-export interface MegaMenuItemProps
```
mark-tate, line 24: H. aria-current. Tests assert "does not persist item active state after selection" — but a navigational mega menu should mark the active route with aria-current="page". This is consumer-controlled today (via render={<RouterLink>}), but the component offers nothing. Consider an active or current prop on MegaMenuItem (matches NavigationItem).

packages/lab/src/mega-menu/index.ts
mark-tate, line 1: CoPilot comments on Context 

1. panelId could be derived in MegaMenu (const id = useId() once) and passed down read-only, removing the setPanelId round-trip and the mount-cycle race where aria-controls is briefly absent.
2. focusFirstItemOnOpen is effectively imperative; it should live in a ref, not state. Today it triggers re-renders for a value that no rendered output depends on.
3. setOpen's useCallback dependency list omits setOpenState (from useControlled); the existing eslint rule allows it because it's stable in Salt's hook, but it'd be safer to include.

packages/lab/src/mega-menu/useMegaMenuNavigation.ts
```diff
@@ -0,0 +1,376 @@
...
+        return true;
+      }
+      case "ArrowRight": {
+        if (colIndex < columns.length - 1) {
+          firstFocusable(columns[colIndex + 1])?.focus();
+        } else if (
+          rowIndex === cells.length - 1 &&
+          focusNextTriggerAndClose(context)
+        ) {
+          // Last item of the last column with a next trigger: exit + close.
```
mark-tate: Smelly empty condition ?

could be something like .

  if (rowIndex === cells.length - 1 && focusNextTriggerAndClose(context)) return true;
  focusTrigger(context);

packages/lab/src/mega-menu/index.ts
mark-tate, line 1: The mega menu doesn't participate in FloatingTree (unlike Menu which wraps in <FloatingTree> when there's no parent node id). If a mega menu were ever nested in another floating overlay (e.g. inside a modal), useDismiss outside-press behavior could fight with the parent. Probably out-of-scope, but worth a note in the component's docstring.

Maybe need a document update to make it clear where it can/should be used. (if not already there?)

site/docs/components/mega-menu/usage.mdx
mark-tate, line 1: Docs say (twice — for triggers and for link lists):
"Use a `<ul>`, since the triggers have no inherent order. Reserve <ol> for cases where the sequence is meaningful…"
Every site example, every Storybook story, and the e2e tests use:
`<StackLayout as="ol" direction="row" gap={1} …>`
`  <li>…</li>`
And MegaMenuItemList itself renders `<ul>` (✅ that half matches the link-list guidance — though MegaMenuItemList has a render prop allowing override, which is fine).
Action: pick one. Either flip all examples to as="ul", or update the docs to match what the examples actually show.

mark-tate, line 1: Copilot says 

Shift + Tab (first item)
"moves focus to the previous focusable element outside of the mega menu"
Wrong — code focuses the trigger and keeps the menu open (verified by MegaMenu.accessibility.cy.tsx test "returns focus to the trigger when Shift+Tab passes the first item, keeping the menu open")

mark-tate, line 1: Shift + Tab (on open trigger)
Not documented
Closes menu, lets browser default move to previous focusable
❌ Missing

mark-tate, line 1: Up arrow (stacked)
Not documented
handleLinear falls back to ArrowUp = previous, ArrowUp on first = trigger
❌ Missing

mark-tate, line 1: Left arrow between triggers
Not documented
Trigger keydown handler closes current menu, moves to previous trigger
❌ Missing

mark-tate, line 1: Home / End (stacked)
Not documented; doc text implies group-scoped
Code in handleLinear: cells[0]/cells[cells.length - 1] of whole panel, not group
❌ Inaccurate when stacked
Self-consuming controls (input/combobox/etc.)
Not documented
Code yields arrows/Home/End to them
❌ Missing (consumers placing an input inside MegaMenuSupportingContent will wonder why arrows behave differently)
Static/non-interactive supporting content
Doc says "The custom region itself is not focusable"
✅ correct, but doesn't say static content contributes no cells
⚠️ partial

mark-tate, line 1: 4.2 "Full width" prose vs example
Docs:
"use the following CSS: --saltMegaMenuPanel-inlineSize: var(--saltMegaMenuPanel-availableWidth)"
index.module.css:
.fullWidthPanel { --saltMegaMenuPanel-inlineSize: var(--saltMegaMenuPanel-availableWidth); }
.fullWidthPanel :global(.saltMegaMenuGroups) {
  --saltMegaMenuGroups-columnWidth: 16rem;
  align-self: stretch;
}
The example also widens columns to 16rem and stretches the grid. Docs imply the one variable is enough; reality requires the column-width override to actually fill the panel.

mark-tate, line 1: Edge to edge" prose vs example — biggest documentation gap
Docs say:
--saltMegaMenuPanel-inlineSize: 100vw;
--saltMegaMenuPanel-maxInlineSize: 100vw;
index.module.css actually requires four additional pieces the docs don't mention:
.edgeToEdgeWrapper { --saltMegaMenuPanel-pageMargin: 0; }   /* load-bearing */
.edgeToEdgePanel :global(.saltMegaMenuBody) { max-inline-size: 100%; }
.edgeToEdgePanel :global(.saltMegaMenuGroups) {
  --saltMegaMenuGroups-columnWidth: 212px;
  max-inline-size: 920px;
  margin-inline: auto;
}

mark-tate, line 1: 4.5 "With adornment" example — dead CSS
index.module.css:
.menuItemFullWidth > a { inline-size: 100%; }
The adornment items in WithAdornment.tsx (e.g. Telemedicine, Production planning, Training, In-person, Release notes, Community forum) omit both render and href, so they render as <button>. The > a selector never matches; only adornment items that do have a render={<Link>} or href get the full-width treatment. The other items collapse to content width and the badge/tag lands beside the text rather than at the row's end.

mark-tate, line 1: 7. Documented behaviours that depend on undocumented contracts
Several pieces of behaviour described in examples.mdx/accessibility.mdx work only because consumers wrap their triggers a specific way. None of this is in the docs:
<li> wrapping is required for inter-trigger arrow keys and "exit + next trigger" behaviour. getAdjacentTrigger and focusNextTriggerAndClose both call closest("li"). If a consumer follows the docs' <ul> recommendation in 3.1 but drops the <li> wrappers, arrow navigation between triggers silently breaks.
Source order determines MegaMenuSupportingContent left/right position. Documented for examples but the panel's slot contract is not stated in the panel's own description.
MegaMenuSupportingActions must live inside MegaMenuBody. Documented in "With link" but not in the MegaMenuSupportingActions props description.
⚠️ These three contracts should be hoisted into usage.mdx ("Structure" section) and into the relevant component's prop description (which today is just "The content of …").

packages/lab/src/mega-menu/usePageMargin.ts
```diff
@@ -0,0 +1,26 @@
+import { useDensity, useIsomorphicLayoutEffect } from "@salt-ds/core";
+import { useState } from "react";
+
+/**
+ * Resolves the panel's page-margin token to pixels for floating-ui padding, to
+ * handle token (`--saltMegaMenuPanel-pageMargin`) override by consumers
+ * (edge to edge example).
+ */
+export function usePageMargin(
```
mark-tate, line 9: is this needed ?

1. Replace usePageMargin with the size middleware — biggest win
Today usePageMargin.ts injects a hidden <div> next to the trigger on every density change, reads getBoundingClientRect().width, and feeds that pixel value into flip({ padding }) and shift({ padding }). This is the single smelliest piece of the package.
Floating-ui's size middleware gives you the real available width/height as computed by the layout engine, with no DOM probe:

```
import { size, offset, flip, shift, limitShift } from "@floating-ui/react";

useFloatingUI({
  rootContext: floatingRootContext,
  open: isOpen,
  placement,
  middleware: [
    offset(1),
    flip({ padding: { left: pageMarginPx, right: pageMarginPx } }), // or drop entirely
    shift({ padding: { left: pageMarginPx, right: pageMarginPx }, limiter: limitShift() }),
    size({
      padding: pageMarginPx, // <- accepts a `Padding`; could be a number from a CSS var read once
      apply({ availableWidth, availableHeight, elements }) {
        Object.assign(elements.floating.style, {
          maxInlineSize: `${availableWidth}px`,
          maxBlockSize: `${Math.min(availableHeight, window.innerHeight * 0.9)}px`,
        });
      },
    }),
  ],
});
```

packages/lab/src/mega-menu/index.ts
mark-tate, line 1: Did we consider matching the menu behaviour?

 Add useTypeahead — accessibility parity with Menu
Floating-ui's useTypeahead gives first-letter / multi-character matching. The core Menu uses it; mega-menu doesn't. For a navigation pattern with ≥20 links across columns, typing "p" to jump to "Patient management" is a real win.

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,83 @@
...
+
+Updated `MegaMenu` with a revised component structure and API.
+
+`MegaMenuPanel` lays its children out as a single horizontal row. `MegaMenuBody`
+is the center navigation area; an optional `MegaMenuSupportingContent` renders
+alongside it — to the left when placed before `MegaMenuBody`, to the right when
+placed after. Position is derived from source order alone — there are no
+placement props.
+
+- **`MegaMenuBody`** — the center navigation area. It stacks a `MegaMenuGroups`
```
joshwooding: I'm not really sure about the "body" terminology, we've not used it in the past.

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,83 @@
...
+  `<li>`.
+- **`MegaMenuItem`** — the focusable item. It renders an `<a>` when given an
+  `href` or a `render` element (such as `react-router`'s `Link`), and a
+  `<button>` otherwise; it closes the menu on activation.
+- **`MegaMenuSupportingContent`** — side content beside `MegaMenuBody`, placed to
+  the left or right by source order.
+- **`MegaMenuSupportingActions`** — a row of supporting actions inside
+  `MegaMenuBody`, beneath the groups and spanning their width.
+
+Responsive behaviour is left to consumers. `MegaMenuGroups` lays its columns out
```
joshwooding: Has the figma component been updated to do the same? @honey-chang

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  extends HTMLAttributes<HTMLUListElement> {
+  /**
+   * The items of the group, typically `MegaMenuItem` components. Each child is
+   * wrapped in its own `<li>`.
+   */
+  children?: ReactNode;
+  /**
+   * Render prop to customize the underlying list element. Defaults to a `<ul>`;
+   */
+  render?: RenderPropsType["render"];
```
joshwooding: Why do we need a render prop here?

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+  type HTMLAttributes,
+  type ReactNode,
+} from "react";
+import { useMegaMenuGroup } from "./MegaMenuGroupContext";
+import megaMenuItemListCss from "./MegaMenuItemList.css";
+
+const withBaseName = makePrefixer("saltMegaMenuItemList");
+
+export interface MegaMenuItemListProps
+  extends HTMLAttributes<HTMLUListElement> {
```
joshwooding: We should be using ComponentPropsWithoutRef for consistency in out prop interfaces.

packages/lab/src/mega-menu/MegaMenuItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  const { headingId } = useMegaMenuGroup();
+
+  return renderProps("ul", {
+    className: clsx(withBaseName(), className),
+    "aria-labelledby": clsx(headingId, ariaLabelledBy) || undefined,
+    ref,
+    render,
+    ...rest,
+    children: Children.map(children, (item, index) => (
```
joshwooding: Not a huge fan of mapping children but if this is the nicest API, maybe it's fine

packages/lab/src/mega-menu/MegaMenuPanel.css
```diff
@@ -1,14 +1,18 @@
...
+     authors need no wrapper scaffolding. */
+  --saltMegaMenuPanel-availableWidth: calc(100vw - var(--salt-spacing-600));
   display: flex;
-  flex-wrap: wrap;
+  align-items: stretch;
   inline-size: var(--saltMegaMenuPanel-inlineSize, fit-content);
-  max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth, calc(100vw - var(--salt-spacing-600))));
-  max-height: 90%;
+  max-inline-size: var(--saltMegaMenuPanel-maxInlineSize, var(--saltMegaMenuPanel-availableWidth));
+  max-height: 90dvh;
```
joshwooding: dvh is not supported by our minimum browser

packages/lab/src/mega-menu/MegaMenuPanel.tsx
```diff
@@ -49,56 +51,41 @@ export const MegaMenuPanel = forwardRef<HTMLDivElement, MegaMenuPanelProps>(
...
-        host.appendChild(probe);
-        const width = probe.getBoundingClientRect().width;
-        probe.remove();
-        setPageMargin((prev) => (prev === width ? prev : width));
-      };
-      measure();
-      targetWindow.addEventListener("resize", measure);
-      return () => targetWindow.removeEventListener("resize", measure);
-    }, [targetWindow, floatingRootContext]);
+    const referenceEl = floatingRootContext.elements.domReference as
```
joshwooding: Rather than casting we could use a typeguard

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,83 @@
...
+with grid `auto-fit` and `minmax(0, …)`, so they shrink and wrap to the
+available width rather than overflowing; there is no built-in viewport
+breakpoint. To stack the panel on small screens, set `flex-direction: column` on
+`MegaMenuPanel` at your own breakpoint.
+
+The panel width is controlled with the `--saltMegaMenuPanel-inlineSize` (default
+`fit-content`) and `--saltMegaMenuPanel-maxInlineSize` (default
+`--saltMegaMenuPanel-availableWidth`) custom properties.
+
+Keyboard navigation follows the panel layout. Up and Down move within a column,
```
joshwooding: Has this changed?

.changeset/mega-menu-api-updates.md
joshwooding, line 1: This changelog doesn't seem to focus on what actually changed and more gives you a lay of the land

packages/lab/src/mega-menu/MegaMenuItem.tsx
```diff
@@ -47,37 +44,30 @@ export const MegaMenuItem = forwardRef<HTMLLIElement, MegaMenuItemProps>(
...
-            ),
-          )}
-        </ItemAction>
-      </li>
+      <MegaMenuItemAction
+        className={clsx(withBaseName(), className)}
+        ref={ref}
+        href={href}
+        onClick={handleClick}
+        render={render ?? (isLink ? undefined : <button type="button" />)}
```
joshwooding: I think we should just simplify this and make mega menu items links with the option to change with render, the main purpose is navigation and links would be more correct.

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
joshwooding: It's likely we don't need to do this @jake-costa

packages/lab/src/mega-menu/index.ts
joshwooding: I think it's fine this is out of scope for now

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: renamed, also refactored components structure for a clear responsibility split and more transparent api, have a look

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: refactored to make it polymorphic

packages/lab/src/mega-menu/MegaMenu.tsx
kriskoder: reduced placement to "bottom" | "bottom-start" | "bottom-end", agree, top placement can be added later if needed

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
kriskoder: collapsed structure and changed component to render `<h->` tag instead of `<div>`, set default `h` level to 3 , let me know what do you think

packages/lab/src/mega-menu/MegaMenuPanel.css
kriskoder: changed to `vh` and to `max-block-size` for consistency

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: added `active` prop - matches `NavigationItem`, let me know what you think

packages/lab/src/mega-menu/index.ts
kriskoder: 1. MegaMenu now generates the default id itself
2. moved to `ref`
3. it follows convention from core package, `useState` is referentially stable, there's no need to implement that recommendation

packages/lab/src/mega-menu/useMegaMenuNavigation.ts
kriskoder: fixed

site/docs/components/mega-menu/usage.mdx
kriskoder: in docs, moved to one place **Accessibility semantics**
all examples are using `<ul>` now

kriskoder: updated to make `Shift + Tab` interaction more transparent and aligned with code

kriskoder: added that case to docs

kriskoder: updated selector to `.menuItemFullWidth > :global(.saltMegaMenuListItem-wrapper)`

packages/lab/src/mega-menu/index.ts
kriskoder: because mega-menu is using a custom nav engine, using `useTypeahead` would complicate things. I can try re-writing the engine to use `useTypeahead ` or implement custom implementation, that will be relatively small update with current keyboard navigation engine

site/docs/components/mega-menu/usage.mdx
kriskoder: updated both in `usage.mdx` and in props

kriskoder: Documented

kriskoder: Documented

packages/lab/src/mega-menu/index.ts
kriskoder: caveat added to the MegaMenu JSDoc

site/docs/components/mega-menu/usage.mdx
kriskoder: All three documented

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: good shout, simplified as suggested

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: The render prop was removed in a later refactor — it's now a polymorphic as prop. We're keeping element customization (ul/ol) to meet an accessibility requirement

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: removed by earlier refactor

packages/lab/src/mega-menu/MegaMenuItemList.tsx
kriskoder: good catch, implemented here and across other files

packages/lab/src/mega-menu/MegaMenuPanel.tsx
kriskoder: fixed

packages/lab/src/mega-menu/usePageMargin.ts
kriskoder: The probe resolves the density-aware page-margin token (`--salt-layout-page-margin`) to px for `flip`/`shift` `padding`, which keeps a symmetric gutter from the viewport edges. It's hard to remove because floating-ui `padding` needs a number, Salt has no JS token values, and `getComputedStyle` can't resolve a `var()`/`calc()` chain reliably — so a layout probe is the only robust token→px path. `size` middleware doesn't help (its `padding` needs the same px). The only clean removal is to split axes — floating-ui for vertical, CSS `margin-inline` for the horizontal gutter — but that requires the panel to be page-aligned rather than trigger-anchored, which is a design decision. So it's a positioning-model refactor + that decision, not a quick change. Suggest keeping the probe and doing this as a follow-up.

**What I tried**: a `Dialog`-style `size` prop (`content`/`available`/`edge-to-edge`) to replace the CSS-var sizing contract, with the gutter moved into CSS (dropping the `flip`/`shift` `padding` and reserving the margin via `max-inline-size`). The prop cleaned up the width API, but it didn't remove the probe — width modes and the edge gutter are separate problems. And the gutter-in-CSS regressed: with no `shift` `padding`, near a viewport edge the panel slid flush to one side (gutter on one side, zero on the other) instead of staying symmetric. So I reverted the whole thing.

.changeset/mega-menu-api-updates.md
kriskoder: Rewrote changeset into a focused "what changed"

.changeset/mega-menu-api-updates.md
kriskoder: renamed:
MegeMenuBody -> MegeMenuContent
MegaMenuSupportingActions -> MegaMenuActions
MegaMenuSupportingContent -> MegaMenuAside

.changeset/mega-menu-api-updates.md
kriskoder: trimmed changelog to show only what changed

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: Added a disabled prop. Follows the core convention (MenuItem, Option, etc.): aria-disabled rather than native disabled - **to be confirmed with design**

kriskoder: got reply from design, disabled is not supported for now

site/docs/components/mega-menu/usage.mdx
kriskoder: updated

kriskoder: updated

packages/lab/src/mega-menu/MegaMenuTrigger.tsx
```diff
@@ -113,12 +116,11 @@ export const MegaMenuTrigger = forwardRef<HTMLElement, MegaMenuTriggerProps>(
...
-          // For NavigationItem, sync via the `expanded` prop (which maps to aria-expanded internally).
-          // For all other elements, set aria-expanded directly.
+          // NavigationItem maps `expanded` to aria-expanded; others get it directly.
           ...(shouldSyncExpanded
             ? { expanded: openState }
             : { "aria-expanded": openState }),
-          ...(panelId && openState ? { "aria-controls": panelId } : null),
+          ...(openState ? { "aria-controls": panelId } : null),
           onKeyDown: handleKeyDown,
           ...rest,
```
wemmyo: Spreading rest last will override everything before, is this intended?

packages/lab/src/mega-menu/MegaMenuTrigger.tsx
```diff
@@ -14,11 +14,12 @@ import {
   useCallback,
 } from "react";
 import { useMegaMenu } from "./useMegaMenu";
-import { FOCUSABLE_SELECTOR, focusFirstItem } from "./useMegaMenuKeyboard";
+import { FOCUSABLE_SELECTOR, focusFirstItem } from "./useMegaMenuNavigation";
 
 export interface MegaMenuTriggerProps {
```
wemmyo, line 19: There's a type mismatch, only children declared here but we still spread props and consumers can pass the props in as well

packages/lab/src/mega-menu/MegaMenuItem.tsx
kriskoder: had to be removed -> confirmed with design, not supporting because of accessibility

packages/lab/src/mega-menu/MegaMenu.tsx
```diff
@@ -51,13 +61,14 @@ export function MegaMenu({
 
   const [reference, setReference] = useState<HTMLElement | null>(null);
   const [floating, setFloating] = useState<HTMLElement | null>(null);
-  const [focusFirstItemOnOpen, setFocusFirstItemOnOpen] = useState(false);
-  const [panelId, setPanelId] = useState<string | undefined>(undefined);
+  const focusFirstItemOnOpenRef = useRef(false);
+  const generatedId = useIdMemo();
```
wemmyo: Any reason for using `useIdMemo` over `useId`?

packages/lab/src/mega-menu/MegaMenuListItem.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  useComponentCssInjection({
+    testId: "salt-mega-menu-list-item",
+    css: megaMenuListItemCss,
+    window: targetWindow,
+  });
+
+  const handleClick: MouseEventHandler<HTMLAnchorElement> = (event) => {
+    onClick?.(event);
+    megaMenu.setOpen(false);
```
wemmyo, line 64: I believe you mentioned we're not supporting `disabled` but is it possible to prevent the mega menu from closing on click?

packages/lab/src/mega-menu/useMegaMenuNavigation.ts
```diff
@@ -0,0 +1,376 @@
...
+  const refFocusable = firstFocusable(reference) ?? reference;
+
+  const nextLi = refFocusable?.closest("li")?.nextElementSibling;
+  const nextSibling =
+    nextLi instanceof HTMLElement ? firstFocusable(nextLi) : null;
+
+  const nextOutside =
+    nextSibling ||
+    (() => {
+      const allFocusable = queryFocusables(
```
wemmyo, line 265: Co-pilot feedback: This walks every focusable in the page, then `indexOf` the trigger. On large pages (data grids, complex layouts) this is hundreds to thousands of nodes, all to find one element. It also runs inside two nested [requestAnimationFrame](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) calls, so the cost is reflowed.

Fix — use a TreeWalker that stops at the first match after the trigger, and short-circuit when we find it:

packages/lab/src/mega-menu/MegaMenu.tsx
kriskoder: `useIdMemo` returns a defined `id` on the first render, which we need because it seeds `useState`. `useId` returns `undefined` on the first render in React 16,17

kriskoder: **RISK** - `useIdMemo's` global counter is not SSR-safe (hydration-mismatch risk) - `React.useId` is, let me know if we're considering that

joshwooding: Do you actually need to seed useState here? Wouldn't a more robust way of doing this be having derived state for the final id? e.g.

const panelId = idProp ?? generatedId;

This way if for some reason the idProp is removed (edge case) the generated one will be used. Right now if that happens a stale id will be used (the old idProp).

Side note: we should be supporting SSR.

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
```diff
@@ -0,0 +1,61 @@
...
+  extends ComponentPropsWithoutRef<"h2"> {
+  /**
+   * The content of the mega menu group heading.
+   */
+  children?: ReactNode;
+  /**
+   * Heading level — renders the matching `<h1>`–`<h6>` element.
+   * @default 3
+   */
+  level?: 1 | 2 | 3 | 4 | 5 | 6;
```
wemmyo: Not a big fan of using the level prop, seems less composable. A more common approach we've used in other components is: 
```
<MegaMenuGroupHeading as="h2">Text</MegaMenuGroupHeading>
```
Makes it a bit clearer it maps to a `h2`. What do you think?

packages/lab/src/mega-menu/MegaMenuTrigger.tsx
kriskoder: good catch, I dropped `rest` operator entirely

packages/lab/src/mega-menu/MegaMenuTrigger.tsx
kriskoder: removed `rest` operator

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,44 @@
...
+- **Narrowed `MegaMenu`'s `placement` prop** to the new exported `MegaMenuPlacement` type (`"bottom" | "bottom-start" | "bottom-end"`); other floating-ui placements are no longer accepted.
+- **`MegaMenuGroups` columns now shrink to fit.** The grid uses `minmax(0, …)`, so columns shrink and wrap to the available width instead of overflowing.
+- **Adjusted the panel shadow** to `--salt-overlayable-shadow-modal` (was `--salt-overlayable-shadow-popout`).
+- **Reworked keyboard navigation.** The panel's navigation engine was rewritten, with these changes from the previous behaviour:
+  - `ArrowUp` at the top of a column now returns focus to the trigger, instead of wrapping to the previous column.
+  - `ArrowRight` on the rightmost column now exits to the next trigger (closing the menu) only from the last item, and otherwise returns focus to the trigger with the menu kept open.
+  - `MegaMenuActions` now navigates as a horizontal row (`ArrowLeft`/`ArrowRight` within it, `ArrowUp` back to the last column), rather than as a vertical column.
+  - `Shift+Tab` on an open trigger now closes the menu.
```
wemmyo: We can exclude implementation details from the changeset

.changeset/mega-menu-api-updates.md
```diff
@@ -0,0 +1,44 @@
+---
+"@salt-ds/lab": minor
+---
+
+Updated `MegaMenu` with a revised structure and API:
+
+- **Renamed `MegaMenuHeader` to `MegaMenuGroupHeading`.** It now renders a semantic heading (`<h3>` by default, configurable with `level`) instead of a `<div>`.
+- **Renamed `MegaMenuItem` to `MegaMenuListItem`.** It renders its own `<li>` wrapping the focusable action — an `<a>` by default, overridable with a `render` element (such as `react-router`'s `Link`). Its `ref` and DOM props target the `<li>`. It also adds an `active` prop that sets `aria-current="page"` for the current route.
```
wemmyo: Renaming the component is a breaking change making this a `major` not `minor` change

site/docs/components/mega-menu/accessibility.mdx
wemmyo, line 11: The best practices needs some work, not everything here is relevant. A couple of things that are relevant:
- Wrapping the mega menu in `<nav>`, but it shouldn't be under 'Triggering mega menu'
- Use of `ul` and `ol`, but accessibility semantics as the title seems vague
- Providing an accessible name for the mega menu panel region, the title should make that clear. Not just 'Mega menu panel'

site/docs/components/mega-menu/examples.mdx
wemmyo, line 1: Seeing a couple of code snippets within the docs. The docs are generally intended for both designers and engineers, the code snippets won't be relevant to designers so would be good to keep it minimal if necessary, the actual code preview should contain everything required. Since we're open source, engineers can look through the code anyway

.changeset/mega-menu-api-updates.md
kriskoder: don't we include changes to user facing behaviours?

wemmyo: Yes but not implementation detail. For example, "`Shift+Tab` closing the menu" can be framed as a "The menu now closes when focus leaves the trigger backwards" or something similar

packages/lab/src/mega-menu/useMegaMenuNavigation.ts
kriskoder: This only runs on a single Tab keypress in an edge case (last cell of the panel, when the trigger is the last item in the nav bar). `querySelectorAll` is native and the cost is sub-frame even on large pages, so I'd rather keep the simpler version that reuses our `queryFocusables` helper than add a TreeWalker (a new pattern here) to shave microseconds off a cold path. Let me know what do you think

packages/lab/stories/mega-menu/mega-menu.qa.stories.tsx
joshwooding, line 1: It's fine for now but we're missing a lot of scenarios in the visual regression story

packages/lab/src/__tests__/__e2e__/mega-menu/MegaMenu.cy.tsx
joshwooding, line 1: One minor comment, we should really see if we can use our storybook stories in our test files

.changeset/mega-menu-api-updates.md
kriskoder: @joshwooding what I figured out, convention here has been to ship breaking changes as `minor` — the only `major` in lab's history is the initial release, and the previous `MegaMenu` rename (this same changeset file) went out as `minor`. So I've kept `minor` for consistency. Let me know

packages/lab/src/mega-menu/MegaMenuGroupHeading.tsx
kriskoder: fixed

packages/lab/src/mega-menu/MegaMenuListItem.tsx
kriskoder: Not supported right now — mega menu is navigation-first, so close-on-click is the intended default. We can add an opt-out later without a breaking change.

site/docs/components/mega-menu/accessibility.mdx
kriskoder: improved

site/docs/components/mega-menu/examples.mdx
kriskoder: improved

packages/lab/src/mega-menu/MegaMenu.tsx
kriskoder: Good call — moved id ownership into MegaMenuPanel (useId(idProp), SSR-safe) and made MegaMenu just relay it, mirroring Collapsible; no more useState seeding or stale id, and useIdMemo is gone.

packages/lab/stories/mega-menu/mega-menu.qa.stories.tsx
kriskoder: good shout, I'm happy to do that after merging this PR

packages/lab/src/__tests__/__e2e__/mega-menu/MegaMenu.cy.tsx
kriskoder: good shout, I'm happy to do that after merging this PR

joshwooding (dismissed):

tycup (dismissed):

joshwooding (approved):

wemmyo (approved):

PR #6636, opened 2026-06-08
Toolbar Fixes | Combobox focus behaviour

tycup:
- Toolbar was not respecting a multiselect combobox internal keyboard navigation; in general i don't like how its done across other input types too because it becomes brittle to change
- Use stopPropagation but only to suppress re-entry onFocus event (focus temporarily always goes to the first focusable element in the toolbar and the toolbar decides to instantly move it to the remembered focus; we (carefully and intentionally) supress bubbling here (e.g. from the PillInput inside Combobox).

### DIFF
--- a/packages/lab/src/toolbar-next/toolbarNextKeyboardUtils.ts
+++ b/packages/lab/src/toolbar-next/toolbarNextKeyboardUtils.ts
@@ -466,7 +466,8 @@ function getToolbarNextKeyboardPolicy(
   if (
     target.isContentEditable ||
     isPlainTextInput(target) ||
-    isComboBoxInput(target)
+    isComboBoxInput(target) ||
+    isPillInputPill(target)
   ) {
     return {
       preserveHorizontalArrows: true,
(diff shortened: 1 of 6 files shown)


joshwooding (dismissed):
Do we need a test?

joshwooding (dismissed):

packages/lab/src/toolbar-next/toolbarNextKeyboardUtils.ts
```diff
@@ -571,14 +580,40 @@ export function isToolbarNextFocusable(
   return target.getClientRects().length > 0;
 }
 
+// Composite controls whose focus root (the element toolbar focus restore lands
+// on) differs from the inner element that handles keyboard interaction.
+// Restoring focus to the root would leave ArrowDown/Enter/Space targeting a
+// non-interactive wrapper, so we redirect to the interactive element
+const toolbarNextFocusEntryOverrides = [
```
tycup, line 587: Even thought its only got one element, this is here more to create some sort of separation and for future components that we might want to make compatible in toolbar

wemmyo (approved):

joshwooding (approved):

PR #6639, opened 2026-06-08
[FIX] - fix build | (date-components/date-picker): export overlay provider types

kriskoder:
as in PR title


joshwooding:
I raised this to @mark-tate the other day, I don't think we should be exporting context. It should ideally be removed before we merge to core

joshwooding (changes requested):
This change needs a changeset

PR #6655, opened 2026-06-10
Align AG Grid theme with Salt Table and fix Row Group Panel pill colors

### LINKED ISSUE #5391: Extended Styles for AG Grid Headers
We've received feedback from designers that there is a need for extended styles for Data Grid headings, ideally including: 

- Background variants
- Font weight variants

> since moving to Salt, we've gotten a lot of feedback from senior leadership, users, and PO's about the styling of the table headers. The feedback is that it's hard to distinct from the cell and row content, since the font stylings are so similar. Are there options here to customize these/make these more distinct/darken the header row?

> A few things that would help is making the header labels either bold or small in font style. It would help to also help to change the vertical height of the header so it is not identical to the table rows. 

> For example in macOS they do the following to differentiate:
The header is taller than the rows. A shorter header could also be appropriate as long as it is different size from the rows. If row size and header are too similar they feel like the same thing.
A more distinct division between the header and the rows. In the apple case there is a divider and the rest of the rows do not have dividers.
There is a space between the header than the rest of the items
The font in the header is smaller and lighter than the rows except for the one column that is the current sort order.
Another thing that helps that AG grid does in their demos is having a gray background to the header.

### LINKED ISSUE #6196: [AG Grid Theme] Row Group Panel pill text uses incorrect design token
# [AG Grid Theme] Row Group Panel pill text uses incorrect design token

## Description

The pill text in the AG Grid Row Group Panel is currently using an incorrect Salt Design System token, resulting in the wrong foreground color being applied.

## Current Behavior

The row group panel pill text color resolves to:

`--ag-secondary-foreground-color: var(--salt-content-primary-foreground);`

This applies a content foreground color to the pill, which does not accurately reflect the Salt tokens used in Salt Pill.

## Expected Behavior

The row group panel pill text and icons should use the actionable bold foreground token:

`--pill-color: var(--salt-actionable-bold-foreground);`

Since the row group pills are interactive/actionable elements, they should align with the Salt Design System's actionable bold foreground token to ensure visual consistency with other actionable components.

## Steps to Reproduce

1. Open the AG Grid Row Group Panel story in Storybook: [Row Group Panel Story](https://storybook.saltdesignsystem.com/?path=/story/ag-grid-ag-grid-theme--row-group-panel&globals=viewport.value:1515-896)
2. Inspect the pill element used for row grouping.
3. Observe that the text color is derived from `--salt-content-primary-foreground` instead of `--salt-actionable-bold-foreground`.

## Suggested Fix

Update the AG Grid theme CSS to map the pill text and icon color to the correct token:

Remove: `--ag-secondary-foreground-color: var(--salt-content-primary-foreground);`

Replace with/ or equivalent of: `--pill-color: var(--salt-actionable-bold-foreground);`

## Impact

- Pill text and icon color is inconsistent with other actionable elements in the Salt Design System.
- Causes visual inconsistencies across light/dark modes if the two tokens resolve to different values.

joshwooding:
Closes #6196
Closes #5391


packages/ag-grid-theme/css/parts/ag-buttons.css
```diff
@@ -17,15 +17,15 @@
...
+  background-color: var(--salt-actionable-subtle-background-hover);
+  color: var(--salt-actionable-subtle-foreground-hover);
 }
 
 .ag-theme-salt-light .ag-ltr .ag-filter-apply-panel-button,
 .ag-theme-salt-dark .ag-ltr .ag-filter-apply-panel-button,
 .ag-theme-salt-compact-light .ag-ltr .ag-filter-apply-panel-button,
 .ag-theme-salt-compact-dark .ag-ltr .ag-filter-apply-panel-button {
-  margin-left: 8px;
+  margin-left: var(--salt-spacing-100);
```
joshwooding, line 28: @bhoppers2008 is this change correct?

yunjungyeh (dismissed):

site/docs/components/table/examples.mdx
```diff
@@ -42,6 +42,10 @@ The `variant` prop can also be applied to the table header. This allows you to c
 
 The `divider` prop can be used to adjust the appearance of the dividers in the table. It accepts: `primary`, `secondary`, `tertiary` or `none`.
 
+### Best practices
+
+- Avoid using a less prominent header variant than the body variant, as this can create confusion for users.
```
tycup: nit; i just stumbled a few times reading the current version but could be me. "That is" might be clearer than "that's" even if goes against language guidelines.
```suggestion
- Avoid using a header variant that's less prominent than the body variant, as this can create confusion for users.
```

tycup (dismissed):

packages/ag-grid-theme/css/parts/ag-header-variants.css
```diff
@@ -0,0 +1,15 @@
+/* --- Background --- */
+div[class*="ag-theme-salt"].ag-theme-salt-header-primary {
```
tycup: robot helped: beats variant-* rules via order, but lose to zebra combos in `ag-row.css` which have higher specificity and also set --ag-header-background-color. So `variant-secondary + zebra + header-tertiary` would render a secondary heade

packages/ag-grid-theme/stories/ag-grid-theme.stories.tsx
```diff
@@ -33,6 +33,7 @@ export {
   DragRowOrder,
   FloatingFilter,
   HDCompact,
+  HeaderVariants,
```
tycup, line 36: should we add this also to the QA story?

packages/ag-grid-theme/css/parts/ag-header-variants.css
```diff
@@ -0,0 +1,15 @@
...
+  --ag-header-background-color: var(--salt-container-primary-background);
+  --ag-header-cell-hover-background-color: var(--ag-header-background-color);
```
tycup: do we need to do this if ag-root-var already aliases it

yunjungyeh:
Should the Header variants example in storybook be multi-select? it currently can only select one row

tycup (approved):

yunjungyeh (approved):

PR #6675, opened 2026-06-11
Toolbar | Move Toolbar to Core

tycup:


.changeset/mean-cars-poke.md
```diff
@@ -0,0 +1,42 @@
+---
+"@salt-ds/core": patch
```
joshwooding:
```suggestion
"@salt-ds/core": minor
```

.changeset/four-nails-jog.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/lab": minor
```
kriskoder, line 2: should it be major?

.changeset/mean-cars-poke.md
```diff
@@ -0,0 +1,42 @@
+---
+"@salt-ds/core": minor
```
kriskoder, line 2: should it be major?

kriskoder:
move looks clean, left 2 small comments

.changeset/mean-cars-poke.md
tycup: nope - we're not removing or breaking any existing functionality or exports - minor as new addition

.changeset/four-nails-jog.md
tycup: nope but more by nature of labs being labs - feels like it would be major because technically there is a breaking change but down to the way we treat labs as staging to core

kriskoder (dismissed):

site/docs/components/toolbar/usage.mdx
```diff
@@ -26,29 +26,29 @@ data:
...
-- When a position needs multiple `TooltrayNext` components, dividers, custom spacing, or content-level props, wrap them in `ToolbarContentNext` and set the `position` prop.
+- Always place controls inside `Tooltray`. Both composition models require it.
+- For simple toolbars, place `Tooltray` directly inside `Toolbar`. Use the `align` prop to position each tray at the `start`, `center`, or `end`.
+- When a position needs multiple `Tooltray` components, dividers, custom spacing, or content-level props, wrap them in `ToolbarContent` and set the `position` prop.
 - Use one composition model per toolbar.
 
 ## Import
 
-To import `Toolbar` from the lab Salt package, use:
+To import `Toolbar` from the core Salt package, use:
```
joshwooding:
```suggestion
To import `Toolbar` and related components from the core Salt package, use:
```

packages/core/src/toolbar/Tooltray.tsx
```diff
@@ -1,31 +1,31 @@
...
 import { useWindow } from "@salt-ds/window";
 import { clsx } from "clsx";
 import { type ComponentPropsWithoutRef, forwardRef } from "react";
+import { makePrefixer } from "../utils";
 
-import tooltrayCss from "./TooltrayNext.css";
+import tooltrayCss from "./Tooltray.css";
 
-export type TooltrayNextOverflowMode = "none" | "independent" | "grouped";
+export type TooltrayOverflowMode = "none" | "independent" | "grouped";
```
joshwooding: Minor: Do we need to export this out the root e.g. import { TooltrayOverflowMode } from "@salt-ds/core"

Or should it only be used internally or through TooltrayProps?

packages/core/src/toolbar/ToolbarContent.tsx
```diff
@@ -0,0 +1,42 @@
+import { useComponentCssInjection } from "@salt-ds/styles";
+import { useWindow } from "@salt-ds/window";
+import { clsx } from "clsx";
+import { type ComponentPropsWithoutRef, forwardRef } from "react";
+import { makePrefixer } from "../utils";
+
+import toolbarContentCss from "./ToolbarContent.css";
+
+export type ToolbarContentPosition = "start" | "center" | "end";
```
joshwooding: Minor: Do we need to export this out the root e.g. import { ToolbarContentPosition } from "@salt-ds/core"

Or should it only be used internally or through ToolbarContentProps?

packages/core/src/toolbar/Tooltray.tsx
tycup: Would keep exported out of root - user would want these (maybe for config or wrapper)

packages/core/src/toolbar/ToolbarContent.tsx
tycup: Would keep exported out of root - user would want these (maybe for config or wrapper)

packages/core/src/toolbar/Tooltray.tsx
joshwooding: @tycup our other components provide the same but through the props interface of the component

tycup: Fair - was looking at `FormFieldLabelPlacement` and `FlexAlignment` and some bits in Stepper. Are those inconsistent/older api?

packages/core/src/toolbar/ToolbarContent.tsx
tycup: updated

packages/core/src/toolbar/Tooltray.tsx
tycup: updated

joshwooding: Yeah, those are inconsistent. We've currently got a massive public API that we'll reduce in the next breaking change.

joshwooding (approved):

kriskoder (approved):

PR #6684, opened 2026-06-15
Highcharts | Add zoom reset styling

tycup:
- Looks to (imperfectly) match the style of Button
- Its an approximation because we can't pass the same properties to SVGs as we can otherwise (e.g. gradients)

### DIFF
--- a/.changeset/slimy-shrimps-wave.md
+++ b/.changeset/slimy-shrimps-wave.md
@@ -1,0 +1,7 @@
+---
+"@salt-ds/highcharts-theme": patch
+---
+
+Styled the reset zoom button to (imperfectly) match Salt Button solid neutral appearance, including hover & focus states.
+
+Salt defaults now set `chart.zooming.resetButton.theme` and adjust `accessibility.keyboardNavigation.order` so the zoom proxy sits before the chart SVG, enabling focus styling without event hooks.
--- a/.changeset/slimy-shrimps-wave.md
+++ b/.changeset/slimy-shrimps-wave.md
@@ -2,6 +2,6 @@
 "@salt-ds/highcharts-theme": patch
 ---
 
-Styled the reset zoom button to (imperfectly) match Salt Button solid neutral appearance, including hover & focus states.
+Styled the reset zoom button to (imperfectly) match Salt Button solid neutral appearance, including hover and focus states.
 
-Salt defaults now set `chart.zooming.resetButton.theme` and adjust `accessibility.keyboardNavigation.order` so the zoom proxy sits before the chart SVG, enabling focus styling without event hooks.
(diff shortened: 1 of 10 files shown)


joshwooding (dismissed):

packages/highcharts-theme/src/useChart.ts
```diff
@@ -82,6 +83,12 @@ export const useChart = (
     setMergedOptions(getMergedOptions(elementUsed));
   }, [chartRef, getMergedOptions, mode, targetWindow, theme]);
 
+  useEffect(() => {
+    const chart = chartRef.current?.chart as Highcharts.Chart | undefined;
```
joshwooding: Is it possible to use a typeguard instead of casting?

packages/highcharts-theme/src/accessibility-enhancements.ts
```diff
@@ -0,0 +1,83 @@
...
+const canListenForBlur = (value: unknown): value is EventTarget =>
+  typeof (value as EventTarget | undefined)?.addEventListener === "function";
+
+// ensures we only call setState on blur if the button still exists - reset button is destroyed when
+// zoom is cleared or chart redraws and blur could fire after that so this guards against a stale ref
+const resetZoomButtonState = (
+  resetButton: ResetZoomButton | undefined,
+  state: number,
+) => {
+  if (resetButton?.element) {
```
joshwooding, line 52: I don't know enough about this, but should we be checking isConnected() too?

joshwooding (dismissed):

packages/highcharts-theme/src/useChart.ts
joshwooding: Turns out the type is already correct so we don't need any of the casts in this file

.changeset/slimy-shrimps-wave.md
```diff
@@ -0,0 +1,7 @@
+---
+"@salt-ds/highcharts-theme": patch
+---
+
+Styled the reset zoom button to (imperfectly) match Salt Button solid neutral appearance, including hover and focus states.
```
wemmyo, line 5: Should the reset button adjust based on density? The chart and legend does

packages/highcharts-theme/src/accessibility-enhancements.ts
tycup: checking just the element is the pattern that Highchart recommends itself so think fine to keep just this

.changeset/slimy-shrimps-wave.md
tycup: Nope - e.g. touch density buttons are too large so we decided to stick with medium across the board

joshwooding (approved):

mark-tate (approved):

wemmyo (approved):

PR #6775, opened 2026-06-29
Fix BorderLayout not working properly if a header and footer is omitted

joshwooding:


mark-tate (approved):

packages/core/stories/border-layout/border-layout.stories.tsx
```diff
@@ -166,6 +166,36 @@ const NoHeaderTemplate: StoryFn<typeof BorderLayout> = (args) => {
...
+      <BorderItem position="west">
+        <div className="border-item border-left">
+          <p>West</p>
+        </div>
+      </BorderItem>
+      <BorderItem position="center">
+        <div
+          className="border-item border-main"
+          style={{
+            minWidth: 100,
```
tycup, line 181: Is this unit-less on purpose?

tycup (approved):

packages/core/stories/border-layout/border-layout.stories.tsx
joshwooding: Not really, but in JSX it's valid :)

PR #6790, opened 2026-07-02
Fix file drop zone drag handling

### LINKED ISSUE #5265: Add an example showing a custom text label to FileDropZoneTrigger
### Related page(s)

https://www.saltdesignsystem.com/salt/components/file-drop-zone/examples

### Kind of issue

Missing information

### Description

We should add an example to show custom text for the trigger

### Browser

- [ ] Chrome
- [ ] Safari
- [ ] Firefox
- [ ] Microsoft Edge

### Operating system

- [ ] macOS
- [ ] Windows
- [ ] Linux
- [ ] iOS
- [ ] Android

joshwooding:
Closes #5265

### DIFF
--- a/.changeset/file-drop-zone-drop-behaviour.md
+++ b/.changeset/file-drop-zone-drop-behaviour.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed `FileDropZone` drag-and-drop handling so disabled drop zones ignore dropped files while preventing browser file navigation, and non-file drags preserve native behavior. `FileDropZoneTrigger` now composes consumer `onClick` handlers and calls `onChange` before resetting the input value.
--- a/.changeset/file-drop-zone-drop-behaviour.md
+++ b/.changeset/file-drop-zone-drop-behaviour.md
@@ -2,4 +2,4 @@
 "@salt-ds/core": patch
 ---
 
-Fixed `FileDropZone` drag-and-drop handling so disabled drop zones ignore dropped files while preventing browser file navigation, and non-file drags preserve native behavior. `FileDropZoneTrigger` now composes consumer `onClick` handlers and calls `onChange` before resetting the input value.
+Fixed `FileDropZone` drag-and-drop handling so disabled drop zones ignore dropped files while preventing browser file navigation, and non-file drops are cancelled without invoking the file `onDrop` callback. `FileDropZoneTrigger` now composes consumer `onClick` handlers and calls `onChange` before resetting the input value.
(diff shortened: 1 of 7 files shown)


packages/core/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -60,17 +60,19 @@ export const FileDropZoneTrigger = forwardRef<
     buttonRef.current?.focus();
   };
 
-  const handleClick = (event: SyntheticEvent<HTMLButtonElement>) => {
+  const handleClick = (event: MouseEvent<HTMLButtonElement>) => {
+    onClick?.(event);
```
tycup, line 64: do we need to respect preventDefault though i guess on a button it doesnt do much?

tycup (dismissed):

packages/core/src/file-drop-zone/FileDropZoneTrigger.tsx
joshwooding: We don't typically, but currently onClick would have replaced the default logic

tycup (dismissed):

dplsek (dismissed):

wemmyo (dismissed):

packages/core/src/file-drop-zone/FileDropZoneTrigger.tsx
```diff
@@ -60,17 +60,19 @@ export const FileDropZoneTrigger = forwardRef<
...
     event.stopPropagation();
     fileInputRef.current?.click();
   };
 
   const handleChange = (event: ChangeEvent<HTMLInputElement>) => {
     const files = Array.from((event.target as HTMLInputElement).files ?? []);
+    onChange?.(event, files);
+
     // Allow selecting the same file multiple times - #3591
     // User would still be able to put back the value in onChange, if necessary
```
wemmyo: I guess we can remove this comment now

.changeset/file-drop-zone-drop-behaviour.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed `FileDropZone` drag-and-drop handling so disabled drop zones ignore dropped files while preventing browser file navigation, and non-file drags preserve native behavior. `FileDropZoneTrigger` now composes consumer `onClick` handlers and calls `onChange` before resetting the input value.
```
mark-tate: nit: so no longer calls 'onDrop' for non-files, that could create new behaviour for some non-file content.
ie. links dragged will now navigate to the new page, whilst they were previously ignored.

packages/core/src/file-drop-zone/FileDropZoneTrigger.tsx
joshwooding: Why? it's still relevant no? the bit helping is resetting the event.target.value

joshwooding: Ah I see, that bottom line

.changeset/file-drop-zone-drop-behaviour.md
joshwooding: I've updated the behaviour so preventDefault is always called

tycup (approved):

mark-tate (approved):

PR #6806, opened 2026-07-06
Add baseline.css to normalize browser defaults

mark-tate:


packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+ *   - Body background / color / min-height / font-smoothing.
+ *   - `list-style: none` on `<ul>` / `<ol>`.
+ *   - `prefers-reduced-motion` global overrides.
+ *   - Mobile-only tweaks (`-webkit-text-size-adjust`, tap highlight).
+ */
+
+/* Modern box-sizing — every element (and pseudo-element) uses
+ * border-box so widths / heights include padding and border. */
+*,
+*::before,
```
joshwooding: All Salt component will still have to handle this

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+ul,
+ol,
+fieldset,
+hr {
+  margin: 0;
+  margin-block: 0;
+  margin-inline: 0;
+  padding: 0;
+  padding-block: 0;
+  padding-inline: 0;
```
joshwooding: The shorthand CSS here isn't supported by our minimum browsers

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+/* Prevent the browser's oversized default `<h1>` etc. from breaking
+ * layouts. Apps that want an editorial hierarchy should use Salt's
+ * typography components (e.g. `<H1>`, `<Text>`). */
+h1,
+h2,
+h3,
+h4,
+h5,
+h6 {
+  font-size: inherit;
```
joshwooding: Not sure about this?

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+video,
+canvas,
+svg {
+  display: block;
+  max-width: 100%;
+}
+
+/* Form controls — inherit typography so bare inputs / buttons don't
+ * fall back to the browser's system font. */
+input,
```
joshwooding: Or this

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+button,
+textarea,
+select {
+  font: inherit;
+  color: inherit;
+}
+
+/* Bare `<button>` — strip UA chrome. Salt's `<Button>` component
```
joshwooding: I don't think we should care about this? It's more likely to break things than fix

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+  padding: 0;
+  cursor: pointer;
+}
+
+button:disabled {
+  cursor: not-allowed;
+}
+
+/* Fieldsets ship with a browser-default border that is almost
+ * never wanted. Padding is already zeroed above. */
```
joshwooding: Feels like a level of opinionation we don't need to have?

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+fieldset {
+  border: 0;
+}
+
+/* Collapse table borders by default so bare tables don't render
+ * with the browser's "double line" cell borders. */
+table {
```
joshwooding: We already have a table component

packages/theme/css/baseline.css
```diff
@@ -0,0 +1,146 @@
...
+  border-collapse: collapse;
+  border-spacing: 0;
+}
+
+/* Anchor color inherits from the surrounding text so unthemed
+ * `<a>` elements do not render as the browser's default bright
+ * blue. Salt's `<Link>` component still applies its own treatment. */
+a {
```
joshwooding: I also don't think we should do this either

packages/theme/src/css/baseline.css
kriskoder, line 1: how about squashing selectors to something like that:
`/*
 * Salt CSS baseline
 *
 * An optional, deliberately narrow reset for applications using Salt.
 * Import it before the Salt theme and component styles. Salt components remain
 * self-contained and do not depend on this file.
 */

body,
h1,
h2,
h3,
h4,
h5,
h6,
p {
  margin: 0;
}`

site/docs/getting-started/developing.mdx
```diff
@@ -218,6 +218,31 @@ Refer to the [SaltProvider](../components/salt-provider) docs for the full API.
...
+Salt's text styles control typography, but browsers also add margins to semantic elements such as headings and paragraphs. Those margins remain when Markdown content is rendered with `Text` using the semantic `as` prop; changing its visual style with `styleAs` does not change the element's spacing. This mixes browser-defined spacing with Salt's spacing system.
+
+To remove that browser spacing, `@salt-ds/theme` ships an opt-in baseline reset. Import it once at the root of your app, before the theme CSS:
+
+```tsx
+import "@salt-ds/theme/css/baseline.css";
+import "@salt-ds/theme/css/global.css";
+import "@salt-ds/theme/css/theme-next.css";
+```
+
```
kriskoder, line 232:
```suggestion
On the Legacy theme, import it before @salt-ds/theme/index.css instead.

```

site/docs/getting-started/developing.mdx
```diff
@@ -218,6 +218,31 @@ Refer to the [SaltProvider](../components/salt-provider) docs for the full API.
 
 Salt injects component styles at runtime by default, so you don’t need to import component CSS for the setup above. If you need static component CSS or your application uses a Content Security Policy that requires a nonce for injected styles, see [Style injection](./style-injection).
 
+#### (Optional) Apply the CSS baseline reset
+
+Salt's text styles control typography, but browsers also add margins to semantic elements such as headings and paragraphs. Those margins remain when Markdown content is rendered with `Text` using the semantic `as` prop; changing its visual style with `styleAs` does not change the element's spacing. This mixes browser-defined spacing with Salt's spacing system.
```
kriskoder: I would suggest dropping the Markdown promise, it's not fixing it.

Markdown headings and paragraphs already have their margins set explicitly — our own MDX layer does that with tokens. They win over p { margin: 0 } on specificity, so the reset is a no-op for them. What Markdown does leak is ul, ol and blockquote.

```suggestion
alt's text styles control typography, but browsers also add margins to semantic elements such as headings and paragraphs. Those margins remain when Text is rendered with the semantic as prop; changing its visual style with styleAs does not change the element's spacing. This mixes browser-defined spacing with Salt's spacing system.
```

kriskoder (changes requested):

site/docs/getting-started/developing.mdx
tycup: shortened the entire section :)

kriskoder (dismissed):

joshwooding (dismissed):

joshwooding:
We should add it to our site/ storybook

kriskoder (approved):

joshwooding (approved):

PR #6829, opened 2026-07-13
Fix empty read-only logic and align Dropdown validation precedence

joshwooding:
- Input: treat numeric 0 as a real value when read-only instead of showing the empty read-only marker
- PillInput: same read-only empty-value handling for consistency (internal)
- Dropdown: FormField validation status now takes precedence over the local validationStatus prop, matching the other form controls

### DIFF
--- a/.changeset/dropdown-validation-precedence.md
+++ b/.changeset/dropdown-validation-precedence.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed `Dropdown` validation status precedence to align with the other form controls, so the `FormField` validation status now takes precedence over the `Dropdown`'s own `validationStatus` prop.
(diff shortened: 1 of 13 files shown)


tycup (dismissed):

wemmyo (dismissed):

.changeset/dropdown-validation-precedence.md
wemmyo, line 1: Considering this could cause a visual change from form field now changing the validation status, should it be a minor change instead of patch?

joshwooding: I would still consider it a bug fix, which can also cause visual changes. All the other form controls have this setup only dropdown was the inconsistency

kriskoder (approved):

wemmyo (approved):

PR #6832, opened 2026-07-14
Optimize large list controls

joshwooding:


packages/core/src/list-control/ListControlOptionStore.ts
```diff
@@ -0,0 +1,158 @@
...
+      const previousSnapshot = this.getSnapshot(option.id);
+      this.#options.delete(option.id);
+      this.#deleteValueId(option.value, option.id);
+      if (previousSnapshot !== 0) this.#notify(option.id);
+    };
+  }
+
+  getSnapshot(
+    id: string,
+    fallbackOption?: OptionValue<Item>,
```
tycup, line 91: not sure about this being called fallbackOption since it wins over registered data

packages/core/src/list-control/ListControlOptionContext.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <ListControlOptionContext.Provider
+      value={value as unknown as ListControlOptionContextValue<unknown>}
+    >
+      {children}
+    </ListControlOptionContext.Provider>
+  );
+}
+
+export function useListControlOptionContext<Item>() {
+  const context = useContext(ListControlOptionContext);
```
tycup, line 58: Could this use the repo createContext

packages/core/src/list-box/ListBox.tsx
tycup, line 116: Since we use `(startIndex + offset) % options.length` in `findOptionFromSearch` i dont think the second lookup here (and in Dropdown) will find anything the first didnt

tycup (dismissed):

packages/core/src/list-control/ListControlState.ts
```diff
@@ -71,20 +142,31 @@ export function useListControl<Item>(props: ListControlProps<Item>) {
...
   const listRef = useRef<HTMLDivElement>(null);
   const [listElement, setListElement] = useState<HTMLDivElement | null>(null);
   const setListRef = useForkRef<HTMLDivElement>(listRef, setListElement);
 
   const [focusedState, setFocusedState] = useState(false);
-  const [focusVisibleState, setFocusVisibleState] = useState(false);
+  const [focusVisibleState, setFocusVisibleStateInternal] = useState(false);
+  const setFocusVisibleState = useCallback((newValue: boolean) => {
+    setFocusVisibleStateInternal(newValue);
+  }, []);
```
kriskoder: I would suggest accessing that setter directly, no need to wrap in callback, unless there's something I missed here.

```suggestion
  const [focusVisibleState, setFocusVisibleState] = useState(false);
```

packages/core/src/list-control/ListControlState.ts
```diff
@@ -71,20 +142,31 @@ export function useListControl<Item>(props: ListControlProps<Item>) {
...
 
   useEffect(() => {
     // remove focus when controlling disabled
     if (disabled && focusedState) {
       setFocusedState(false);
       setFocusVisibleState(false);
     }
-  }, [disabled, focusedState]);
+  }, [disabled, focusedState, setFocusVisibleState]);
```
kriskoder: related to comment above: 
```suggestion
}, [disabled, focusedState]);
```

packages/core/src/list-control/ListControlRegistry.ts
```diff
@@ -0,0 +1,174 @@
...
+}
+
+interface RegistryEntry<Item> extends RegisteredOption<Item> {
+  generation: number;
+}
+
+export function createCoalescedRebuild(
+  rebuild: () => void,
+  enqueue: (callback: () => void) => void = (callback) => {
+    // Promise jobs are supported by Chrome 86.
```
kriskoder, line 15:
```suggestion
```

.changeset/fast-option-lists.md
kriskoder, line 1: NIT:
I would suggest putting more emphasis on what consumers need to know and less on technical stuff.
Something like:
Improved Dropdown, ComboBox, and ListBox performance for large, non-virtualized option lists. Fixed active/highlighted styling to track the active Option by id instead of value, so Options that share the same value no longer all appear active together; selection is unaffected and remains value-based.

kriskoder (commented):
nice time complexity optimisation, only a few small things

kriskoder (approved):

tycup (approved):

PR #6836, opened 2026-07-15
Route Day.js operations through injected factory

joshwooding:


packages/date-adapters/src/dayjs-adapter/index.ts
```diff
@@ -16,8 +16,8 @@ import {
 
 type Constructor = {
   (...args: Parameters<typeof defaultDayjs>): Dayjs;
-  tz?: (value: Parameters<typeof defaultDayjs>[0], timezone: string) => Dayjs;
-  utc?: (value?: Parameters<typeof defaultDayjs>[0]) => Dayjs;
+  tz?: typeof defaultDayjs.tz;
```
tycup: if i have skipLibCheck: false which is a default, id get a build fail just by importing AdapterDayjs (uness i already import both plugin declarations), since this relies on module augments from dayjs/plugin/timezone and ../utc and declaration gen removes those imports - bit of an edge case but yeah

tycup: also wondering if this getting narrower could break something for consumers that match the previous signature

tycup (approved):

packages/date-adapters/src/dayjs-adapter/index.ts
```diff
@@ -16,7 +16,11 @@ import {
 
 type Constructor = {
   (...args: Parameters<typeof defaultDayjs>): Dayjs;
-  tz?: (value: Parameters<typeof defaultDayjs>[0], timezone: string) => Dayjs;
+  tz?: {
+    (value: Parameters<typeof defaultDayjs>[0], timezone: string): Dayjs;
+    guess?: () => string;
+    setDefault?: (timezone?: string) => void;
```
wemmyo, line 22: I don't see `setDefault` being used anywhere, is this intentional?

packages/date-adapters/src/dayjs-adapter/index.ts
```diff
@@ -88,7 +92,7 @@ export class AdapterDayjs implements SaltDateAdapter<Dayjs, string> {
       return undefined;
     }
     if (timezone === "system") {
-      return defaultDayjs.tz.guess();
+      return this.dayjs.tz?.guess?.() ?? defaultDayjs.tz.guess();
```
wemmyo, line 95: Is it worth having a separate function to guess system timezone?

packages/date-adapters/src/dayjs-adapter/index.ts
joshwooding: It's just the typing on what is in tz. it's for correctness

packages/date-adapters/src/dayjs-adapter/index.ts
joshwooding: I don't think so? It's only used a few times

.changeset/dayjs-injected-factory.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/date-adapters": patch
+---
+
+Fixed the Day.js adapter to consistently use an injected Day.js factory for timezone resolution, UTC/timezone date creation, `today`, and `now`.
```
mark-tate, line 5: does this adequatelyexplain  the additional props and any docs need updating ?

kriskoder (approved):

PR #6839, opened 2026-07-15
Improve component CSS injection lifecycle

joshwooding:


packages/styles/src/__tests__/__e2e__/useStyleInject.cy.tsx
```diff
@@ -39,6 +42,79 @@ const RemovableTest = () => {
...
+      : '[data-marker="dynamic-example-one"]',
+  );
+
+  return (
+    <div>
+      <Button onClick={() => setUseSecondMarker(true)}>
+        Move insertion point
+      </Button>
+      <InsertionPointProvider insertionPoint={insertionPoint}>
+        <TestComponent
```
tycup, line 108: nit: what if i have two components (maybe if i was doing ssr safe inits?) that share the stylesheet, count 2, insterted before  - would count ever get to 0? would be rare since most apps pick the css insert location once and not change it

tycup: (i just asked the robot and it said its unlikely that anyone would do this)

tycup (dismissed):

tycup (approved):

kriskoder (approved):

PR #6845, opened 2026-07-20
Update AG Grid pagination examples

tycup:
- Use Salt components in the ag-grid Pagination examples
- Design chosen to be similar to the default (i would prefer to move the page size Dropdown to the left side to slightly prevent layout jumps as the label changes)

### DIFF
--- a/packages/ag-grid-theme/src/examples/Pagination.tsx
+++ b/packages/ag-grid-theme/src/examples/Pagination.tsx
@@ -1,7 +1,29 @@
+import {
+  Button,
+  CompactInput,
+  CompactPaginator,
+  Dropdown,
+  FlexLayout,
+  Option,
+  Pagination,
+  StackLayout,
+  Text,
+} from "@salt-ds/core";
+import { FirstIcon, LastIcon } from "@salt-ds/icons";
+import type { GridApi, PaginationChangedEvent } from "ag-grid-community";
 import { AgGridReact, type AgGridReactProps } from "ag-grid-react";
+import {
+  type SyntheticEvent,
+  useCallback,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 import dataGridExampleColumns from "../dependencies/dataGridExampleColumns";
 import dataGridExampleData from "../dependencies/dataGridExampleData";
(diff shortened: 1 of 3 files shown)


joshwooding (changes requested):
Looks good but I think we should add a note to the docs saying we're using a custom pagination in the docs just for clarity

packages/ag-grid-theme/src/examples/Pagination.tsx
```diff
@@ -16,21 +38,140 @@ const generateData = (states: typeof dataGridExampleData) =>
...
+        justify="end"
+        gap={3}
+        style={{
+          borderTop:
+            "var(--salt-size-fixed-100) var(--salt-borderStyle-solid) var(--salt-separable-secondary-borderColor)",
+          paddingBlock: "var(--salt-spacing-100)",
+          paddingInline: "var(--salt-spacing-100)",
+        }}
+      >
+        <FlexLayout align="center" gap={1}>
```
joshwooding: Is there a reason we're not using a form field here

packages/ag-grid-theme/src/examples/Pagination.tsx
```diff
@@ -16,21 +38,140 @@ const generateData = (states: typeof dataGridExampleData) =>
...
+          <Text styleAs="label">Page Size</Text>
+          <Dropdown
+            aria-label="Page Size"
+            selected={[pageSize]}
+            onSelectionChange={handlePageSizeChange}
+            bordered
+            style={{ width: "calc(var(--salt-size-base) * 3)" }}
```
joshwooding: Can you try a different width? e.g. ch based

site/docs/components/ag-grid-theme/examples.mdx
```diff
@@ -148,7 +148,7 @@ This overlay tells the user that there is no information to display.
 
 ## Pagination
 
-Users can move between pages using the controls in the lower right.
+Users can move between pages using the controls in the lower right. The example below suppresses AG Grid's built-in pagination panel and composes Salt components to handle pagination instead.
```
joshwooding: We've discussed moving the addition to a new line

joshwooding (approved):

kriskoder (approved):

PR #6860, opened 2026-07-22
Added custom carousel slide example

ThusharaJ07:
JIRA: CIBDESIGNSYSTEMJIRA-468


site/docs/components/carousel/examples.mdx
```diff
@@ -82,3 +82,9 @@ The Embla API enables programmatic control, such as the visible slide.
 </Callout>
 
 <LivePreview componentName="carousel" exampleName="ControlledCarousel" />
+
+## Custom slide
+
+You can render fully custom slide content, such as `Avatar` components, by composing your own slide elements inside `CarouselSlides`.
```
mark-tate: Fully custom slides ?  any concerns about convergence @ivan-calderon ?

I'm not sure why we would have a carousel of Avatars, feels like an edge case rather than something that adds value to the doc. Typically examples are real-world use-cases that you copy and paste into code, anything less than that, should stay in Storybook as a test rig or edge-case example.

I think maybe we go with just an example in Storybook and prove it's possible without UX issues (see other comment) then update our internal skill to handle migration of avatar edge cases, based on the proven storybook code.

If this stays...
At least we would say, what the use-case is - is it for bios or something ?

mark-tate:
This code has a UX issue, click on any avatar and it scroll to that avatar before it allows you to drag it.

site/docs/components/carousel/examples.mdx
ThusharaJ07: I can move it storybook to test it out.

wemmyo: @ivan-calderon Also this looks like some variant of Avatar group which we decided not to support unless the avatars overlap with each other within group. So might be worth aligning on the design side

ThusharaJ07:
Fixed another story based issue:
Card-based Carousel stories were hiding Link actions because the CSS rule relied on a class (.carouselSlideIsSnapped) that was never added to the DOM. The [embla-carousel-class-names] plugin — which produces that class — wasn't registered on any story, even though it is used correctly in the site docs example.

packages/embla-carousel/stories/carousel.stories.tsx
```diff
@@ -37,6 +40,7 @@ const CarouselCardExample: StoryFn<CarouselProps & { ariaVariant: string }> = ({
     <Carousel
       aria-labelledby={`${carouselId}-title`}
       className="carousel"
+      emblaPlugins={[ClassNames({ snapped: "carouselSlideIsSnapped" })]}
```
joshwooding, line 43: Is this actually used?

joshwooding (dismissed):

joshwooding:
Waiting for @jake-costa's review

joshwooding (dismissed):

joshwooding (approved):

wemmyo (approved):

PR #6883, opened 2026-07-30
Update app header pattern images

honey-chang:
Updated app header pattern images


site/public/img/patterns/app-header/app-header-vertical-navigation-dark.png
joshwooding, line 1: NIT: The lo-fi vertical navigation isn't consistent with other vertical navigation appearances. But from what I can see we don't have an updated vertical navigation

joshwooding (approved):

amritadesmet (approved):

PR #6884, opened 2026-07-30
[STORY] collapsible vertical navigation

kriskoder:
add example with collapsible vertical navigation,
nested links are outside the scope of this example 
has to be merged before https://github.com/jpmorganchase/salt-ds/pull/6885


packages/core/stories/vertical-navigation/vertical-navigation.stories.tsx
```diff
@@ -508,6 +510,125 @@ export const WithIcon: StoryFn<typeof VerticalNavigation> = (args) => {
...
+  },
+  {
+    title: "About Us",
+    href: "/about",
+    icon: <UserGroupSolidIcon aria-hidden />,
+  },
+  {
+    title: "Support",
+    href: "/support",
+    icon: <HelpSolidIcon aria-hidden />,
```
tycup: This icon is deprecated - would `HelpCircleSolidIcon` work here?

packages/core/stories/vertical-navigation/vertical-navigation.stories.tsx
```diff
@@ -508,6 +510,125 @@ export const WithIcon: StoryFn<typeof VerticalNavigation> = (args) => {
...
+    >
+      <div className="collapsibleNavigationExample-sidebar">
+        <Tooltip
+          content={collapsed ? "Expand navigation" : "Collapse navigation"}
+          placement="right"
+        >
+          <Button
+            appearance="transparent"
+            aria-expanded={!collapsed}
+            aria-label={collapsed ? "Expand navigation" : "Collapse navigation"}
```
tycup: since aria-expanded already announces the state, could the aria label be something like "Toggle Navigation" ?

joshwooding: It needs to be changed

packages/core/stories/vertical-navigation/vertical-navigation.stories.tsx
```diff
@@ -508,6 +510,125 @@ export const WithIcon: StoryFn<typeof VerticalNavigation> = (args) => {
...
+                  content={item.title}
+                  disabled={!collapsed}
+                  placement="right"
+                >
+                  <MockedTrigger to={item.href}>
+                    {item.icon}
+                    {/* Hidden once collapsed, but kept in the DOM for the
+                        item's accessible name. */}
+                    <VerticalNavigationItemLabel
+                      className={showLabels ? undefined : "salt-visuallyHidden"}
```
joshwooding: Can we add this styling to the example please

kriskoder: @joshwooding I updated css file, let me know if that solves the issue.

packages/core/stories/vertical-navigation/vertical-navigation.stories.tsx
```diff
@@ -508,6 +510,125 @@ export const WithIcon: StoryFn<typeof VerticalNavigation> = (args) => {
...
+  return (
+    <div
+      className={clsx("collapsibleNavigationExample", {
+        "collapsibleNavigationExample-collapsed": collapsed,
+        "collapsibleNavigationExample-animating": animating,
+      })}
+    >
+      <div className="collapsibleNavigationExample-sidebar">
+        <Tooltip
+          content={collapsed ? "Expand navigation" : "Collapse navigation"}
```
kriskoder: @tycup should this be changed to "Toggle Navigation" too, same as in line 574?

tycup: Run it through a screenreader :) Currently getting "Toggle Navigation Collapse navigation, expanded, button" which is a bit convoluted

packages/core/stories/vertical-navigation/vertical-navigation.stories.css
```diff
@@ -1,3 +1,93 @@
 .saltVerticalNavigation {
   min-width: 30ch;
 }
+
+/* CollapsibleNavigation example */
+
+.collapsibleNavigationExample {
+  --collapsibleNavigationExample-width: clamp(18ch, 50vw, 30ch);
```
kriskoder, line 8: @joshwooding do we want the side panel width to adjust to smaller viewport or prefer to keep width fixed?

joshwooding: I don't mind. On smaller viewports things can scroll in one direction so we might not need to make the width smaller. I would just want to example to still be representative at those smaller screen sizes

kriskoder:
@joshwooding @jake-costa
VerticalNavigation will need some updates due to it's limitations in properly handling that scenario from an accessibility standpoint.
Accessibility improved after Jake's reviews, good to go for now.

jake-costa (approved):

joshwooding (approved):

PR #6900, opened 2026-08-06
Fix Overlay gap and props type

joshwooding:


packages/core/src/overlay/Overlay.tsx
```diff
@@ -72,35 +84,55 @@ export const Overlay = ({
     useDismiss(context),
   ]);
 
+  const floatingElement = elements.floating;
+  const floatingHeight = floatingElement?.offsetHeight;
+  const floatingWidth = floatingElement?.offsetWidth;
```
kriskoder: These reads can be moved into UseMemo and read from `context`, that will keep the values up to date and prevent reflow on every render.

packages/core/src/overlay/Overlay.tsx
```diff
@@ -72,35 +84,55 @@ export const Overlay = ({
...
+
   const floatingStyles = useMemo(() => {
     return {
```
kriskoder, line 87: related to: https://github.com/jpmorganchase/salt-ds/pull/6900/changes#r3728438730
```suggestion
    return {
    const floatingElement = context.elements.floating;
```

packages/core/src/overlay/Overlay.tsx
```diff
@@ -72,35 +84,55 @@ export const Overlay = ({
...
       top: y ?? 0,
       left: x ?? 0,
       position: strategy,
-      width: elements.floating?.offsetWidth,
-      height: elements.floating?.offsetHeight,
+      width: floatingWidth,
+      height: floatingHeight,
```
kriskoder: related to: https://github.com/jpmorganchase/salt-ds/pull/6900/changes#r3728438730
```suggestion
      width: floatingElement?.offsetWidth,
      height: floatingElement?.offsetHeight,
```

packages/core/src/overlay/Overlay.tsx
```diff
@@ -72,35 +84,55 @@ export const Overlay = ({
...
     };
-  }, [elements.floating, strategy, x, y]);
+  }, [floatingHeight, floatingWidth, strategy, x, y]);
```
kriskoder: related to: https://github.com/jpmorganchase/salt-ds/pull/6900/changes#r3728438730
```suggestion
  }, [context, strategy, x, y]);
```

packages/core/src/overlay/Overlay.tsx
```diff
@@ -48,22 +52,30 @@ export const Overlay = ({
...
+    () => [
+      offset(8),
+      flip(),
+      shift({ limiter: limitShift() }),
+      arrow({ element: arrowRef }),
+    ],
+    [],
+  );
 
   const { x, y, strategy, context, elements, floating, reference } =
```
kriskoder, line 70: related to https://github.com/jpmorganchase/salt-ds/pull/6900/changes#r3728438730
```suggestion
const { x, y, strategy, context, floating, reference } =
```

packages/core/src/overlay/Overlay.tsx
```diff
@@ -48,22 +52,30 @@ export const Overlay = ({
     state: "open",
   });
 
-  const handleOpenChange = (newOpen: boolean) => {
-    setOpenState(newOpen);
-    onOpenChange?.(newOpen);
-  };
+  const handleOpenChange = useCallback(
```
kriskoder: Nice idea with `useCallback, but it's identity only goes into `useFloatingUI` that recreates its own wrapper on every render, stability is lost there anyway. It can be reverted to plain function and have the same effect.

kriskoder (changes requested):

packages/core/src/overlay/Overlay.tsx
joshwooding: The issue is that useMemo isn't guaranteed to run on render, but in this scenario it probably will when we need it to

joshwooding: But I think we can still make the change

kriskoder (approved):

dplsek (approved):

PR #6908, opened 2026-08-10
fix useClassNameInjection dropping props

mark-tate:


joshwooding (approved):

packages/styles/src/useClassNameInjection.tsx
kriskoder, line 140:
```suggestion
 * - Returning undefined leaves keys untouched; returning a string (even "") removes them.
 */
```

kriskoder, line 22:
```suggestion
/**
 * Inspects a component's declared props and optionally returns a class to inject.
 * Return `undefined` to opt out and leave the props untouched; return a string
 * (including `""`) to opt in, merging the class and stripping the declared keys.
 */
export type ClassNameInjector<Props, Keys extends keyof Props> = (
```

kriskoder (commented):
all good, just 2 JSDoc suggestions

joshwooding (approved):

kriskoder (approved):

PR #6914, opened 2026-08-11
Add Link button

joshwooding:


site/docs/components/link-button/examples.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+Use the `underline` prop to compare the default underline with `underline="never"`.
+
+<LivePreview componentName="link-button" exampleName="Underline" />
+
+## With icon
+
+Add a leading or trailing icon when it helps reinforce the destination or direction of travel.
+
+<LivePreview componentName="link-button" exampleName="WithIcon" />
```
amritadesmet, line 41: Can we add a second example of Link button with a leading icon since we mention both options are available?

amritadesmet: If the preference is to have trailing icon example only and leading icons are reserved for added affordance, that could be mentioned

site/docs/components/link-button/usage.mdx
```diff
@@ -0,0 +1,54 @@
...
+### When to use
+
+- When the primary action is navigation, such as opening a new page or external site.
+- When a button-like appearance is needed for visibility and interaction.
+- When the link is essential to the user flow but should stand out from inline text links.
+
+### When not to use
+
+- Do not use a link button for on-page actions like opening a modal, submitting a form, or toggling UI elements.
+- If the action does not navigate the user elsewhere, use [`Button`](../button) instead.
```
amritadesmet: Consider adding a bullet for when consumers are recommended to use Link instead of Link button as you have with Button

dplsek (approved):

pseys (approved):

PR #6916, opened 2026-08-12
Avatar | Add --saltAvatar-size CSS API

tycup:
Proposed approach to sizing and scaling Avatar so it works with typography alignment 
Hardcoding the ratios is far from ideal - very open to suggestion - the reason why the token based division wasn't used was because firefox does not support division with units

### DIFF
--- a/packages/core/src/avatar/Avatar.css
+++ b/packages/core/src/avatar/Avatar.css
@@ -2,8 +2,10 @@
 .saltAvatar {
   --avatar-size-multiplier: var(--saltAvatar-size-multiplier, 1);
   --avatar-base-size: var(--salt-size-base);
-  --avatar-base-fontSize: var(--saltAvatar-fontSize, var(--salt-text-notation-fontSize));
-  --avatar-container-size: calc(var(--avatar-base-size) * var(--avatar-size-multiplier));
+  /* 10 / 28 (--salt-text-notation-fontSize / --salt-size-base at medium density)  preserves the initials-to-container ratio */
+  --avatar-direct-fontSize: calc(var(--saltAvatar-size) * (10 / 28) / var(--avatar-size-multiplier));
+  --avatar-base-fontSize: var(--saltAvatar-fontSize, var(--avatar-direct-fontSize, var(--salt-text-notation-fontSize)));
+  --avatar-container-size: var(--saltAvatar-size, calc(var(--avatar-base-size) * var(--avatar-size-multiplier)));
   --avatar-fontSize: calc(var(--avatar-base-fontSize) * var(--avatar-size-multiplier));
   /* Icon styling */
   --saltIcon-size: calc(var(--avatar-container-size) / 2);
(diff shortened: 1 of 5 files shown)


joshwooding (dismissed):

packages/core/src/avatar/Avatar.css
```diff
@@ -1,9 +1,31 @@
+/* TODO: Remove hardcoded values - unitless notation-fontSize / size-base because Firefox can't do unit division  */
```
kriskoder, line 1: I would suggest addressing that 😄

packages/core/src/avatar/Avatar.css
```diff
@@ -1,9 +1,31 @@
...
+}
+
 /* Default variables applied to the root element */
 .saltAvatar {
   --avatar-size-multiplier: var(--saltAvatar-size-multiplier, 1);
   --avatar-base-size: var(--salt-size-base);
-  --avatar-base-fontSize: var(--saltAvatar-fontSize, var(--salt-text-notation-fontSize));
-  --avatar-container-size: calc(var(--avatar-base-size) * var(--avatar-size-multiplier));
+  --avatar-direct-fontSize: calc(var(--saltAvatar-size) * var(--avatar-fontSize-ratio) / var(--avatar-size-multiplier));
+  --avatar-base-fontSize: var(--saltAvatar-fontSize, var(--avatar-direct-fontSize, var(--salt-text-notation-fontSize)));
```
kriskoder, line 27: if `--saltAvatar-fontSize` is defined, it gets multiplied inside `--avatar-fontSize`, line 29, is that intended?
just as king because once `--saltAvatar-size` is defined, it doesn't get multiplied

so we might end up:

`
  <Avatar style={{ "--saltAvatar-size": "18px", "--saltAvatar-fontSize": "7px" }} />
`
-> 18px box, 14px text (7 × the default size of 2)

packages/core/src/avatar/Avatar.css
kriskoder, line 1: how about adding something like that to cypress tests?
`import type { CSSProperties } from "react";

const directSizeStyle = {
  "--saltAvatar-size": "18px",
} as CSSProperties;

it("should cancel the size multiplier when --saltAvatar-size is set", () => {
  cy.mount(
    <>
      <Default name="Juanito Jones" style={directSizeStyle} />
      <Default name="Juanito Jones" size={4} style={directSizeStyle} />
    </>,
  );

  cy.get(".saltAvatar").then(($avatars) => {
    const [baseline, withSizeProp] = [$avatars.eq(0), $avatars.eq(1)];

    expect(withSizeProp.css("width")).to.equal("18px");
    expect(withSizeProp.css("height")).to.equal("18px");
    expect(withSizeProp.css("font-size")).to.equal(baseline.css("font-size"));
  });
});`

kriskoder (changes requested):

packages/core/src/avatar/Avatar.css
tycup: this is valid but a niche combination rather than a bug - wouldn't have it as blocking for now

kriskoder (approved):

joshwooding (approved):

PR #6923, opened 2026-08-14
Avatar | Fix style override of size multiplier

tycup:


.changeset/tasty-hornets-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": minor
```
joshwooding:
```suggestion
"@salt-ds/core": patch
```

joshwooding (approved):

kriskoder (changes requested):
all good, but I would add small regerssion test in case someone decides to flip the order in the future

tycup:
> all good, but I would add small regerssion test in case someone decides to flip the order in the future

I wouldn't out of granularity :) I don't think we test for this anywhere else, and if someone flipped it again, it would cause visual regression to fail anyway

kriskoder (approved):

PR #6944, opened 2026-08-28
add header, content and footer components to drawer

kriskoder:
add Header block and content to Drawer. Using Header and Content blocks inside Drawer becomes our recommended approach. In scope: components, stories and tests.

### DIFF
--- a/.changeset/olive-moons-dismiss.md
+++ b/.changeset/olive-moons-dismiss.md
@@ -1,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed `disableDismiss` on `Drawer` also preventing <kbd>Escape</kbd> from closing the drawer. It now only prevents dismissal on click away, matching its documentation and `Dialog`.
--- a/.changeset/thick-eagles-tickle.md
+++ b/.changeset/thick-eagles-tickle.md
@@ -2,7 +2,7 @@
 "@salt-ds/core": minor
 ---
 
-Added `DrawerHeader` and `DrawerContent`, implementing header block in core `Drawer`. `DrawerHeader` takes optional `header`, `preheader`, `description` and `actions`, and displays an accent bar unless `disableAccent` is set. Heading names the drawer and `description` describes it, so `aria-labelledby` and `aria-describedby` no longer need to be set manually. `DrawerCloseButton` should now be passed to `actions`. `DrawerHeader` stays pinned to the top, `DrawerContent` is scrollable. Drawers without `DrawerHeader` and `DrawerContent` are deprecated.
+Added `DrawerHeader` and `DrawerContent`, implementing header block in core `Drawer`. `DrawerHeader` takes optional `header`, `preheader`, `description` and `actions`, and displays an accent bar unless `disableAccent` is set. `header` prop from `DrawerHeader` names `Drawer` and `description` prop from `DrawerHeader` describes `Drawer`, so `aria-labelledby` and `aria-describedby` no longer need to be set manually. `DrawerCloseButton` should be passed to `actions`. `DrawerHeader` stays pinned to the top, `DrawerContent` is scrollable. Drawers without `DrawerHeader` and `DrawerContent` are deprecated.
 
 ```tsx
 <Drawer open={open} onOpenChange={setOpen}>
--- a/.changeset/thick-eagles-tickle.md
+++ b/.changeset/thick-eagles-tickle.md
@@ -1,0 +1,12 @@
+---
+"@salt-ds/core": minor
+---
+
+Added `DrawerHeader` and `DrawerContent`, implementing header block in core `Drawer`. `DrawerHeader` takes a required `header`, with optional `preheader`, `description` and `actions`, and displays an accent bar unless `disableAccent` is set. Heading names the drawer, so `aria-labelledby` no longer needs to be set manually. `DrawerHeader` stays pinned to the top, `DrawerContent` is scrollable. Drawers without a `DrawerHeader` are unchanged.
+
+```tsx
+<Drawer open={open} onOpenChange={setOpen}>
+  <DrawerHeader header="Check deposit #1278" actions={closeButton} />
+  <DrawerContent>{content}</DrawerContent>
+</Drawer>
+```
--- a/.changeset/thick-eagles-tickle.md
+++ b/.changeset/thick-eagles-tickle.md
@@ -2,7 +2,7 @@
 "@salt-ds/core": minor
 ---
 
-Added `DrawerHeader` and `DrawerContent`, implementing header block in core `Drawer`. `DrawerHeader` takes a required `header`, with optional `preheader`, `description` and `actions`, and displays an accent bar unless `disableAccent` is set. Heading names the drawer, so `aria-labelledby` no longer needs to be set manually. `DrawerHeader` stays pinned to the top, `DrawerContent` is scrollable. Drawers without a `DrawerHeader` are unchanged.
+Added `DrawerHeader` and `DrawerContent`, implementing header block in core `Drawer`. `DrawerHeader` takes optional `header`, `preheader`, `description` and `actions`, and displays an accent bar unless `disableAccent` is set. Heading names the drawer, so `aria-labelledby` no longer needs to be set manually. `DrawerHeader` stays pinned to the top, `DrawerContent` is scrollable. Drawers without a `DrawerHeader` are unchanged.
 
 ```tsx
 <Drawer open={open} onOpenChange={setOpen}>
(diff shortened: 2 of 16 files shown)


packages/core/src/drawer/index.ts
```diff
@@ -1,2 +1,5 @@
 export * from "./Drawer";
 export * from "./DrawerCloseButton";
+export * from "./DrawerContent";
+export * from "./DrawerContext";
```
joshwooding:
```suggestion
```

packages/core/src/__tests__/__e2e__/drawer/drawer.browser.test.tsx
```diff
@@ -138,3 +152,171 @@ describe("GIVEN a Drawer", () => {
...
+    await renderWithSalt(<Header />);
+    await page.getByRole("button", { name: "Open Drawer" }).click();
+
+    const closeButton = page.getByRole("button", { name: "Close Drawer" });
+    await expect.element(closeButton).toHaveFocus();
+    await closeButton.click();
+    await expect.element(page.getByRole("dialog")).not.toBeInTheDocument();
+  });
+
+  it("exposes overflowing content as a region reachable by keyboard", async () => {
```
joshwooding: NIT: This isn't really related to header

packages/core/src/__tests__/__e2e__/drawer/drawer.browser.test.tsx
```diff
@@ -138,3 +152,171 @@ describe("GIVEN a Drawer", () => {
...
+        </DrawerContent>
+      </Drawer>,
+    );
+
+    const region = page.getByRole("region", { name: "Notifications" });
+    await expect.element(region).toBeVisible();
+    await expect.element(region).toHaveAttribute("tabindex", "0");
+  });
+
+  it("hands its padding to DrawerContent when there is no DrawerHeader", async () => {
```
joshwooding: NIT: This should be a visual regression test

packages/core/src/__tests__/__e2e__/drawer/drawer.browser.test.tsx
```diff
@@ -138,3 +152,171 @@ describe("GIVEN a Drawer", () => {
...
+    await expect
+      .element(page.getByRole("button", { name: "Close Drawer" }))
+      .toBeVisible();
+  });
+});
+
+describe("GIVEN a Drawer with no heading in its header", () => {
+  const longText = "Pending transaction review. ".repeat(200);
+
+  it("falls back to the drawer's aria-label to name the content region", async () => {
```
joshwooding: This is really related to testing the content rather than the header

packages/core/src/drawer/DrawerCloseButton.css
```diff
@@ -1,11 +1,4 @@
 .saltDrawerCloseButton-container {
-  position: relative;
```
joshwooding, line 2: Why was this removed? This will break existing users.

packages/core/src/drawer/DrawerCloseButton.css
```diff
@@ -3,9 +3,3 @@
   display: flex;
   justify-content: flex-end;
 }
-
```
kriskoder, line 6: @joshwooding 
This looks to me like a leftover after moving `DrawerCloseButton` away from `variant` in 2024, and it only targets `secondary`. I would remove it, but double checking with you first.

jake-costa (dismissed):

packages/core/src/drawer/DrawerCloseButton.css
joshwooding: It needs to be kept, removing it is a breaking change

kriskoder: reverted

navkaur76 (dismissed):

packages/core/src/drawer/DrawerHeader.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+export interface DrawerHeaderProps extends ComponentPropsWithoutRef<"div"> {
+  /**
+   * Hides the accent bar in the Drawer Header. Defaults to `false`.
+   */
+  disableAccent?: boolean;
+  /**
+   * Displays the header at the top of the Drawer
+   */
+  header?: ReactNode;
```
wemmyo, line 24:
```suggestion
  header: ReactNode;
```
Making this required will simplify some of the logic below and is consistent with Dialog. When Header is used we would expect the `header` to be provided, so we can remove `hasHeading`

.changeset/olive-moons-dismiss.md
```diff
@@ -0,0 +1,5 @@
+---
+"@salt-ds/core": patch
+---
+
+Fixed `disableDismiss` on `Drawer` also preventing <kbd>Escape</kbd> from closing the drawer. It now only prevents dismissal on click away, matching its documentation and `Dialog`.
```
joshwooding, line 5:
```suggestion
Fixed `disableDismiss` on `Drawer` also preventing Escape from closing the drawer. It now only prevents dismissal on click away, matching its documentation and `Dialog`.
```

.changeset/thick-eagles-tickle.md
```diff
@@ -0,0 +1,26 @@
...
+  Drawer,
+  DrawerCloseButton,
+  DrawerContent,
+  DrawerHeader,
+} from "@salt-ds/core";
+
+<Drawer open={open} onOpenChange={setOpen}>
+  <DrawerHeader
+    header="Check deposit #1278"
+    actions={<DrawerCloseButton onClick={() => setOpen(false)} />}
```
joshwooding, line 22:
```suggestion
    actions={<Button appearance="transparent" onClick={() => setOpen(false)} />}
```

packages/core/stories/drawer/drawer.stories.tsx
```diff
@@ -54,9 +59,15 @@ export const Default: StoryFn<DrawerProps> = (args) => {
         {...args}
         open={openPrimary}
         onOpenChange={(newOpen) => setOpenPrimary(newOpen)}
-        style={{ width: 200 }}
+        style={{ width: 300 }}
       >
-        <DrawerCloseButton onClick={() => setOpenPrimary(false)} />
+        <DrawerHeader
+          header="Primary drawer"
+          actions={<DrawerCloseButton onClick={() => setOpenPrimary(false)} />}
```
joshwooding, line 66:
```suggestion
          actions={<Button appearance="transparent" onClick={() => setOpenPrimary(false)} />}
```

packages/core/src/drawer/DrawerHeader.tsx
```diff
@@ -0,0 +1,119 @@
...
+      window: targetWindow,
+    });
+
+    const { setHeaderId, setDescriptionId } = useDrawerContext();
+    const headingId = useId();
+    const descriptionId = useId();
+    const hasHeading = Boolean(header || preheader);
+    const hasDescription = Boolean(description);
+
+    useIsomorphicLayoutEffect(() => {
```
wemmyo, line 64: Can we replace both with useEffect?

joshwooding (changes requested):

packages/core/src/drawer/DrawerHeader.tsx
kriskoder: <DrawerHeader> is an element of updated default composition approach. Users should be able to omit header prop and use <DrawerHeader> `action` prop to render just a close button for example.

packages/core/src/drawer/DrawerHeader.tsx
kriskoder: I would keep it as it feeds id to aria labeling before screen reader reaches it, impact on browser's paint is negligible. We use this approach for setting IDs more often.

packages/core/src/drawer/DrawerCloseButton.tsx
```diff
@@ -10,6 +10,9 @@ import drawerCloseButtonCss from "./DrawerCloseButton.css";
 
 const withBaseName = makePrefixer("saltDrawerCloseButton");
 
+/**
+ * @deprecated Use `Button` in `DrawerHeader`'s `actions` instead.
```
joshwooding:
```suggestion
 * @deprecated since 1.71.0. Use `Button` in `DrawerHeader`'s `actions` instead.
```

joshwooding (dismissed):

navkaur76 (dismissed):

joshwooding (dismissed):

joshwooding (approved):

wemmyo (approved):

PR #6945, opened 2026-08-28
add footer to drawer

kriskoder:
Add Footer to the Drawer. In scope: components, tests and stories.

### DIFF
--- a/packages/core/src/drawer/DrawerFooter.css
+++ b/packages/core/src/drawer/DrawerFooter.css
@@ -2,7 +2,6 @@
   padding: var(--salt-spacing-300);
   padding-bottom: 0;
   display: flex;
-  flex-direction: row;
   justify-content: flex-end;
   align-items: center;
   gap: var(--salt-spacing-100);
--- a/packages/core/src/drawer/DrawerFooter.css
+++ b/packages/core/src/drawer/DrawerFooter.css
@@ -1,6 +1,7 @@
 .saltDrawerFooter {
   background: var(--drawer-background);
   padding: var(--salt-spacing-300);
+  padding-bottom: 0;
   display: flex;
   flex-direction: row;
   justify-content: flex-end;
@@ -8,8 +9,4 @@
   gap: var(--salt-spacing-100);
   flex-shrink: 0;
   box-sizing: border-box;
-}
-
-.saltDrawerContent + .saltDrawerFooter {
-  padding-top: 0;
 }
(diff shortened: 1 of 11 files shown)


tycup:
If you do end up adding docs, might be worth aligning on how to document these structural components - Card uses a "Sections" heading (and I updated Overlay to do the same). Not saying its the perfect way of doing it, but if it lends itself well, it could be worth doing here too

jake-costa (dismissed):

joshwooding (dismissed):

joshwooding (dismissed):

navkaur76 (dismissed):

joshwooding (dismissed):

packages/core/src/drawer/DrawerFooter.css
```diff
@@ -0,0 +1,10 @@
+.saltDrawerFooter {
+  padding: var(--salt-spacing-300);
+  padding-bottom: 0;
```
joshwooding:
```suggestion
```

Drawer footer styling is wrong

packages/core/src/drawer/DrawerFooter.css
```diff
@@ -0,0 +1,10 @@
...
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  gap: var(--salt-spacing-100);
+  flex-shrink: 0;
+  box-sizing: border-box;
+}
```
joshwooding, line 9: There's currently double padding between content and footer, it's likely you need to remove the top padding on footer when content is before it

site/src/examples/drawer/BottomDrawer.tsx
```diff
@@ -205,16 +206,16 @@ export const BottomDrawer = (): ReactElement => {
...
-                Cancel
-              </Button>
-              <Button sentiment="accented" onClick={handleClose}>
-                Update Tier
-              </Button>
-            </FlowLayout>
           </StackLayout>
         </DrawerContent>
+        <DrawerFooter>
+          <Button appearance="transparent" onClick={handleClose}>
```
wemmyo: @navkaur76 Do we expect `DrawerFooter` to be used with bottom drawer, I didn't see any designs on Figma

wemmyo: Also, If we're using multiple buttons it would most likely need to follow button bar pattern

kriskoder: good shout, docs updates are coming in 3rd PR in this stack, updated there

navkaur76 (approved):

joshwooding (approved):

PR #6950, opened 2026-09-02
Add Status Indicator with Text example

tycup:
- Would be nice if the consumer didnt have to copy the first-line alignment style; i also tried adding a parent container which would hold the prev StatusIndicator and render children alongside it but then the ref would change which is breaking.
- Alternatively, could create a new component which owns the above and composes 'StatusIndicator' (and maybe later deprecate StatusIndicator in favour of this component).

### DIFF
--- a/packages/core/stories/status-indicator/status-indicator.stories.tsx
+++ b/packages/core/stories/status-indicator/status-indicator.stories.tsx
@@ -1,6 +1,7 @@
 import {
   FlowLayout,
   StatusIndicator,
+  Text,
   VALIDATION_NAMED_STATUS,
 } from "@salt-ds/core";
 import type { Meta, StoryFn } from "@storybook/react-vite";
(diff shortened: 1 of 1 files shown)


packages/core/stories/status-indicator/status-indicator.stories.tsx
```diff
@@ -38,3 +39,32 @@ const Template: StoryFn<typeof StatusIndicator> = (args) => {
 
 export const Default = Template.bind({});
 Default.args = { status: VALIDATION_NAMED_STATUS[3] };
+
+export const WithText: StoryFn<typeof StatusIndicator> = (args) => {
+  const { status = "error" } = args;
+
+  return (
+    <div
```
joshwooding, line 47: Is there a reason why we didn't use a layout component here?

packages/core/stories/status-indicator/status-indicator.stories.tsx
```diff
@@ -38,3 +39,32 @@ const Template: StoryFn<typeof StatusIndicator> = (args) => {
...
+      style={{
+        display: "flex",
+        gap: "var(--salt-spacing-75)",
+        maxWidth: 180,
```
joshwooding, line 51: Where did the value come from?

joshwooding (dismissed):

packages/core/stories/status-indicator/status-indicator.stories.tsx
```diff
@@ -38,3 +40,26 @@ const Template: StoryFn<typeof StatusIndicator> = (args) => {
...
+        {...args}
+        aria-hidden
+        status={status}
+        style={{
+          marginTop:
+            "calc((var(--salt-text-lineHeight) - max(var(--salt-size-icon), 12px)) / 2)",
+        }}
+      />
+      <Text color={status}>
+        <i>Request could not be submitted. Please try again later.</i>
```
kriskoder: NIT: i would suggest using css styling instead of <i>, same way docs example does.
If you prefer to stick to <i>, then use <Text as="i">.
Whichever option you choose, I would suggest using same technique in story and example.

kriskoder (commented):
all good, just one NIT

kriskoder (approved):

joshwooding (approved):

PR #6956, opened 2026-09-02
Improve multiline input character count example

### LINKED ISSUE #4566: Make it easier to provide an accessibility experience when using a character counter
### Description

Having a character limit/count is a popular requirement and can be tricky to implement in a way that meets out standards for screen reader users. We should either add more logic to improve the default experience or create an example to show how it is done.

### Expected behavior

.

### Alternatives and examples

.

### Context

.

joshwooding:
Closes #4566


jake-costa:
@joshwooding 
Took a look at your changes. I like that you added `aria-invalid`, that's a great call.

One thing I wanted to run by you. Now that we're going through the aria announcer provider, I think we'd want to tweak how we're handling the `aria-describedby`.

Right now it points at the counter label, and since that text changes on every keystroke, screen readers pick up the change and read it out. So in NVDA you end up hearing "3/10", "4/10", "5/10" in between each character you type. The debounce on `useAriaAnnouncer` doesn't help with it either, since the description is its own separate mechanism from the live region. You also end up getting the count twice, once from the description and once from the announcer.

What I'd suggest is pointing `aria-describedby` at static text that just states the limit, something like "You can enter up to 10 characters," and letting the announcer handle the running count. Then hide the visible counter from screen readers so it's not a third copy of the same thing. That's how [GOV.UK](https://design-system.service.gov.uk/components/character-count/) and the [GC Design System](https://github.com/cds-snc/gcds-components/issues/1110) both landed on it, and there's good background from [Adrian Roselli](https://adrianroselli.com/2022/04/accessible-description-exposure.html) on why changing descriptions are unreliable in general.

The other thing I noticed is that the assertive announcement for the character limit being exceeded is working correctly, that fires exactly like it should. But any character typed after that doesn't trigger anything, so it goes quiet at 12, 13, 14 while the counter keeps climbing on screen. Not required, but I'd highly recommend we re-trigger the error announcement for every character typed past the max.

jake-costa (dismissed):

jake-costa (approved):

tycup (approved):

PR #6957, opened 2026-09-03
Add Overlay Footer

tycup:


packages/core/src/overlay/OverlayFooter.css
```diff
@@ -0,0 +1,12 @@
+.saltOverlayFooter {
+  padding: var(--saltOverlay-footer-padding, var(--salt-spacing-100));
+  display: flex;
+  align-items: center;
```
kriskoder, line 4: to match placement of action buttons in footer in dialog and drawer
```suggestion
    align-items: center;
    justify-content: flex-end;
```

packages/core/src/overlay/OverlayFooter.css
```diff
@@ -0,0 +1,12 @@
...
+  gap: var(--salt-spacing-100);
+  box-sizing: border-box;
+}
+
+.saltOverlayHeader + .saltOverlayFooter,
+.saltOverlayPanelContent + .saltOverlayFooter {
```
kriskoder, line 10:
```suggestion
.saltOverlayHeader ~ .saltOverlayFooter,
.saltOverlayPanelContent ~ .saltOverlayFooter {
```

site/docs/components/overlay/examples.mdx
```diff
@@ -20,15 +20,17 @@ Use the `placement` prop to position overlay content to the `top`, `bottom`, `le
...
 <LivePreview componentName="overlay" exampleName="Placement" />
 
-## Header
+## Sections
 
-`OverlayHeader` follows our standardized header for container components and app regions, and it can be added as a child of the `OverlayPanel` component to provide a structured header for overlay. The header includes a title and actions that follows our [Header Block pattern](/salt/patterns/header-block). This approach is recommended over using the `OverlayPanelCloseButton` separately.
+Use `OverlayHeader`, `OverlayPanelContent` and `OverlayFooter` to create distinct padded sections within `OverlayPanel`. Sections need to be direct children of `OverlayPanel` so their spacing rules apply; adjacent sections share a single `--salt-spacing-100` gap.
 
-<LivePreview componentName="overlay" exampleName="WithHeader" />
+`OverlayHeader` follows the [Header Block pattern](/salt/patterns/header-block). Pass the title to `header`, add `preheader` or `description` when needed and put controls such as the close button in `actions`.
```
kriskoder: isn't the close button in the header deprecated?
`and put controls such as the close button in `actions`.

site/src/examples/overlay/WithFooter.tsx
kriskoder, line 1: i would rename it to `WithSections` as it shows more than just footer and would be consistent with docs section name

site/docs/components/overlay/examples.mdx
```diff
@@ -20,15 +20,17 @@ Use the `placement` prop to position overlay content to the `top`, `bottom`, `le
...
+
+<LivePreview componentName="overlay" exampleName="WithFooter" />
 
 ## Long content
 
-When overlay content extends beyond the set height, the content defaults to scroll. The close button remains sticky and accessible at all times.
+When content in `OverlayPanelContent` extends beyond its set height, it scrolls while the header and its close button stay in place.
```
kriskoder: this is another example that uses close button, that's marked as deprecated below

site/src/examples/overlay/CloseButton.tsx
```diff
@@ -27,10 +27,13 @@ export const CloseButton = (): ReactElement => {
       <OverlayPanel aria-labelledby={id}>
         <OverlayPanelCloseButton onClick={handleClose} />
         <OverlayPanelContent>
-          <h3 className={styles.contentHeading} id={id}>
+          <H3
```
kriskoder, line 30: I would suggest using OverlayHeader instead of `<H3>`

site/src/examples/overlay/Default.tsx
```diff
@@ -20,17 +21,18 @@ export const Default = (): ReactElement => {
       </OverlayTrigger>
       <OverlayPanel aria-labelledby={id}>
         <OverlayPanelContent>
-          <h3 className={styles.contentHeading} id={id}>
+          <H3
```
kriskoder, line 24: I would suggest using OverlayHeader instead of `<H3>`

site/src/examples/overlay/HideArrow.tsx
```diff
@@ -19,10 +19,13 @@ export const HideArrow = (): ReactElement => {
       </OverlayTrigger>
       <OverlayPanel aria-labelledby={id}>
         <OverlayPanelContent>
-          <h3 className={styles.contentHeading} id={id}>
+          <H3
```
kriskoder, line 22: I would suggest using OverlayHeader instead of `<H3>`

site/src/examples/overlay/Placement.tsx
```diff
@@ -22,34 +24,29 @@ export const OverlayTemplate = (props: OverlayProps): ReactElement => {
       </OverlayTrigger>
       <OverlayPanel aria-labelledby={id}>
         <OverlayPanelContent>
-          <h3 id={id} className={styles.contentHeading}>
+          <H3
```
kriskoder, line 27: I would suggest using OverlayHeader instead of `<H3>`

site/src/examples/overlay/WithActions.tsx
```diff
@@ -72,38 +74,42 @@ const WithActionsContent = ({ id, onClose }: WithActionsContentProps) => {
...
-          {checkboxesData.map((data) => (
-            <Checkbox key={data.value} {...data} />
-          ))}
-        </CheckboxGroup>
-        <Divider variant="secondary" />
-        <Button
-          style={{ float: "right", marginRight: 2 }}
-          onClick={handleExport}
+      <OverlayPanelContent>
+        <H3
```
kriskoder, line 78: I would suggest using OverlayHeader instead of `<H3>`

site/src/examples/overlay/CloseButton.tsx
```diff
@@ -27,10 +27,13 @@ export const CloseButton = (): ReactElement => {
       <OverlayPanel aria-labelledby={id}>
         <OverlayPanelCloseButton onClick={handleClose} />
```
kriskoder, line 28: how about moving that button into header's actions? I wouldn't suggest showing an example of deprecated code.

site/docs/components/overlay/examples.mdx
```diff
@@ -20,15 +20,17 @@ Use the `placement` prop to position overlay content to the `top`, `bottom`, `le
...
 
 <LivePreview componentName="overlay" exampleName="LongContent" />
```
kriskoder, line 35: I would remove close button from header section, as this example has. a bit different purpose.

site/src/examples/overlay/index.module.css
```diff
@@ -1,18 +0,0 @@
...
-
-.dividerContainer {
-  display: flex;
-  align-items: center;
-  width: 100%;
-  padding-top: 8px;
-  padding-bottom: 8px;
-}
-
-.divider {
```
kriskoder, line 15: divider needs some tweaking to be aligned with similar components like Dialog.

Currently top divider has a delay when showing on the screen. Bottom divider needs to be added in case there are all sections used and it has long content.
Have a look at `Dialog`.

packages/core/stories/overlay/overlay.stories.tsx
kriskoder, line 110: I know that's not part of this PR, but a11y reflow can be improved, there's no vertical scroll, and in general I would align it with how Dialog handles it in a11y reflow.

kriskoder (changes requested):
I would suggest aligning new composition approach in all stories, qa stories and docs -> moving standalone headers and close buttons into `OverlayHeader` components, replacing raw `div's` around text element with `StackedLayout` and replacing raw html text elements with our `Text` component

site/src/examples/overlay/WithHeader.tsx
kriskoder: this example is no longer used, got replaced by `WithFooter`

kriskoder, line 1: this example is no longer used, got replaced by `WithFooter`

kriskoder (changes requested):

packages/core/src/overlay/OverlayPanel.css
```diff
@@ -17,4 +17,14 @@
...
+  max-height: 100vh;
+  box-sizing: border-box;
+}
+
+.saltOverlayPanel-content {
+  max-width: inherit;
+  max-height: inherit;
+  overflow-y: auto;
+  box-sizing: border-box;
 }
```
kriskoder, line 30: footer isn't sticky without manually sized  `OverlayPanelContent` I suggest we use same approach as in Drawer

packages/core/src/overlay/OverlayFooter.css
```diff
@@ -0,0 +1,13 @@
+.saltOverlayFooter {
+  padding: var(--saltOverlay-footer-padding, var(--salt-spacing-100));
```
kriskoder, line 2: to keep it consistent with `.saltOverlayHeader` in `OverlayHeader.css` I suggest adding `--saltOverlay-header-padding` in overlay header css, as we use that approach in Card, Drawer and other elements of Overlay.

kriskoder (changes requested):
just 2 things left

site/src/examples/overlay/WithActions.tsx
kriskoder, line 1: in this example let's render children of  OverlayPanel directly, without wrapping them in `WithActionsContent`. the hasOverlaySecion won't work with wrapper, even if it's such a plain one as this one. User cannot see the bug because css selector catches it. Same goes for WithActions story. Drawer is affected in the same way, I'll provide bugfix in separate PR.

kriskoder (changes requested):
good updates, a few small things left, as in the code comment

packages/core/src/overlay/OverlayPanelContent.css
kriskoder, line 1: I would update the padding logic to match drawer, where header and footer they set their own padding in 4 directions and content sets only horizontal padding -> that's where all 3 composition elements are used.
If there's no header or footer, then content will add that missing padding conditionally.

kriskoder (changes requested):
I added one more thing, now it should be all :)

packages/core/src/overlay/OverlayPanelContent.css
tycup: i've pushed it more towards this direction, since i agree the encapsulation is a bit nicer but ive kept the content section starting off as 4 sided padding but i didnt align exactly with Drawer to prevent a breaking change on what --saltOverlay-content-padding does and the exported close button

we should also probably look to consolidate and align with Card

kriskoder: agree, all those sectioned components need a review session, as Dialog is another component where the approach drifts away.

kriskoder (dismissed):
approved

site/docs/components/overlay/examples.mdx
```diff
@@ -20,15 +20,17 @@ Use the `placement` prop to position overlay content relative to the trigger. Th
 
 <LivePreview componentName="overlay" exampleName="Placement" />
 
-## Header
+## Sections
 
-`OverlayHeader` follows our standardized header for container components and app regions, and it can be added as a child of the `OverlayPanel` component to provide a structured header for overlay. The header includes a title and actions that follows our [Header Block pattern](/salt/patterns/header-block). This approach is recommended over using the `OverlayPanelCloseButton` separately.
+Use `OverlayHeader`, `OverlayPanelContent` and `OverlayFooter` to create distinct padded sections within `OverlayPanel`. Sections need to be direct children of `OverlayPanel` so their spacing rules apply; adjacent sections share a single `--salt-spacing-100` gap. When `OverlayPanel` has a constrained height, `OverlayHeader` and `OverlayFooter` stay in place while `OverlayPanelContent` scrolls.
```
joshwooding: I wonder if we should avoid mentioning the spacing token used

joshwooding (approved):

kriskoder (approved):

PR #6962, opened 2026-09-08
normalise css in stories

kriskoder:
PR touches only stories and qa stories changed in the target branch.

### DIFF
--- a/packages/core/stories/salt-provider/salt-provider.stories.tsx
+++ b/packages/core/stories/salt-provider/salt-provider.stories.tsx
@@ -23,12 +23,10 @@ export const Default = () => {
   return (
     <SaltProvider density="high" mode="light">
       <Card>
-        <div>
-          <h1 style={{ marginBottom: "var(--salt-spacing-300)" }}>
-            This is Card
-          </h1>
+        <StackLayout>
+          <h1>This is Card</h1>
           <span>Using Nested DOM Elements</span>
-        </div>
+        </StackLayout>
       </Card>
     </SaltProvider>
   );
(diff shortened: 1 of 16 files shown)


packages/core/stories/salt-provider/salt-provider.stories.tsx
```diff
@@ -23,12 +23,10 @@ export const Default = () => {
   return (
     <SaltProvider density="high" mode="light">
       <Card>
-        <div>
-          <h1 style={{ marginBottom: "var(--salt-spacing-300)" }}>
-            This is Card
-          </h1>
+        <StackLayout>
+          <h1>This is Card</h1>
```
joshwooding, line 27: Can we change this to use a Salt text component

packages/core/stories/salt-provider/salt-provider.stories.tsx
joshwooding, line 1: Actually the headers in this file too please

packages/core/stories/text/text.stories.tsx
```diff
@@ -115,57 +114,60 @@ export const Display = FigureTextComponent.bind({});
...
-    </H3>
-    <br />
-    <H4>
-      This is header 4 <strong>emphasis high</strong>
-    </H4>
-    <H4>
-      This is header 4 <small>emphasis low</small>
-    </H4>
-  </>
+  <StackLayout>
```
joshwooding, line 117: I think we should maybe make this gap 6

joshwooding:
Nice work, a few things but should be good after

joshwooding (changes requested):

packages/core/stories/text/text.stories.tsx
kriskoder: yeah, better separation especially around H1

packages/core/stories/salt-provider/salt-provider.stories.tsx
kriskoder: changed both `h` and `p `

PR #6975, opened 2026-09-15
Feat/chris/normalise css all stories

kriskoder:


packages/core/stories/border-item/border-item.stories.tsx
```diff
@@ -16,23 +16,23 @@ const Template: StoryFn<typeof BorderItem> = (args) => {
   return (
     <BorderLayout columnGap={1} rowGap={1} className="layout-container">
       <BorderItem position="north" className="border-item">
-        <p>North</p>
+        <Text>North</Text>
```
joshwooding, line 19: NIT: we should probably use as="p" on these

kriskoder: yeah, agree. I'll do full sweep through all stories as main got updated with new stories recently.

PR #6976, opened 2026-09-15
refactor: normalise css in docs

kriskoder:
normalising css in docs and examples


site/src/examples/flex-layout/Spacing.tsx
```diff
@@ -42,7 +43,7 @@ export const Spacing = (): ReactElement => {
               key={`item-${index + 1}`}
               padding={2}
             >
-              <p>{index + 1}</p>
+              <Text>{index + 1}</Text>
```
joshwooding, line 46: We should be consistent and use as="p", like we have in the other places.

kriskoder: good catch. I'll do full sweep through all stories as main got updated with new stories recently.

